PR #201, opened 2014-08-28
Set 'time' column in InfluxDB driver

vishh:
This will preserve the per-second granularity even if cAdvisor were to batch writes


mboussaa:
How can I change the default timestamp granularity? by default it is 1s right?

PR #371, opened 2014-12-17
Add Docker's hostname as one of the cAdvisor aliases for the container

hinesmr:
What's the "right" way to take metadata from "docker inspect" and put it into cAdvisor?

We really need the hostname to help identify the statistics provided by Cadvisor. This 1-line patch fixes the problem.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -116,7 +116,7 @@ func newDockerContainerHandler(
 	// Add the name and bare ID as aliases of the container.
 	handler.aliases = append(handler.aliases, strings.TrimPrefix(ctnr.Name, "/"))
 	handler.aliases = append(handler.aliases, id)
-    handler.aliases = append(handler.aliases, ctnr.Config.Hostname)
+	handler.aliases = append(handler.aliases, ctnr.Config.Hostname)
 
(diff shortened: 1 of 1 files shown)


vmarmol:
Any updates on this? :) Were you able to reuse Docker's name field as an identifier?

hinesmr:
No, that's not possible - it's already used by our management software stack for other purposes. And it's made worse by the fact that Docker does not (yet) support arbitrary key/name value pairs upon creating new containers, so there's no way to store additional metadata inside Docker to be consumed later.

emmanuel:
@hinesmr gliderlabs/registrator reads environment variables from (other) containers as a form of arbitrary metadata. Metadata in that form is not necessarily easily accessible (depending on how you want to access it), but the approach might be worth a look.

hinesmr:
Very interesting. How do you install those environment variables on a per-container basis? Not through libcontainer or docker? Do you install them before the container starts or somewhere in the container filesystem? (I would like to avoid reaching into the container filesystem if at all possible).

timstclair:
This PR looks abandoned. Please reopen it if that's not the case.

PR #375, opened 2014-12-18
Add cpu frequency to machine info.

rjnagal:


mboussaa:
@rjnagal how did you do finally to get the %cpu usage of one container? 
Did you consider cpu frequency to compute it?

Issue #414, opened 2015-01-03
failed to get cache information for node error on VPS


richp10:
On a VPS with two cores:
# ls /sys/devices/system/cpu
# cpu0  cpu1  kernel_max (etc)

Run ./cadvisor
F0103 14:13:33.988492 00589 cadvisor.go:76] Failed to create a Container Manager: failed to get cache information for node 0: open /sys/devices/system/cpu/cpu2/cache: no such file or directory

For some reason it seems to assume more cpu cores than exist on the system.  Tried to figure out the code but my go isn't good enough..

rjnagal:
I think I know what's going on. Will send a fix in the evening.

rjnagal:
@richp10 Can you please try jnagal/cadvisor:canary image to verify the patch? Thanks.

richp10:
Worked perfectly, thanks..

R

----------------------------
Richard Phillips
----------------------------

rjnagal:
Thanks for verifying @richp10 . We'll cut a release this week.

PR #415, opened 2015-01-04
Fix reading cache info for machines with unused packages.

rjnagal:
Use thread ids instead of core-ids to identify caches.


rjnagal:
Fixes #414

PR #419, opened 2015-01-05
Cleanup of libcontainer state generation.

vmarmol:


rjnagal:
LGTM.

This looks fine. I am slightly nervous of mis-using state for anything other than network devices. We should look for a way to hide that from rest of the code.

PR #420, opened 2015-01-05
Don't limit gauges width. Also make them smaller.

vmarmol:


rjnagal:
I liked the bigger ones better. But these are fine too - that's just probably my age :)

PR #422, opened 2015-01-05
Use strconv.ParseInt where possible.

vmarmol:


utils/sysfs/sysfs.go
```diff
@@ -150,9 +150,8 @@ func (self *realSysFs) GetCacheInfo(id int, name string) (CacheInfo, error) {
 	if err != nil {
 		return CacheInfo{}, err
 	}
-	var level int
-	n, err = fmt.Sscanf(string(out), "%d", &level)
-	if err != nil || n != 1 {
+	level, err := strconv.ParseInt(string(out), 10, 32)
```
rjnagal: Some versions of Parse* wasn't handling these without stripping spaces because of /t and /n returned by sysfs.

vmarmol: I'll just close this PR, seems minor and not necessary.

PR #425, opened 2015-01-08
Handle initial zero value of lastError.

vmarmol:
Also reduce logging to once every minute.


vmarmol:
Ping @rjnagal this didn't let us log anything since the initial value was always zero.

vmarmol:
Open to discussion on whether a minute or more/less often. An hour seemed way too long :)

vishh:
LGTM

rjnagal:
doh. One minute is fine for now.

PR #426, opened 2015-01-08
Updating to latest libcontainer.

vmarmol:


vmarmol:
Ping @rjnagal @vishh.

@kateknister this will fix the issue you were seeing before.

vishh:
LGTM

PR #427, opened 2015-01-08
Add percentiles utility methods.

rjnagal:
This is copied verbatim for our older prototype. We'll use it as
a starting point for building stats summary.

### DIFF
--- a/utils/percentiles.go
+++ b/utils/percentiles.go
@@ -1,0 +1,135 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"math"
+	"sort"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/cadvisor/info"
+)
+
+const milliSecondsToNanoSeconds = 1000000
+const secondsToMilliSeconds = 1000
+
+type uint64Slice []uint64
+
+func (a uint64Slice) Len() int           { return len(a) }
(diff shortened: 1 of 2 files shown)


vishh:
LGTM

utils/percentiles.go
```diff
@@ -0,0 +1,126 @@
...
+	// Average over the collected sample.
+	Mean uint64 `json:"mean"`
+	// Max seen over the collected sample.
+	Max uint64 `json:"max"`
+	// 90th percentile over the collected sample.
+	Ninety uint64 `json:"ninety"`
+}
+
+// Get 90th percentile of the provided samples. Round to integer.
+func Get90Percentile(samples uint64Slice) uint64 {
```
vmarmol: what about making it a method on the object (uint64Slice)?

utils/percentiles.go
```diff
@@ -0,0 +1,126 @@
...
+	index := int(idx)
+	percentile := float64(samples[index-1])
+	if index > 1 || index < count {
+		percentile += frac * float64(samples[index]-samples[index-1])
+	}
+	return uint64(percentile)
+}
+
+// Add new sample to existing average.
+func GetMean(mean float64, value uint64, count uint64) float64 {
```
vmarmol: WDYT:

func NewMean(size uint64) {
  // .. create ...
}

type Mean struct {
  size uint64
  mean uint64
}

func (self *Mean) Add(value uint64) {
  // Do what you did here
}

utils/percentiles.go
```diff
@@ -0,0 +1,126 @@
...
+	if count < 1 {
+		return 0
+	}
+	c := float64(count)
+	v := float64(value)
+	return (mean*(c-1) + v) / c
+}
+
+func GetPercentiles(stats []*info.ContainerStats) (Percentiles, Percentiles) {
```
vmarmol: Can we add a comment saying what output is what?

vmarmol: And this is memory usage and total CPU usage

vmarmol: Ah, it's working set :)

utils/percentiles.go
rjnagal: Made it a method.

utils/percentiles.go
rjnagal: Yeah, that's nicer. Changed :)

utils/percentiles.go
rjnagal: Updated comment.

vmarmol:
LGTM

PR #428, opened 2015-01-09
Add missing copyright notices.

rjnagal:
We'll add a travis rule when this is in.


vmarmol:
LGTM

vmarmol:
I am not proud to say that all those were files I added :P

PR #431, opened 2015-01-10
Remove sched stats scraping utils.

rjnagal:
sched_debug is getting the wrong load information. Runnable tasks list
in the output include running and sleeping tasks. We only need to look
at nr_running for each scheduling entity to figure out load. We also
don't need per-core stats.

I am going to redo these to derive per-cgroup load from nr_running.


vmarmol:
LGTM

PR #433, opened 2015-01-12
Adding documentation for enabling memory cgroup in Debian.

### LINKED ISSUE #432: cAdvisor does not return memory data on Debian
I have had some troubles making cAdvisor work on Debian, especially returning data on the memory, and I wanted to share the solution! The use case is a [Docker based PaaS dedicated to Java/JEE applications] (http://cloudunit.fr/).

By default, the Cgroup memory controller is in the state missing on Debian, so cAdvisor does not collect the relevant data.

The solution simply consists in adding the line:
```bash
GRUB_CMDLINE_LINUX="cgroup_enable=memory"
```
to the file `/etc/default/grub`. Then, one has to run: `sudo update-grub2` and reboot.

[A more detailed article](http://blog.treeptik.fr/2014/12/cadvisor-sous-debian/) is available on our blog.

vmarmol:
Fixes #432.


rjnagal:
LGTM

PR #434, opened 2015-01-12
cAdvisor 0.8.0

vmarmol:


vishh:
LGTM.

rjnagal:
Do we still have the issue with cAdvisor not able to get to docker socket
on Redhat?

vmarmol:
Note sure...we should test :)

PR #437, opened 2015-01-13
Add a utility to read cpu load stats.

rjnagal:


utils/cpuload/conn.go
```diff
@@ -0,0 +1,89 @@
...
+type Connection struct {
+	fd   int
+	pid  uint32
+	seq  uint32
+	addr syscall.SockaddrNetlink
+	rbuf *bufio.Reader
+}
+
+// Create and bind a new netlink socket.
+func NewConnection() (*Connection, error) {
```
vmarmol: Do we want to have a separate netlink in util? It seems like we're starting to make an interface for it. I'm fine leaving here for now and making all private.

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/golang/glog"
+)
+
+type CpuLoadReader struct {
+	familyId uint16
+	conn     *Connection
+}
+
+func NewCpuLoadReader() (*CpuLoadReader, error) {
```
vmarmol: Can just be New() since it is the only thing exposed by the cpuload package :)

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+	conn, err := NewConnection()
+	if err != nil {
+		fmt.Printf("failed to create new connection")
```
vmarmol: glog? or was this for testing? Here and elsewhere.

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (self *CpuLoadReader) Close() {
+	if self.conn != nil {
+		self.conn.Close()
+	}
+}
+
+// This mirrors kernel internal structure.
+type LoadStats struct {
+	NrSleeping       uint64 // Number of tasks sleeping
```
vmarmol: Maybe add the json info for these?

utils/cpuload/netlink.go
```diff
@@ -0,0 +1,232 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"syscall"
+)
+
+const (
+	GENL_ID_CTRL            = syscall.NLMSG_MIN_TYPE
```
vmarmol: I personally like these consts like this, but Go says that they should be camel-case and start with lowercase (since they're private)

utils/cpuload/netlink.go
```diff
@@ -0,0 +1,232 @@
...
+	GENL_VERSION_CTRL       = 0x1
+	TASKSTATS_GENL_NAME     = "TASKSTATS"
+	CGROUPSTATS_CMD_ATTR_FD = 0x1
+	CTRL_ATTR_FAMILY_ID     = 0x1
+	CTRL_ATTR_FAMILY_NAME   = 0x2
+	CTRL_CMD_GETFAMILY      = 0x3
+)
+
+type GenMsghdr struct {
```
vmarmol: I'm guessing all these structs are package-private? Should be lowercase if that's the case.

utils/cpuload/netlink.go
```diff
@@ -0,0 +1,232 @@
...
+// Adds attribute info (length and type), followed by the data and necessary padding.
+// Can be called multiple times to add attributes. Only fixed size and string type
+// attributes are handled. We don't need nested attributes for task stats.
+func addAttribute(buf *bytes.Buffer, attrType uint16, data interface{}, dataSize int) {
+	attr := syscall.RtAttr{
+		Len:  syscall.SizeofRtAttr,
+		Type: attrType,
+	}
+	attr.Len += uint16(dataSize)
+	binary.Write(buf, binary.LittleEndian, attr)
```
vmarmol: TODO to handle big-endian? We may not need to worry for a while, but just have some way to catch it if it happens :P

utils/cpuload/netlink.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	err = binary.Read(buf, binary.LittleEndian, &m.Stats)
+	if err != nil {
+		return m, err
+	}
+	return m, err
+}
+
+// Verify and return any error reported by kernel.
+func VerifyHeader(msg syscall.NetlinkMessage) error {
```
vmarmol: Do we mean for this to be public?

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+	NrRunning        uint64 // Number of tasks running
+	NrStopped        uint64 // Number of tasks in stopped state
+	NrUinterruptible uint64 // Number of tasks in uninterruptible state
+	NrIoWait         uint64 // Number of tasks waiting on IO
+}
+
+// Returns instantaneous number of running tasks in a group.
+// Caller can use historical data to calculate cpu load.
+// path is a cgroup path under cpu hierarchy eg. "/sys/fs/cgroup/cpu/cadvisor"
```
vmarmol: "path is an absolute filesystem path for a container under the CPU cgroup hierarchy"?

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+// NOTE: non-hierarchical load is returned. It does not include load for subcontainers.
+func (self *CpuLoadReader) GetCpuLoad(path string) (uint64, error) {
+	if len(path) == 0 {
+		return 0, fmt.Errorf("Cgroup path can not be empty!")
```
vmarmol: nit: errors start with lowercase. Here and below.

utils/cpuload/cpuload.go
```diff
@@ -0,0 +1,83 @@
...
+	NrStopped        uint64 // Number of tasks in stopped state
+	NrUinterruptible uint64 // Number of tasks in uninterruptible state
+	NrIoWait         uint64 // Number of tasks waiting on IO
+}
+
+// Returns instantaneous number of running tasks in a group.
+// Caller can use historical data to calculate cpu load.
+// path is a cgroup path under cpu hierarchy eg. "/sys/fs/cgroup/cpu/cadvisor"
+// NOTE: non-hierarchical load is returned. It does not include load for subcontainers.
+func (self *CpuLoadReader) GetCpuLoad(path string) (uint64, error) {
```
vmarmol: WDYT of just returning all of LoadStats?

vmarmol:
This looks awesome! Mainly nits on my part :)

utils/cpuload/conn.go
rjnagal: I didn't do that yet, because we only implemented the two interfaces we needed for getting cpu load. If we need more, I think we can export a more elaborate netlink implementation. There are few in github, but none of those directly expose the task stats stuff.

I moved most of the netlink related stuff in a single file, so we can move/replace it if needed.

utils/cpuload/netlink.go
rjnagal: Moved the endianness to a global var for now. I need to test what it does on big-endian machines, but I think the kernel might be reporting these in a specific order. We can always check the machine and set up endianness globally if that's not the case. Added a TODO.

utils/cpuload/cpuload.go
rjnagal: done.

utils/cpuload/cpuload.go
rjnagal: Done.

utils/cpuload/cpuload.go
rjnagal: For testing. Merged with error.

utils/cpuload/cpuload.go
rjnagal: Now that these are returned, added json info.

utils/cpuload/netlink.go
rjnagal: Having the same name as kernel consts makes it easier to lookup in kernel sources :(

Changed it to go fmt, but added kernel equivalent as comments.

utils/cpuload/netlink.go
rjnagal: Done.

utils/cpuload/netlink.go
rjnagal: no :)

utils/cpuload/cpuload.go
rjnagal: gah. I am going to add another hook to catch these :)

utils/cpuload/cpuload.go
rjnagal: I was debating about it. Maybe it'll be nice for UI display. Updated.

vmarmol:
LGTM, thanks again @rjnagal!

Issue #436, opened 2015-01-13
Host network stats don't work in CoreOS


vmarmol:
The network devices cAdvisor finds are:

```
vmarmol@vmarmol-coreos / $ ifconfig
docker0: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1460
        inet 172.17.42.1  netmask 255.255.0.0  broadcast 0.0.0.0
        inet6 fe80::5484:7aff:fefe:9799  prefixlen 64  scopeid 0x20<link>
        ether 56:84:7a:fe:97:99  txqueuelen 0  (Ethernet)
        RX packets 7361  bytes 12847469 (12.2 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 7449  bytes 707220 (690.6 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

ens4v1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1460
        inet 10.240.79.9  netmask 255.255.255.255  broadcast 10.240.79.9
        inet6 fe80::4001:aff:fef0:4f09  prefixlen 64  scopeid 0x20<link>
        ether 42:01:0a:f0:4f:09  txqueuelen 1000  (Ethernet)
        RX packets 272018  bytes 80617348 (76.8 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 264873  bytes 72187423 (68.8 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 0  (Local Loopback)
        RX packets 1351  bytes 399896 (390.5 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 1351  bytes 399896 (390.5 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

None are `eth` devices which is what we expect.

rjnagal:
We can scan all ethernet devices and use the first one, instead of limiting to eth*. Can we ignore docker*, or shall we report them too? The current public API limits us to exporting only one i/f.

vmarmol:
We're probably fine ignoring veth,docker, and loopback. We can then pick the first of those.

[...]

PR #439, opened 2015-01-14
Handle root network device names other than eth.

rjnagal:
New logic ignores veth, docker, and loopback devices.

### DIFF
--- a/utils/sysinfo/sysinfo.go
+++ b/utils/sysinfo/sysinfo.go
@@ -77,8 +77,16 @@ func GetNetworkDevices(sysfs sysfs.SysFs) ([]info.NetInfo, error) {
 	netDevices := []info.NetInfo{}
 	for _, dev := range devs {
 		name := dev.Name()
-		// Only consider ethernet devices for now.
-		if !strings.HasPrefix(name, "eth") {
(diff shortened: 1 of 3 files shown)


rjnagal:
Fixes #436 Verified on CoreOS.

utils/sysinfo/sysinfo.go
```diff
@@ -77,8 +77,16 @@ func GetNetworkDevices(sysfs sysfs.SysFs) ([]info.NetInfo, error) {
 	netDevices := []info.NetInfo{}
 	for _, dev := range devs {
 		name := dev.Name()
-		// Only consider ethernet devices for now.
-		if !strings.HasPrefix(name, "eth") {
+		// Ignore docker, loopback, and veth devices.
+		ignoredDevices := []string{"lo", "veth", "docker"}
```
vmarmol: nit: can we use a map instead :)

rjnagal: I thought it won't help because we have to match just the prefix?

vmarmol: Ah, good point.

vmarmol:
LGTM

PR #443, opened 2015-01-16
Add free number of bytes to filesystem stats of machine api

arkadijs:
Probably some tests should be added or changes, but at least it now matches UI output that is able to display filesystem free space.

### DIFF
--- a/info/machine.go
+++ b/info/machine.go
@@ -17,9 +17,10 @@ package info
 type FsInfo struct {
 	// Block device associated with the filesystem.
 	Device string `json:"device"`
-
 	// Total number of bytes available on the filesystem.
 	Capacity uint64 `json:"capacity"`
+	// Free number of bytes available on the filesystem.
+	Free uint64 `json:"free"`
 }
 
 type Node struct {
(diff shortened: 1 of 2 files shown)


info/machine.go
```diff
@@ -17,9 +17,10 @@ package info
 type FsInfo struct {
 	// Block device associated with the filesystem.
 	Device string `json:"device"`
-
 	// Total number of bytes available on the filesystem.
 	Capacity uint64 `json:"capacity"`
+	// Free number of bytes available on the filesystem.
+	Free uint64 `json:"free"`
```
vishh: This should probably go into "root" cgroup stats. I am curious as to why this is necessary since you can compute 'free' from 'capacity' which is here and 'usage' which is in "root" cgroup stats.

arkadijs:
Strictly speaking you're right - usage goes under /container but /machine describes provisioned capacity. I wanted to display basic machine info without going to two places.

vishh:
'/machine' is meant to provide static information only. If looking up two different APIs is an issue, I would suggest placing capacity along with usage in '/container'.

arkadijs:
I understand.
/containers also provide `capacity`.
Thank you for looking into this.

PR #445, opened 2015-01-16
Add task load stats to containers.

rjnagal:
The stats are only populated when cAdvisor is running outside network namespaces.
We'll add a different backend to retrieve the same data from within namespaces.

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -25,6 +25,7 @@ import (
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/info"
 	"github.com/google/cadvisor/storage"
+	"github.com/google/cadvisor/utils/cpuload"
 )
 
 // Housekeeping interval.
@@ -43,6 +44,7 @@ type containerData struct {
 	info                 containerInfo
 	storageDriver        storage.StorageDriver
 	lock                 sync.Mutex
+	loadReader           *cpuload.CpuLoadReader
 	housekeepingInterval time.Duration
 	lastUpdatedTime      time.Time
 	lastErrorTime        time.Time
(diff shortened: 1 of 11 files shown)


manager/container.go
```diff
@@ -234,6 +237,17 @@ func (c *containerData) updateStats() error {
 	if stats == nil {
 		return nil
 	}
+	if c.loadReader != nil {
+		path, err := c.handler.GetCgroupPath("cpu")
+		if err == nil {
+			loadStats, err := c.loadReader.GetCpuLoad(path)
+			if err != nil {
+				glog.V(1).Infof("failed to get load stat for %q - path %q, error %s", c.info.Name, path, err)
```
vmarmol: Why not error like we do for other parts of the stats?

manager/manager.go
```diff
@@ -119,12 +120,21 @@ type manager struct {
...
 	dockerContainersRegexp *regexp.Regexp
+	loadReader             *cpuload.CpuLoadReader
 }
 
 // Start the container manager.
 func (self *manager) Start() error {
+	// Create cpu load reader.
+	cpuLoadReader, err := cpuload.New()
+	if err != nil {
+		glog.Infof("could not initialize cpu load reader: %s", err)
```
vmarmol: nit: Warning? :)

vmarmol:
LGTM, only minor nits. (and the Travis failure)

manager/manager.go
rjnagal: This is going to show up as a warning in all namespaced containers. We can update it to warning once we have the sched_debug backend.

vmarmol: SGTM, TODO?

manager/container.go
rjnagal: done.

manager/manager.go
rjnagal: fine, added...

rjnagal:
Updated the failing tests :)

vmarmol:
LGTM

jzy1688:
@rjnagal Hello. Could you tell me how could I enable load_stats? I always get "has_load =flase"

rjnagal:
@jzy1688 We disabled load stats as we are re-working the implementation. It needs a few more tweaks before we can enable it again.

jzy1688:
Thanks

PR #446, opened 2015-01-16
Add scheduler type to disk info.

rjnagal:
Type can be one of "none", "noop", "deadline", "cfq".
For block devices that don't use scheduler (like dm), the type will be "none".
We'll also report "none" for partitions when we start reporting those.

### DIFF
--- a/info/machine.go
+++ b/info/machine.go
@@ -93,6 +93,9 @@ type DiskInfo struct {
 
 	// Size in bytes
 	Size uint64 `json:"size"`
+
+	// I/O Scheduler - one of "none", "noop", "cfq", "deadline"
+	Scheduler string `json:"scheduler"`
 }
 
 type NetInfo struct {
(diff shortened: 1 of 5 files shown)


rjnagal:
Sample output:

"disk_map":{
"252:0": {"name":"dm-0","major":252,"minor":0,"size":107374182400,"scheduler":"none"},
"252:1": {"name":"dm-1","major":252,"minor":1,"size":10737418240,"scheduler":"none"},
"8:0": {"name":"sda","major":8,"minor":0,"size":10737418240,"scheduler":"cfq"}
}

info/machine.go
```diff
@@ -93,6 +93,9 @@ type DiskInfo struct {
 
 	// Size in bytes
 	Size uint64 `json:"size"`
+
+	// Scheduler - one of "none", "noop", "cfq", "deadline"
```
vmarmol: nit: I/O scheduler?

vmarmol:
LGTM, feel free to merge :)

info/machine.go
rjnagal: just the comment, or field name, or json tag? I am assuming its just the comment :)

vmarmol: Yeah, just the comment :) thanks for putting up with my nits! Merging.

PR #447, opened 2015-01-17
Add I/O scheduler information to validator.

rjnagal:
Sample output:

Block device setup: [Supported and recommended]
	Atleast one device supports 'cfq' I/O scheduler. Some disk stats can be reported.
	 Disk "dm-1" Scheduler type "none".
	 Disk "sda" Scheduler type "cfq".
	 Disk "dm-0" Scheduler type "none".

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -235,6 +235,29 @@ func validateCgroupMounts() (string, string) {
 	return Supported, out
 }
 
+func validateIoScheduler(containerManager manager.Manager) (string, string) {
+	var desc string
+	mi, err := containerManager.GetMachineInfo()
+	if err != nil {
+		return Unknown, "Machine info not available\n\t"
+	}
+	cfq := false
+	for _, disk := range mi.DiskMap {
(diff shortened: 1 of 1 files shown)


validate/validate.go
```diff
@@ -235,6 +235,29 @@ func validateCgroupMounts() (string, string) {
...
+	for _, disk := range mi.DiskMap {
+		desc += fmt.Sprintf("\t Disk %q Scheduler type %q.\n", disk.Name, disk.Scheduler)
+		if disk.Scheduler == "cfq" {
+			cfq = true
+		}
+	}
+	// Since we get lot of random block devices, report recommended if
+	// at least one of them is on cfq. Report Supported otherwise.
+	if cfq {
+		desc = fmt.Sprintf("Atleast one device supports 'cfq' I/O scheduler. Some disk stats can be reported.\n") + desc
```
vishh: why use `fmt.Sprintf` here and below when they are just raw strings?

rjnagal: copy/paste :) fixed.

vishh:
LGTM

PR #448, opened 2015-01-18
Move network graphs above filesystem info for root.

rjnagal:
This is also consistent with the other container pages.


vmarmol:
LGTM

PR #449, opened 2015-01-18
Add blkio cgroup to list of supported subsystems

arkadijs:
I'm not sure this is intended omission, so looks like blkio should be there?
https://groups.google.com/forum/#!topic/google-containers/OyX_B_Qu-4c


vishh:
LGTM

vmarmol:
Not an intended omission. Thanks for the patch @arkadijs!

Issue #452, opened 2015-01-20
cAdvisor from HEAD crashes while calculating cpu load


vishh:
Here is the stack trace:

I0120 04:55:23.919359 21096 cadvisor.go:150] Starting cAdvisor version: "0.8.0" on port 8080
panic: runtime error: slice bounds out of range

goroutine 66 [running]:
runtime.panic(0x8d0100, 0xf04b2f)
        /usr/local/go/src/pkg/runtime/panic.c:279 +0xf5
bufio.(*Reader).Read(0xc208004f60, 0xc2080f8000, 0xf5, 0xf5, 0xfffffffffffffff0, 0x0, 0x0)
        /usr/local/go/src/pkg/bufio/bufio.go:184 +0x3b3
github.com/google/cadvisor/utils/cpuload.(*Connection).ReadMessage(0xc208117e30, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/vishnuk/go/src/github.com/google/cadvisor/utils/cpuload/conn.go:90 +0x168
github.com/google/cadvisor/utils/cpuload.getLoadStats(0xc208160014, 0xb, 0xc208117e30, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/vishnuk/go/src/github.com/google/cadvisor/utils/cpuload/netlink.go:228 +0x175
github.com/google/cadvisor/utils/cpuload.(*CpuLoadReader).GetCpuLoad(0xc2081371c0, 0xc20816ac00, 0x5a, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/vishnuk/go/src/github.com/google/cadvisor/utils/cpuload/cpuload.go:67 +0x2b8
github.com/google/cadvisor/manager.(*containerData).updateStats(0xc2081a8ea0, 0x0, 0x0)
        /home/vishnuk/go/src/github.com/google/cadvisor/manager/container.go:243 +0x1f5
github.com/google/cadvisor/manager.(*containerData).housekeepingTick(0xc2081a8ea0)
        /home/vishnuk/go/src/github.com/google/cadvisor/manager/container.go:205 +0x2d
github.com/google/cadvisor/manager.(*containerData).housekeeping(0xc2081a8ea0)
        /home/vishnuk/go/src/github.com/google/cadvisor/manager/container.go:167 +0x1d1
created by github.com/google/cadvisor/manager.(*containerData).Start
[...]

PR #453, opened 2015-01-20
Add error checks for netlink read.

rjnagal:


rjnagal:
Couldn't repro #452 but this might help get some error message instead of panic.

vmarmol:
LGTM

PR #463, opened 2015-01-26
Add load from all task groups to root.

rjnagal:
This makes our root load much closer to the one reported by top.


vmarmol:
This is different with the netlink implementation no? Are we sure we want to have those be different? Seems like we won't be able to have a stable return in the API if that's the case.

rjnagal:
That's true. I am trying to figure out a way to make netlink implementation
hierarchical. One way is to run a similar housekeeping loop in there to
discover all containers and return cached stats.

Without this fix, we can't use root load information anywhere up the stack.
How about switching to sched debug as the primary way till we get
hierarchical behavior working with netlink?

vmarmol:
SGTM, especially since most of our users run us in a container. Doing the housekeeping in netlink doesn't sound so bad either.

rjnagal:
Closing this. Sent another PR to make scheddebug the preferred option. Let's do another one to make all load stats hierarchical under scheddebug.

PR #465, opened 2015-01-26
docs: broken relative link fixes

domluna:
A few links were broken... no longer.


vmarmol:
Thanks for the fixes @domluna! LGTM

PR #468, opened 2015-01-27
Compute smoothed cpu load for containers.

rjnagal:


info/container.go
```diff
@@ -197,6 +197,8 @@ type CpuStats struct {
 		// Unit: nanoseconds
 		System uint64 `json:"system"`
 	} `json:"usage"`
+	// Number of runnable threads x 1000.
+	// We multiply by thousand to avoid using floats, but preserving precision.
```
vishh: Curious: Why will the number of runnable threads be a float?

rjnagal: This is a smoothed avg of number of running threads. I'll update the comment to make it clear.

vishh:
LGTM. Can you take a look at the integration test failures?

rjnagal:
vmarmol@ said that there's some flakes he's fixing. possibly related.

vmarmol:
The integration test failure is a flake. I need to fix that.

info/container.go
```diff
@@ -197,6 +197,8 @@ type CpuStats struct {
 		// Unit: nanoseconds
 		System uint64 `json:"system"`
 	} `json:"usage"`
+	// Smoothed average of number of runnable threads x 1000.
+	// We multiply by thousand to avoid using floats, but preserving precision.
 	Load int32 `json:"load"`
```
vmarmol: Only concern I have here is that all of the above stats are instantaneous except this one. We should at least document over what time period or how this is smoothed. Idk if we should export smoothing over different intervals. As long as we do this one over a small enough one we should be okay.

vmarmol: Milliload? :)

rjnagal: Nooo. Not milliload :(

rjnagal: We smooth it over last 10s. If the load changes from 2 to 50 and stays there, this load value should get to 50 within 10s. The instantaneous value is very spiky and not useful for analysis. If a tool wants to do its own calculation or look for bursts, the value is still available in loadstats.NrRunning.

I can change the field to LoadAverage and add the 10s interval in comments.

vmarmol: I think the name change and the comment sounds good. Thanks for pointing out that we also export the instantaneous one. SGTM :)

vmarmol:
LGTM

PR #469, opened 2015-01-27
Expect user + cpu to be within 500ms of total.

vmarmol:
This should be a sane default.


vishh:
Curious: Why is there a deviation between these stats?

vmarmol:
Because we don't gather them atomically and because one is accurate to nanoseconds and the other to tens of milliseconds (kernel configurable).

vishh:
Got it. Thanks for clarifying @vmarmol

PR #470, opened 2015-01-28
Rename load to load average.

rjnagal:
Load is now LoadAverage. Also added comments about smoothing interval.


rjnagal:
Missed out the final comment update in the last PR.

vishh:
LGTM

vmarmol:
LGTM

vishh:
Travis is unhappy.

rjnagal:
Updated. Added a TODO to handle load in integration test. The current assert of load being zero is no longer good :)

vmarmol:
LGTM

PR #473, opened 2015-01-28
Enable check for cpu load in integration tests.

rjnagal:


vmarmol:
LGTM

vmarmol:
Integration test is failing, for some reason the command is failing. Wonder if we're killing it? I'd also expect to not have the Run return though since it is running in the foreground.

rjnagal:
It works with local docker daemon. I guess the ssh command invocation is messing up the script. I'll take a look.

rjnagal:
Withdrawing this for now as we integrate #495

PR #481, opened 2015-02-01
Initial types for stats summary.

rjnagal:


rjnagal:
@vishh I think this captures what we discussed around stats summary.

info/container.go
```diff
@@ -405,3 +405,34 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
 	}
 	return cur - prev
 }
+
+type Percentiles struct {
+	// Average over the collected sample.
+	Mean uint64 `json:"mean"`
```
vishh: How about moving the comment as a 'description' as part of the json name?

info/container.go
```diff
@@ -405,3 +405,34 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
...
+	// Max seen over the collected sample.
+	Max uint64 `json:"max"`
+	// 90th percentile over the collected sample.
+	Ninety uint64 `json:"ninety"`
```
vishh: Will `uint64` be appropriate for all stats? What prevents us from providing more precision?

info/container.go
```diff
@@ -405,3 +405,34 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
...
+}
+
+type Usage struct {
+	// Indicates amount of data available [0-100].
+	// If we have data for half a day, we'll still process DayUsage,
+	// but set PercentComplete to 50.
+	PercentComplete int32 `json:"percent_complete"`
+	// Mean, Max, and 90p cpu rate value in milliCpus/seconds. Converted to milliCpus to avoid floats.
+	CpuUsage Percentiles `json:"cpu_usage"`
```
vishh: How about map[ResourceName]ResourceQuantity instead of explicit fields? Take a look at [type quantity](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/api/resource/quantity.go#L86) that is part of kubernetes. That might help avoiding having to infer scaling factors from comments or json description.

info/container.go
```diff
@@ -405,3 +405,34 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
...
+	// Mean, Max, and 90p memory size in bytes.
+	MemoryUsage Percentiles `json:"memory_usage"`
```
vishh: nit: Why not just `Memory` instead of `MemoryUsage`? Ditto for `CpuUsage`.

vishh:
Overall LGTM. Just a few comments.

info/container.go
vmarmol: The quantity stuff sounds pretty good, but I'm unsure about the generic map. We've always preferred strong over weak typing and having the fields spelled out here is stronger and easier to read (than specifying what string values are valid in the map)

vishh: Having strong type requires us to also specify whether a field is valid.
With a map we don't have to do that.

vmarmol: I buy that, we don't have an "is present" in JSON. +1

(well theres the omitifempty but that doesn't work anywhere close so what we'd like it to work like)

vishh: We need something like 'NaN'. If we use float we can do that.

vmarmol: I don't think we need to go to floats. Those always lead to trouble as
different machines/languages treat them in different ways.

info/container.go
rjnagal: Done.

info/container.go
rjnagal: I am following the convention in this whole file. Don't want to update just this one.

info/container.go
rjnagal: I think this is the biggest precision we need. We can always use different units.

info/container.go
rjnagal: I agree to the part of presence checks and it is probably the best approach when we have some passthrough components that just move the data around. For this one, I have two issues:
- It hurts readability quite a bit. No one can figure out what data is being returned just looking at the info file.
- Every component we have in the final summary pipeline actually looks at the data. We'd add up more lookup charges without simplifying much.

rjnagal:
@vishh added valid field to generic percentile type.

info/container.go
```diff
@@ -405,3 +405,36 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
 	}
 	return cur - prev
 }
+
+type Percentiles struct {
+	// Indicates whether the stats are present or not.
+	valid bool `json:"valid"`
```
vmarmol: Capital Valid? WDYT: Present instead?

rjnagal: ok, Present it is.

info/container.go
```diff
@@ -405,3 +405,36 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
 	}
 	return cur - prev
 }
+
+type Percentiles struct {
+	// Indicates whether the stats are present or not.
```
vmarmol: nit: Add: If true, the values below do not have data.

vmarmol:
LGTM, waiting on @vishh to LGTM

info/container.go
```diff
@@ -405,3 +405,36 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
...
+	Present bool `json:"present"`
```
vishh: Do we foresee a scenario where we will have some stats but not all?

info/container.go
```diff
@@ -405,3 +405,36 @@ func calculateCpuUsage(prev, cur uint64) uint64 {
...
+	Max uint64 `json:"max"`
+	// 90th percentile over the collected sample.
+	Ninety uint64 `json:"ninety"`
+}
+
+type Usage struct {
+	// Indicates amount of data available [0-100].
+	// If we have data for half a day, we'll still process DayUsage,
+	// but set PercentComplete to 50.
+	PercentComplete int32 `json:"percent_complete"`
```
vishh: If one of the resources here is not available or is partially available, what will PercentComplete refer to? % of Cpu or % of Memory or aggregate?

Also how about renaming this to accuracy?

rjnagal: Resources available or not are tracked in Percentiles through Present. PercentComplete only tracks how many Usage samples we gathered irrespective of what type is available.

vishh: Wouldn't Percent then vary for every resource? How can a user then
interpret this value?

rjnagal: PercentComplete just indicates how many samples cadvisor has collected for that time duration. The primary use for present is when the system is configured to say not have memory isolation or cpu load, for example.

Are you concerned about cadvisor failing sometimes to collect a single resource data? We'll handle that internally in cadvisor either by ignoring the sample of extrapolating. It should not bubble up to the clients.

vishh: My concern is the latter. When we say 50% data available, is it 50% for all
stats, which is how I interpret the current API, or 50% of one of the
stats? To make it unambiguous, we can move the PercentComplete to
'Percentiles'.

rjnagal: It's 50% of all stats. That's why it is part of Usage.

vishh: Ok. If we are restricting ourselves to just basic resources which cannot dynamically appear or disappear, this model should work.
LGTM

vishh:
LGTM

PR #489, opened 2015-02-05
Add container creation time to ContainerSpec.

vmarmol:


rjnagal:
TestGetContainerInfo is failing - unexpected container info.

vmarmol:
Fixed, PTAL

rjnagal:
still failing...
--- FAIL: TestGetSubcontainersInfo (0.01s)
	client_test.go:179: received unexpected ContainerInfo
	client_test.go:182: received unexpected ContainerInfo
	client_test.go:185: received unexpected ContainerInfo

container/raw/handler.go
```diff
@@ -180,6 +182,20 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
...
+	lowestTime := now
+	for _, cgroupPath := range self.cgroupPaths {
+		// The modified time of the cgroup directory is when the container was created.
+		fi, err := os.Stat(cgroupPath)
+		if err == nil && fi.ModTime().Before(lowestTime) {
+			lowestTime = fi.ModTime()
+		}
+	}
+	if lowestTime != now {
+		spec.CreationTime = lowestTime
```
vishh: Is `CreationTime` being set elsewhere to `now`?

vmarmol: No, we take it from when the cgroup hierarchies were created.

container/raw/handler.go
```diff
@@ -180,6 +182,20 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
...
 	// The raw driver assumes unified hierarchy containers.
 
+	// Get the lowest creation time from all hierarchies as the container creation time.
+	now := time.Now()
+	lowestTime := now
+	for _, cgroupPath := range self.cgroupPaths {
+		// The modified time of the cgroup directory is when the container was created.
+		fi, err := os.Stat(cgroupPath)
+		if err == nil && fi.ModTime().Before(lowestTime) {
+			lowestTime = fi.ModTime()
```
rjnagal: Would this work for things like systemd containers that just join cpu and names hierarchies? Won't it always pick the machine boot time in that case?

vmarmol: I think it would. Those two are also created around the same time as the container would (I don't think they're created ahead of time).

container/raw/handler.go
vishh: It looks like `spec.CreationTime` is not being set if the condition is not satisfied. Is it being initialized elsewhere?

vmarmol: No, it would be uninitialized (zero) since we don't know when it was created.

vishh:
LGTM

vishh:
Why are we adding this?

vmarmol:
Think I finally got it :) it was passing locally but not in Travis for some reason. Those Eq() checks we're doing are hurting more than helping IMO.

vmarmol:
While working on exposing metrics to GCM when we expose cumulative metrics (like CPU) it it useful to understand when the counter started. This gives us that :)

container/raw/handler.go
rjnagal: So if you open a ssh to ubuntu/RH box, it will report boot time as container creation even though we just created cpu and systemd cgroups with right timestamp.

I don't think we care that much for these sessions, so its okay to ignore.

rjnagal:
LGTM

PR #490, opened 2015-02-05
Basic event handler to monitor events as they occur and grab events that have already happened

kateknister:
Adding preliminary events handler

### DIFF
--- a/events/handler.go
+++ b/events/handler.go
@@ -1,0 +1,253 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package events
+
+import (
+	"errors"
+	"sort"
+	"strings"
+	"time"
+)
+
+// EventManager is implemented by Events. It provides two ways to monitor
+// events and one way to add events
+type EventManager interface {
+	// Watch checks if events fed to it by the caller of AddEvent satisfy the
+	// request and if so sends the event back to the caller on outChannel
+	WatchEvents(outChannel chan *Event, request *Request) error
+	// GetEvents() returns a slice of all events detected that have passed
+	// the *Request object parameters to the caller
+	GetEvents(request *Request) (EventSlice, error)
+	// AddEvent allows the caller to add an event to an EventManager
+	// object
+	AddEvent(e *Event) error
+}
+
+// Events  holds a slice of *Event objects with a potential field
+// that caps the number of events held.  It is an implementation of the
+// EventManager interface
+type events struct {
+	// eventlist holds the complete set of events found over an
+	// EventManager events instantiation.
+	eventlist EventSlice
+	// the slice of watch pointers allows the EventManager access to channels
+	// linked to different calls of WatchEvents. When new events are found that
+	// satisfy the request of a given watch object in watchers, the event
+	// is sent over the channel to that caller of WatchEvents
+	watchers []*watch
+}
+
+// initialized by a call to WatchEvents(), a watch struct will then be added
+// to the events slice of *watch objects. When AddEvent() finds an event that
+// satisfies the request parameter of a watch object in events.watchers,
+// it will send that event out over the watch object's channel. The caller that
+// called WatchEvents will receive the event over the channel provided to
+// WatchEvents
+type watch struct {
+	// request specifies all the parameters that events sent through the
+	// channel must satisfy. Specified by the creator of the watch object
+	request *Request
+	// a channel created by the caller through which events satisfying the
+	// request are sent to the caller
+	channel chan *Event
+}
+
+// typedef of a slice of Event pointers
+type EventSlice []*Event
+
+// Event contains information general to events such as the time at which they
+// occurred, their specific type, and the actual event. Event types are
+// differentiated by the EventType field of Event.
+type Event struct {
+	// the absolute container name for which the event occurred
+	ContainerName string
+	// the time at which the event occurred
+	Timestamp time.Time
+	// the type of event. EventType is an enumerated type
+	EventType EventType
+	// the original event object and all of its extraneous data, ex. an
+	// OomInstance
+	EventData EventDataInterface
+}
+
+// Request holds a set of parameters by which Event objects may be screened.
+// The caller may want events that occurred within a specific timeframe
+// or of a certain type, which may be specified in the *Request object
+// they pass to an EventManager function
+type Request struct {
+	// events falling before StartTime do not satisfy the request. StartTime
+	// must be left blank in calls to WatchEvents
+	StartTime time.Time
+	// events falling after EndTime do not satisfy the request. EndTime
+	// must be left blank in calls to WatchEvents
+	EndTime time.Time
+	// EventType is a map that specifies the type(s) of events wanted
+	EventType map[EventType]bool
+	// allows the caller to put a limit on how many
+	// events they receive. If there are more events than MaxEventsReturned
+	// then the most chronologically recent events in the time period
+	// specified are returned
+	MaxEventsReturned int
+	// the absolute container name for which the event occurred
+	ContainerName string
+	// if IncludeSubcontainers is false, only events occurring in the specific
+	// container, and not the subcontainers, will be returned
+	IncludeSubcontainers bool
+}
+
+// EventType is an enumerated type which lists the categories under which
+// events may fall. The Event field EventType is populated by this enum.
+type EventType int
+
+const (
+	TypeOom EventType = iota
+	TypeContainerCreation
+	TypeContainerDeletion
+)
+
+// a general interface which populates the Event field EventData. The actual
+// object, such as an OomInstance, is set as an Event's EventData
+type EventDataInterface interface {
+}
+
+// returns a pointer to an initialized Events object
+func NewEventManager() *events {
+	return &events{
+		eventlist: make(EventSlice, 0),
+		watchers:  []*watch{},
+	}
+}
+
+// returns a pointer to an initialized Request object
+func NewRequest() *Request {
+	return &Request{
+		EventType:            map[EventType]bool{},
+		IncludeSubcontainers: false,
+	}
+}
+
+// returns a pointer to an initialized watch object
+func newWatch(request *Request, outChannel chan *Event) *watch {
+	return &watch{
+		request: request,
+		channel: outChannel,
+	}
+}
+
+// function necessary to implement the sort interface on the Events struct
+func (e EventSlice) Len() int {
+	return len(e)
+}
+
+// function necessary to implement the sort interface on the Events struct
+func (e EventSlice) Swap(i, j int) {
+	e[i], e[j] = e[j], e[i]
+}
+
+// function necessary to implement the sort interface on the Events struct
+func (e EventSlice) Less(i, j int) bool {
+	return e[i].Timestamp.Before(e[j].Timestamp)
+}
+
+// sorts and returns up to the last MaxEventsReturned chronological elements
+func getMaxEventsReturned(request *Request, eSlice EventSlice) EventSlice {
+	sort.Sort(eSlice)
+	n := request.MaxEventsReturned
+	if n >= eSlice.Len() || n <= 0 {
+		return eSlice
+	}
+	return eSlice[eSlice.Len()-n:]
+}
+
+// If the request wants all subcontainers, this returns if the request's
+// container path is a prefix of the event container path.  Otherwise,
+// it checks that the container paths of the event and request are
+// equivalent
+func checkIfIsSubcontainer(request *Request, event *Event) bool {
+	if request.IncludeSubcontainers == true {
+		return strings.HasPrefix(event.ContainerName+"/", request.ContainerName+"/")
+	}
+	return event.ContainerName == request.ContainerName
+}
+
+// determines if an event occurs within the time set in the request object and is the right type
+func checkIfEventSatisfiesRequest(request *Request, event *Event) bool {
+	startTime := request.StartTime
+	endTime := request.EndTime
+	eventTime := event.Timestamp
+	if !startTime.IsZero() {
+		if startTime.After(eventTime) {
+			return false
+		}
+	}
+	if !endTime.IsZero() {
+		if endTime.Before(eventTime) {
+			return false
+		}
+	}
+	if request.EventType[event.EventType] != true {
+		return false
+	}
+	if request.ContainerName != "" {
+		return checkIfIsSubcontainer(request, event)
+	}
+	return true
+}
+
+// method of Events object that screens Event objects found in the eventlist
+// attribute and if they fit the parameters passed by the Request object,
+// adds it to a slice of *Event objects that is returned. If both MaxEventsReturned
+// and StartTime/EndTime are specified in the request object, then only
+// up to the most recent MaxEventsReturned events in that time range are returned.
+func (self *events) GetEvents(request *Request) (EventSlice, error) {
+	returnEventList := EventSlice{}
+	for _, e := range self.eventlist {
+		if checkIfEventSatisfiesRequest(request, e) {
+			returnEventList = append(returnEventList, e)
+		}
+	}
+	returnEventList = getMaxEventsReturned(request, returnEventList)
+	return returnEventList, nil
+}
+
+// method of Events object that maintains an *Event channel passed by the user.
+// When an event is added by AddEvents that satisfies the parameters in the passed
+// Request object it is fed to the channel. The StartTime and EndTime of the watch
+// request should be uninitialized because the purpose is to watch indefinitely
+// for events that will happen in the future
+func (self *events) WatchEvents(outChannel chan *Event, request *Request) error {
+	if !request.StartTime.IsZero() || !request.EndTime.IsZero() {
+		return errors.New(
+			"for a call to watch, request.StartTime and request.EndTime must be uninitialized")
+	}
+	newWatcher := newWatch(request, outChannel)
+	self.watchers = append(self.watchers, newWatcher)
+	return nil
+}
+
+// method of Events object that adds the argument Event object to the
+// eventlist. It also feeds the event to a set of watch channels
+// held by the manager if it satisfies the request keys of the channels
+func (self *events) AddEvent(e *Event) error {
+	self.eventlist = append(self.eventlist, e)
+	for _, watcher := range self.watchers {
+		watchRequest := watcher.request
+		inChannel := watcher.channel
+		if checkIfEventSatisfiesRequest(watchRequest, e) {
+			inChannel <- e
+		}
+	}
+	return nil
+}
--- a/events/handler_test.go
+++ b/events/handler_test.go
@@ -1,0 +1,190 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package events
+
+import (
+	"testing"
+	"time"
+)
+
+func createOldTime(t *testing.T) time.Time {
+	const longForm = "Jan 2, 2006 at 3:04pm (MST)"
+	linetime, err := time.Parse(longForm, "Feb 3, 2013 at 7:54pm (PST)")
+	if err != nil {
+		t.Fatalf("could not format time.Time object")
+	} else {
+		return linetime
+	}
+	return time.Now()
+}
+
+// used to convert an OomInstance to an Event object
+func makeEvent(inTime time.Time, containerName string) *Event {
+	return &Event{
+		ContainerName: containerName,
+		Timestamp:     inTime,
+		EventType:     TypeOom,
+	}
+}
+
+// returns EventManager and Request to use in tests
+func initializeScenario(t *testing.T) (*events, *Request, *Event, *Event) {
+	fakeEvent := makeEvent(createOldTime(t), "/")
+	fakeEvent2 := makeEvent(time.Now(), "/")
+
+	return NewEventManager(), NewRequest(), fakeEvent, fakeEvent2
+}
+
+func checkNumberOfEvents(t *testing.T, numEventsExpected int, numEventsReceived int) {
+	if numEventsReceived != numEventsExpected {
+		t.Fatalf("Expected to return %v events but received %v",
+			numEventsExpected, numEventsReceived)
+	}
+}
+
+func ensureProperEventReturned(t *testing.T, expectedEvent *Event, eventObjectFound *Event) {
+	if eventObjectFound != expectedEvent {
+		t.Errorf("Expected to find test object %v but found a different object: %v",
+			expectedEvent, eventObjectFound)
+	}
+}
+
+func TestCheckIfIsSubcontainer(t *testing.T) {
+	myRequest := NewRequest()
+	myRequest.ContainerName = "/root"
+
+	sameContainerEvent := &Event{
+		ContainerName: "/root",
+	}
+	subContainerEvent := &Event{
+		ContainerName: "/root/subdir",
+	}
+	differentContainerEvent := &Event{
+		ContainerName: "/root-completely-different-container",
+	}
+
+	if !checkIfIsSubcontainer(myRequest, sameContainerEvent) {
+		t.Errorf("should have found %v and %v had the same container name",
+			myRequest, sameContainerEvent)
+	}
+	if checkIfIsSubcontainer(myRequest, subContainerEvent) {
+		t.Errorf("should have found %v and %v had different containers",
+			myRequest, subContainerEvent)
+	}
+
+	myRequest.IncludeSubcontainers = true
+
+	if !checkIfIsSubcontainer(myRequest, sameContainerEvent) {
+		t.Errorf("should have found %v and %v had the same container",
+			myRequest.ContainerName, sameContainerEvent.ContainerName)
+	}
+	if !checkIfIsSubcontainer(myRequest, subContainerEvent) {
+		t.Errorf("should have found %v was a subcontainer of %v",
+			subContainerEvent.ContainerName, myRequest.ContainerName)
+	}
+	if checkIfIsSubcontainer(myRequest, differentContainerEvent) {
+		t.Errorf("should have found %v and %v had different containers",
+			myRequest.ContainerName, differentContainerEvent.ContainerName)
+	}
+}
+
+func TestWatchEventsDetectsNewEvents(t *testing.T) {
+	myEventHolder, myRequest, fakeEvent, fakeEvent2 := initializeScenario(t)
+	myRequest.EventType[TypeOom] = true
+	outChannel := make(chan *Event, 10)
+	myEventHolder.WatchEvents(outChannel, myRequest)
+
+	myEventHolder.AddEvent(fakeEvent)
+	myEventHolder.AddEvent(fakeEvent2)
+
+	startTime := time.Now()
+	go func() {
+		time.Sleep(5 * time.Second)
+		if time.Since(startTime) > (5 * time.Second) {
+			t.Errorf("Took too long to receive all the events")
+			close(outChannel)
+		}
+	}()
+
+	eventsFound := 0
+	go func() {
+		for event := range outChannel {
+			eventsFound += 1
+			if eventsFound == 1 {
+				ensureProperEventReturned(t, fakeEvent, event)
+			} else if eventsFound == 2 {
+				ensureProperEventReturned(t, fakeEvent2, event)
+				close(outChannel)
+				break
+			}
+		}
+	}()
+}
+
+func TestAddEventAddsEventsToEventManager(t *testing.T) {
+	myEventHolder, _, fakeEvent, _ := initializeScenario(t)
+
+	myEventHolder.AddEvent(fakeEvent)
+
+	checkNumberOfEvents(t, 1, myEventHolder.eventlist.Len())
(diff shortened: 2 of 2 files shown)


events/events_handler.go
```diff
@@ -0,0 +1,309 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
vmarmol: nit: just handler.go/handler_test.go since we're already in the events directory

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/google/cadvisor/utils/oomparser"
+)
+
+// EventManager is implemented by Events. It provides three main ways to monitor
+// events
+type EventManager interface {
+	// Watch() provides a channel to which EventManager types are streamed as
+	// they are detected, but before being fed to the channel may be screened
+	// by parameter defined in the passed *Request object
+	Watch(eventChannel chan *Event, request *Request)
```
vmarmol: All these functions should return an error (and GetEvents should return a []Event too)

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	GetEvents(request *Request)
+	// AddEvent allows the caller to add an event to an EventManager
+	// instantiation
+	AddEvent(e *Event)
+}
+
+// Events  holds a slice of *Event objects with a potential field
+// that caps the number of events held.  It is an implementation of the
+// EventManager interface
+type Events struct {
```
vmarmol: nit: this can be lowercase as we can make it private. With a New() function we can return an EventManager (of which this will be a type)

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	maxEventsHeld int
+}
+
+// Event contains information general to events such as the time at which they
+// occurred, their specific type, and the actual event. Event types are
+// differentiated by the EventType field of Event.
+type Event struct {
+	ContainerName string
+	TimeOfEvent   time.Time
+	EventType     EventCategory
```
vmarmol: nit: Type and Data

We can also cann EventCategory EventType

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	EventData     EventDataInterface
+}
+
+// Request holds a set of parameters by which Event objects may be screened.
+// The caller may want events that occurred within a specific timeframe
+// or of a certain type, which may be specified in the *Request object
+// they pass to an EventManager function
+type Request struct {
```
vmarmol: nit: lets add some docs on some of these and their interactions :)

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+
+// EventCategory is an enumerated type which lists the categories under which
+// events may fall. The Event field EventType is populated by this enum.
+type EventCategory int
+
+const (
+	OomInstanceCat EventCategory = iota
+	ContainerCreation
+	ContainerDeletion
+	NoSpecifiedEventCategory
```
vmarmol: We can probably omiy this one

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+)
+
+// a general interface which populates the Event field EventData. The actual
+// object, such as an OomInstance, is set as an Event's EventData
+type EventDataInterface interface {
+}
+
+// returns a pointer to an initialized Events object
+func NewEventHolder() *Events {
```
vmarmol: nit: just NewEventManager()

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+		EventList: []*Event{},
+	}
+}
+
+// returns a pointer to an initialized Request object
+func NewRequest() *Request {
+	return &Request{
+		StartTime: time.Time{},
+		EndTime:   time.Time{},
+		EventType: map[EventCategory]bool{
```
vmarmol: We only need to make the map as long as we assume that missing => false

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+}
+
+// determines if an event occurs within the time set in the request object
+func checkIfEventSatisfiesTimeRequest(request *Request, e *Event) bool {
+	startTime := request.StartTime
+	endTime := request.EndTime
+	eventTime := e.TimeOfEvent
+	if !startTime.IsZero() {
+		if startTime.Sub(eventTime) > 0 {
```
vmarmol: How about: startTime.After(eventTime) and endTime.Before(eventTime)?

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+
+	}
+	return true
+}
+
+// sorts and grabs the last N events, as specified in the request object,
+// or if there are fewer than N events, returns them all
+func (self *Events) getLastNEvents(request *Request) error {
+	n := request.LastNEvents
+	if n >= self.Len() || n <= 0 {
```
vmarmol: We can probably treat it as: return up to N events

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+
+// called by Watch to gather ContainerDeletion events as they occur
+// populate the Events.EventList slice with them
+func (self *Events) monitorDeletedContainers(eventChannel chan *Event, request *Request) error {
+	return nil
+}
+
+// called by Watch to gather OomInstanceCat events as they occur
+// populate the Events.EventList slice with them
+func (self *Events) monitorOoms(eventChannel chan *Event, request *Request) error {
```
vmarmol: We shouldn't be monitoring for OOMs ourselves. Let another component do that and feed them to you with AddEvent() :)

events/events_handler.go
```diff
@@ -0,0 +1,309 @@
...
+
+// called by GetEvents to gather ContainerDeletion events that have already
+// occurred and populate the Events.EventList slice with them
+func (self *Events) gatherDeletedContainers(request *Request) error {
+	return nil
+}
+
+// called by GetEvents to gather oomInstance events that have already
+// occurred and populate the Events.EventList slice with them
+func (self *Events) gatherOoms(request *Request) error {
```
vmarmol: Similar to monitor below?

events/events_handler.go
kateknister: Since these are methods called on structs like Events, which contains an []*Event field, should GetEvents just return an error and update the []*Event field of the Event object it's called on?

vmarmol: Hmmm I was seeing this more like there is one "Events" in the system and everyone uses that instead of there being more than one. I also viewed the internal as private to it. What else did you have in mind?

events/events.md
```diff
@@ -0,0 +1,32 @@
+## Events Handler
+
+
+EventManager is an interface implemented by the events struct.  Instantiations
+of event manager have 3 methods.  GetEvents returns events found by the manager
+object.  Watch looks for events as they occur and returns them through a channel
+to the caller.  AddEvent is how the caller adds an event to the manager.
+
+
+The events struct holds a slice of event objects in a field called eventlist
```
vmarmol: Don't worry about putting implementation details here. As they change we'd need to update here which we will forget to do for sure :) Only documenting the public interface should be fine.

events/handler.go
```diff
@@ -0,0 +1,238 @@
...
+func NewEventManager() *events {
+	return &events{
+		eventlist: []*Event{},
+	}
+}
+
+// returns a pointer to an initialized Request object
+func NewRequest() *Request {
+	return &Request{
+		StartTime:   time.Time{},
```
vmarmol: we can remove the zero inits, that's done by default by the Go runtime

events/handler.go
```diff
@@ -0,0 +1,238 @@
...
+		EndTime:     time.Time{},
+		EventType:   map[EventCategory]bool{},
```
vmarmol: WDYT of having an EventType be a slice instead of a map? Then we won't need a New for it.

vmarmol:
ok to test

events/handler.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	returnEventList = getLastNEvents(request, returnEventList)
+	return returnEventList, nil
+}
+
+// method of Events object that maintains an *Event channel passed by the user.
+// When an event occurs that satisfies the parameters in the passed Request
+// object it is fed to the channel
+func (self *events) WatchEvents(eventChannel chan *Event) error {
```
kateknister: There are a few different ways WatchEvents can be implemented depending on how it’s used

1. func (self *events) WatchEvents(eventChannel chan *Event) error
The current implementation.  WatchEvents is called in a goroutine in the manager when the manager is started.  The EventManager is a field of the manager struct.  When events occur, they are fed to the eventChannel.  WatchEvents adds the events it receives in the channel to the EventManager.  AddEvents and WatchEvents are redundant here.

2. func (self *events) WatchEvents(eventChannel chan *Event, request *Request) error
WatchEvents is called in a goroutine in the manager when the manager is started.  The EventManager is a field of the manager struct.  When events occur, they are fed to the eventChannel.  In this implementation, WatchEvents would only add events to the eventlist of self if they satisfy the request. To make this effective, one of two things would have to occur.
a. The caller has multiple EventManager objects, each of which differs by request object.  The manager struct would have a field map[*Request]*events. It creates a unique channel and spins up a different goroutine for each WatchEvents call with a different Request argument for each
b. The *events object that implements EventManager has a collection of eventlists rather than one. The field would look like map[*Request]EventSlice.

This has the potential downfall of storing every event multiple times in every eventlist corresponding to all the Requests the caller asks for that it satisfies.

3. func (self *events) WatchEvents(inChannel chan *Event, outChannel chan *Event, request *Request) error
The caller creates an input and output channel. WatchEvents is called in a goroutine in the manager when the manager is started.  The EventManager is a field of the manager struct.  When events occur, they are fed to the inChannel.  If they satisfy the request, they are fed back to the outChannel and received by the caller.  Then the events may be added by the caller to the EventManager held by the manager struct.

4. WatchEvents is not implemented in the events handler. A goroutine is called when manager starts that has a channel it reads from.  When events are created after either watchForNewContainers or watchForNewOoms, both in manager, they are fed to the channel. Within the goroutine AddEvents is called to update the manager struct’s EventManager.

5. The oomparser and container libraries implement the 3 functions of the EventManager interface and *events and its implementations are removed from the events handler.

4 seems like the least redundant solution. Because space is not an issue, 2.b. seems like it could be a good choice.  3 is the closest to the previous commit that was on github, with the main change that events no longer depends on oomparser.

vmarmol: I agree that 4 is the most likely way to go, but I also realized that you refer to WatchEvents() as a way to take events in, it should be a way to get events out. GetEvents() lets you query historical data, while WatchEvents() lets you wait for new events that will occur. The only way to create events are through AddEvent().

The manager.Manager will have an `eventManager` as a member of its struct. On startup the manager.Manager will watchForOoms() and new/deleted containers. When either of those are encountered, it calls eventManager.AddEvent(). I tried adding more detail to the EventManager interface you have to see if that make it clearer.

```Go
type EventManager interface {
	// Report the occurrence of an event. This event is not available to GetEvents() 
	// requests. If there are any outstanding WatchEvents() looking for this type of event,
	// this new event will be delivered to them.
	AddtEvent(event Event) error

	// Allows the called to wait for events to occur. The caller with use the outChannel to 
	// receive any new events of the specified types.
	WatchEvents(outChannel chan Event, types map[EventCategory]struct{}) error

	// Allows the caller to fetch historical events. The request argument allows the called to 
	// filter the events returned.
	GetEvents(request Request) []Event
}
```

kateknister:
Updated the handler and tests so that WatchEvents can send events out.

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+		if checkIfEventSatisfiesRequest(request, event) {
+			outChannel <- event
+		}
+	}
+	return nil
+}
+
+// method of Events object that adds the argument Event object to the
+// Events.eventlist
+func (self *events) AddEvent(e *Event, passToWatch chan *Event) error {
```
vmarmol: We can abstract away the passToWatch and remove it from WatchEvents() by storing them locally. When someone registers a watch we add it to a slice of chan *Event. We can then look through those when we add an event. Would make it simpler for the caller so they don't have to remember the channel

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+type Request struct {
+	StartTime   time.Time
+	EndTime     time.Time
+	EventType   map[EventCategory]bool
+	LastNEvents int
+}
+
+// EventCategory is an enumerated type which lists the categories under which
+// events may fall. The Event field EventType is populated by this enum.
+type EventCategory int
```
vmarmol: nit: EventType?

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+// typedef of a slice of Event pointers
+type EventSlice []*Event
+
+// Event contains information general to events such as the time at which they
+// occurred, their specific type, and the actual event. Event types are
+// differentiated by the EventType field of Event.
+type Event struct {
+	ContainerName string
+	TimeOfEvent   time.Time
```
vmarmol: nit: Timestamp

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+// EventManager is implemented by Events. It provides three main ways to monitor
+// events
+type EventManager interface {
+	// Watch() provides a channel to which EventManager types are streamed as
+	// they are detected, but before being fed to the channel may be screened
+	// by parameter defined in the passed *Request object
+	WatchEvents(inChannel chan *Event, outChannel chan *Event, request *Request) error
+	// give make it a watch event with a channel, event, and container set
```
vmarmol: ?

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+// limitations under the License.
+
+package events
+
+import (
+	"sort"
+	"time"
+)
+
+// EventManager is implemented by Events. It provides three main ways to monitor
```
vmarmol: nit: two ways to monitor, one is to add :)

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+// attribute and if they fit the parameters passed by the Request object,
+// adds it to a slice of *Event objects that is returned
+func (self *events) GetEvents(request *Request) (EventSlice, error) {
+	returnEventList := EventSlice{}
+	for _, e := range self.eventlist {
+		if checkIfEventSatisfiesRequest(request, e) {
+			returnEventList = append(returnEventList, e)
+		}
+	}
+	returnEventList = getLastNEvents(request, returnEventList)
```
vmarmol: What do you think of saying you either:
- Specify a time range
- Specify last N events
It's not clear how both interact, did he mean give him the last N events in that range or the last N events that have occurred?

events/handler.go
kateknister: Should I change the name to something else?

vmarmol: I was suggesting EventType since the documentation calls it that :)

events/handler.go
kateknister: I was thinking if the user specified both, it would return up to N events that fell in the time range.  I could change the field name LastNEvents to MaxEventsReturned or GetNEvents. In case there's a ton of events, it could be helpful to the user to limit the number of events returned.

events/handler.go
```diff
@@ -0,0 +1,177 @@
...
+        // object
+        AddEvent(e *Event) error
+}
+
+// Events  holds a slice of *Event objects with a potential field
+// that caps the number of events held.  It is an implementation of the
+// EventManager interface
+type events struct {
+        eventlist     EventSlice
+        maxEventsHeld int
```
vmarmol: This one is unused, we can remove.

events/handler.go
```diff
@@ -0,0 +1,196 @@
...
+        }
+        sort.Sort(returnEventList)
+        returnEventList = getMaxEventsReturned(request, returnEventList)
+        return returnEventList, nil
+}
+
+// method of Events object that maintains an *Event channel passed by the user.
+// When an event occurs that satisfies the parameters in the passed Request
+// object it is fed to the channel
+func (self *events) WatchEvents(outChannel chan *Event, request *Request) error {
```
vmarmol: The StartTime and EndTime parameters may not make sense for a watch. We should document that they must be empty for a watch and check that here.

events/handler.go
```diff
@@ -0,0 +1,196 @@
...
+        if self.watchers[request] == nil {
```
vmarmol: Unfortunately we may not be able to make this a map since we may have more than one user with the same request. Imagine two calls to Watch() that only want OOM events, they should be handled as separate things. Maybe a slice with a request and a channel?

events/handler.go
```diff
@@ -0,0 +1,196 @@
...
+                incomingChannel := make(chan *Event)
+                self.watchers[request] = incomingChannel
+                go func() {
```
vmarmol: Remember that we only want new events for watch, we shouldn't need to send the old ones. That means that we shouldn't need to start any go routines here since we can use the outChannel as the channel that is in self.watchers

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+// and StartTime/EndTime are specified in the request object, then only
+// up to the most recent MaxEventsReturned events in that time range are returned.
+func (self *events) GetEvents(request *Request) (EventSlice, error) {
+	returnEventList := EventSlice{}
+	for _, e := range self.eventlist {
+		if checkIfEventSatisfiesRequest(request, e) {
+			returnEventList = append(returnEventList, e)
+		}
+	}
+	sort.Sort(returnEventList)
```
vmarmol: nit: Do we need to sort? We sort with getMaxEventsReturned

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+// When an event is added by AddEvents that satisfies the parameters in the passed
+// Request object it is fed to the channel. The StartTime and EndTime of the watch
+// request should be uninitialized because the purpose is to watch indefinitely
+// for events that will happen in the future
+func (self *events) WatchEvents(outChannel chan *Event, request *Request) error {
+	if !request.StartTime.IsZero() || !request.EndTime.IsZero() {
+		return errors.New(
+			"for a call to watch, request.StartTime and request.EndTime must be uninitialized")
+	}
+	newWatcher := newWatch(request)
```
vmarmol: this should probably take the outChannel and reuse that.

events/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+	"time"
+
+	"github.com/google/cadvisor/utils/oomparser"
+)
+
+func createOldTime(t *testing.T) time.Time {
+	const longForm = "Jan 2, 2006 at 3:04pm (MST)"
+	linetime, err := time.Parse(longForm, "Feb 3, 2013 at 7:54pm (PST)")
+	if err != nil {
+		t.Error("could not format time.time object")
```
vmarmol: nit: Fatal instead of Error

events/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+		ContainerName: currentOomInstance.ContainerName,
+		Timestamp:     currentOomInstance.TimeOfDeath,
+		EventType:     TypeOom,
+		EventData:     currentOomInstance,
+	}
+}
+
+// returns EventManager and Request to use in tests
+func initializeScenario(t *testing.T) (eventHolder *events, request *Request) {
+	returnEventHolder := NewEventManager()
```
vmarmol: nit: can do:

```Go
return NewEventManager(), NewRequest()
```

events/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+	returnRequest := NewRequest()
+
+	return returnEventHolder, returnRequest
+}
+
+func TestCheckIfIsSubcontainer(t *testing.T) {
+	myRequest := NewRequest()
+	myRequest.ContainerName = "/root"
+
+	sameDirectoryEvent := &Event{
```
vmarmol: use subcontainer instead of Directory?

events/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+		ContainerName: "/root",
+	}
+	subDirectoryEvent := &Event{
+		ContainerName: "/root/subdir",
+	}
+
+	if checkIfIsSubcontainer(myRequest, sameDirectoryEvent) != true {
```
vmarmol: nit: !checkIfIsSubcontainer(...) instead of != true.

Here and below

events/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}()
+}
+
+func TestAddEventAddsEventsToEventManager(t *testing.T) {
+	myEventHolder, _ := initializeScenario(t)
+	fakeOom := createFakeOom(createOldTime(t))
+	fakeEvent := oomInstanceToEvent(fakeOom)
+	myEventHolder.AddEvent(fakeEvent)
+	if myEventHolder.eventlist.Len() != 1 {
+		t.Errorf("Expected to return 1 events but received %v",
```
vmarmol: use t.Fatal. That stops the test. Then below you can assume that we only have one element and thus not have the if statement

kateknister:
Thanks, I think I fixed all the comments for the last commit, if you want to take another look!

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	EventType     EventType
+	EventData     EventDataInterface
+}
+
+// Request holds a set of parameters by which Event objects may be screened.
+// The caller may want events that occurred within a specific timeframe
+// or of a certain type, which may be specified in the *Request object
+// they pass to an EventManager function
+type Request struct {
+	StartTime            time.Time
```
vmarmol: Lets document these fields a bit more.

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// typedef of a slice of Event pointers
+type EventSlice []*Event
+
+// Event contains information general to events such as the time at which they
+// occurred, their specific type, and the actual event. Event types are
+// differentiated by the EventType field of Event.
+type Event struct {
+	ContainerName string
```
vmarmol: Lets document these fields a bit more.

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	return eSlice[eSlice.Len()-n:]
+}
+
+// If the request wants all subcontainers, this returns if the request's
+// container path is a prefix of the event container path.  Otherwise,
+// it checks that the container paths of the event and request are
+// equivalent
+func checkIfIsSubcontainer(request *Request, event *Event) bool {
+	if request.IncludeSubcontainers == true {
+		return strings.HasPrefix(event.ContainerName, request.ContainerName)
```
vmarmol: We probably need to append an "/" to request.ContainerName, else this can slip by:

request.ContainerName = "/container"
event.ContainerName = "/container-that-makes-you-think-is-a-subcontainer"

events/handler.go
```diff
@@ -0,0 +1,230 @@
...
+// request should be uninitialized because the purpose is to watch indefinitely
+// for events that will happen in the future
+func (self *events) WatchEvents(outChannel chan *Event, request *Request) error {
+	if !request.StartTime.IsZero() || !request.EndTime.IsZero() {
+		return errors.New(
+			"for a call to watch, request.StartTime and request.EndTime must be uninitialized")
+	}
+	newWatcher := newWatch(request, outChannel)
+	self.watchers = append(self.watchers, newWatcher)
+	for event := range newWatcher.channel {
```
vmarmol: We can just return, the user will do the actual listening (and I think this will listen and send on the same channel no? Since newWatch() takes outChannel)

events/handler_test.go
```diff
@@ -0,0 +1,247 @@
...
+	myRequest.ContainerName = "/root"
+
+	sameContainerEvent := &Event{
+		ContainerName: "/root",
+	}
+	subContainerEvent := &Event{
+		ContainerName: "/root/subdir",
+	}
+
+	if !checkIfIsSubcontainer(myRequest, sameContainerEvent) {
```
vmarmol: Add a test for the case we mentioned above.

events/handler_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+func TestGetEventsForOneEvent(t *testing.T) {
+	myEventHolder, myRequest := initializeScenario(t)
+	myRequest.MaxEventsReturned = 1
+	myRequest.EventType[TypeOom] = true
+	fakeOom := createFakeOom(createOldTime(t))
+	fakeEvent := oomInstanceToEvent(fakeOom)
+	fakeOom2 := createFakeOom(time.Now())
+	fakeEvent2 := oomInstanceToEvent(fakeOom2)
+	myEventHolder.eventlist = append(myEventHolder.eventlist, fakeEvent)
```
vmarmol: Use the normal Add API here :)

events/handler_test.go
```diff
@@ -0,0 +1,247 @@
...
+			receivedEvents.Len())
+	}
+	eventObjectFound := receivedEvents[0]
+	if eventObjectFound != fakeEvent2 {
+		t.Errorf("Expected to find test object %v but found a different object: %v",
+			fakeEvent, eventObjectFound)
+	}
+}
+
+func TestGetEventsForTimePeriod(t *testing.T) {
```
vmarmol: Add a test that uses a time range with Start and End

events/handler_test.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	eventObjectFound := receivedEvents[0]
+	if eventObjectFound != fakeEvent2 {
+		t.Errorf("Expected to find test object %v but found a different object: %v",
+			fakeEvent, eventObjectFound)
+	}
+}
+
+func TestGetEventsForNoTypeRequested(t *testing.T) {
+	myEventHolder, myRequest := initializeScenario(t)
```
vmarmol: Looks like we do a lot of this in common, any way to make a helper for it?

events/handler_test.go
```diff
@@ -0,0 +1,247 @@
...
+		t.Errorf("should have found %v and %v had the same directory",
+			myRequest.ContainerName, sameContainerEvent.ContainerName)
+	}
+	if !checkIfIsSubcontainer(myRequest, subContainerEvent) {
+		t.Errorf("should have found %v was a subdirectory of %v",
+			subContainerEvent.ContainerName, myRequest.ContainerName)
+	}
+}
+
+func TestWatchEventsAddsEventsToEventManager(t *testing.T) {
```
vmarmol: Can we simplify this test by registering the watcher and then adding events?

kateknister:
I got rid of the redundancies in the tests and in the WatchEvents/AddEvent functions that I forgot to clean up earlier.

events/handler.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// typedef of a slice of Event pointers
+type EventSlice []*Event
+
+// Event contains information general to events such as the time at which they
+// occurred, their specific type, and the actual event. Event types are
+// differentiated by the EventType field of Event.
+type Event struct {
+	// the complete container directory for which the event occurred
```
vmarmol: nit: Absolute container name, we don't use directory to refer to them usually.

events/handler.go
```diff
@@ -0,0 +1,253 @@
...
+	// must be left blank in calls to WatchEvents
+	EndTime time.Time
+	// EventType is a map that specifies the type(s) of events wanted
+	EventType map[EventType]bool
+	// allows the caller to put a limit on how many
+	// events they receive. If there are more events than MaxEventsReturned
+	// then the most chronologically recent events in the time period
+	// specified are returned
+	MaxEventsReturned int
+	// the complete container directory for which the event occurred
```
vmarmol: Remove use of "directory" here and in the one below

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+	return &oomparser.OomInstance{
+		ContainerName: "/",
+		Pid:           5,
+		ProcessName:   "fakeOomInstance",
+		TimeOfDeath:   inTime,
+	}
+}
+
+// used to convert an OomInstance to an Event object
+func oomInstanceToEvent(currentOomInstance *oomparser.OomInstance) *Event {
```
vmarmol: We can simplify this by not using an oomparser.OomInstance. Since events doesn't care what's inside we can take ContainerName and Timestamp as arguments, make the EventType TypeOom and make EventData int(0)

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+	outChannel := make(chan *Event, 10)
+	myEventHolder.WatchEvents(outChannel, myRequest)
+
+	myEventHolder.AddEvent(fakeEvent)
+	myEventHolder.AddEvent(fakeEvent2)
+
+	startTime := time.Now()
+	go func() {
+		i := 0
+		for i == 0 {
```
vmarmol: nit: you can just do:

```Go
for {
  ...
}
```

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+	myEventHolder, myRequest, fakeEvent, fakeEvent2 := initializeScenario(t)
+	myRequest.EventType[TypeOom] = true
+	outChannel := make(chan *Event, 10)
+	myEventHolder.WatchEvents(outChannel, myRequest)
+
+	myEventHolder.AddEvent(fakeEvent)
+	myEventHolder.AddEvent(fakeEvent2)
+
+	startTime := time.Now()
+	go func() {
```
vmarmol: Since we don't check anything, can we just sleep the 5s and then error?

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+			subContainerEvent.ContainerName, myRequest.ContainerName)
+	}
+	if checkIfIsSubcontainer(myRequest, differentContainerEvent) {
+		t.Errorf("should have found %v and %v had different directories",
+			myRequest.ContainerName, differentContainerEvent.ContainerName)
+	}
+}
+
+func TestWatchEventsAddsEventsToEventManager(t *testing.T) {
+	myEventHolder, myRequest, fakeEvent, fakeEvent2 := initializeScenario(t)
```
vmarmol: Much nicer looking test thank you :)

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+		t.Errorf("should have found %v was a subdirectory of %v",
+			subContainerEvent.ContainerName, myRequest.ContainerName)
+	}
+	if checkIfIsSubcontainer(myRequest, differentContainerEvent) {
+		t.Errorf("should have found %v and %v had different directories",
+			myRequest.ContainerName, differentContainerEvent.ContainerName)
+	}
+}
+
+func TestWatchEventsAddsEventsToEventManager(t *testing.T) {
```
vmarmol: Can we rename? The name is slightly misleading.

vmarmol:
Just nits, I think this is functionally ready :)

kateknister:
Would this be okay to merge now?

events/handler_test.go
vmarmol: Rename to: TestWatchEvents since it doesn't add events

events/handler_test.go
```diff
@@ -0,0 +1,207 @@
...
+	myEventHolder.WatchEvents(outChannel, myRequest)
+
+	myEventHolder.AddEvent(fakeEvent)
+	myEventHolder.AddEvent(fakeEvent2)
+
+	startTime := time.Now()
+	go func() {
+		i := 0
+		for i == 0 {
+			if time.Since(startTime) > (5 * time.Second) {
```
vmarmol: just sleep and then fail?

vmarmol:
LGTM, thanks @kateknister! Merging

kateknister:
@dchen1107 Here is the events handler that will be used to manager oom events

PR #493, opened 2015-02-06
cAdvisor 0.9.0

vmarmol:


vmarmol:
Ping @vishh @rjnagal all integration tests pass on our usual setups.

rjnagal:
Can we check resource usage on a systemd setup? I was wondering if reporting cpu load hierarchically makes dynamic housekeeping slowdown less likely.

vmarmol:
From my tests 3-6% on a system with O(30) containers. Mainly 3-4% with a
spike every min to 6%. I think that's within bounds, we may need to do an
optimization round in the near future.

rjnagal:
Do we know what it was with 0.8.0?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.9.0 (2015-04-06)
+- Added CPU load information.
+- Support for more network devices.
+- Support for more partition types.
+- More robust handling of stats failures.
```
rjnagal: I think we also started reporting diskio stats.

vmarmol: Ah good point, added.

vmarmol:
Will test and report back. We should report usage as part of the release candidates.

vmarmol:
Uuuu, ouch. So the current usage is ~1% with minute spikes to 3%. That's definitely a big jump in CPU usage. I will try to verify if this is due to CPU load. If so, we should add a flag to allow it to be disabled for the users that worry about that.

You made a good point about the load maybe making it less likely that housekeeping doesn't get lowered. I will check into that as well.

rjnagal:
If the load is causing the housekeeping to change, we should just disable
it rather than add a new flag.
I think we can fix that easily by changing the dynamic housekeeping check
to look at cpu and memory usage specifically rather than comparing the
whole set.

vmarmol:
Regathered the numbers and removed the UI from the equation:
- 0.8.0: ~0.7%
- 0.9.0 without CPU load: ~1.2%
- 0.9.0 with load: ~1.5% (there are bigger peaks more often than in the two above though)

Dynamic housekeeping did not seem to affect the results. So we're using roughly twice as much CPU. Digging into why without CPU load the usage went up so high.

vmarmol:
Alright, found the culprits. So relative to 0.8.0:
- +0.4% CPU usage for Disk IO stats
- +0.3% CPU usage for CPU load stats

vmarmol:
Removed load, PTAL.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.9.0 (2015-02-06)
+- Support for more network devices.
+- Support for more partition types.
```
vishh: Can we be specific here instead of more? I guess we support btrfs in addition to existing "ext4" support?

rjnagal:
LGTM

vishh:
LGTM excepting for that one comment.

vmarmol:
Done, merge after the other PRs.

rjnagal:
merging...

PR #492, opened 2015-02-06
Detect Docker containers by asking the Docker.

vmarmol:
No longer do deduction from the cgroup path (only to get the Docker
container ID).

Fixes #450.

PR #494, opened 2015-02-07
Use shared_cpu_map instead of shared_cpu_list to get cache hierarchy.

rjnagal:


vmarmol:
LGTM

Should this address #492?

rjnagal:
Yes it should. It looks like shared_cpu_map is supported on older kernels. I didn't see a kernel that supports just shared_cpu_list and not shared_cpu_map. If that happens, se could check for both files.

PR #495, opened 2015-02-09
Disable cpu load tracking temporarily.

rjnagal:
I'd like to understand its resource usage and improve accuracy
before we ship it in an official release.


vmarmol:
LGTM

PR #496, opened 2015-02-09
Refactoring API version handling.

vmarmol:
This should make it easier to add future API versions and makes it a bit
simpler to look through the API.

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -21,6 +21,8 @@ import (
 	"io"
 	"net/http"
 	"path"
+	"regexp"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


api/handler.go
```diff
@@ -58,145 +52,63 @@ func RegisterHandlers(m manager.Manager) error {
...
-	}
-	if apiElement != "api" {
-		return fmt.Errorf("invalid API request format %q", request)
+	if len(requestElements) == 0 {
+		return fmt.Errorf("malformed request %q", request)
 	}
-	if _, ok := supportedApiVersions[version]; !ok {
+
+	// Check supported versions.
+	version := requestElements[1]
```
vishh: How about defininf consts for various parts of the api request to make the code more readable?

vishh:
LGTM expect for the one comment.

api/handler.go
vmarmol: I think I did what you meant :) PTAL

api/handler.go
```diff
@@ -58,145 +52,62 @@ func RegisterHandlers(m manager.Manager) error {
...
-	if apiElement != "api" {
-		return fmt.Errorf("invalid API request format %q", request)
+	// Verify that we have all the elements we expect:
+	// /<version>/<request type>[/<args...>]
+	requestElements := apiRegexp.FindStringSubmatch(request)
+	if len(requestElements) == 0 {
+		return fmt.Errorf("malformed request %q", request)
 	}
-	if _, ok := supportedApiVersions[version]; !ok {
+	version := requestElements[1]
```
vishh: I was referring to something like the following:
`const (
apiVersion = iota + 1
apiRequestType
apiRequestArgs
)`, instead of using numeric indexes.
Sorry if that was not clear in the previous comment.

vmarmol: Done.

vishh:
LGTM. @rjnagal any comments from your side?

vishh:
Travis is failing @vmarmol

vmarmol:
Merging to get into 0.9.0. I can address any other comments in a separate PR :) @kateknister is also working on a new API export endpoint so this will be useful.

PR #497, opened 2015-02-09
Also print a 60s average CPU usage for self-logging.

vmarmol:


vmarmol:
Helps while examining CPU usage since it is so fickle :P

manager/container.go
```diff
@@ -181,20 +181,27 @@ func (c *containerData) housekeeping() {
...
+				usageCpuNs := uint64(0)
+				for i := range stats {
+					if i > 0 {
+						usageCpuNs += (stats[i].Cpu.Usage.Total - stats[i-1].Cpu.Usage.Total)
+					}
+				}
+				usageMemory := stats[numSamples-1].Memory.Usage
 
-				usageInCores := float64(usageCpuNs) / float64(stats[1].Timestamp.Sub(stats[0].Timestamp).Nanoseconds())
+				instantUsageInCores := float64(stats[numSamples-1].Cpu.Usage.Total-stats[numSamples-2].Cpu.Usage.Total) / float64(stats[numSamples-1].Timestamp.Sub(stats[numSamples-2].Timestamp).Nanoseconds())
```
vishh: Why is this useful? This is cpu usage calculated every 60 seconds right?

vmarmol: This prints the instantaneous usage and then we show the average over the last 60s.

It is useful since it gives us an idea of the current usage (and how stable that is) as well as what it means in the "medium" (60s) term overall.

vishh: Instead, how about printing max and maybe 90%ile which might be more
useful. The reason I say is that this instantaneous value only looks at 1
second period every 60 seconds and does not care about cadvisor's cpu usage
during the other 59 seconds in the same period.

vmarmol: This actually prints the usage every second. So it will print the usage over the last second as well as the average usage over the last 60s.

Max is probably okay, 90%ile is probably overkill. I don't think we need these for now, so I'll push back unless you feel strongly :)

vishh: Ahh my bad. I misread the code. The current implementation makes sense if it is printed every second.

vishh:
LGTM

vmarmol:
On the plus side, once we have the avg and 90%ile stats we can just use that :)

Issue #505, opened 2015-02-16
v0.9.0 fails to start on Centos6


difro:
Running v0.9.0 (latest at docker hub) on Centos6 fails with following error.
```
F0216 04:47:21.812981 00001 cadvisor.go:76] Failed to create a Container Manager: open /sys/class/net/bonding_masters/address: not a directory
goroutine 1 [running]:
github.com/golang/glog.stacks(0x4c2080ae700, 0x0, 0x0, 0x0)
        /usr/local/google/home/vmarmol/go/src/github.com/google/cadvisor/Godeps/_workspace/src/github.com/golang/glog/glog.go:726 +0xcd
github.com/golang/glog.(*loggingT).output(0xf014a0, 0x4c200000003, 0x4c2080bae40)
        /usr/local/google/home/vmarmol/go/src/github.com/google/cadvisor/Godeps/_workspace/src/github.com/golang/glog/glog.go:677 +0x24c
github.com/golang/glog.(*loggingT).printf(0xf014a0, 0x4c200000003, 0x9fe4d0, 0x28, 0x4c2080d3f38, 0x1, 0x1)
        /usr/local/google/home/vmarmol/go/src/github.com/google/cadvisor/Godeps/_workspace/src/github.com/golang/glog/glog.go:635 +0x19c
github.com/golang/glog.Fatalf(0x9fe4d0, 0x28, 0x4c2080d3f38, 0x1, 0x1)
        /usr/local/google/home/vmarmol/go/src/github.com/google/cadvisor/Godeps/_workspace/src/github.com/golang/glog/glog.go:1033 +0x64
main.main()
        /usr/local/google/home/vmarmol/go/src/github.com/google/cadvisor/cadvisor.go:76 +0x5fe
```

/sys/class/net/ does contain a file called ```bonding_masters```
```
ls -l /sys/class/net/
...
lrwxrwxrwx 1 root root    0 2015-01-30 10:01 bond0 -> ../../devices/virtual/net/bond0
-rw-r--r-- 1 root root 4096 2015-01-30 10:01 bonding_masters
lrwxrwxrwx 1 root root    0 2015-01-30 10:01 docker0 -> ../../devices/virtual/net/docker0
```

Maybe should skip over non-directory, non-symlink files?

rjnagal:
Thanks for reporting @difro. We should filter these out.

pireslaert:
When is that going to be released  and be part of the docker image?

Cheers,

vmarmol:
[...]

PR #506, opened 2015-02-16
Filter out non-netdev entries in sysfs/class/net

difro:
fixes #505

PR #507, opened 2015-02-16
Ignore regular files in /sys/class/net directory.

rjnagal:
Only consider symlinks and directories as potential network devices.


rjnagal:
Fixes #505

rjnagal:
Didn't notice #506 that's already out.

PR #511, opened 2015-02-17
Don't use screen to run cAdvisor in the background.

vmarmol:
Some distros don't have screen. Here we wait for cAdvisor to come up
before continuing.

### DIFF
--- a/integration/common/gce.go
+++ b/integration/common/gce.go
@@ -1,0 +1,50 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


integration/common/gce.go
```diff
@@ -0,0 +1,50 @@
...
+var gceInternalIpRegexp = regexp.MustCompile(" +ip +\\| +([0-9.:]+) +")
+var gceExternalIpRegexp = regexp.MustCompile("external-ip +\\| +([0-9.:]+) +")
+
+// Gets the IP of the specified GCE instance.
+func GetGceIp(hostname string) (string, error) {
+	if hostname == "localhost" {
+		return "127.0.0.1", nil
+	}
+
+	out, err := exec.Command("gcutil", "getinstance", hostname).CombinedOutput()
```
vishh: One other option is to run `gcloud compute instances list <instance-name> --format=yaml` which makes it easy to extract the IP.

vishh:
LGTM.

PR #512, opened 2015-02-17
Use pkill instead of killall.

vmarmol:
pkill is available on more systems.


vmarmol:
Also testing if the new machine added to the tests works :)

rjnagal:
LGTM

vmarmol:
All tests pass, we now have Jenkins testing against the CoreOS beta too :) I'll try to add more distros throughout the day.

PR #514, opened 2015-02-18
Map request for 'api/v1.2/docker/' to 'api/v1.2/docker'

rjnagal:
Presently, we just fail this request:
$ curl 0:8080/api/v1.2/docker/
failed to get Docker container "" with error: unable to find Docker container ""


vmarmol:
LGTM, merging since the Jenkins errors are me trying something new before a flight and breaking it. Tests pass on all current distros we have in the cluster.

PR #515, opened 2015-02-18
Expose a minimal 2.0 API with the first summary endpoint.

rjnagal:
Right now, we just do raw container name. More types to follow.
We'll re-do the older endpoint in 2.0 before we publish it.

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -28,6 +28,7 @@ const (
 	subcontainersApi = "subcontainers"
 	machineApi       = "machine"
 	dockerApi        = "docker"
+	summaryApi       = "summary"
 )
 
 // Interface for a cAdvisor API version
@@ -47,7 +48,8 @@ func getApiVersions() []ApiVersion {
 	v1_0 := &version1_0{}
 	v1_1 := newVersion1_1(v1_0)
 	v1_2 := newVersion1_2(v1_1)
-	return []ApiVersion{v1_0, v1_1, v1_2}
+	v2_0 := newVersion2_0(v1_2)
+	return []ApiVersion{v1_0, v1_1, v1_2, v2_0}
 }
 
 // API v1.0
@@ -64,8 +66,8 @@ func (self *version1_0) SupportedRequestTypes() []string {
 }
 
 func (self *version1_0) HandleRequest(requestType string, request []string, m manager.Manager, w http.ResponseWriter, r *http.Request) error {
-	switch {
-	case requestType == machineApi:
+	switch requestType {
+	case machineApi:
 		glog.V(2).Infof("Api - Machine")
 
 		// Get the MachineInfo
@@ -78,7 +80,7 @@ func (self *version1_0) HandleRequest(requestType string, request []string, m ma
 		if err != nil {
 			return err
 		}
-	case requestType == containersApi:
+	case containersApi:
 		containerName := getContainerName(request)
 		glog.V(2).Infof("Api - Container(%s)", containerName)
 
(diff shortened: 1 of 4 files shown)


api/versions.go
```diff
@@ -219,3 +221,43 @@ func (self *version1_2) HandleRequest(requestType string, request []string, m ma
...
+func (self *version2_0) Version() string {
+	return "v2.0"
+}
+
+func (self *version2_0) SupportedRequestTypes() []string {
+	return append(self.baseVersion.SupportedRequestTypes(), summaryApi)
+}
+
+func (self *version2_0) HandleRequest(requestType string, request []string, m manager.Manager, w http.ResponseWriter, r *http.Request) error {
+	switch {
```
vishh: why not `switch requestType { case summaryApi: ... }`

vishh:
LGTM.

api/versions.go
vmarmol: This is the style of the code above (maybe we should change all of it though).

api/versions.go
```diff
@@ -219,3 +221,43 @@ func (self *version1_2) HandleRequest(requestType string, request []string, m ma
 		return self.baseVersion.HandleRequest(requestType, request, m, w, r)
 	}
 }
+
+// v2.0 builds on v1.2
+type version2_0 struct {
+	baseVersion *version1_2
+}
+
+func newVersion2_0(v *version1_2) *version2_0 {
```
vmarmol: I think v2.0 shouldn't be based on v1.X, it'll probably be separate completely.

Maybe we keep this for now since it is not complete though. Although having an incomplete WIP API will probably discourage others from using it until it is ready :)

api/versions.go
```diff
@@ -219,3 +221,43 @@ func (self *version1_2) HandleRequest(requestType string, request []string, m ma
...
+	case requestType == summaryApi:
+		containerName := getContainerName(request)
+		glog.V(2).Infof("Api - Summary(%v)", containerName)
+
+		stats, err := m.GetContainerDerivedStats(containerName)
+		if err != nil {
+			return err
+		}
+
+		err = writeResult(stats, w)
```
vmarmol: nit: return writeResult()

manager/manager.go
```diff
@@ -364,6 +367,20 @@ func (self *manager) containerDataSliceToContainerInfoSlice(containers []*contai
 	return output, nil
 }
 
+func (self *manager) GetContainerDerivedStats(containerName string) (info.DerivedStats, error) {
+	var ok bool
+	var cont *containerData
+	func() {
```
vmarmol: unrelated and not for this PR: we do this everywhere, we should get a helper...

vmarmol:
LGTM, just some discussion points and nits

manager/manager.go
vishh: +1. I was thinking the same thing.

api/versions.go
rjnagal: Done here and everywhere else.

api/versions.go
rjnagal: I want to keep it based off 1.2 and replace whatever APIs we need to change. My guess is that some of the APIs will stay the same. If we rip out everything, we can remove the base dependency. 

I am fine going the other way and adding the base versions once we know that some of the endpoints will be replicated.

api/versions.go
rjnagal: done

manager/manager.go
rjnagal: I was too lazy this time around. Will do a separate PR :)

vmarmol:
LGTM

PR #516, opened 2015-02-18
Add InTimeRange() to StatsBuffer.

vmarmol:
This will allow us to accept queries over a narrower window of time.

### DIFF
--- a/storage/memory/stats_buffer.go
+++ b/storage/memory/stats_buffer.go
@@ -15,6 +15,9 @@
 package memory
 
 import (
+	"sort"
+	"time"
+
 	"github.com/google/cadvisor/info"
 )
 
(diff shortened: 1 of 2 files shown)


storage/memory/stats_buffer.go
```diff
@@ -43,7 +46,65 @@ func (self *StatsBuffer) Add(item *info.ContainerStats) {
...
+// Results are from first to last. maxResults of -1 means no limit.
+func (self *StatsBuffer) InTimeRange(start, end time.Time, maxResults int) []*info.ContainerStats {
+
+	// NOTE: Since we store the elments in descending timestamp order "start" will
+	// be a higher index than "end".
+
+	var startIndex int
+	if start.IsZero() {
+		// None specified, start at the beginning.
+		startIndex = self.size - 1
```
rjnagal: Can this go to -1? Do we correct for it later?

vmarmol: Only if the size is 0. It does get handled it seems (although by no design of mine :P). When size == 0, start = -1 and end = 0, which means numResults = 0.

Open to suggestions on making this behavior more clear :)

rjnagal: this might not work if the user asks for an end time. Might be better to return early if we don't have any samples yet.

vmarmol: If the user doesn't report an end time it will set endIndex to 0 too, but good idea on just returning early if we're empty. Makes it much clearer. Added that.

rjnagal:
LGTM

PR #517, opened 2015-02-18
Make explicit the use of InMemoryStorage in Manager.

vmarmol:
The current structure is a remnant of when the in memory storage was one
of the options for backing store. Today, we always have the memory
storage with an optional "backend storage" plugin. This commit makes the
InMemoryStorage use explicit.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -61,7 +61,7 @@ func main() {
 
 	setMaxProcs()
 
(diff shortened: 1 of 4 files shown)


vmarmol:
Not ready yet :) will report back when it is.

vmarmol:
Ready for review!

rjnagal:
LGTM

PR #520, opened 2015-02-19
Events branch

kateknister:
Made changes to events library to make it thread safe.

### DIFF
--- a/events/handler.go
+++ b/events/handler.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -47,6 +48,10 @@ type events struct {
 	// satisfy the request of a given watch object in watchers, the event
 	// is sent over the channel to that caller of WatchEvents
 	watchers []*watch
+	// lock that blocks eventlist from being accessed until a writer releases it
+	eventsLock sync.RWMutex
+	// lock that blocks watchers from being accessed until a writer releases it
+	watcherLock sync.RWMutex
 }
 
 // initialized by a call to WatchEvents(), a watch struct will then be added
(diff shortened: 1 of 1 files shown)


events/handler.go
```diff
@@ -47,6 +48,7 @@ type events struct {
 	// satisfy the request of a given watch object in watchers, the event
 	// is sent over the channel to that caller of WatchEvents
 	watchers []*watch
+	eventManagerLock sync.RWMutex
```
rjnagal: Add a comment that it protects only eventlist and not watchers, if that is the intent.

vmarmol:
ok to test

vmarmol:
Looks good for the events list. We should also make access to the watchers list thread-safe though.

events/handler.go
```diff
@@ -47,6 +48,10 @@ type events struct {
 	// satisfy the request of a given watch object in watchers, the event
 	// is sent over the channel to that caller of WatchEvents
 	watchers []*watch
+	// lock that blocks eventlist from being accessed until a writer releases it
+	eventManagerLock sync.RWMutex
```
vmarmol: nit: rename to eventsLock?

events/handler.go
```diff
@@ -233,15 +240,25 @@ func (self *events) WatchEvents(outChannel chan *Event, request *Request) error
...
 // eventlist. It also feeds the event to a set of watch channels
 // held by the manager if it satisfies the request keys of the channels
 func (self *events) AddEvent(e *Event) error {
-	self.eventlist = append(self.eventlist, e)
+	self.updateEventList(e)
+	self.watcherLock.RLock()
+	defer self.watcherLock.RUnlock()
 	for _, watcher := range self.watchers {
 		watchRequest := watcher.request
 		inChannel := watcher.channel
```
vmarmol: We should probably not send with the lock (it may take a long time depending on the user) and can lead to deadlocks. Lets create a slice of the channels we're gonna send on (with the lock) and then send to those channels (after dropping the lock).

events/handler.go
vmarmol: Rename this :)

vmarmol:
Can we do that rename, squash commits, and check the failure (in building it seems)

vmarmol:
LGTM

PR #522, opened 2015-02-19
Added an oom monitor to manager

kateknister:


vmarmol:
ok to test

manager/manager.go
```diff
@@ -623,3 +625,28 @@ func (self *manager) watchForNewContainers(quit chan error) error {
 	}()
 	return nil
 }
+
+func (self *manager) watchForNewOoms() error {
+	outStream := make(chan *oomparser.OomInstance)
```
vmarmol: Lets make this a buffered channel, we can start with 10? See how that goes :)

manager/manager.go
```diff
@@ -623,3 +625,28 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+	oomLog, err := oomparser.New()
+	if err != nil {
+		return err
+	}
+	go func() {
+		err := oomLog.StreamOoms(outStream)
+		glog.Infof("Streaming ooms in manager")
+		if err != nil {
+			glog.Errorf("%v", err)
```
vmarmol: Lets add some extra info: "Error streaming OOMs: %v"

vmarmol: Actually, can we move this outside of the Go routine? We can just pass the channel to the Go routine.

manager/manager.go
```diff
@@ -623,3 +625,28 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+		}
+		for oomInstance := range outStream {
+			newEvent := &events.Event{
+				ContainerName: oomInstance.ContainerName,
+				Timestamp:     oomInstance.TimeOfDeath,
+				EventType:     events.TypeOom,
+				EventData:     oomInstance,
+			}
+			glog.Infof("Created an oom event: %v", newEvent)
```
vmarmol: lets make this a V(1) log:

```Go
glog.V(1).Infof(...)
```

vmarmol:
Added some small comments. Let me know when those are addressed :)

manager/manager.go
kateknister: Did you mean move the call to StreamOoms outside the Go routine here?  I made a change in oomparser so that that function StreamOoms will return rather than hang by putting its monitoring call in another goroutine.

manager/manager.go
```diff
@@ -623,3 +625,27 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+
+func (self *manager) watchForNewOoms() error {
+	outStream := make(chan *oomparser.OomInstance, 10)
+	oomLog, err := oomparser.New()
+	if err != nil {
+		return err
+	}
+	err = oomLog.StreamOoms(outStream)
+	if err != nil {
+		glog.Errorf("error streaming OOMs: %v", err)
```
vmarmol: here we can return the error

vmarmol:
LGTM

PR #523, opened 2015-02-20
Added a way for oom events to be handled by the manager

kateknister:


vmarmol:
Looks like it needs a rebase :)

vmarmol:
ok to test

kateknister:
Does it still need a rebase?

manager/manager.go
```diff
@@ -645,7 +655,21 @@ func (self *manager) watchForNewOoms() error {
...
+			}
 		}
+		return nil
 	}()
 	return nil
 }
+
+// can be called by the api which will take events returned on the channel
+func (self *manager) WatchForEvents(request *events.Request, passedChannel chan *events.Event) error {
+	err := self.eventHandler.WatchEvents(passedChannel, request)
```
vmarmol: We can just return the result here

manager/manager.go
```diff
@@ -645,7 +655,21 @@ func (self *manager) watchForNewOoms() error {
 				EventData:     oomInstance,
 			}
 			glog.V(1).Infof("Created an oom event: %v", newEvent)
+			err := self.eventHandler.AddEvent(newEvent)
```
vmarmol: Glog and error instead. We probably want this to run forever

vmarmol:
Yep, shows one commit now :)

vmarmol:
LGTM

PR #524, opened 2015-02-23
Add instant stat to derived result.

rjnagal:
Before this change, we do not return anything meaningful till cadvisor has been up
for a minute. Also, having instant usage in the same stat gives a more complete
picture.

### DIFF
--- a/info/container.go
+++ b/info/container.go
@@ -464,9 +464,19 @@ type Usage struct {
 	Memory Percentiles `json:"memory"`
 }
 
+// latest sample collected for a container.
+type InstantUsage struct {
+	// cpu rate in cpu milliseconds/second.
+	Cpu uint64 `json:"cpu"`
+	// Memory usage in bytes.
+	Memory uint64 `json:"memory"`
+}
+
 type DerivedStats struct {
 	// Time of generation of these stats.
 	Timestamp time.Time `json:"timestamp"`
+	// Latest instantaneous sample.
+	LatestUsage InstantUsage `json:"latest_usage"`
 	// Percentiles in last observed minute.
 	MinuteUsage Usage `json:"minute_usage"`
 	// Percentile in last hour.
--- a/summary/percentiles.go
+++ b/summary/percentiles.go
@@ -17,6 +17,7 @@
 package summary
 
 import (
+	"fmt"
 	"math"
 	"sort"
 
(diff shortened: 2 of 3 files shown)


vmarmol:
Integrations tests pass mainly, failing on CoreOS but it seems unrelated since the other PR fails too. Will take a look there.

summary/percentiles.go
```diff
@@ -147,23 +148,30 @@ func getPercentComplete(stats []*secondSample) (percent int32) {
...
 }
 
+// Calculate cpurate from two consecutive total cpu usage samples.
+func getCpuRate(latest, previous secondSample) (uint64, error) {
+	var elapsed int64
+	elapsed = latest.Timestamp.Sub(previous.Timestamp).Nanoseconds()
+	if elapsed < 10*milliSecondsToNanoSeconds {
+		return 0, fmt.Errorf("elapsed time too small: %d ns: time now %s last %s", elapsed, latest.Timestamp.String(), previous.Timestamp.String())
+	}
+	cpuRate := (latest.Cpu - previous.Cpu) * secondsToMilliSeconds / uint64(elapsed)
```
vmarmol: Shouldn't this just be `/ uint64(elapsed)` The difference in CPU is in nanoseconds and we'll divide by the interval (in ns) to get the number of cores used in that interval.

summary/percentiles.go
```diff
@@ -147,23 +148,30 @@ func getPercentComplete(stats []*secondSample) (percent int32) {
...
+	if cpuRate < 0 {
```
vmarmol: cpuRate is unsigned, we shouldn't need this check.

summary/percentiles.go
vmarmol: I realize we had the code before, so I'm probably just missing something :)

info/container.go
```diff
@@ -467,6 +467,8 @@ type Usage struct {
 type DerivedStats struct {
 	// Time of generation of these stats.
 	Timestamp time.Time `json:"timestamp"`
+	// Latest instantaneous sample.
+	LatestUsage Usage `json:"latest_usage"`
```
vmarmol: I'm okay adding this, do you think it is confusing that we expose the average, 90, max even though they are all the same?

Would it be okay to say to users, if you want instantaneous data go to the /stats endpoint? WDYT?

summary/percentiles.go
rjnagal: we are reporting rate in millicores. :)

vmarmol: Aaaaaah that's right :D can we add a comment so I don't bother you a third time? (I think I had the same issue last review lol)

info/container.go
rjnagal: I was debating putting in a new type here with just one value. We can do that - one uint64 each for cpu and memory.

The goal was to extend the derived stats from day, hour, minute down to second. When cadvisor starts, you just receive a bunch of zeroes for a minute. With seconds data, caller can start seeing usage stat earlier.

vmarmol: I can see what you mean. The single value approach would make it clearer that we actually don't have Max,Mean,90p data that low. I'm +1 to doing that if you prefer.

rjnagal: SG, let me do that.

vmarmol:
Tests are passing now :)

info/container.go
rjnagal: Added a new type.

summary/percentiles.go
rjnagal: Moved the check before cpuRate calculations to drop bad samples.

summary/percentiles.go
rjnagal: Added a comment.

vmarmol:
LGTM

PR #526, opened 2015-02-23
Add extra logging for cAdvisor invocation.

vmarmol:


rjnagal:
LGTM :)

vmarmol:
Thanks :) I think I found the issue with this.

PR #530, opened 2015-02-23
Don't fail startup if OOM watching failed.

### LINKED ISSUE #527: Don't fail cAdvisor startup due to OOM parser not being able to start
We should fail gracefully

vmarmol:
Fixes #527.


vmarmol:
Not sure why this wasn't caught by the test there...

rjnagal:
LGTM

PR #532, opened 2015-02-23
Sort subcontainers in ContainerSpec.

### LINKED ISSUE #531: sort subcontainers list in webui
Currenly `containers/SUB` provides a list of Subcontainers in which (as far as I can tell) is arbitrary order.  It would be nice if these were sorted lexogrphically by name (which woud also sort anything named with a time based UUID).

vmarmol:
Fixes #531.


rjnagal:
LGTM

PR #535, opened 2015-02-25
Added a basic api interface for events

kateknister:


kateknister:
Hey Victor, could you look at this when you get the chance?  I'm putting in the streaming write function now, but I tried to break it up a bit by not putting it in this commit.

vmarmol:
ok to test

api/versions.go
```diff
@@ -50,7 +52,9 @@ func getApiVersions() []ApiVersion {
 	v1_1 := newVersion1_1(v1_0)
 	v1_2 := newVersion1_2(v1_1)
 	v2_0 := newVersion2_0(v1_2)
```
vmarmol: nit: make v2_0 depend on v1_3

api/versions.go
```diff
@@ -270,3 +274,63 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+func (self *version1_3) Version() string {
+	return "v1.3"
+}
+
+func (self *version1_3) SupportedRequestTypes() []string {
+	return append(self.baseVersion.SupportedRequestTypes(), eventsApi)
+}
+
+func (self *version1_3) HandleRequest(requestType string, request []string, m manager.Manager, w http.ResponseWriter, r *http.Request) error {
+	switch {
```
vmarmol: nit:

```Go
switch requestType {
case eventsApi:
...
}
```

api/versions.go
```diff
@@ -270,3 +274,63 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+
+		if eventsFromAllTime {
+			allEvents, err := m.GetPastEvents(query)
+			if err != nil {
+				return err
+			}
+			return writeResult(allEvents, w)
+		} else {
+			// every time URL is entered to watch, a channel is created here
+			eventChannel := make(chan *events.Event)
```
vmarmol: Make it buffered?

api/versions.go
```diff
@@ -270,3 +274,63 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+			err = m.WatchForEvents(query, eventChannel)
+
+			go func() {
```
vmarmol: No need to make this a separate Go routine, one is spun up per HTTP request so it is okay to block it.

api/versions.go
```diff
@@ -270,3 +274,63 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+				defer close(eventChannel)
+				currentEventSet := make(events.EventSlice, 0)
+				for ev := range eventChannel {
+					// todo: implement write-as-received writeResult method
+					currentEventSet = append(currentEventSet, ev)
+					err = writeResult(currentEventSet, w)
+					if err != nil {
+						glog.Errorf("Error writing result %v was %v", currentEventSet, err)
```
vmarmol: we can return the error

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
 	return &query, nil
 }
 
+func getEventRequest(request []string) (*events.Request, bool, error) {
+	if len(request) <= 0 || len(request[0]) == 0 {
```
vmarmol: Maybe add some comments on what format you're expecting from the user

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+		return events.NewAllEventsRequest(), true, nil
+	} else if len(request) > 1 {
+		return nil, false, errors.New("Should have received one url rawQuery but got multiple")
```
vmarmol: nit: errors start with lowercase

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+	}
+	rawQuery := strings.Trim(request[0], "/ ")
+
+	query := events.NewRequest()
+	getEventsFromAllTime := false
+
+	urlMap, err := url.ParseQuery(rawQuery)
```
vmarmol: you can do `r.URL.Query()`

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+	if err != nil {
+		glog.Errorf("url parse error of %v is %v", rawQuery, err)
+		return nil, false, err
+	}
+
+	valueMap := make(map[string]urlData)
```
vmarmol: isn't the return a url.Values?

vmarmol: Ah I see, why don't we do the parsing per field? Since once we get the key we know what type we expect.

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+			continue
+		}
+		newBool, err := strconv.ParseBool(val[0])
+		if err == nil {
+			valueMap[key] = urlData{boolVal: newBool}
+		}
+	}
+
+	getEventsFromAllTime = valueMap["allTime"].boolVal
+	query.IncludeSubcontainers = valueMap["IncludeSubcontainers"].boolVal
```
vmarmol: These are usually in lowercase with `_` so:
- `subcontainers=true`
- `oom_events=true`
- `creation_events=true`
- `deletion_events=true`
- `start_time=TIME`
- `end_time=TIME`

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+			valueMap[key] = urlData{intVal: int(newInt)}
+			continue
+		}
+		newBool, err := strconv.ParseBool(val[0])
+		if err == nil {
+			valueMap[key] = urlData{boolVal: newBool}
+		}
+	}
+
+	getEventsFromAllTime = valueMap["allTime"].boolVal
```
vmarmol: Lets not initially include the "all events".

api/handler.go
```diff
@@ -142,6 +152,57 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+	query.IncludeSubcontainers = valueMap["IncludeSubcontainers"].boolVal
+	query.EventType[events.TypeOom] = valueMap["TypeOom"].boolVal
+	query.EventType[events.TypeContainerCreation] = valueMap["TypeContainerCreation"].boolVal
+	query.EventType[events.TypeContainerDeletion] = valueMap["TypeContainerDeletion"].boolVal
+	query.MaxEventsReturned = valueMap["MaxEventsReturned"].intVal
+
+	if mapVal, ok := valueMap["StartTime"]; ok {
+		query.StartTime = time.Now().Add(time.Duration(-1 * mapVal.intVal))
```
vmarmol: The times we should take as RFC3339 which is what we [parse normally in Go](http://golang.org/pkg/time/#Parse) too (it is the same format that Go prints to naturally).

vmarmol:
Looks good overall :) only minor comments.

api/handler.go
kateknister: Is there an advantage to converting the string to a url and calling query on it rather than the current implementation?

api/handler.go
```diff
@@ -142,6 +147,84 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+	getEventsFromAllTime := false
+
+	urlMap, err := url.ParseQuery(rawQuery)
+	if err != nil {
+		glog.Errorf("url parse error of %v is %v", rawQuery, err)
+		return nil, false, err
+	}
+
+	for key, val := range urlMap {
+		if key == "all_time" {
```
vmarmol: What about we check each of these instead of checking in a loop:

```Go
if val, ok := urlMap["all_time"]; ok {
  getEventsFromAllTime = strconv.ParseBool(val[0])
}
```

api/handler.go
vmarmol: The URL is an object built into the HTTPRequest and saves you from having to do some of the parsing yourself

vmarmol: You also shouldn't need to do most of the checking you're doing above. The r is the HTTPRequest object btw

api/handler.go
```diff
@@ -142,6 +147,84 @@ func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, er
...
+			newBool, err := strconv.ParseBool(val[0])
```
vmarmol: can we use "historical" instead of "all_time"?

events/handler.go
```diff
@@ -143,6 +143,18 @@ func NewRequest() *Request {
 	}
 }
 
+// returns a pointer to an initialized Request object
+func NewAllEventsRequest() *Request {
```
vmarmol: Lets put this in the tests file since we only use it there.

api/versions_test.go
```diff
@@ -0,0 +1,91 @@
...
+package api
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/google/cadvisor/events"
+)
+
+func TestGetEventRequestBasicRequest(t *testing.T) {
```
vmarmol: consider using "github.com/stretchr/testify/assert" This is used by the integration tests and it makes a lot of thing much easier :)
e.g.:
```Go
assert.False(t, shouldGetAllEvents)
assert.NotNil(t, err)
```

vmarmol:
Tests look good :)

api/versions_test.go
```diff
@@ -0,0 +1,94 @@
...
+	receivedQuery, getHistoricalEvents, err := getEventRequest(r)
+
+	if !reflect.DeepEqual(expectedQuery, receivedQuery) {
+		t.Errorf("expected %v but received %v", expectedQuery, receivedQuery)
+	}
+	assert.False(t, getHistoricalEvents)
+	assert.Nil(t, err)
+}
+
+func TestGetEventEmptyStringRequest(t *testing.T) {
```
vmarmol: We can probably remove this test, dupe of the above.

kateknister:
I removed the extra test

manager/manager.go
```diff
@@ -612,15 +616,16 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+			case event := <-eventsChannel:
 				switch {
 				case event.EventType == container.SubcontainerAdd:
 					err = self.createContainer(event.Name)
 				case event.EventType == container.SubcontainerDelete:
 					err = self.destroyContainer(event.Name)
-				}
-				if err != nil {
-					glog.Warning("Failed to process watch event: %v", err)
+					if err != nil {
```
vmarmol: Looks like this should be moved outside the switch.

vmarmol:
LGTM

PR #536, opened 2015-02-25
cAdvisor 0.10.0.

vmarmol:
Left out some work that is in progress (summary stats) as well as the 2.0 API (since we don't want that used just yet).


rjnagal:
LGTM

PR #539, opened 2015-02-26
Added container event monitoring

kateknister:


kateknister:
This will allow container events (creation and deletion) to be tracked by the manager's event manager.

vmarmol:
ok to test

manager/manager.go
```diff
@@ -509,6 +521,18 @@ func (m *manager) destroyContainer(containerName string) error {
...
+
+	newEvent := &events.Event{
+		EventType:     events.TypeContainerDeletion,
+		Timestamp:     time.Now(),
+		ContainerName: containerName,
+		EventData:     cont,
+	}
+	err = m.eventHandler.AddEvent(newEvent)
+	if err != nil {
+		glog.Errorf("Failed to add event %v, got error: %v", newEvent, err)
```
vmarmol: remove log line since we are returning the error

manager/manager.go
```diff
@@ -478,6 +478,18 @@ func (m *manager) createContainer(containerName string) error {
 	// Start the container's housekeeping.
 	cont.Start()
 
+	newEvent := &events.Event{
```
vmarmol: We don't want to do this here since we will create an even each time we restart cAdvisor which is not what we want. We should do this only in detectSubcontainers() or watchForNewContainers(). In detectSubcontainers() we should also only do it if the container creation time was before the cAdvisor start time. Lets start with just watchForNewContainers() and move to detectSubcontainers() in a followup PR.

vmarmol:
ok to test

manager/manager.go
```diff
@@ -475,9 +477,31 @@ func (m *manager) createContainer(containerName string) error {
...
+
+	if contSpecs.CreationTime.After(m.startupTime) {
+		contRef, err := cont.handler.ContainerReference()
+		if err != nil {
+			return err
+		}
+
+		newEvent := &events.Event{
+			ContainerName: contRef.Name,
+			EventData:     cont,
```
vmarmol: We probably don't want to return our internal container data structure here :) How about we return ContainerSpec instead since that is a user-facing structure.

manager/manager.go
```diff
@@ -509,6 +533,22 @@ func (m *manager) destroyContainer(containerName string) error {
...
 	glog.Infof("Destroyed container: %q (aliases: %v, namespace: %q)", containerName, cont.info.Aliases, cont.info.Namespace)
+
+	contRef, err := cont.handler.ContainerReference()
+	if err != nil {
+		return err
+	}
+
+	newEvent := &events.Event{
+		ContainerName: contRef.Name,
+		EventData:     cont,
```
vmarmol: We also shouldn't use cont here. What data do you think would be useful here? Spec would be nice, but that's gone by now. Maybe the ContainerReference? That's just the full names of the container. Maybe just blank and leave as is?

kateknister: The information from cont.handler.GetSpec() pertains more to the container when it was alive, but we could still return it here as event data in case the user wanted to see information about the container before it was stopped.

vmarmol: Ah good point. We should return that then.

manager/manager.go
```diff
@@ -138,10 +138,12 @@ type manager struct {
 	dockerContainersRegexp *regexp.Regexp
 	loadReader             cpuload.CpuLoadReader
 	eventHandler           events.EventManager
+	startupTime            time.Time
 }
 
 // Start the container manager.
 func (self *manager) Start() error {
+	self.startupTime = time.Now()
```
vmarmol: Actually, can we move this to New?

manager/manager.go
```diff
@@ -475,9 +477,31 @@ func (m *manager) createContainer(containerName string) error {
 	}
 	glog.Infof("Added container: %q (aliases: %v, namespace: %q)", containerName, cont.info.Aliases, cont.info.Namespace)
 
+	contSpecs, err := cont.handler.GetSpec()
```
vmarmol: lets move this inside the if since we only use it in that case.

vmarmol: Ah right we need the creation time haha, nvm :)

manager/manager.go
```diff
@@ -509,6 +533,27 @@ func (m *manager) destroyContainer(containerName string) error {
 		})
 	}
 	glog.Infof("Destroyed container: %q (aliases: %v, namespace: %q)", containerName, cont.info.Aliases, cont.info.Namespace)
+
+	contRef, err := cont.handler.ContainerReference()
+	if err != nil {
+		return err
+	}
+
+	contSpecs, err := cont.handler.GetSpec()
```
vmarmol: This will fail since the container doesn't exist anymore. Lets leave it blank for now. We can consider caching the last spec we got and returning that.

vmarmol:
LGTM thanks @kateknister :)

PR #540, opened 2015-02-26
Handle dynamic housekeeping in derived stats.

rjnagal:
Use timestamps instead of sample counts for generating summary.

### DIFF
--- a/summary/summary.go
+++ b/summary/summary.go
@@ -68,11 +68,21 @@ func (s *StatsSummary) AddSample(stat info.ContainerStats) error {
 	s.secondSamples = append(s.secondSamples, &sample)
 	s.updateLatestUsage()
 	// TODO(jnagal): Use 'available' to avoid unnecessary computation.
-	if len(s.secondSamples) == 60 {
-		// Make a minute stat.
+	numSamples := len(s.secondSamples)
+	elapsed := time.Nanosecond
+	if numSamples > 1 {
+		start := s.secondSamples[0].Timestamp
+		end := s.secondSamples[numSamples-1].Timestamp
+		elapsed = end.Sub(start)
+	}
+	if elapsed > 60*time.Second {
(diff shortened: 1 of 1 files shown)


summary/summary.go
```diff
@@ -68,11 +68,21 @@ func (s *StatsSummary) AddSample(stat info.ContainerStats) error {
 	s.secondSamples = append(s.secondSamples, &sample)
 	s.updateLatestUsage()
 	// TODO(jnagal): Use 'available' to avoid unnecessary computation.
-	if len(s.secondSamples) == 60 {
-		// Make a minute stat.
+	numSamples := len(s.secondSamples)
+	elapsed := int64(0)
```
vmarmol: nit: time.Duration and don't use .Nanoseconds() below

summary/summary.go
```diff
@@ -68,11 +68,21 @@ func (s *StatsSummary) AddSample(stat info.ContainerStats) error {
...
+	if numSamples > 1 {
+		start := s.secondSamples[0].Timestamp
+		end := s.secondSamples[numSamples-1].Timestamp
+		elapsed = end.Sub(start).Nanoseconds()
+	}
+	if elapsed > 60*time.Second.Nanoseconds() {
```
vmarmol: Should we worry if this is >> 60s, for example if it is 120s we should in theory keep the stats over the 60s. I'm guessing if we add one stat at a time this is not a problem though. Is that what we expect?

rjnagal: yeah, we add one stat at a time and also max dynamic housekeeping is currently limited to 1minute (with a flag). The only way this can go way beyond 60s is if we  increase the max allowed housekeeping time. Testing on empty containers generate 7 and 2 samples every other minute (due to backoff).

summary/summary.go
rjnagal: much better. Done.

PR #541, opened 2015-02-27
disabling oomparser due to CPU usage

kateknister:


vmarmol:
ok to test

vmarmol:
Have we verified the CPU usage with this?

Issue #502, opened 2015-02-10
Prometheus integration


beorn7:
After our VC discussion, I thought I open an issue here to track the options and efforts for Prometheus integration.

I have looked a bit at the cadvisor code, and it seems to me that it should be relatively easy to (optionally) export a sub-set of metrics directly by the cadvisor binary (via an additional HTTP endpoint) using https://github.com/prometheus/client_golang . The sub-set should be defined by metrics that fit well into the Prometheus data model and are interesting to have available together with metrics from instrumented applications running in the same cluster. Technically, you would probably implement a prometheus.Collector that takes info.ContainerInfo objects and extracts Prometheus metrics from it on scrape time (in the Collect method). I'm happy to assist with questions, or I could provide a prototype implementation. I think, this approach makes more sense than implementing and running a separate binary that scrapes cadvisor's JSON and re-exports it as Prometheus metrics. 

Note that there is https://github.com/docker-infra/container_exporter already, which might serve as a source of inspiration. (Alternatively, one might simply decide to run a container_exporter in addition to cadvisor.)

I'm a bit lost when trying to figure out which parts of cadvisor are Docker specific, which are general container things, and which are Kubernetes specific. So I can't really judge what the best course of action would be here...

vmarmol:
This approach makes sense, I agree that we'd rather export the metrics here than have a separate collector process. We've been looking to export other cAdvisor metrics (latency and usage data) through Prometheus so this lines up nicely. It looks like we can register the custom Collector as well as the other collectors we'd want to use, nice! Looking forward to trying this out :) We'll certainly be poking you with questions.

cAdvisor is primarily container-agnostic, the API exports all containers on the machine. We're able to gather stats on Docker and non-Docker containers simultaneously. We do have some custom code for Docker containers to get some extra information and export those in a separate namespace in the API (since it's a very common use).

On a slight side note, the other portion of this is in [Heapster](https://github.com/GoogleCloudPlatform/heapster). There we'd like to collect Prometheus metrics from clients and possibly re-export them to a Prometheus backend. It seems like there is a library for the [extraction](https://github.com/prometheus/client_golang/tree/master/extraction) (awesome!) and a custom Collector as you mentioned may help with re-exporting.

vishh:
Is there a reason why the Prometheus server doesn't support a 'push' based
model? The reason I ask is that we can very easily have a cadvisor side-car
that can periodically push metrics to prometheus server. This is the model
we are trying to move towards.

[...]

PR #545, opened 2015-02-27
Provide cadvisor stats as prometheus metrics

discordianfish:
This is ~~WIP~~ ready to get merged ~~and provided for discussion:~~

- ~~name is aliases joined by ',' and definitely not what we want: Ideas to solve that? We can~~
 - We're using Aliases[0], if exists, as name
- ~~It doesn't support image name as metric dimension since I'm not sure how to get this~~
 - Out of scope of this PR but should be added eventually
- ~~Unsure about package structure: I didn't use prometheus/ but exporter/ because a) it might not be the only exporter? b) the import path in main would clash with client_golang/prometheus~~
  - we decided for metrics/
- ~~No tests; Will look into that once we agreed on the general form~~
  - Tests will provided once https://github.com/prometheus/client_golang/issues/58 is fixed
- ~~Usage is in seconds, not percent. Not sure what's better~~ Using seconds is fine

Implements #502


exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+	networkRxBytes   *prometheus.CounterVec
+	networkRxPackets *prometheus.CounterVec
+	networkRxErrors  *prometheus.CounterVec
+	networkRxDropped *prometheus.CounterVec
+	networkTxBytes   *prometheus.CounterVec
+	networkTxPackets *prometheus.CounterVec
+	networkTxErrors  *prometheus.CounterVec
+	networkTxDropped *prometheus.CounterVec
+
+	tasksSleeping        *prometheus.GaugeVec
```
brian-brazil: This should all be one metric I think, as it sounds like a partition on the tasks

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+		manager: manager,
+		errors: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "errors_total",
+			Help:      "Errors while exporting container metrics.",
+		}),
+		lastSeen: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "last_seen",
+			Help:      "Last time a container was seen by the exporter",
```
brian-brazil: This sounds like a gauge

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+		cpuUsageSeconds: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "cpu_usage_seconds_total",
+			Help:      "Total seconds of cpu time consumed.",
+		},
+			[]string{"name", "id", "image", "type"},
+		),
+		cpuUsageSecondsPerCPU: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "cpu_usage_per_cpu_seconds_total",
```
brian-brazil: I'd only have one cpu metric, and split it out by cpu.

You shouldn't export multiple versions of a metric with different labels unless you're running into performance problems.

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+			Namespace: namespace,
+			Name:      "memory_working_set_bytes",
+			Help:      "Current working set in bytes.",
+		},
+			[]string{"name", "id", "image"},
+		),
+		memoryFailures: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "memory_failures_total",
+			Help:      "Number of times memory usage hits limits.",
```
brian-brazil: s/hits/hit/

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+			Namespace: namespace,
+			Name:      "network_receive_errors_total",
+			Help:      "Cumulative count of errors encountered",
+		},
+			[]string{"name", "id", "image"},
+		),
+		networkTxBytes: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "network_transmit_bytes_total",
+			Help:      "Cumulative count of bytes transmitd",
```
brian-brazil: transmitted, and below

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// Collect fetches the stats from all containers and delivers them as
+// Prometheus metrics. It implements prometheus.PrometheusCollector.
+func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
+	c.mutex.Lock() // To protect metrics from concurrent collects.
+	defer c.mutex.Unlock()
+
+	for _, v := range c.vectors {
+		v.Reset()
```
brian-brazil: I think you'd be better off creating the metrics afresh each time. It avoids locking too.

exporter/prometheus.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	containers, err := c.manager.SubcontainersInfo("/", &info.ContainerInfoRequest{NumStats: 1})
+	if err != nil {
+		c.errors.Inc()
```
brian-brazil: Usually this is done as a gauge that's 1/0 depending on whether this particular scrape worked.

brian-brazil:
> name is aliases joined by ',' and definitely not what we want: Ideas to solve that? We can:

What I'd do is have an additional metric called 'names', with a label called 'name' and a timeseries for each name with value 1.

> Usage is in seconds, not percent. Not sure what's better

Seconds is what you want as you should leave the math to prometheus

exporter/prometheus.go
discordianfish: Good point, will do that.

exporter/prometheus.go
discordianfish: It's always hard to answer those things.. Any other opinions? But I'm fine with using one metric.

exporter/prometheus.go
discordianfish: You mean instead of creating them via NewXX in NewPrometheusCollector() I would create them in here? But I assume I need to create them before Describe() gets called?

brian-brazil: Yes.

I'm used to the other clients where there is no Describe() :)

discordianfish:
@brian-brazil I don't think we want such 'map metric' in this case. The docker container name in aliases is what you are usually use when refering to a container. Always through through that map sounds painful.

beorn7:
General remark about usage of client_golang: If you have metric vec's that you are resetting and filling on each scrape, it should also work to just keep the Desc's around (and return them in Describe()) and then use NewConstMetric to create a throw-away metric to send to the channel in Collect().

Perhaps it's also a good strategy to leave it conceptionally as it is now, and I'll try the approach described above in a separate PR (once we know that the metrics collected are sane).

brian-brazil:
> I don't think we want such 'map metric' in this case. The docker container name in aliases is what you are usually use when refering to a container. Always through through that map sounds painful.

It sounds like the names/aliases aren't static, so it'd not wise to use them as label values on everything as every change to them will break your graphs. Having all the aliases may be usable for particular use cases such as Docker but not work for the general case.

This incidentally is where the more advanced label features and console templates come into play.

discordianfish:
@vmarmol Do you have some input before I work on the remaining bits and look into getting a CLA signed?

vmarmol:
@discordianfish sorry, it's been on my list to take a look and I've been bad at keeping myself to that. I will take a look today :)

exporter/prometheus.go
vmarmol: "tasks_count" broken down by type sounds good.

Although I agree it is a slippery slope. network_bytes broken down by tx, rx. Memory broken down by usage, working set, etc. Those can stay as is :)

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+	vectors []vector
+}
+
+// New returns a new PrometheusCollector
+func NewPrometheusCollector(manager manager.Manager) *PrometheusCollector {
+	c := &PrometheusCollector{
+		manager: manager,
+		errors: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "errors_total",
```
vmarmol: nit: maybe metric_errors_total? Not to confuse with some other kind of error.

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Help:      "Last time a container was seen by the exporter",
+		},
+			[]string{"name", "id", "image"},
+		),
+		cpuUsageSeconds: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "cpu_usage_seconds_total",
+			Help:      "Total seconds of cpu time consumed.",
+		},
+			[]string{"name", "id", "image", "type"},
```
vmarmol: It's more useful to expose CPU usage total not broken down since it is significantly more accurate than broken down. We can either have type=total,user,system or have a separate total and breakdown metric. I'm guessing the first is better? If we go that route we can rename to cpu_usage_seconds

exporter/prometheus.go
vmarmol: These are also in nanoseconds

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Namespace: namespace,
+			Name:      "fs_limit_bytes",
+			Help:      "Number of bytes that can be consumed by the container on this filesystem.",
+		},
+			[]string{"name", "id", "image", "device"},
+		),
+		fsUsage: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "fs_usage_bytes",
+			Help:      "Number of bytes that is consumed by the container on this filesystem.",
```
vmarmol: nit: s/is/are/

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Namespace: namespace,
+			Name:      "fs_reads_total",
+			Help:      "Number of reads completed",
+		},
+			[]string{"name", "id", "image", "device"},
+		),
+		fsReadsSectors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "fs_sector_reads_total",
+			Help:      "Number of sectors reads completed",
```
vmarmol: nit: s/sectors/sector/

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Namespace: namespace,
+			Name:      "fs_writes_total",
+			Help:      "Number of writes completed",
+		},
+			[]string{"name", "id", "image", "device"},
+		),
+		fsWritesSectors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "fs_sector_writes_total",
+			Help:      "Number of sectors writes completed",
```
vmarmol: nit: s/sectors/sector/

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Namespace: namespace,
+			Name:      "network_receive_packets_dropped_total",
+			Help:      "Cumulative count of bytes received",
+		},
+			[]string{"name", "id", "image"},
+		),
+		networkRxErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "network_receive_errors_total",
+			Help:      "Cumulative count of errors encountered",
```
vmarmol: "...while receiving"

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			Namespace: namespace,
+			Name:      "network_transmit_packets_total",
+			Help:      "Cumulative count of packets transmitted",
+		},
+			[]string{"name", "id", "image"},
+		),
+		networkTxDropped: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "network_transmit_packets_dropped_total",
+			Help:      "Cumulative count of bytes transmitd",
```
vmarmol: "Cumulative count of packets dropped"

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+		},
+			[]string{"name", "id", "image"},
+		),
+		networkTxErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "network_transmit_errors_total",
```
vmarmol: "...while transmitting"

exporter/prometheus.go
vmarmol: This is probably ignorance on my part :) but what does Reset() do here?

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+	if err != nil {
+		c.errors.Inc()
+		glog.Warning("Couldn't get containers: %s", err)
+		return
+	}
+	for _, container := range containers {
+		id := container.Name
+		name := strings.Join(container.Aliases, ",")
+		stats := container.Stats[0]
+		image := "undefined" // FIXME:
```
vmarmol: I think its fine to not use image. We never really expose it in cAdvisor.

exporter/prometheus.go
```diff
@@ -0,0 +1,454 @@
...
+			}
+			for metric, value := range map[*prometheus.GaugeVec]uint64{
+				c.fsIoInProgress: stat.IoInProgress,
+				c.fsLimit:        stat.Limit,
+				c.fsUsage:        stat.Usage,
+			} {
+				metric.WithLabelValues(name, id, image, stat.Device).Set(float64(value))
+			}
+		}
+	}
```
vmarmol: Nice compact collector :)

exporter/prometheus.go
brian-brazil: Why do you expect them to be less accurate when broken down? If is this is the same sort of cpu stat as in /proc/stat I'd expect it to all add up.

vmarmol:
Thanks for the PR @discordianfish! This looks awesome. +1 to keeping in seconds. As per the name/alias and labels, today we only ever have two so we can split by "id", "name", and "alias" unless there is a better solution.

exporter/prometheus.go
brian-brazil: > These are also in nanoseconds

They're converted down in line 385

exporter/prometheus.go
brian-brazil: Removes all samples from that metric.

exporter/prometheus.go
vmarmol: Yeah I removed my other comment and not this one...ignore :)

exporter/prometheus.go
brian-brazil: tasks_state sounds like a good name.

http://prometheus.io/docs/practices/naming/#metric-names has guidelines on when to use labels.

exporter/prometheus.go
vmarmol: These are cgroup stats. We get total and per-cpu usage with nanosecond precision. System and user breakdown has jiffy precision which is typically ~10ms.

brian-brazil:
> today we only ever have two so we can split by "id", "name", and "alias" unless there is a better solution.

Is that the case for every potential user of this code?

Prometheus as of today has the features to work with a single timeseries that has all the potentially useful labels.

vmarmol:
@brian-brazil for Docker containers yes, for non-Docker containers no (we'll only have an ID for those). I'm not sure I understand your second sentence?

exporter/prometheus.go
brian-brazil: Sounds like two different metrics then.

Naming will be fun on those as you want to avoid "by" in metric names as a label will not always be present after aggregation.

brian-brazil:
> I'm not sure I understand your second sentence?

I'd propose to have metrics such as:
```
container_names{id="myid",name="name1"} 1
container_names{id="myid",name="name2"} 1
container_names{id="myid",name="name3"} 1
other_metrics{id="myid"} 7
```

Using the `group_left`/`group_right` modifiers you can join the names on to the metrics when you need the names, without the disadvantages of dealing with names changing over time.

vmarmol:
@brian-brazil we can try that. It does sound slightly painful to write queries for though :) I don't think the names really change over time. Theoretically the alias can, although I haven't seen it in practice.

brian-brazil:
> I don't think the names really change over time. Theoretically the alias can, although I haven't seen it in practice.

Okay, sounds like we can get away with the name as a label on everything and an alias as I suggest.

exporter/prometheus.go
discordianfish: Just realized that CpuStats's PerCpu doesn't include system vs user usage. Only the overall cpu usage is split into user / system.

- Should we keep both metrics?
- Ignore the overall usage and live without system usage?
- Change CpuStats to include system / user usage in PerCpu (obviously the hardest option to implement)

exporter/prometheus.go
discordianfish: It is? Because this kind of error isn't about processing some items for which a error rate would be useful but something affecting overall scrape? We probably should add that to the best practices in the prometheus docs.

exporter/prometheus.go
discordianfish: Keep in mind it's still prefixed by container, so would be 'container_errors_total'. Still, probably not the best name but maybe something like 'container_exporter_errors_total' is better than metric_errors? Other opinions? @brian-brazil usually has a good feeling for those things :)

exporter/prometheus.go
discordianfish: Ah, then ignore my comment above and let's just keep both metrics for now? Seem the best we can do. @brian-brazil Naming, do you have some suggestion?

exporter/prometheus.go
discordianfish: Hrmm I think having the image as a dimension is actually very useful. Usually people have their version numbers / git-shas in the image tag, so you often want to group by that to compare one version of your containers to another version.  Do you think it's hard to get the image name with the current design of cadvisor? If so, we can probably leave the image for now, but I would like to see that in the long term.

exporter/prometheus.go
brian-brazil: http://prometheus.io/docs/practices/instrumentation/#collectors :)

discordianfish: yay :)

exporter/prometheus.go
discordianfish: I've changed it to (container_)export_errors (removed the _total) because it's not a Gauge as suggestd by @brian-brazil

exporter/prometheus.go
discordianfish: I've added it back now. Let me know if you want me to change the name.

discordianfish:
>>  I don't think the names really change over time. Theoretically the alias can, although I haven't seen it in practice.
> Okay, sounds like we can get away with the name as a label on everything and an alias as I suggest.

@brian-brazil What's exactly wrong with changing label values? We still have the id in here which exists as long as the container exists, aka this gets stale after the container is gone. But I assume that's okay, isn't it?:

discordianfish:
Okay, I've changed the PR according to the discussion. I also changed it as @beorn7 suggested and did some moving around.
@vmarmol Are you okay with the package structure as it is right now?

brian-brazil:
> But I assume that's okay, isn't it?:

It's not okay. You want your timeseries to have the same labels over their lifetime, you'll run into all sorts of problems if this is not the case.

exporter/prometheus.go
```diff
@@ -0,0 +1,350 @@
...
+	descs []*prometheus.Desc
+}
+
+// NewPrometheusCollector returns a new PrometheusCollector
+func NewPrometheusCollector(manager manager.Manager) *PrometheusCollector {
+	c := &PrometheusCollector{
+		manager: manager,
+		errors: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "export_errors",
```
brian-brazil: We call this `scrape_error` elsewhere

exporter/prometheus.go
```diff
@@ -0,0 +1,350 @@
...
+		ch <- d
+	}
+}
+
+// Collect fetches the stats from all containers and delivers them as
+// Prometheus metrics. It implements prometheus.PrometheusCollector.
+func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
+	containers, err := c.manager.SubcontainersInfo("/", &info.ContainerInfoRequest{NumStats: 1})
+	if err != nil {
+		c.errors.Inc()
```
brian-brazil: .Set(1) would be clearer

discordianfish:
@brian-brazil Hrm.. maybe this is a misunderstanding. The timeseries, for example 'memory usage of container X' will have the same labels for the lifetime of 'container X'. But if you, for example, delete and recreate the container with the same name etc the id and therfor the label changes. I assume this is okay because it's in fact a new container.

exporter/prometheus.go
discordianfish: But it's not really an error while scraping? But I'm fine with changing it.

brian-brazil:
> The timeseries, for example 'memory usage of container X' will have the same labels for the lifetime of 'container X'. But if you, for example, delete and recreate the container with the same name etc the id and therfor the label changes. I assume this is okay because it's in fact a new container.

That's fine, and what you want.

The issue arises if the labels change during the lifetime of container X, such as if the name or aliases change.

exporter/prometheus.go
```diff
@@ -0,0 +1,350 @@
...
+			Name:      "export_errors",
+			Help:      "Errors while exporting container metrics.",
+		}),
+		lastSeen: prometheus.NewDesc(
+			"container_last_seen",
+			"Last time a container was seen by the exporter",
+			[]string{"name", "id"},
+			nil),
+		cpuUsageSeconds: prometheus.NewDesc(
+			"container_cpu_usage_seconds_total",
```
brian-brazil: You're missing the type label here, and the names of metrics need to be unique.

I don't suppose we can just ignore the user vs. system metric?

exporter/prometheus.go
discordianfish: I've changed it now to scrape_error with same description as in node-exporter/textfile

exporter/prometheus.go
discordianfish: Good catch, I've fixed it. I think it's important to differentiate user vs system so would rather keep it. On the long run we could try to get the user vs system dimension into the per_cpu metric but not sure that's even exposed by the kernel. For now, I think it's fine to keep both.

exporter/prometheus.go
```diff
@@ -0,0 +1,350 @@
...
+		networkTxErrors: prometheus.NewDesc(
+			"container_network_transmit_errors_total",
+			"Cumulative count of errors encountered while transmitting",
+			[]string{"name", "id"},
+			nil),
+
+		tasks: prometheus.NewDesc(
+			"container_tasks_state",
+			"Number of tasks in given state",
+			[]string{"type", "name", "id"},
```
juliusv: Does `type` indicate the state? Should it be named `state` then?

exporter/prometheus.go
brian-brazil: Okay, I never came across a good way to name things like this.

container_cpu_usageclass_seconds_total maybe?

exporter/prometheus.go
```diff
@@ -0,0 +1,350 @@
...
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/manager"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const namespace = "container"
```
beorn7: You are using this in only one place now. So you could get rid of it. (Or include it in the Desc generation...)

beorn7:
NewConstMetric usage looks good to me.

exporter/prometheus.go
discordianfish: Oh sure, thanks!

exporter/prometheus.go
```diff
@@ -0,0 +1,348 @@
...
+func NewPrometheusCollector(manager manager.Manager) *PrometheusCollector {
+	c := &PrometheusCollector{
+		manager: manager,
+		errors: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "container",
+			Name:      "scrape_error",
+			Help:      "1 if there was an error while getting container metrics, 0 otherwise",
+		}),
+		lastSeen: prometheus.NewDesc(
+			"container_last_seen",
```
vmarmol: Don't we not need "container" in the name of all of these due to the namespace?

exporter/prometheus.go
vmarmol: cpu_usage_type_seconds there are 3 labels: total, user, system no?

vmarmol:
GitHub lost the other comment threads so I'll reply to some here (let me know if I missed anything).

About the image. It is awkward to get them with the API today. We can put this in as is today and see how to get the image data in.

For the package structure, can we rename "exporter" to "metrics". We will probably use it for more metrics.

exporter/prometheus.go
brian-brazil: > cpu_usage_type_seconds there are 3 labels: total, user, system no?

The label would be 'type' the value would be 'user' or 'system'

You shouldn't mix total in with breakdowns, as that breaks aggregation.

exporter/prometheus.go
beorn7: The namespace is only one of the options in `...Opts`, which are used to construct fully fledged metric objects for Prometheus. Those are meant to be used for "native" Prometheus metrics. What we are doing here is copying existing metrics into Prometheus metrics. For that, we create some of the Prometheus metrics on the fly during scraping, which does not follow the 'high level' path with the Opts object, but assembles the necessary parts one level further down (in this case a metric descriptor is created, which doesn't feature setting a namespace via Opts - so the "container" part has to be included in the name).

metrics/prometheus.go
```diff
@@ -0,0 +1,348 @@
...
+			[]string{"name", "id", "device"},
+			nil),
+		fsIoTime: prometheus.NewDesc(
+			"container_fs_io_time_seconds_total",
+			"Number of seconds spent doing I/Os",
+			[]string{"name", "id", "device"},
+			nil),
+		fsWeightedIoTime: prometheus.NewDesc(
+			"container_fs_io_time_weighted_seconds_total",
+			"Number of seconds spent doing I/Os",
```
juliusv: Should this and some of the counters above also start with "Cumulative count of...", like the counters below? That makes it clear to the user that they are indeed cumulative and not over some arbitrary period of time.

juliusv:
:+1: It's potentially possible to simplify the metric instantiation/mapping code in the future, but looks good for now.

vmarmol:
This looks good @discordianfish! Can we rebase, and I think we can merge :)

vmarmol:
LGTM, will merge on green. A million thanks @discordianfish :)

metrics/prometheus.go
```diff
@@ -0,0 +1,348 @@
...
+			"Last time a container was seen by the exporter",
+			[]string{"name", "id"},
+			nil),
+		cpuUsageSeconds: prometheus.NewDesc(
+			"container_cpu_usage_seconds_total",
+			"Cumulative cpu time consumed in seconds.",
+			[]string{"name", "id", "type"},
+			nil),
+		cpuUsageSecondsPerCPU: prometheus.NewDesc(
+			"container_cpu_usage_seconds_total",
```
brian-brazil: This needs to be named differently to the above metric, otherwise prometheus won't be able to parse the metrics

vmarmol: "container_per_cpu_usage_seconds_total"?

brian-brazil: That'll be confusing when aggregated up, i'd suggest container_cpu_usagetype_seconds_total for the other

discordianfish: Good catch, seems like I missed that when rebasing. I'm not very happy with container_cpu_(usagetype)_seconds_total but it's probably better the best option.

vmarmol:
ok to test

vmarmol:
Merged! Thanks @discordianfish :D

PR #549, opened 2015-02-27
Revised oomparser to not use all the cpu

kateknister:


utils/oomparser/oomparser.go
```diff
@@ -38,6 +38,9 @@ var firstLineRegexp *regexp.Regexp = regexp.MustCompile(
 // struct to hold file from which we obtain OomInstances
 type OomParser struct {
 	systemFile string
+
+	// Inotify event watcher.
+	watcher *inotify.Watcher
```
rjnagal: Do we know if watching is better? For many systems, /var/log/messages gets updated more than once per second.

kateknister:
I added an inotify watcher to the oomparser struct to not constantly query the file, which makes it use much less cpu.

utils/oomparser/oomparser.go
kateknister: Do you mean there might be a backlog of IN_MODIFY notices to the watcher if there are many changes to the file all at once?

rjnagal: /var/log/messages gets debug logs from many sources. The watcher might get invoked more than once per second.

Were we doing a periodic poll earlier? Can we just check once every 10s and keep an open fd?
(Sorry, I haven't looked at all details of oom-parser. But I feel like watching might not help some of the machines).

kateknister:
Oomparser on with 100ms delay for every loop call: (average: 0.114 cores), 10.01 GB of memory
No oomparser: (average: 0.119 cores), 10.04 GB of memory
The delay seems effective at making the oomparser's cpu usage negligible

utils/oomparser/oomparser.go
```diff
@@ -162,8 +156,18 @@ func (self *OomParser) StreamOoms(outStream chan *OomInstance) error {
...
 	}
+	ioreader := bufio.NewReader(file)
+
+	// Process the events received from the kernel.
 	go func() {
-		self.analyzeLines(file, outStream)
+		// this call allows the historical kernel log to be scanned for ooms
+		for {
+			err := self.analyzeLines(ioreader, outStream)
+			if err != nil && err.Error() != "EOF" {
```
vmarmol: can we move this logic to analyzeLines instead?

utils/oomparser/oomparser.go
```diff
@@ -104,15 +103,10 @@ func checkIfStartOfOomMessages(line string) (bool, error) {
 // lines are checked against a regexp to check for the pid, process name, etc.
 // At the end of an oom message group, AnalyzeLines adds the new oomInstance to
 // oomLog
-func (self *OomParser) analyzeLines(outPipe io.ReadCloser, outStream chan *OomInstance) {
-	ioreader := bufio.NewReader(outPipe)
+func (self *OomParser) analyzeLines(ioreader *bufio.Reader, outStream chan *OomInstance) {
 	line, err := ioreader.ReadString('\n')
 	for err == nil {
```
vmarmol: This code is getting a bit hard to read, let's try to simplify it. We will never break out of this loop so lets make it an infinite loop with a read at the beginning. We should also only wait while the error is EOF:

```Go
var line string
var err error
for true {
    // Wait for more content to show up in the file.
    for line, err = ioreader.ReadString('\n'); err != nil && err == io.EOF; {
        time.Sleep(100 * time.Millisecond)
    }

    // Do work...
}
```

vmarmol:
LGTM

PR #550, opened 2015-02-28
Always use alias as display name if its available.

rjnagal:
This is useful for identifying k8 containers on /docker handler.
Since k8 handles are pretty long, I arbitrarily decided to truncate the display.
Let me know if you don't like it.


vmarmol:
LGTM

PR #552, opened 2015-03-02
Report status of hierarchical memory accounting in validate output.

rjnagal:
We still report 'Supported and Recommended' even if hierarchical accounting is disabled.
Most current distros don't enable it by default.

### DIFF
--- a/validate/validate.go
+++ b/validate/validate.go
@@ -133,6 +133,31 @@ func areCgroupsPresent(available map[string]int, desired []string) (bool, string
 	return true, ""
 }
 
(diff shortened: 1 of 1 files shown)


vmarmol:
LGTM, ick thanks for checking this. Wonder if we have it on in Container VMs.

vmarmol:
And the answer is no, sigh....will work to get that in.

rjnagal:
Even if its not enabled on root, we'd have to enable it on custom cgroup layout (when that gets it) to get pod usage.

PR #555, opened 2015-03-03
Register container handler factories in manager.New().

vmarmol:
This lets us simplify the startup of the manager.


rjnagal:
LGTM

PR #559, opened 2015-03-04
Move derived stats to v2. Add v2 container spec.

rjnagal:


rjnagal:
@vmarmol Let me know what you think about adding conversion layer vs changing the internal structs.

info/v2/container.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"time"
+)
+
+type CpuSpec struct {
+	// Requested cpu shares. Default is 1024.
+	Limit uint64 `json:"limit"`
+	// Requested cpu hard limit. Default is unlimited (0)
+	MaxLimit uint64 `json:"max_limit"`
```
vmarmol: Can we add units?

info/v2/container.go
```diff
@@ -0,0 +1,97 @@
...
+	// Cpu affinity mask.
+	Mask string `json:"mask,omitempty"`
```
vmarmol: Maybe add a TODO to make this a set of sorts? WDYT?

vmarmol:
If we go with a different internal struct we have future flexibility but conversions to v1 and v2. If we go with v2 we only need to convert to v1 which should go away over time. I tend to favor the second since I'm not sure how much more different we could make an internal struct. WDYT?

info/v2/container.go
rjnagal: I feel like string is more compact and manageable.  Its easier for people looking for visual input and straightforward for advanced users to convert.

info/v2/container.go
rjnagal: done. Looks like we don't fill it in yet.

rjnagal:
Ok. I think I am going to build v2 with v1 conversion and slowly wean off internal usage from v1 to v2.

info/v2/container.go
vmarmol: SGTM, can we add a TODO to write a small library to do this? :) We already have it, we just need to export it here.

vmarmol:
Plan SGTM

vmarmol:
LGTM

PR #560, opened 2015-03-04
Add system uuid and machine id information

simon3z:


vmarmol:
ok to test

vmarmol:
LGTM thanks for the change @simon3z! When you have time, could you please sign the CLA? Once that's done we can merge.

simon3z:
@vmarmol I am covered by the Red Hat CLA as far as I know, I commit to kubernetes already (still it seems that the bot has a problem recognizing me)

vmarmol:
Thanks for letting me know! I didn't see you in the RedHat org.

rjnagal:
@simon3z Can I change the semantic for machine-id a bit:

- Make the default value for machine-id as empty.
- if user sets a machine-id flag, use that.
- Use /etc/machine-id if flag is not set.
- if /etc/machine-id doesn't exist, use /var/lib/dbus/machine-id
- return empty if none of these files exist.

This will add a machine id to other distros like ubuntu without requiring flag to be changed. SG?

vmarmol:
WDYT of making the flag accept a list of values where to look for the machine-id and make the default values /etc/machine-id and /var/lib/dbus/machine-id. Thus we check both and take the first, if none exists we leave empty.

rjnagal:
@vmarmol that would work too. Let me do that.

PR #561, opened 2015-03-04
Make machine-id sources flag a comma-separated list.

rjnagal:
Use "/etc/machine-id,/var/lib/dbus/machine-id" as default.

### DIFF
--- a/manager/machine.go
+++ b/manager/machine.go
@@ -41,7 +41,7 @@ var nodeRegExp = regexp.MustCompile("physical id\\t*: +([0-9]+)")
 var CpuClockSpeedMHz = regexp.MustCompile("cpu MHz\\t*: +([0-9]+.[0-9]+)")
 var memoryCapacityRegexp = regexp.MustCompile("MemTotal: *([0-9]+) kB")
 
-var machineIdFilePath = flag.String("machine_id_file", "/etc/machine-id", "File containing the machine-id")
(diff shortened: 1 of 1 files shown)


manager/machine.go
```diff
@@ -210,12 +210,17 @@ func getTopology(sysFs sysfs.SysFs, cpuinfo string) ([]info.Node, int, error) {
...
 		return ""
 	}
-	return strings.TrimSpace(string(id))
+	for _, file := range strings.Split(*machineIdFilePath, ",") {
+		id, err := ioutil.ReadFile(file)
+		if err == nil {
+			return strings.TrimSpace(string(id))
+		}
+	}
+	glog.Infof("Couldn't collect machine-id from any of the files in %q", *machineIdFilePath)
```
vishh: nit: Why not report an error instead of info?

vishh:
LGTM. The build is failing on CoreOS.

vmarmol:
LGTM

manager/machine.go
rjnagal: I am not sure if we have machine-id on all distro. It might be expected to fail on some hosts.

rjnagal:
e2e test seems to be flaky. Running it again fixed it.

vmarmol:
Grrr need to tackle the flakes. LGTM

PR #564, opened 2015-03-05
Add stubs for non-Linux builds of statvfs and netlink.

vmarmol:
This will allow the code to build when cross-compiled. The existing
functionality will be missing and we will degrade gracefully.


rjnagal:
discussed offline. closing.

PR #565, opened 2015-03-05
Update documentation on running integration tests.

rjnagal:
runner.go doesn't work on localhost as it tries to ssh to the local machine :)


vmarmol:
LGTM

PR #567, opened 2015-03-09
Lowering log level for frequent events.

vmarmol:
Lowering all frequent normal logs to v=3. Kubelet runs by default on
debug of v=2 and we don't want to log these events in that case.


rjnagal:
LGTM.

The tests have been flaky lately. I don't think the failure is related. We should track the test flakiness in another bug.

PR #570, opened 2015-03-10
Don't fail requests if System UUID is not available.

vmarmol:
We won't have access to it if we're not running as root.


vishh:
LGTM

PR #571, opened 2015-03-10
Use mux to register HTTP handlers.

vmarmol:
This will allow us to register handlers on the non-default HTTP handler.


vishh:
LGTM.

vmarmol:
Kicked Jenkins, seems weird that it could not find the godep

PR #572, opened 2015-03-10
Don't fail if the raw cgroup factory fails to register.

vmarmol:
This will allow us to degrade gracefully in the environments where no
cgroups exist, but we'd still like cAdvisor to be running.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -127,7 +127,7 @@ func New(memoryStorage *memory.InMemoryStorage, sysfs sysfs.SysFs) (Manager, err
 	// Register the raw driver.
 	err = raw.Register(newManager)
 	if err != nil {
-		return nil, fmt.Errorf("registration of the raw container factory failed: %v", err)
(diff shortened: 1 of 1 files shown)


vmarmol:
This is really for use by the Kubelet. Under certain integration tests we run without cgroups. We don't do too much useful work, but it seems easier to just handle these cases by staying up.

An alternative is to use build labels for those cases, but that starts to get more complicated I think...

vishh:
Can we bubble this error up through the '/validate' handler?

vishh:
What other kernel interfaces will we end up stubbing for cadvisor to work in such environments?

vmarmol:
This is already surfaced in /validate. I am taking these one at a time as I find them. This is the last one I've found, but I thought that before too :-\

vishh:
Ack. LGTM

vmarmol:
Thanks for the super fast reviews @vishh :) this has really unblocked me.

+1 to your suggestion of dropping Travis. Esp, at times like this :P

PR #573, opened 2015-03-11
Add /storage endpoint to 2.0 API.

rjnagal:
/storage returns {device, mountpoint, capacity, usage} for all filesystems.
In addition, it also detect and applies label for each filesystem - currently two - "root", "docker-images".

/storage/<label> returns info about the filesystem with specific label. eg. /storage/root returns info for root filesystem.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -28,6 +28,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/container/libcontainer"
+	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/utils"
 )
@@ -62,6 +63,10 @@ func UseSystemd() bool {
 	return useSystemd
 }
 
+func RootDir() string {
+	return *dockerRootDir
+}
+
 type dockerFactory struct {
 	machineInfoFactory info.MachineInfoFactory
 
@@ -72,6 +77,9 @@ type dockerFactory struct {
 
 	// Information about the mounted cgroup subsystems.
 	cgroupSubsystems libcontainer.CgroupSubsystems
+
+	// Information about mounted filesystems.
+	fsInfo fs.FsInfo
 }
 
 func (self *dockerFactory) String() string {
@@ -87,6 +95,7 @@ func (self *dockerFactory) NewContainerHandler(name string) (handler container.C
 		client,
 		name,
 		self.machineInfoFactory,
+		self.fsInfo,
 		*dockerRootDir,
 		self.usesAufsDriver,
 		&self.cgroupSubsystems,
@@ -151,7 +160,7 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
 }
 
 // Register root container before running this function!
-func Register(factory info.MachineInfoFactory) error {
+func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 	client, err := docker.NewClient(*ArgDockerEndpoint)
 	if err != nil {
 		return fmt.Errorf("unable to communicate with docker daemon: %v", err)
@@ -213,6 +222,7 @@ func Register(factory info.MachineInfoFactory) error {
 		client:             client,
 		usesAufsDriver:     usesAufsDriver,
 		cgroupSubsystems:   cgroupSubsystems,
+		fsInfo:             fsInfo,
 	}
 	container.RegisterContainerHandlerFactory(f)
 	return nil
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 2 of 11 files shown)


rjnagal:
Example /storage output ( I kept minimal information here. Let me know if you think something else should be added):

ubuntu w/ 1 disk:
[
{"device":"/dev/disk/by-uuid/0e4daed1-3084-4d1f-be9594c1d5ceaf97",
"mountpoint":"/",
"capacity":10534313984,
"usage":8768860160,
"labels":["root","docker-images"]},
{"device":"/dev/mapper/docker-8:1-1407781a6ff5d93152e8781d5cd39095c6711d0b4c9aa0d5a22cf341d56f40e2098e6c",
"mountpoint":"/var/lib/docker/devicemapper/mnt/1a6ff5d93152e8781d5cd39095c6711d0b4c9aa0d5a22cf341d56f40e2098e6c",
"capacity":10434699264,
"usage":26365952,
"labels":[]}]

ubuntu w/ 2 disks:
[
{"device":"/dev/disk/by-uuid/0e501679-6d73-47fa-9cf9-558872950040",
"mountpoint":"/",
"capacity":10534313984,
"usage":6669844480,
"labels":["root"]},
{"device":"/dev/sdb1",
"mountpoint":"/var/lib/docker",
"capacity":10534313984,
"usage":3308961792,
"labels":["docker-images"]}]

core-os:
[{
"device":"/dev/sda9","mountpoint":"/","capacity":7176433664,"usage":2078629888,"labels":["root","docker-images"]},{"device":"/dev/sda3","mountpoint":"/usr","capacity":1031946240,"usage":322973696,"labels":[]},
{"device":"/dev/sda6","mountpoint":"/usr/share/oem","capacity":113229824,"usage":39419904,"labels":[]}]

vmarmol:
WDYT of having the endpoint do: `/storage?label=<label>` instead of `/storage/<label>`. I think that in the future we may want to do `/storage/<device>` or `/storage/<path>` and it'll be easier to support that way.

Reviewing the rest :)

container/docker/factory.go
```diff
@@ -62,6 +63,10 @@ func UseSystemd() bool {
 	return useSystemd
 }
 
+func RootDir() string {
+	return *dockerRootDir
```
vmarmol: not blocking this PR, but we need to detect this root dir better if we're going to start relying on it. Today we assume it is always /var/lib/docker, but users do change it.

fs/fs.go
```diff
@@ -41,6 +42,11 @@ import (
 
 var partitionRegex = regexp.MustCompile("^(:?(:?s|xv)d[a-z]+\\d*|dm-\\d+)$")
 
+const (
+	systemRoot   = "root"
```
vmarmol: TODO to make these public and part of the API so people can use them

can we also append "Label" to them?

fs/fs.go
```diff
@@ -49,14 +55,22 @@ type partition struct {
 
 type RealFsInfo struct {
 	partitions map[string]partition
+	labels     map[string]string
```
vmarmol: nit: can document the field? It's a map from label to mount point of the filesystem that has that label?

fs/fs.go
```diff
@@ -68,7 +82,79 @@ func NewFsInfo() (FsInfo, error) {
...
+			if _, ok := self.labels[systemRoot]; !ok {
+				self.labels[systemRoot] = src
+			}
+		}
+		self.updateDockerImagesPath(src, p.mountpoint, dockerPaths)
+		// TODO(rjnagal): Add label for docker devicemapper pool.
+	}
+}
+
+func getDockerImagePaths(context Context) []string {
```
vmarmol: nit: can we document this method and the one below?

info/v2/container.go
```diff
@@ -97,3 +97,20 @@ type DerivedStats struct {
...
 }
+
+type FsInfo struct {
+	// The block device name associated with the filesystem.
+	Device string `json:"device"`
+
+	// Path where the filesystem is mounted.
+	Mountpoint string `json:"mountpoint"`
+
+	// Filesystem usage.
```
vmarmol: nit: units

vmarmol:
Just small comments. This is gonna be nice to use :)

container/docker/factory.go
rjnagal: I did some refactoring to get it from docker factory. The docker base dir is chosen by -g flag, but none of the docker APIs return the path they are using. We can either parse docker command line, or infer it from docker client by parsing graph driver info. Easiest way would be to enhance docker to put the base dir somewhere.

I tried doing it in docker factory, but it was too ugly. Will revisit soon.

fs/fs.go
rjnagal: Updated and made public.

fs/fs.go
rjnagal: Added. It's from label to source device. We can use the output of this to find the partition info in partitions map.

fs/fs.go
rjnagal: Added.

info/v2/container.go
rjnagal: Added :)

rjnagal:
Changed the API to be /storage?label=<label>. I took the lazy way first as labels were the only static thing users would have. But querying by mountpoint sounds plausible for fs that we don't auto-label.

container/docker/factory.go
vmarmol: We can also look at the Docker config file. Customization is usually done there.

vmarmol:
LGTM, thanks @rjnagal!

PR #576, opened 2015-03-11
Provides a way to remove unused channels from an eventHandler's watchers list

kateknister:


kateknister:
@rjnagal , @vmarmol  could you review this change if you have a chance? Thanks!

api/handler.go
```diff
@@ -151,8 +151,9 @@ func streamResults(results chan *events.Event, w http.ResponseWriter, r *http.Re
 	for {
 		select {
 		case <-cn.CloseNotify():
-			return nil
-		case ev := <-results:
+			eventChannel.Stop()
+			return m.CloseDeadChannels()
```
vmarmol: We need to keep on waiting for data here otherwise we'll deadlock (if no one is reading while the writer wants to write)

events/handler.go
```diff
@@ -259,7 +283,7 @@ func (self *events) findValidWatchers(e *Event) []*watch {
 	defer self.watcherLock.RUnlock()
 	for _, watcher := range self.watchers {
 		watchRequest := watcher.request
-		if checkIfEventSatisfiesRequest(watchRequest, e) {
+		if checkIfEventSatisfiesRequest(watchRequest, e) && !watcher.eventChannel.stopSignalReceived {
```
vmarmol: We can remove channels here as we find that they are stopped.

events/handler.go
```diff
@@ -273,7 +297,18 @@ func (self *events) AddEvent(e *Event) error {
 	self.updateEventList(e)
 	watchesToSend := self.findValidWatchers(e)
 	for _, watchObject := range watchesToSend {
-		watchObject.channel <- e
+		watchObject.eventChannel.GetChannel() <- e
+	}
+	return nil
+}
+
+func (self *events) RemoveDeadChannels() error {
```
vmarmol: I'd rather us not have to have this is we can. To remove the channels in a timely manner we may need to have the event manager spin up a thread that listens to all stops and does the removal.

So we'd start a cleanup thread that listens on a cleanup channel, when someone calls Stop() it sends a request to the cleanup thread to remove that channel. The thread wakes up and removed it.

api/handler.go
```diff
@@ -151,8 +151,9 @@ func streamResults(results chan *events.Event, w http.ResponseWriter, r *http.Re
 	for {
 		select {
 		case <-cn.CloseNotify():
-			return nil
-		case ev := <-results:
+			glog.V(3).Infof("Received CloseNotify event")
+			m.StopWatch(eventChannel.GetWatchID())
```
vmarmol: Shouldn't we just call Stop() on the eventChannel?

events/handler.go
```diff
@@ -47,11 +51,17 @@ type events struct {
...
 	// lock that blocks eventlist from being accessed until a writer releases it
 	eventsLock sync.RWMutex
 	// lock that blocks watchers from being accessed until a writer releases it
 	watcherLock sync.RWMutex
+	// receives notices when a watch event ends and needs to be removed from
+	// the watchers list
+	deadWatchAlert chan int
+	// Scans through watchers list and removes those whose EventChannel
+	// fields have received a stop signal
+	last_id int
```
vmarmol: Go uses camelcase: "lastId"

events/handler.go
```diff
@@ -66,7 +76,9 @@ type watch struct {
 	request *Request
 	// a channel created by the caller through which events satisfying the
 	// request are sent to the caller
-	channel chan *Event
+	eventChannel *EventChannel
+	//
+	id int
```
vmarmol: missing description :)

events/handler.go
```diff
@@ -144,13 +169,25 @@ func NewRequest() *Request {
...
+func newWatch(request *Request, eventChannel *EventChannel) *watch {
 	return &watch{
-		request: request,
-		channel: outChannel,
+		request:      request,
+		eventChannel: eventChannel,
 	}
 }
 
+func (self *events) StopWatch(watch_id int) {
```
vmarmol: We can replace this with a Stop() on the EventChannel() that way we don't have to go all the way to the manager to call StopWatch().

events/handler.go
```diff
@@ -273,7 +313,19 @@ func (self *events) AddEvent(e *Event) error {
 	self.updateEventList(e)
 	watchesToSend := self.findValidWatchers(e)
 	for _, watchObject := range watchesToSend {
-		watchObject.channel <- e
+		watchObject.eventChannel.GetChannel() <- e
 	}
 	return nil
 }
+
+func (self *events) CleanupWatcher() {
```
vmarmol: We don't call this anywhere yet btw.

events/handler.go
```diff
@@ -273,7 +313,19 @@ func (self *events) AddEvent(e *Event) error {
...
+	for i := range self.deadWatchAlert {
+		_, ok := self.watchers[i]
+		if !ok {
+			glog.Infof("Could not find watcher instance %v", i)
+		}
+		self.watcherLock.RLock()
```
vmarmol: you want a write lock, so .Lock(), you probably want to lock over all this too, so we use a pattern like this:

```Go
func() {
  self.watcherLock.Lock()
  defer self.watcherLock.Unlock()

  ...work here
}()
```

events/handler.go
```diff
@@ -234,16 +271,19 @@ func (self *events) GetEvents(request *Request) (EventSlice, error) {
...
-		return errors.New(
+		return nil, errors.New(
 			"for a call to watch, request.StartTime and request.EndTime must be uninitialized")
 	}
-	newWatcher := newWatch(request, outChannel)
 	self.watcherLock.Lock()
 	defer self.watcherLock.Unlock()
-	self.watchers = append(self.watchers, newWatcher)
-	return nil
+	new_id := self.last_id + 1
```
vmarmol: ID generation looks good.

events/handler.go
kateknister: I called it in a goroutine in manager's Start function

events/handler.go
kateknister: since CleanupWatcher is called in a goroutine and the channel deadWatchAlert does not terminate, I thought by making a defer call to Unlock() would mean it would never release the watcherLock

api/handler.go
kateknister: In that case should I also get rid of the EventManager's cleanup thread?
It might be difficult to alert the EventManager's cleanup thread that this particular watcher is no longer needed if we only tell the EventChannel to stop. But I could change it back to the way it was originally written, so that I tell the EventChannel to stop and then, rather than putting the CleanupWatcher in a goroutine and having it always run, just call it on a case by case basis here to search for and remove any stopped EventChannel objects.

events/handler.go
```diff
@@ -271,9 +301,23 @@ func (self *events) findValidWatchers(e *Event) []*watch {
...
 	for _, watchObject := range watchesToSend {
-		watchObject.channel <- e
+		watchObject.eventChannel.GetChannel() <- e
 	}
 	return nil
 }
+
+// Removes a watch instance from the EventManager's watchers map
+func (self *events) StopWatch(watchId int) {
+	_, ok := self.watchers[watchId]
```
vmarmol: This should also be inside the lock below

events/handler.go
```diff
@@ -271,9 +301,23 @@ func (self *events) findValidWatchers(e *Event) []*watch {
 // held by the manager if it satisfies the request keys of the channels
 func (self *events) AddEvent(e *Event) error {
 	self.updateEventList(e)
+	self.watcherLock.Lock()
```
vmarmol: I think this can be a read lock.

vmarmol:
LGTM, thanks @kateknister!

PR #578, opened 2015-03-12
WIP: Add stats API for 2.0.

rjnagal:
First cut on 2.0 stats API. Main change in returned stats is that presence checks (eg. HasCPU) are embedded in each stat.
Added support to handle querying container with name.

Current syntax:
/stats?name=/&count=1,
/stats?name=docker/928c058ce260ac2a55972b18cb991fa0475fbbb7bc15bd295e62b76964d05fe6 [default count: 64]

Other handlers to include: dockerid, dockeralias. We can make subcontainers inclusion as an option too.

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -17,6 +17,7 @@ package api
 import (
 	"fmt"
 	"net/http"
+	"strconv"
 
 	"github.com/golang/glog"
 	"github.com/google/cadvisor/events"
@@ -31,6 +32,7 @@ const (
 	machineApi       = "machine"
 	dockerApi        = "docker"
 	summaryApi       = "summary"
+	statsApi         = "stats"
 	specApi          = "spec"
 	eventsApi        = "events"
 	storageApi       = "storage"
@@ -314,6 +316,22 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 		}
 
 		return writeResult(stats, w)
+	case statsApi:
+		name := getContainerName(request)
+		sr, err := getStatsRequest(name, r)
+		if err != nil {
+			return err
+		}
+		glog.V(2).Infof("Api - Stats: Looking for stats for container %q, options %+v", name, sr)
+		query := info.ContainerInfoRequest{
+			NumStats: sr.Count,
+		}
+		cont, err := m.GetContainerInfo(name, &query)
+		if err != nil {
+			return fmt.Errorf("failed to get container %q: %v", name, err)
+		}
+		contStats := convertStats(cont)
+		return writeResult(contStats, w)
 	case specApi:
 		containerName := getContainerName(request)
 		glog.V(2).Infof("Api - Spec(%v)", containerName)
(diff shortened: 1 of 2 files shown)


rjnagal:
Looking for suggestions on /stats format.

Also I took the lazy option of including v1 into v2 to avoid writing deep conversion methods. Let me know if this is unacceptable :)

vmarmol:
WDYT of this option:

`/stats/container_name/maybe_hierarchical?namespace=docker&count=1`

Mainly that namespace (or pick another name) is the type of name we have. It lines up nicely with how we structure it today. We can do "dockerid", "dockername", etc or we can just do the raw "docker" or empty (raw) as we do today. Plus side is no need to know whether it is a name of an ID and it gives us flexibility with pods. Roughly:

`/stats/<container name>?namespace=<type of container name>`

Using v1 SGTM, agreed that we should move off of it eventually :) thankfully it should be transparent to users so no worry there.

WDYT of having a `recursive=true` option instead of the /subcontainers of today? Can probably be an option for all our resources. For another PR of course :)

Will review more thoroughly tomorrow.

vishh:
+1 on the recursive option.
Why not embed the namespace in the API itself? `/stats/<container_type>/<container_name>?recursive=true`? Most of our users are docker users too.

vmarmol:
Only issue with embedding the type is that by default it is usually "" which means "raw". We'd force users to have to say "raw". We can try to do more smart matching (if the first one after the / is not a known factory then it is raw), but that gets brittle and harder to use IMO.

rjnagal:
I was going to add recursive=true [default:false] when converting subcontainers.

I can go with Victor's naming suggestion. Although I think we need a different term than namespace as it might diverge from the 'namespace' we use internally in the code. eg dockeralias. I don't like 'type' that much either.

vmarmol:
Hmmm: filter, container_type, kind?

vishh:
@vmarmol: Just curious.. How many of our users really care about raw cgroups? Aren't most of them are docker users? 
Also why not drop `container` and just name it `type`?

vmarmol:
I don't think we have good data either way :) I don't doubt that most are Docker users though.

@rjnagal didn't like type :P

rjnagal:
OK, I'll go with type. There's still about a week to change the name before
we start using 2.0 :)

rjnagal:
Updated the API and introduced 'type'.

Get root container stats as:

/stats/
/stats/?type=name
/stats/?type=name&count=1

Next additions:
/stats/81fcc6b7b0b1c2367cb791?type=dockerid&count=1 (matches partial id)
/stats/docker?type=name&count=1&recursive=true
/stats/grumpy-netwon?type=dockeralias

As always, open to better keywords and option names suggestions :)

api/versions.go
```diff
@@ -365,3 +383,59 @@ func convertSpec(specV1 info.ContainerSpec) v2.ContainerSpec {
...
+		if stat.HasDiskIo {
+			stat.DiskIo = val.DiskIo
+		}
+		// TODO(rjnagal): Handle load stats.
+		stats = append(stats, stat)
+	}
+	return stats
+}
+
+func getStatsRequest(id string, r *http.Request) (v2.StatsRequest, error) {
```
vmarmol: nit: s/id/name/

vmarmol:
LGTM except that nit :)

Why do we want to do dockerid vs dockeralias btw? Why not just have a docker which can handle both? It'd be easier for the user (and for us with how we handle it today).

api/versions.go
rjnagal: Used id here as it won't always be the name once we extend options to include docker.

rjnagal:
we can combine id and alias into one. My only argument was better error returns, but that's not a big deal.

vmarmol:
SGTM

PR #579, opened 2015-03-12
Start() exits gracefully when there are no factories.

vmarmol:
This allows us to handle the case where there are no factories as
happens when testing in a non-container environment. We will still serve
the machine information that is available.


rjnagal:
LGTM

PR #582, opened 2015-03-12
Don't fail fetching machine information when not available

vmarmol:


vmarmol:
After this, the Travis build with cAdvisor passes.

rjnagal:
LGTM.

I thought this was the only information we wanted to depend on. But I guess not anymore after kublet integration.

PR #585, opened 2015-03-13
Clean up code around Prometheus metrics, more DRY.

juliusv:
With this change, all definitions and functionality for a given metric
are in a single place only instead of being distributed all over the
file. This makes it easier to inspect the code for correctness and
adding/changing metrics.


juliusv:
@discordianfish @vmarmol

vmarmol:
ok to test

vmarmol:
LGTM, thanks again!

PR #586, opened 2015-03-13
Add handling of docker container to 2.0 stats API.

rjnagal:
Next: Apply similar handling to spec and summary endpoints.

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/google/cadvisor/events"
@@ -36,6 +37,8 @@ const (
 	specApi          = "spec"
 	eventsApi        = "events"
 	storageApi       = "storage"
+	typeName         = "name"
+	typeDocker       = "docker"
 )
 
 // Interface for a cAdvisor API version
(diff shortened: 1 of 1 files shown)


api/versions.go
```diff
@@ -326,12 +327,37 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 		query := info.ContainerInfoRequest{
 			NumStats: sr.Count,
 		}
-		cont, err := m.GetContainerInfo(name, &query)
-		if err != nil {
-			return fmt.Errorf("failed to get container %q: %v", name, err)
+		if sr.IdType == "name" {
```
vmarmol: nit: select :)

api/versions.go
```diff
@@ -326,12 +327,37 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+			}
+			contStats := convertStats(cont)
+			return writeResult(contStats, w)
 		}
-		contStats := convertStats(cont)
-		return writeResult(contStats, w)
+		if sr.IdType == "docker" {
+			contStats := make(map[string][]v2.ContainerStats, 0)
+			if name == "/" {
+				// special case: get all docker containers.
```
vmarmol: I'm +1 if you want to remove this behavior :) "/" is only valid with recursive == true. Should be fine since it is in the 2.0 API anyways

vmarmol:
Small comments, LGTM overall

api/versions.go
rjnagal: switched :)

api/versions.go
rjnagal: I think its important to have a way to get all docker containers. The other option:

/stats/docker?type=name&recursive=true

is going to stop working after custom cgroups.

We can make it require recursive=true when its added.

vmarmol: What I mean is

/stats/?type=docker&recursive=true

Gives you all containers


/stats/?type=docker

Fails with: no Docker container specified.

It should still work with custom cgroups since it'd be like today in systemd systems.

api/versions.go
```diff
@@ -326,12 +329,38 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+				// TODO(rjnagal): require recursive=true to be set?
+				containers, err := m.AllDockerContainers(&query)
+				if err != nil {
+					return fmt.Errorf("failed to get all docker containers: %v", err)
+				}
+				for name, cont := range containers {
+					contStats[name] = convertStats(&cont)
+				}
+			} else {
+				name = path.Base(name)
```
vmarmol: lets to strings.TrimPrefix(name, "/") so that we don't ignore if the users gives us: /this/is/bad/<docker ID>

rjnagal: done.

api/versions.go
rjnagal: agreed, this is what I meant.

PR #588, opened 2015-03-13
Add /machine endpoint for 2.0

rjnagal:
Only difference from v1 /machine is that both h/w and s/w states are included.

### DIFF
--- a/info/v2/machine.go
+++ b/info/v2/machine.go
@@ -1,0 +1,80 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v2
+
+import (
+	// TODO(rjnagal): Move structs from v1.
+	"github.com/google/cadvisor/info/v1"
+)
+
+type Attributes struct {
+	// Kernel version.
+	KernelVersion string `json:"kernel_version"`
+
+	// OS image being used for cadvisor container, or host image if running on host directly.
+	ContainerOsVersion string `json:"container_os_version"`
+
+	// Docker version.
+	DockerVersion string `json:"docker_version"`
+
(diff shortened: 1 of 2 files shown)


info/v2/machine.go
```diff
@@ -0,0 +1,80 @@
...
+	KernelVersion string `json:"kernel_version"`
+
+	// OS image being used for cadvisor container, or host image if running on host directly.
+	ContainerOsVersion string `json:"container_os_version"`
+
+	// Docker version.
+	DockerVersion string `json:"docker_version"`
+
+	// cAdvisor version.
+	CadvisorVersion string `json:"cadvisor_version"`
```
vishh: Should this really belong in `/machine`?

vishh:
LGTM except for that one question.

info/v2/machine.go
rjnagal: I read /machine as machine global information, rather than h/w information. I was thinking that one endpoint to get all machine-wide attributes would be useful. But I am fine splitting it into two endpoints.

vishh: To me, `/machine` seems to imply the hardware config. Should we have a `/version` endpoint that provides all the software version info?

vmarmol: +1 to /version just since there could be multiple cAdvisors. I'm okay putting some software stuff here too.

rjnagal: or we can rename ```/machine``` to ```/attributes```

vmarmol: I like machine more :) but that's just me being picky

vishh: I like `/attributes'. But I would still vote for a `/machine` with static info and a separate `/attributes` which provide more dynamic info.

rjnagal:
you guys are confusing me :) 

So do I hear three? /machine and /version as it is today and /attribute that combines the two and adds more stuff. /attribute would be the more dynamic one?

vishh:
I am suggesting `/attributes` instead of `/version` for all software stuff like kernel version, distribution, etc.
`/version` can continue to return just cadvisor version.

rjnagal:
Ah, I see. I'd still like /attribute to return cpu_frequency, topology, and uuids, if not the capacity fields.
We don't have a /version today, but we can add the one as you suggested.

vmarmol:
Lets go with one overall /attributes since we all seem to agree that's fine :) Having a separate /version is nice so we can use it directly in clients and scripts.

vishh:
+1 on `/attributes`

rjnagal:
ok, new set of changes:
- /machine : same as v1.2
- /version : returns cadvisor version
- /attributes : returns the originally proposed /machine (union of versionInfo and machineInfo)

vmarmol:
LGTM

PR #593, opened 2015-03-15
Return unique containers in Subcontainers().

### LINKED ISSUE #592: `manager.SubcontainersInfo` reports Docker containers multiple times
When I ran cAdvisor locally against some Docker containers, I noticed that the exported Prometheus metrics were present three times for every Docker container (error reporting for this triplication in the Prometheus client library was turned off by default, otherwise it would have resulted in an error).

For example, `container_cpu_usage_seconds_total` under `/metrics`:

```
container_cpu_usage_seconds_total{cpu="cpu00",id="/docker/35aa95c76ec978c0a293233ddbb307853848eb0c7f7d14c7c3d84e648af3c376",name="pensive_poitras"} 192.631525408
container_cpu_usage_seconds_total{cpu="cpu00",id="/docker/35aa95c76ec978c0a293233ddbb307853848eb0c7f7d14c7c3d84e648af3c376",name="pensive_poitras"} 192.631525408
container_cpu_usage_seconds_total{cpu="cpu00",id="/docker/35aa95c76ec978c0a293233ddbb307853848eb0c7f7d14c7c3d84e648af3c376",name="pensive_poitras"} 192.631525408
```

Digging deeper, I found that even at version `0.10.1`, cAdvisor reports every Docker container three times in `/api/v1.1/subcontainers`. The output of:

```bash
curl http://localhost:8080/api/v1.1/subcontainers/ | jsonpipe | sort
```

...is here:

https://gist.githubusercontent.com/juliusv/dd22acd9818841e3b073/raw/subcontainersinfo.jsonpipe

If you search for `name	"/docker/35aa95c76ec978c0a293233ddbb307853848eb0c7f7d14c7c3d84e648af3c376"` in there, you'll see three entries for this container name instead of just one.

Is this known and expected behavior or a bug?

Some infos about my system:

```
$ docker version
Client version: 1.5.0
Client API version: 1.17
Go version (client): go1.4.1
Git commit (client): a8a31ef
OS/Arch (client): linux/amd64
Server version: 1.5.0
Server API version: 1.17
Go version (server): go1.4.1
Git commit (server): a8a31ef
```

I'm on Ubuntu 14.04. This is the output of `find /sys/fs/cgroups`:

https://gist.github.com/juliusv/7e2cb00069e621a7ef2a

vmarmol:
Before this we'd return an instance for each alias we'd saved.

Fixes #592.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -378,22 +378,26 @@ func (self *manager) containerDataToContainerInfo(cont *containerData, query *in
 }
 
 func (self *manager) SubcontainersInfo(containerName string, query *info.ContainerInfoRequest) ([]*info.ContainerInfo, error) {
-	var containers []*containerData
+	var containersMap map[string]*containerData
 	func() {
 		self.containersLock.RLock()
 		defer self.containersLock.RUnlock()
-		containers = make([]*containerData, 0, len(self.containers))
+		containersMap = make(map[string]*containerData, len(self.containers))
 
-		// Get all the subcontainers of the specified container
+		// Get all the unique subcontainers of the specified container
 		matchedName := path.Join(containerName, "/")
 		for i := range self.containers {
 			name := self.containers[i].info.Name
 			if name == containerName || strings.HasPrefix(name, matchedName) {
-				containers = append(containers, self.containers[i])
+				containersMap[self.containers[i].info.Name] = self.containers[i]
 			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


juliusv:
This fixes #592 for me and looks good, thanks! :+1:

rjnagal:
LGTM

Thanks for the fix @vmarmol. I noticed the same while making 2.0 changes, where I just used a map.

PR #599, opened 2015-03-18
Add recursive and type options (raw/docker) to spec and summary endpoints.

rjnagal:
spec, stats, and summary now all support the same options (except count,
which is only parsed for stats).

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -63,11 +63,14 @@ type Manager interface {
 	// Gets information about a specific Docker container. The specified name is within the Docker namespace.
 	DockerContainer(dockerName string, query *info.ContainerInfoRequest) (info.ContainerInfo, error)
 
-	// Gets spec for a container.
-	GetContainerSpec(containerName string) (v2.ContainerSpec, error)
+	// Gets spec for all containers based on request options.
+	GetContainerSpec(containerName string, options v2.RequestOptions) (map[string]v2.ContainerSpec, error)
 
-	// Get derived stats for a container.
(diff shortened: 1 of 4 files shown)


vmarmol:
btw:

```
$ ./deploy/check_gofmt.sh .
api/versions.go
```

manager/manager.go
```diff
@@ -66,8 +66,11 @@ type Manager interface {
 	// Gets spec for a container.
 	GetContainerSpec(containerName string) (v2.ContainerSpec, error)
 
-	// Get derived stats for a container.
-	GetContainerDerivedStats(containerName string) (v2.DerivedStats, error)
+	// Get data for all requested containers based on the request options.
+	GetRequestedContainers(containerName string, options v2.RequestOptions) (map[string]*containerData, error)
```
vmarmol: containerData wasn't meant to be publicly exposed. If we do this route we should make it safe to do so (doesn't have to be in this PR). I'm just worried about its use in Kubelet :)

We can also replace most of the stuff here with a call to this and then getting some data.

rjnagal: We can go back to exposing  manager.GetDerivedStats() and manager.GetSpec() and make containerData private. We'll just need to move that logic from api/ to manager/

Let me do that in this PR.

vmarmol: I'm +1 on exposing a Container object that we can GetSpec() and GetStats() from.

rjnagal:
PTAL. Moved all containerData references back to manager.

vmarmol:
LGTM

PR #600, opened 2015-03-19
adding systemd oomparser functionality

kateknister:


vmarmol:
Ping @rjnagal

utils/oomparser/oomparser.go
```diff
@@ -153,12 +153,37 @@ func (self *OomParser) analyzeLines(ioreader *bufio.Reader, outStream chan *OomI
...
 			in_oom_kernel_log = false
+			glog.V(1).Infof("Sending an oomInstance: %v", oomCurrentInstance)
 			outStream <- oomCurrentInstance
 		}
 	}
 	glog.Infof("exiting analyzeLines")
 }
 
+func callJournalctl() (io.ReadCloser, error) {
+	cmd := exec.Command("journalctl", "-f")
```
rjnagal: Do we have access to journalctl from within the cadvisor container?
Other option might be to communicate with DBus directly eg like github.com/TheCreeper/JournalCheck

kateknister: Only if it's run as root.

rjnagal: I think we need to do the generic solution. We can just use the pkg I referred to earlier. I am fine with going with journalctl binary approach first as this would work with kubernetes at least.

utils/oomparser/oomparser_test.go
```diff
@@ -116,44 +116,6 @@ func TestCheckIfStartOfMessages(t *testing.T) {
...
-		if *oomCheckInstance != *oomInstance {
-			t.Errorf("wrong instance returned. Expected %v and got %v",
-				oomCheckInstance, oomInstance)
-			t.Errorf("Container of one was %v and the other %v", oomCheckInstance.ContainerName, oomInstance.ContainerName)
-		}
-	case <-timeout:
-		t.Error(
-			"timeout happened before oomInstance was found in test file")
-	}
-}
```
rjnagal: Are there any new equivalent tests that we can add?

utils/oomparser/oominfo/main.go
```diff
@@ -1,44 +0,0 @@
...
-		if err != nil {
-			glog.Errorf("%v", err)
-		}
-		// demonstration of how to get oomLog's list of oomInstances or access
-		// the user-declared oomInstance channel, here called outStream
-		for oomInstance := range outStream {
-			glog.Infof("Reading the buffer. Output is %v", oomInstance)
-		}
-	}
-}
```
rjnagal: Why did we delete this example binary? Can we update instead?

utils/oomparser/oomparser_test.go
kateknister: I deleted these tests because I deleted the function.  The tests for StreamOoms are almost identical to the tests I deleted for analyzeLines

kateknister:
Hey @rjnagal, does this look alright as a first patch? Once I figure out how to communicate with the dbus as you suggested I can replace it with that solution.

rjnagal:
LGTM.

Sounds good. Let's get the other changes in later.

PR #602, opened 2015-03-19
Revert back to progrium/busybox.

### LINKED ISSUE #601: cadvisor won't start dockerized
I wanted to enable prometheus support for cadvisor, which doesn’t appear to be in the latest/beta/canary builds available from the docker registry (the /metrics endpoint redirects to /containers/ causing an error when prometheus scrapes it).

I cloned the repository (changeset: a973ddb) and used the deploy/build.sh script to build and generate the docker image. However running the generated image resulted in:

    FATA[0000] Error response from daemon: Cannot start container <hash>: no such file or directory

I tracked this down to the fact that cadvisor uses cog features, and at least in my environment this links against glibc, whereas the alpine image cadvisor is based off for the docker image uses musl, [related docker-alpine issue](https://github.com/gliderlabs/docker-alpine/issues/12)

Reverting [this change to the Dockerfile](https://github.com/google/cadvisor/commit/e69432f5db51d3dfa25781384c3902e7f0f57d6d#diff-2a33e645dffdb8c6e5dc53153fc455a6) fixes the issue as the old base image contains glibc

Not sure how this should progress: there must have been a reason to switch to alpine, in which case there needs to be an updated build script/instructions, or switch back to busybox.

vmarmol:
Alpine doesn't have glibc which we build with.

Fixes #601

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM gliderlabs/alpine:3.1
+FROM progrium/busybox
 MAINTAINER dengnan@google.com vmarmol@google.com vishnuk@google.com
 
 # Grab cadvisor from the staging directory.


vishh:
LGTM

vishh:
Feel free to self merge once tests pass.

rjnagal:
LGTM

PR #603, opened 2015-03-19
Lazily init DBus only when it is needed

smarterclayton:
OpenShift has some cAdvisor packages in its import chain. The init() here
invokes Dbus start even when cAdvisor is not actively running or being
used (in our packages).  This change makes the systemd check and dbus
initialization and connection lazy, occuring only the first time someone
invokes UseSystemd().

Seemed like the simplest possible tweak.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -21,6 +21,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/docker/libcontainer/cgroups"
 	"github.com/docker/libcontainer/cgroups/systemd"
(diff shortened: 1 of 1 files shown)


smarterclayton:
I have a CLA under Red Hat.

vmarmol:
ok to test

vmarmol:
LGTM, will wait for green. Thanks @smarterclayton

smarterclayton:
Thank you - cAdvisor gives me such pretty numbers to show everyone.  :)

----- Original Message -----

vishh:
Tests are passing. Merging this PR.

PR #604, opened 2015-03-20
Reduce the level for a few common log messages

smarterclayton:
Following the Kubernetes convention of V(2) is normal verbosity (log
each request to a webserver).


vishh:
ok to test

vishh:
LGTM. Will merge once tests pass.

PR #609, opened 2015-03-23
machine: collect node boot id if available

simon3z:


vmarmol:
ok to test

vmarmol:
LGTM, thanks @simon3z!

PR #610, opened 2015-03-23
goauth2 replacement

kargakis:
I am not sure if the way I set up the token is right since I am not very familiar with authentication stuff. Please review the change and let me know what needs to change.


vishh:
ok to test

vmarmol:
Looks like we don't ever get the token.

kargakis:
> Looks like we don't ever get the token.

I changed the generation of the token to be handled from `jwt` as it was before. Can you retest this?

vmarmol:
Thanks @kargakis! This looks good. @rjnagal is there any other testing you'd like to do?

rjnagal:
Let me bring up an instance while travis is still working on it.

vmarmol:
@kargakis looks like something is missing from the deps.

rjnagal:
Hit the same error as travis. @kargakis Can you run `godep save ./...`

kargakis:
Missing deps added and Travis is happy

rjnagal:
testing locally now

rjnagal:
LGTM.

Works great. Thanks @kargakis

PR #611, opened 2015-03-23
Add documentation for 2.0 API.

rjnagal:


docs/api_v2.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Machine Information
+
+The resource name for machine information is as follows:
+
+`/api/v2.0/machine`
+
+The machine information is returned as a JSON object containing:
+
+- Number of schedulable logical CPU cores
```
vishh: Should we include the field name here?

vmarmol:
LGTM, great docs thanks @rjnagal!

docs/api_v2.md
```diff
@@ -0,0 +1,105 @@
...
+
+| Docker container     | Resource Name                             |
+|----------------------|-------------------------------------------|
+| All docker containers| /api/v2.0/stats?type=docker&recursive=true|
+| clever_colden        | /api/v2.0/stats/clever_colden?type=docker |
+| 2c4dee605d22         | /api/v2.0/stats/2c4dee605d22?type=docker  |
+
+The Docker name can be either the UUID or the short name of the container. It returns the information of the specified container(s).
+
+Note that `recursive` is only valid when docker root is specified. It is used to get stats for all docker containers.
```
vishh: What about getting stats for all subcontainers under '/'?

docs/api_v2.md
vmarmol: Honestly, I would even remove this list and just keep a pointer to the MachineInfo struct. We don't keep a list for other API types and the list is manually updated...

vishh: I feel the [docker remote API](https://docs.docker.com/reference/api/docker_remote_api_v1.17/) documentation is pretty useful. That is the basis for my comment.

vmarmol: I agree, I mean that we don't list all the fields here (they don't in the Docker API either). Long term we should move to go-restfull or swagger which auto-generate these :)

rjnagal: The example requests and output do look useful. I'll try to add these.

Removing the field descriptions from machine now.

docs/api_v2.md
rjnagal: It still works with regular combination when container name is used. Only for docker containers, we don't accept docker container id and recursive=true.

vishh: Ahh got it. May be add a sentence about 'recursive' option for regular
cgroup stats as well?

PR #612, opened 2015-03-23
Created client interface for api events handler

kateknister:


api/handler.go
```diff
@@ -163,6 +166,43 @@ func streamResults(results chan *events.Event, w http.ResponseWriter, r *http.Re
...
 }
 
+func requestResults(w http.ResponseWriter, r *http.Request) error {
+	oldUrlString := r.URL.RequestURI()
+	glog.Infof("oldUrlString is %v", oldUrlString)
+	requestElements := clientRegexp.FindStringSubmatch(oldUrlString)
+	if len(requestElements) != 3 {
+		return fmt.Errorf("Could not interpret client request: %v", requestElements)
+	}
+	newUrlString := "http://localhost:8080" + requestElements[1] + "events" + requestElements[2]
```
rjnagal: Can you add comments on why we need to map it to localhost and hardcode port number here?

api/handler.go
```diff
@@ -163,6 +166,43 @@ func streamResults(results chan *events.Event, w http.ResponseWriter, r *http.Re
...
+	glog.Infof("requestElements are %v", requestElements)
+	glog.Infof("newUrlString is %v", newUrlString)
+	req, err := http.NewRequest("GET", newUrlString, nil)
```
rjnagal: Sorry, I don't have enough context as to what client API is supposed to do. From here, it looks like its rewriting the query to events endpoint and just print out the result. I haven't followed the events stuff in detail before. Can you describe it a bit here.

kateknister:
Hey @rjnagal,  I moved the events client to the client library and got rid of the hardcoded url. Rather than printing, now the functions return a struct or a channel through which events may be retrieved. I put in an example main package as well but I can remove that, I wasn't sure if that was useful or not.

client/client.go
```diff
@@ -39,10 +42,32 @@ func NewClient(url string) (*Client, error) {
...
 	return &Client{
-		baseUrl: fmt.Sprintf("%sapi/v1.2/", url),
+		baseUrl: fmt.Sprintf("%sapi/v1.3/", url),
 	}, nil
 }
 
+// Returns all past events that satisfy the request
+func (self *Client) EventStaticInfo(name string) (einfo []*events.Event, err error) {
+	u := self.eventsInfoUrl(name)
+	glog.Infof("got complete url %v", u)
```
rjnagal: Make the log message V(3).

api/handler.go
```diff
@@ -57,6 +57,7 @@ func RegisterHandlers(mux httpMux.Mux, m manager.Manager) error {
 
 // Captures the API version, requestType [optional], and remaining request [optional].
 var apiRegexp = regexp.MustCompile("/api/([^/]+)/?([^/]+)?(.*)")
+var clientRegexp = regexp.MustCompile("(/api/[^/]+/?)client(.*)")
```
rjnagal: Is this used anywhere?

rjnagal:
@kateknister I think this looks good, except the one nit about unused var.

rjnagal:
LGTM

rjnagal:
Looks like one of the tests broke:

 github.com/google/cadvisor/client
client/client.go:50: undefined: events.Event
client/client.go:53: undefined: events.Event
# github.com/google/cadvisor/client
client/client.go:50: undefined: events.Event
client/client.go:53: undefined: events.Event
?   	github.com/google/cadvisor	[no test files]

PR #615, opened 2015-03-25
Return all stats by default.

vmarmol:
Rather than relying on 64 being a "big enough" value.

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -164,11 +164,7 @@ func streamResults(results chan *events.Event, w http.ResponseWriter, r *http.Re
 }
 
 func getContainerInfoRequest(body io.ReadCloser) (*info.ContainerInfoRequest, error) {
-	var query info.ContainerInfoRequest
-
-	// Default stats and samples is 64.
-	query.NumStats = 64
-
+	query := info.DefaultContainerInfoRequest()
 	decoder := json.NewDecoder(body)
 	err := decoder.Decode(&query)
 	if err != nil && err != io.EOF {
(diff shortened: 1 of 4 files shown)


vishh:
Curious: How does this change help deal with the kubelet part of the issue?

vishh:
LGTM

rjnagal:
Returning all stats doesn't seem like a good default, as it can change based on our memory storage/backend setup.

Does it fix a real problem?

vmarmol:
No, it just fixed an expected behavior (in my opinion). I'm fine leaving a default but having it documented. It is better than returning all...will change this PR for that.

@vishh, not for that but there is a followup with your name on it :)

vishh:
We should separate time range queries from the number of stats. Requiring a max number of stats for time range queries is non intuitive personally.

vmarmol:
@vishh I can see the usecase where someone wants to get the 10 stats before a certain time or limit the number of results. But it does feel like a better default for time-based queries is "all" rather than an arbitrary limit. That was the logic behind this PR. It would be simpler if we returned "all" by default in our cases. WDYT @rjnagal? It may be specific to the flags on the memory storage, but it'll provide consistent behavior.

An alternative is to say that the default when specifying both start and end is "all" unless otherwise specified.

rjnagal:
Separating out time and count queries sounds fine. Not receiving a ton of
stats by default would be nice. When we are using an external storage, will
it try to get all stats from backend?

vmarmol:
We no longer fetch stats from external drivers. We just use them for export.

I'm fine with specifying: default NumStats = 60 and if Start and End are specified NumSats is ignored.

rjnagal:
SG.

vmarmol:
Updated to document the limit, set the default, and ignore NumStats for time range queries. PTAL @vishh @rjnagal

rjnagal:
LGTM

PR #617, opened 2015-03-27
v2.0 API does not depend on v1.x.

vmarmol:
The v2.0 API is separate and will evolve to have separate structured
from v1.x.


vishh:
LGTM

PR #622, opened 2015-04-01
Move Event to info since it is a public API object.

vmarmol:


vmarmol:
Heads up @kateknister and @vishh

rjnagal:
LGTM

PR #626, opened 2015-04-01
Fix build. change  event.Event to info.Event.

rjnagal:


rjnagal:
@vmarmol all green.

vmarmol:
Thanks @rjnagal! Sorry for the delay.

PR #629, opened 2015-04-04
Preliminary integration tests for getting events

kateknister:
These integration tests don't work yet but this is the basic setup.  One main issue is that I cannot oom kill the docker containers in a reasonable timeframe.

### DIFF
--- a/integration/tests/api/event_test.go
+++ b/integration/tests/api/event_test.go
@@ -1,0 +1,75 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


rjnagal:
You can use bernardo/stress images.
Memory options:
-m, --vm N         spawn N workers spinning on malloc()/free()
     --vm-bytes B   malloc B bytes per vm worker (default is 256MB)
     --vm-stride B  touch a byte every B bytes (default is 4096)
     --vm-hang N    sleep N secs before free (default is none, 0 is inf)
     --vm-keep      redirty memory instead of freeing and reallocating

You can consistently hit the memory limit of a container with --vm-keep.

integration/tests/api/event_test.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
```
rjnagal: s/2014/2015/

rjnagal:
LGTM.

We can improve it further in separate PRs.

PR #640, opened 2015-04-09
Make EventData a concrete type.

vmarmol:
Having this as an interface was giving us a lot of problems with
unmarshaling. Made into a union of concrete types.


vishh:
LGTM

vmarmol:
The Jenkins failure is the same one we've been seeing and expected. This change should help @kateknister send a fix for the issue so I'd recommend we merge :)

vmarmol:
Thanks for the super quick review @vishh :)

PR #641, opened 2015-04-09
Improved event integration tests

kateknister:


integration/tests/api/event_test.go
```diff
@@ -52,24 +51,25 @@ func TestStreamingEventInformationIsReturned(t *testing.T) {
...
-			require.Nil(t, err)
-			var oomEvent *oomparser.OomInstance
-			err = json.Unmarshal(marshaledData, &oomEvent)
-			require.Nil(t, err)
-			require.True(t, oomEvent.ProcessName == "stress")
+	for {
+		select {
+		case ev := <-einfo:
+			if ev.EventType == 0 {
+				if strings.Trim(ev.ContainerName, "/ ") == strings.Trim("docker/"+containerId, "/ ") {
```
vmarmol: this won't be true in CoreOS machines for example. WDYT of getting the name from a Client().ContainerInfo() and checking that?

kateknister: I am running into issues using the Client().ContainerInfo() and DockerContainer() methods with this image type because I keep getting the error "failed to get container..." with error "unknown container...." I assume this is because I force the container to timeout and generally the client's attempt to get the container information occurs after the container has stopped. However, the container id is returned when the "-d" flag is added to the docker command, so would it be sufficient to check that the event container name contains the returned containerId?

vmarmol: Ah good point. Let's just do a strings.Contains() of the Docker ID then :)

vmarmol:
@cadvisorJenkinsBot please test

rjnagal:
LGTM.

Merging it. @kateknister is going to add the missing piece later.

Issue #643, opened 2015-04-13
Networking stats are reversed


rjnagal:
I was trying to debug networking integration test flakiness, and it seems like the cause is that sometimes the ping container is reporting no Tx packets but multiple Rx packets. I think this means that the numbers are somehow reversed after we started using libcontainer stats. Looking into it.

rjnagal:
Fixed by #644

PR #644, opened 2015-04-13
Fix networking stats by switching Tx and Rx stats.

rjnagal:
A better fix is to directly use stats collected by libcontainer.

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -16,10 +16,7 @@ package libcontainer
 
 import (
 	"fmt"
-	"io/ioutil"
 	"path"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/docker/libcontainer"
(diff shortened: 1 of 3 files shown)


rjnagal:
Fixes #643 for now.

container/libcontainer/helpers.go
```diff
@@ -90,10 +87,21 @@ func GetStats(cgroupManager cgroups.Manager, networkInterfaces []string) (*info.
 	if len(networkInterfaces) != 0 {
 		// ContainerStats only reports stat for one network device.
 		// TODO(rjnagal): Handle multiple physical network devices.
-		stats.Network, err = sysinfo.GetNetworkStats(networkInterfaces[0])
+		// TODO(rjnagal): Use networking stats directly from libcontainer.
+		net, err := sysinfo.GetNetworkStats(networkInterfaces[0])
 		if err != nil {
 			return stats, err
 		}
+		// Ingress for host veth is from the container.
```
vmarmol: I think we need to do this in `container/docker/handler.go`. This function is also used by the raw handler so we'll get flipped stats for host.

rjnagal: Ah, good catch! Fixed. The direction is fine for both host and docker containers now.

vmarmol:
LGTM

vmarmol:
Thanks @rjnagal!

PR #648, opened 2015-04-15
Split OOM event into OOM and OOM Kill

vmarmol:


manager/manager.go
```diff
@@ -876,21 +876,32 @@ func (self *manager) watchForNewOoms() error {
...
 			newEvent := &info.Event{
 				ContainerName: oomInstance.ContainerName,
 				Timestamp:     oomInstance.TimeOfDeath,
 				EventType:     info.EventOom,
+			}
+			err := self.eventHandler.AddEvent(newEvent)
+			if err != nil {
+				glog.Errorf("failed to add OOM event for %q: %v", oomInstance.ContainerName, err)
+			}
+			glog.V(2).Infof("Created an OOM event in container %q at %v", oomInstance.ContainerName, oomInstance.TimeOfDeath)
```
vishh: V(3) or V(4)?

rjnagal:
So every OOM creates two events now? It seems a little confusing.

manager/manager.go
vmarmol: Done

vishh:
LGTM.

vmarmol:
@rjnagal its meant to simplify and allow for watching of OOM kills and OOMs separately.

For example:
Container /a/b watches for OOM kills on itself and its children. Container /a OOMs and kills a process in /a/b/c. Currently, my notification won't fire, with the new interface it will.

It is true that not all OOMs lead to OOM kills (and we don't handle those today, we should).

rjnagal:
LGTM

I understand the rationale, but I worry that people not familiar with OOM working will get confused.
Let's get it in for now as your use-case is legitimate and needs to be covered. The API solution might be a more sophisticated spec when registering for watches.

PR #653, opened 2015-04-15
Added container_name field to event API

kateknister:


rjnagal:
LGTM

vmarmol:
We should be handling this with the resource path similar to what we do in other parts of the API. So: `/api/v1.3/events/container_name?other_params`

Will send a PR.

PR #655, opened 2015-04-16
preliminary changes to make load stats hierarchical

kateknister:


kateknister:
This shouldn't be merged yet, I just wanted some feedback on this approach

kateknister:
@rjnagal

manager/container.go
```diff
@@ -270,6 +271,30 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+			// glog.Infof("Failed to get subcontainer stats with error %v", err)
+			continue
+		}
+		stats.UpdateTaskStats(loadStats)
+		c.updateLoad(loadStats.NrRunning)
+		// convert to 'milliLoad' to avoid floats and preserve precision.
+		stats.Cpu.LoadAverage = int32(c.loadAvg * 1000)
+		// glog.Infof("Added stats from cpuloadpath %v and name %v", subpath, sub.Name)
+	}
+	return stats, nil
```
rjnagal: Doesn't this update every docker containers load multiple times in each housekeeping loop - once when we are processing /docker, and once processing /docker/foo?

rjnagal: Just tracking the time of last update and comparing might be helpful to avoid double checking.

manager/container.go
```diff
@@ -270,6 +271,30 @@ func (c *containerData) updateLoad(newLoad uint64) {
 	}
 }
 
+func (c *containerData) updateSubcontainerStats(stats *info.ContainerStats, path string) (*info.ContainerStats, error) {
+	// type []info.ContainerReference
+	subcontainers, err := c.handler.ListContainers(container.ListRecursive)
```
rjnagal: Can we cache this list and update on each subcontainer create/delete?

manager/container.go
```diff
@@ -270,6 +271,30 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+	// glog.Infof("subcontainers for container %v are %v", c.info.Name, subcontainers)
+	if err != nil {
+		// glog.Infof("returning from subcontainerstats with error %v", err)
+		return stats, err
+	}
+	for _, sub := range subcontainers {
```
rjnagal: Another problem with this approach is that some of the subcontainers might be idle and we would end up checking load for them. Today,  manager's housekeeping slows down stats check for idle containers.

This is more of a concern on systemd systems, where there are billion containers created at startup.

kateknister:
@cadvisorJenkinsBot

manager/container.go
```diff
@@ -270,6 +273,44 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+	}
+	return stats, nil
+}
+
+func (c *containerData) getLoadStats() (info.LoadStats, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	var loadstats info.LoadStats
+	if time.Since(c.lastCpuLoadGrab) > 2*time.Second {
+		loadstats, err := c.loadReader.GetCpuLoad(c.info.Name, c.cgroupPath)
```
rjnagal: Do we need to check if loadReader is nil?

manager/container.go
```diff
@@ -270,6 +273,44 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+		stats.Cpu.LoadAverage = int32(c.loadAvg * 1000)
+	}
+	return stats, nil
+}
+
+func (c *containerData) getLoadStats() (info.LoadStats, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	var loadstats info.LoadStats
+	if time.Since(c.lastCpuLoadGrab) > 2*time.Second {
```
rjnagal: why 2 seconds? Our normal housekeeping is 1 second. so this will always be true.

manager/container.go
rjnagal: ah, we check this in the calling fn.

manager/container.go
```diff
@@ -270,6 +273,44 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+		loadstats, err := c.loadReader.GetCpuLoad(c.info.Name, c.cgroupPath)
+		if err == nil {
+			c.lastCpuLoadGrab = time.Now()
+		}
+		return loadstats, err
+	} else {
+		subStats, statsErr := c.handler.GetStats()
```
rjnagal: Why are we calling GetStats again?

manager/container.go
```diff
@@ -270,6 +273,44 @@ func (c *containerData) updateLoad(newLoad uint64) {
 	}
 }
 
+func (c *containerData) updateSubcontainerStats(stats *info.ContainerStats, path string) (*info.ContainerStats, error) {
```
rjnagal: Should this be recursive? Calling load stats on "/" should sum up stats on all existing containers.

manager/manager.go
```diff
@@ -688,12 +688,54 @@ func (m *manager) createContainer(containerName string) error {
...
 
 	return nil
 }
 
+func (m *manager) getSubcontainerContainerData(containerName string) map[namespacedContainerName]*containerData {
+	subcontainers := make(map[namespacedContainerName]*containerData)
+	m.containersLock.RLock()
+	defer m.containersLock.RUnlock()
+	for namespacedNameKey, containerDataValue := range m.containers {
+		if strings.HasPrefix(namespacedNameKey.Name, containerName) && namespacedNameKey.Name != containerName {
```
rjnagal: Does this add all containers as subcontainers of "/". Do we track the immediate children anywhere? We are only interested in immediate children.

manager/container.go
kateknister: This list of subcontainers is a set of all descendants of the container, based on how it was populated in getSubcontainerContainerData in manager.go 708. But this would need to be recursive if that were to change to only immediate children. Why does the below need to only track immediate children rather than all descendant containers?

rjnagal: Ah, I see what this is doing. In my mind, I was assuming that both load and task stats will be hierarchically reported.
This approach works too, but I am a little worried if we are doing repetitive work unnecessarily. We might also be maintained a large list of subcontainers for root, even though some of them are empty.

Can you measure the amount of work we end up doing for a busy machine? We might be able to make the current solution a bit more efficient if required.

kateknister: Sure, and I'll change it to be recursive. That seems like it would be more efficient, so that we're not always checking the sub-sub containers of a given container.

rjnagal:
I'll take this over, rebase, and re-push. closing for now.

PR #658, opened 2015-04-20
Fix subcontainer queries in subcontainers.

vmarmol:
Queries for root were not being correctly diagnosed.

### DIFF
--- a/events/handler_test.go
+++ b/events/handler_test.go
@@ -67,6 +67,8 @@ func ensureProperEventReturned(t *testing.T, expectedEvent *info.Event, eventObj
 func TestCheckIfIsSubcontainer(t *testing.T) {
 	myRequest := NewRequest()
 	myRequest.ContainerName = "/root"
+	rootRequest := NewRequest()
+	rootRequest.ContainerName = "/"
 
 	sameContainerEvent := &info.Event{
 		ContainerName: "/root",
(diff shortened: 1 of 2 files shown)


events/handler_test.go
```diff
@@ -78,6 +80,10 @@ func TestCheckIfIsSubcontainer(t *testing.T) {
 		ContainerName: "/root-completely-different-container",
 	}
 
+	if checkIfIsSubcontainer(rootRequest, sameContainerEvent) {
+		t.Errorf("should have found %v and %v did not have the same container name",
```
rjnagal: s/should have/should not have/

events/handler_test.go
```diff
@@ -87,8 +93,13 @@ func TestCheckIfIsSubcontainer(t *testing.T) {
...
 	}
 
+	rootRequest.IncludeSubcontainers = true
 	myRequest.IncludeSubcontainers = true
 
+	if !checkIfIsSubcontainer(rootRequest, sameContainerEvent) {
+		t.Errorf("should have found %v and %v had the same container",
+			rootRequest.ContainerName, sameContainerEvent.ContainerName)
+	}
 	if !checkIfIsSubcontainer(myRequest, sameContainerEvent) {
```
rjnagal: I think this check should be checking against subContainerName.

events/handler_test.go
```diff
@@ -87,8 +93,13 @@ func TestCheckIfIsSubcontainer(t *testing.T) {
 			myRequest, subContainerEvent)
 	}
 
+	rootRequest.IncludeSubcontainers = true
 	myRequest.IncludeSubcontainers = true
 
+	if !checkIfIsSubcontainer(rootRequest, sameContainerEvent) {
+		t.Errorf("should have found %v and %v had the same container",
```
rjnagal: The error message is hard to follow. Maybe : "should have found %v as %v is a subcontainer."

events/handler_test.go
vmarmol: Done

events/handler_test.go
vmarmol: Done

events/handler_test.go
vmarmol: I *think* this one is fine, we check against subcontainer below it. We check:
- `/root` subcontainerOf `/root` == `true`
- `/root/sub` subcontainerOf `/root` == `true`
- `/root-different` subcontainerOf `/root` == `false`

rjnagal:
LGTM

PR #664, opened 2015-04-24
Add /var/log/kern.log to kernel log files.

vmarmol:
This file is present in most distros and provides most of the logs we're
looking for. We support the others if this one is not available.


vmarmol:
/cc @vishh @rjnagal 

Verified this fixes the problem in Kubernetes. Will patch this in once we merge :)

rjnagal:
LGTM

vishh:
LGTM

PR #665, opened 2015-04-24
Limit the number of events stored to 24h.

vmarmol:
There is a TODO to make this configurable.

### DIFF
--- a/events/handler.go
+++ b/events/handler.go
@@ -23,20 +23,21 @@ import (
 
 	"github.com/golang/glog"
 	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/utils"
 )
 
-type EventSlice []*info.Event
+type byTimestamp []*info.Event
 
 // functions necessary to implement the sort interface on the Events struct
-func (e EventSlice) Len() int {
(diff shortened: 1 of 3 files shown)


rjnagal:
Why do it by time? The most common case we want to avoid is the crashlooping containers flooding events.

vmarmol:
We've been moving to time based since its easier to reason about (we keep all events for the last 24hrs rather than the last 1000 events). Alternative we can have some policy to rate-limit the errors or place an absolute limit as well as a time limit. WDYT? We can do either of those in another PR since this one makes the current situation strictly better :)

rjnagal:
This is better than what we have. But the memory bounds are still worrisome. I'd say we should place an absolute limit based on amount of memory we need to store these.

rjnagal:
Can you also look at the test failure to see if its related?

vmarmol:
@rjnagal PTAL, all tests pass. The error was on the CI side, the machines had a bunch of leftover crud.

vmarmol:
I will followup this PR with an absolute limit.

rjnagal:
LGTM

PR #670, opened 2015-04-30
Add --docker_only flag to enable tracking for only docker containers & root.

rjnagal:
This reduces unnecessary load on the system and also cleans up the UI clutter.
Currently defaulted to false.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -73,6 +73,9 @@ type Manager interface {
 	// Get info for all requested containers based on the request options.
 	GetRequestedContainersInfo(containerName string, options v2.RequestOptions) (map[string]*info.ContainerInfo, error)
 
+	// Returns true if the named container exists.
+	Exists(containerName string) bool
+
 	// Get information about the machine.
 	GetMachineInfo() (*info.MachineInfo, error)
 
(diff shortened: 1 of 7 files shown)


manager/manager.go
```diff
@@ -619,6 +622,22 @@ func (m *manager) GetVersionInfo() (*info.VersionInfo, error) {
...
 
+func (m *manager) Exists(containerName string) bool {
+	m.containersLock.Lock()
+	defer m.containersLock.Unlock()
+
+	namespacedName := namespacedContainerName{
+		Name: containerName,
+	}
+
+	// Check that the container didn't already exist.
```
vmarmol: stale comment?

vmarmol:
We'll print out each new raw container as a failure. We will also "detect" them every 60s and re-print out the failure. We need to somehow tell the manager to ignore the container. Maybe create it but don't house keep it?

manager/manager.go
rjnagal: removed

rjnagal:
Just ignoring housekeeping still shows the container in all APIs and UI. Its a query-able container that doesn't return anything. 

Uploaded a different fix that distinguish failed vs ignored containers in manager and silently fails the latter. PTAL.

vmarmol:
LGTM, thanks @rjnagal! Merging, the e2e machines are unhappy :(

PR #671, opened 2015-05-01
Add an item limit to TimedStore

vmarmol:
Made the default limit 100,000 elements which is a little over 1 event per second for a day and assuming 250 bytes per event it would take up ~24MiB per event type.

/cc @rjnagal

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -137,7 +137,7 @@ func New(memoryStorage *memory.InMemoryStorage, sysfs sysfs.SysFs) (Manager, err
 	glog.Infof("Version: %+v", newManager.versionInfo)
 
 	// TODO(vmarmol): Make configurable.
-	newManager.eventHandler = events.NewEventManager(24 * time.Hour)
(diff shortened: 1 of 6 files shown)


manager/manager.go
```diff
@@ -137,7 +137,7 @@ func New(memoryStorage *memory.InMemoryStorage, sysfs sysfs.SysFs) (Manager, err
 	glog.Infof("Version: %+v", newManager.versionInfo)
 
 	// TODO(vmarmol): Make configurable.
-	newManager.eventHandler = events.NewEventManager(24 * time.Hour)
+	newManager.eventHandler = events.NewEventManager(24*time.Hour, 100000)
```
rjnagal: Can we make this a flag?

vmarmol: Yes, the TODO is to make it configurable.

rjnagal:
LGTM

manager/manager.go
vmarmol: (since we'd still like to do the per-event limit before we expose it as a flag)

PR #678, opened 2015-05-04
Add Collector executor support

vmarmol:
/cc @rjnagal

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/docker/docker/pkg/units"
 	"github.com/golang/glog"
+	"github.com/google/cadvisor/collector"
 	"github.com/google/cadvisor/container"
 	info "github.com/google/cadvisor/info/v1"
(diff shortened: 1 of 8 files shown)


manager/container.go
```diff
@@ -226,12 +232,24 @@ func (c *containerData) housekeeping() {
 			}
 		}
 
-		// Schedule the next housekeeping. Sleep until that time.
+		// Run custom collectors.
+		nextCollectionTime, err := c.collectorManager.Collect()
```
rjnagal: check for nil. For most cases, we want to have collectorManager as nil.

vmarmol: I was more thinking that it'd never be nil although there may be no collectors. I want to re-implement stats as a collector so there should be at least one collector.

vmarmol: WDYT?

rjnagal: ok, SG.

rjnagal:
LGTM

PR #680, opened 2015-05-05
Make all links relative.

vmarmol:
This will allow cAdvisor to be reverse proxied, and fixes a long-standing feature request :)

Fixes #513
Fixes #566
Fixes #596
Fixes #667

PR #690, opened 2015-05-06
Revert "Make all links relative."

rjnagal:
Reverts google/cadvisor#680


vishh:
LGTM. Lets get '/docker' HTTP endpoint to work and then re-commit this patch.

rjnagal:
/docker is broken as the change treats it as a container endpoint (container/docker).
Github is also not able to render container_js.go file, so something might have gone wrong with the git merge.

Will fix and reapply.

PR #691, opened 2015-05-06
Make all link relative - part 2.

rjnagal:
Added missing docker page handling.


vmarmol:
LGTM

vmarmol:
Thanks for the fix @rjnagal!

PR #695, opened 2015-05-07
Fix handling of time ranges in timed store.

rjnagal:
The current logic assumes that entries are added to the store in
monotonically increasing order for time. This is not true when
we add creation events for existing containers.

Also fixes the bug that we don't delete expired entries if all entries in the buffer are expired.
Modified the test that depended on that behavior.


vmarmol:
LGTM

Nice catch! I remember noticing the issue of us assuming increasing times, but I forgot about it :(

PR #696, opened 2015-05-07
Improve approximation of container creation time.

rjnagal:
Containers with subcontainers always report creation time to be same as the time
of creation of the latest subcontainer.

Still not an ideal solution, but accurate for most practical purposes.


vmarmol:
LGTM

PR #698, opened 2015-05-07
Return partial topology information if cache info is not available.

rjnagal:
Everything other than cache information is available through /proc/cpuinfo.


vmarmol:
LGTM

PR #700, opened 2015-05-08
update documentation for CentOS 7

furlongm:
1) cgroups were getting mounted at /cgroup: not /cgroup
2) incorrect flag was used:

provided but not defined: --privilege
See 'docker run --help'.


furlongm:
I signed it!

rjnagal:
Thanks for trying out the new docs and fixing them @furlongm 

LGTM

PR #704, opened 2015-05-11
Fix WorkingSet calculation

### LINKED ISSUE #685: Node memory is returning 9223372036854775807
9223372036854775807 is the largest int64

vmarmol:
Fixes #685

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -186,10 +186,21 @@ func toContainerStats2(s *cgroups.Stats, ret *info.ContainerStats) {
 		ret.Memory.HierarchicalData.Pgmajfault = v
 	}
 	if v, ok := s.MemoryStats.Stats["total_inactive_anon"]; ok {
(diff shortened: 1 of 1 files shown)


container/libcontainer/helpers.go
```diff
@@ -186,10 +186,21 @@ func toContainerStats2(s *cgroups.Stats, ret *info.ContainerStats) {
...
+			workingSet = 0
+		} else {
+			workingSet -= v
 		}
+
+		if v, ok := s.MemoryStats.Stats["total_inactive_file"]; ok {
+			if workingSet < v {
+				workingSet = 0
+			} else {
+				workingSet -= v
```
rjnagal: you need to check workingSet to be non-negative after this deduction.

vmarmol: It's a `uint64` so unfortunately I can't check. I check whether the deduction would make it negative beforehand and just set it to zero. That should work I think?

rjnagal: ah, this should be okay then.

LGTM, will merge on green

PR #705, opened 2015-05-11
fix working set calculations.

rjnagal:


rjnagal:
@vmarmol got it already :)

vmarmol:
Sorry :)

PR #706, opened 2015-05-11
Add 'bytes available' to fs info.

rjnagal:
The current capacity and usage numbers are insufficient to figure out
actual bytes available for a non-root user for the fs. Available is the
value used by df and the one we need to track to detect low diskspace
condition.


vmarmol:
LGTM

PR #707, opened 2015-05-11
Add an api to support ps/top.

rjnagal:
I need the ps/top output for the UI. Not sure if we should be adding it as an API. WDYT?

Also check the fields that I selected to show. Do we have any issues showing the full command. My guess is that we don't care.

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -38,6 +38,7 @@ const (
 	storageApi       = "storage"
 	attributesApi    = "attributes"
 	versionApi       = "version"
+	psApi            = "ps"
 )
 
 // Interface for a cAdvisor API version
@@ -304,7 +305,7 @@ func (self *version2_0) Version() string {
 }
 
 func (self *version2_0) SupportedRequestTypes() []string {
-	return []string{versionApi, attributesApi, eventsApi, machineApi, summaryApi, statsApi, specApi, storageApi}
+	return []string{versionApi, attributesApi, eventsApi, machineApi, summaryApi, statsApi, specApi, storageApi, psApi}
 }
 
 func (self *version2_0) HandleRequest(requestType string, request []string, m manager.Manager, w http.ResponseWriter, r *http.Request) error {
@@ -391,6 +392,17 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 		return writeResult(fi, w)
 	case eventsApi:
 		return handleEventRequest(request, m, w, r)
+	case psApi:
+		// reuse container type from request.
+		// ignore recursive.
+		// TODO(rjnagal): consider count to limit ps output.
+		name := getContainerName(request)
+		glog.V(4).Infof("Api - Spec for container %q, options %+v", name, opt)
+		ps, err := m.GetProcessList(name, opt)
+		if err != nil {
+			return fmt.Errorf("process listing failed: %v", err)
+		}
+		return writeResult(ps, w)
 	default:
 		return fmt.Errorf("unknown request type %q", requestType)
 	}
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -399,8 +399,7 @@ func (self *rawContainerHandler) ListThreads(listType container.ListType) ([]int
 }
 
 func (self *rawContainerHandler) ListProcesses(listType container.ListType) ([]int, error) {
-	// TODO(vmarmol): Implement
-	return nil, nil
+	return libcontainer.GetProcesses(self.cgroupManager)
 }
 
 func (self *rawContainerHandler) watchDirectory(dir string, containerName string) error {
--- a/info/v2/container.go
+++ b/info/v2/container.go
@@ -166,3 +166,16 @@ type RequestOptions struct {
 	// Whether to include stats for child subcontainers.
 	Recursive bool `json:"recursive"`
 }
+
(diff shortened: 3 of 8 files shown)


manager/manager.go
```diff
@@ -640,6 +643,27 @@ func (m *manager) Exists(containerName string) bool {
...
+	conts, err := m.getRequestedContainers(containerName, options)
+	if err != nil {
+		return nil, err
+	}
+	if len(conts) != 1 {
+		return nil, fmt.Errorf("Expected the request to match only one container")
+	}
+	// TODO(rjnagal): handle count? Only if we can do count by type (eg. top 5 cpu users)
+	ps := []v2.ProcessInfo{}
+	for _, cont := range conts {
```
vmarmol: nit: it's only one do we need the loop?

info/v2/container.go
```diff
@@ -166,3 +166,16 @@ type RequestOptions struct {
...
 	Recursive bool `json:"recursive"`
 }
+
+type ProcessInfo struct {
+	User        string `json:"user"`
+	Pid         int    `json:"pid"`
+	Ppid        int    `json:"parent_pid"`
+	StartTime   string `json:"start_time"`
+	Pcpu        string `json:"percent_cpu"`
+	Rss         string `json:"rss"`
```
vmarmol: nit: RSS and VirtualSize

info/v2/container.go
```diff
@@ -166,3 +166,16 @@ type RequestOptions struct {
 	// Whether to include stats for child subcontainers.
 	Recursive bool `json:"recursive"`
 }
+
+type ProcessInfo struct {
+	User        string `json:"user"`
+	Pid         int    `json:"pid"`
+	Ppid        int    `json:"parent_pid"`
+	StartTime   string `json:"start_time"`
+	Pcpu        string `json:"percent_cpu"`
```
vmarmol: nit: PercentCPU

container/raw/handler.go
```diff
@@ -336,6 +336,10 @@ func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
 	return stats, nil
 }
 
+func (self *rawContainerHandler) GetProcesses() ([]int, error) {
```
vmarmol: Do we need this one? Why not ListProcesses()?

vmarmol:
+1 to adding to UI/API and only querying on demand

info/v2/container.go
```diff
@@ -166,3 +166,16 @@ type RequestOptions struct {
 	// Whether to include stats for child subcontainers.
 	Recursive bool `json:"recursive"`
 }
+
+type ProcessInfo struct {
+	User        string `json:"user"`
+	Pid         int    `json:"pid"`
+	Ppid        int    `json:"parent_pid"`
+	StartTime   string `json:"start_time"`
```
vmarmol: Should we parse these values and get a more concrete API? Happy if we add a TODO :)

rjnagal: I just kept the original strings, as I would have to convert them back to put on the UI :)
will update.

rjnagal: Its actually convoluted to change the time fields back as the output changes based on the duration:

api/v2.0/ps/2ff16760cda11e8568589a9d73fe7e833b83ca3a8f55537157f7ff52195bad70?type=docker
[{"user":"root","pid":7003,"parent_pid":20768,"start_time":"May06","percent_cpu":"0.0","rss":"424","virtual_size":"7308","status":"Ss","running_time":"00:00:00","cmd":"stress"},{"user":"root","pid":7021,"parent_pid":7003,"start_time":"May06","percent_cpu":"99.5","rss":"262284","virtual_size":"269456","status":"R","running_time":"4-23:08:51","cmd":"stress"}]

start time can be May06, 20:11, or 2014. I think we can use a different option to print the whole time, but then that will be hard to condense back to be top like :/

vmarmol: Should be fine for now then :)

manager/manager.go
rjnagal: loop was simpler because conts is a map and containerName isn't necessarily the key.

vmarmol: aaah I didn't realize it was a map.

info/v2/container.go
rjnagal: done

container/raw/handler.go
rjnagal: leftover from before I realized that we already had a ListProcesses() :) removed.

info/v2/container.go
rjnagal: done

api/versions.go
```diff
@@ -391,6 +392,17 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 		return writeResult(fi, w)
 	case eventsApi:
 		return handleEventRequest(request, m, w, r)
+	case psApi:
```
vmarmol: Also add to SupportedRequestTypes(). Realized we can make this interface better and less manual :P

vmarmol:
LGTM, just the nit. And run tests please :) since Travis is being mean.

api/versions.go
rjnagal: done

vmarmol:
LGTM, thanks @rjnagal!

PR #709, opened 2015-05-12
Add process information to the UI.

rjnagal:
For root, we report all processes. Process stats are refreshed every minute.


vmarmol:
Don't we need to add the gCharts mapping in static.go?

rjnagal:
I added it to containers_js.go instead:

const containersJs = gchartsJs + `
function humanize(num, size, units) {...

We still just pull containers_js.  I just separated them out for readability.

vmarmol:
Ah kk SGTM and thank you for splitting them out! It was very hard to read how we had it before (and impossible to diff...)

LGTM

PR #711, opened 2015-05-13
Add the support for redis in cAdvisor's storage

superzhaoyy:
provide redis driver for cAdvisor's storage

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -106,6 +106,14 @@
 			"Rev": "09334c56c63bab2cd6c4ccab924d89e2419a361f"
 		},
 		{
+			"ImportPath": "github.com/garyburd/redigo/internal",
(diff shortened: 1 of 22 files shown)


vmarmol:
ok to test

vmarmol:
One thing I noticed, is this just always over-writing the same key? Such that we only store one value? I'm not as familiar with Redis so I apologize for the noob question :)

superzhaoyy:
hi@vmarmol .
you can pass some additional flags to cAdvisor telling it what's the key when you run a cAdvisor.
when a cAdvisor is running ,the key is unique,we use a list to store the value.
At set intervals, we add the newest values to the list every time .you can use redis's "LRANGE" to get any range of the list. It'sflexible.

for example：
redis> LRANGE cadvisor 0 10      // we can get ten of the latest data

vmarmol:
Thanks for explaining @superzhaoyy!

LGTM, merging.

PR #713, opened 2015-05-15
Register factories in Start() instead of New().

vmarmol:
This will allow for delayed registration for client users (e.g.: Kubelet).

/cc @rjnagal @dchen1107


dchen1107:
LGTM

PR #715, opened 2015-05-19
Add docker status and images info to /docker page.

rjnagal:
Added the basics. Still needs better styling. But its good enough to go.


vmarmol:
LGTM

PR #716, opened 2015-05-19
Improve process table output.

rjnagal:
Use pretty prints, but maintain sorting capabilities.

### DIFF
--- a/pages/static/containers_js.go
+++ b/pages/static/containers_js.go
@@ -25,10 +25,9 @@ function humanize(num, size, units) {
 
 // Following the IEC naming convention
 function humanizeIEC(num) {
-        var ret = humanize(num, 1024, ["TiB", "GiB", "MiB", "KiB", "Bytes"]);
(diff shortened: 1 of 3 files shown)


pages/static/containers_js.go
```diff
@@ -28,7 +28,11 @@ function humanizeIEC(num) {
         var ret = humanize(num, 1024, ["TiB", "GiB", "MiB", "KiB", "Bytes"]);
 	return ret[0].toFixed(2) + " " + ret[1];
 }
-
+// IEC naming convention, with shorter abbreviations
```
vmarmol: Wanna just change the one above to B and use it for both? I don't thinkwe mind :)

rjnagal: ok, changed the original one :) I couldn't find any place on the UI where the numbers are low enough to show up as bytes - other than the ps output.

vmarmol:
LGTM, thanks @rjnagal!

PR #722, opened 2015-05-20
Make a presubmit script to run all checkswq

vmarmol:
This is the first step towards running these checks in Jenkins and not have to wait for Travis anymore...


build/presubmit.sh
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+./build/check_gofmt.sh .
```
vishh: How about a Makefile instead?

vmarmol: I thought about it, but didn't want to take the dependency and encourage other things to be placed in make (instead of the simple Go flow of today)

vmarmol: If our flow becomes more complex (like Heapster's build), then I can't complain about moving to make :)

rjnagal:
LGTM

build/presubmit.sh
vishh: Ack. An advantage with make is that you can remember a few build rule names
and not have to type the entire build, test and vet commands. But as you
say its not a big deal.

PR #726, opened 2015-05-21
Separate inotify handling into its own object.

vmarmol:
This makes it easier to ensure the error conditions are handled well and
that we don't leak watches.

Fixes #703

/cc @rjnagal

PR #728, opened 2015-05-21
cAdvisor 0.14.0

vmarmol:
Should wait for #726 to be merged.

/cc @rjnagal


vmarmol:
This is now good to go @rjnagal

PR #729, opened 2015-05-21
Report the last cgroup destroy event.

vmarmol:
/cc @rjnagal

### DIFF
--- a/container/raw/inotify_watcher.go
+++ b/container/raw/inotify_watcher.go
@@ -27,11 +27,8 @@ type InotifyWatcher struct {
 	// Underlying inotify watcher.
 	watcher *inotify.Watcher
 
-	// Containers being watched.
-	containersWatched map[string]bool
-
-	// Full cgroup paths being watched.
-	cgroupsWatched map[string]bool
+	// Map of containers being watched to cgroup paths watched for that container.
+	containersWatched map[string]map[string]bool
 
 	// Lock for all datastructure access.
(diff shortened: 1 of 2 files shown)


container/raw/inotify_watcher.go
```diff
@@ -55,46 +51,51 @@ func (iw *InotifyWatcher) AddWatch(containerName, dir string) (bool, error) {
...
 }
 
-// Remove watch from the specified directory. Returns if the container was already being watched.
+// Remove watch from the specified directory. Returns if this was the last watch on the specified container.
 func (iw *InotifyWatcher) RemoveWatch(containerName, dir string) (bool, error) {
 	iw.lock.Lock()
 	defer iw.lock.Unlock()
 
-	alreadyWatched := iw.containersWatched[containerName]
+	cgroupsWatched := iw.containersWatched[containerName]
```
rjnagal: We need to check if cgroupsWatched returns nil.

rjnagal:
I think we can redo inotifyWatcher logic to provide a smaller interface later. I am fine with the current code in the meantime.

container/raw/inotify_watcher.go
vmarmol: If it returns nil, all the operations we use here are handled (get, delete, and len). Went ahead and made it explicit though to make it clearer.

rjnagal:
LGTM

PR #730, opened 2015-05-21
Add support for parsing CPU speed from /proc/cpuinfo on Power (ppc64) systems

bpradipt:
Add support for parsing CPU speed from /proc/cpuinfo on Power (ppc64) systems

/proc/cpuinfo format varies between architectures. For example

On Intel the cpu speed is shown like the following
cpu MHz		: 3000

whereas on Power systems its shown like the following
clock		: 3000MHz

This patch enables support for /proc/cpuinfo parsing on Power systems

Signed-off-by: Pradipta Kr. Banerjee <bpradip@in.ibm.com>

### DIFF
--- a/manager/machine.go
+++ b/manager/machine.go
@@ -62,7 +62,12 @@ func getClockSpeed(procInfo []byte) (uint64, error) {
 	// Fall back to /proc/cpuinfo
 	matches := CpuClockSpeedMHz.FindSubmatch(procInfo)
 	if len(matches) != 2 {
-		return 0, fmt.Errorf("could not detect clock speed from output: %q", string(procInfo))
+		//Check if we are running on Power systems which have a different format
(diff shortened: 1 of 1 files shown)


rjnagal:
ok to test

manager/machine.go
```diff
@@ -60,6 +61,9 @@ func getClockSpeed(procInfo []byte) (uint64, error) {
 		return maxFreq, nil
 	}
 	// Fall back to /proc/cpuinfo
+	if strings.Contains(runtime.GOARCH, "ppc64") {
```
rjnagal: I would avoid runtime and just check the speed if the x86 test fails. It will make testing it easier with a fake procinfo file.

bpradipt: Thanks for the review. I'll send a new version removing the arch detection.

bpradipt:
@rjnagal, I have uploaded new version based on your suggestions. Can you please review.
Thanks

rjnagal:
LGTM

Thanks @bpradipt !

PR #739, opened 2015-05-27
Initial client library for statsd backend.

rjnagal:
Restricted to supporting just gauges now as counters or timing are not very useful for cAdvisor.


vmarmol:
LGTM

PR #744, opened 2015-06-01
Remove RecentStats() from all storage drivers except memory.

rjnagal:
We should probably make memory as a cache type rather than storage.
RecentStats() can then be removed from the storage interface.
Will try it out as a separate PR.


rjnagal:
fyi @mnuessler

vmarmol:
We started treating memory as a cache and not a storage driver so it should be easier to make the change :)

vmarmol:
LGTM, nice! removes lots of code.

rjnagal:
yeah, cache seems pretty straightforward. Might get to it today :)

PR #745, opened 2015-06-02
 Separate in-memory cache from storage drivers.

rjnagal:


cache/cache.go
```diff
@@ -0,0 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import info "github.com/google/cadvisor/info/v1"
+
+type StorageDriver interface {
```
vmarmol: WDYT of renaming this to Cache? It may be confusing to have two storage drivers.

vmarmol:
LGTM

cache/cache.go
rjnagal: fixed.

vmarmol:
LGTM, thanks @rjnagal!

PR #750, opened 2015-06-04
Make process listing work when cadvisor is running in docker.

rjnagal:
Use /rootfs/proc to build the process listing.

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -132,26 +132,22 @@ func (c *containerData) getCgroupPath(cgroups string) (string, error) {
 	return string(matches[1]), nil
 }
 
-func (c *containerData) GetProcessList() ([]v2.ProcessInfo, error) {
+func (c *containerData) GetProcessList(cadvisorContainer string, inHostNamespace bool) ([]v2.ProcessInfo, error) {
 	// report all processes for root.
 	isRoot := c.info.Name == "/"
-	pidMap := map[int]bool{}
-	if !isRoot {
-		pids, err := c.handler.ListProcesses(container.ListSelf)
-		if err != nil {
-			return nil, err
-		}
-		for _, pid := range pids {
-			pidMap[pid] = true
-		}
-	}
 	// TODO(rjnagal): Take format as an option?
 	format := "user,pid,ppid,stime,pcpu,pmem,rss,vsz,stat,time,comm,cgroup"
-	args := []string{"-e", "-o", format}
+	args := []string{}
+	command := "ps"
+	if !inHostNamespace {
+		command = "/usr/sbin/chroot"
+		args = append(args, "/rootfs", "ps")
+	}
+	args = append(args, "-e", "-o", format)
 	expectedFields := 12
-	out, err := exec.Command("ps", args...).Output()
+	out, err := exec.Command(command, args...).Output()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


manager/container.go
```diff
@@ -132,7 +132,7 @@ func (c *containerData) getCgroupPath(cgroups string) (string, error) {
 	return string(matches[1]), nil
 }
 
-func (c *containerData) GetProcessList() ([]v2.ProcessInfo, error) {
+func (c *containerData) GetProcessList(inRootContainer bool) ([]v2.ProcessInfo, error) {
 	// report all processes for root.
 	isRoot := c.info.Name == "/"
 	pidMap := map[int]bool{}
```
vmarmol: I think we can remove the pid map now?

manager/container.go
```diff
@@ -187,15 +193,21 @@ func (c *containerData) GetProcessList() ([]v2.ProcessInfo, error) {
 		if err != nil {
 			return nil, fmt.Errorf("invalid virtual size %q: %v", fields[7], err)
 		}
+		cgroup, err := c.getCgroupPath(fields[11])
+		if err != nil {
+			return nil, fmt.Errorf("could not parse cgroup path from %q: %v", fields[10], err)
+		}
+		// Remove the ps command we just ran from cadvisor container.
+		// Not necessary, but makes the cadvisor page looks cleaner.
```
vmarmol: nit: s/looks/look/

manager/container.go
```diff
@@ -187,15 +193,21 @@ func (c *containerData) GetProcessList() ([]v2.ProcessInfo, error) {
...
+		if !inRootContainer && c.info.Name == cgroup && fields[10] == "ps" {
```
vmarmol: the ps was run in the cAdvisor container, not the container we're examining right now no?

manager/manager.go
```diff
@@ -671,7 +671,7 @@ func (m *manager) GetProcessList(containerName string, options v2.RequestOptions
 	// TODO(rjnagal): handle count? Only if we can do count by type (eg. top 5 cpu users)
 	ps := []v2.ProcessInfo{}
 	for _, cont := range conts {
-		ps, err = cont.GetProcessList()
+		ps, err = cont.GetProcessList(m.cadvisorContainer == "/")
```
vmarmol: Sadly I don't think this will work since cAdvisor may not be running inside namespaces (like in Kubernetes). We should check the mount/pid namespace of us vs init?

vmarmol:
Niiice :)

rjnagal:
Fixed the namespace detection logic to look for /rootfs/proc instead.

manager/manager.go
```diff
@@ -130,12 +131,20 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs) (Manager, error)
 	if err != nil {
 		return nil, err
 	}
+
+	// If cAdvisor was started with host's rootfs mounted, assume that its running
+	// in its own namespaces.
+	inHostNamespace := false
+	if _, err := os.Stat("/rootfs/proc"); os.IsNotExist(err) {
```
vmarmol: We should probably make `/rootfs`a flag eventually, fine for now/

vmarmol:
LGTM, thanks @rjnagal!

PR #764, opened 2015-06-09
Only show network statistics for containers with bridge network

AnanyaKumar:
Victor mentioned that there was a 1 line modification that solves the problem, but I couldn't find it. I'm still very new to the cAdvisor code base and to Docker - it took me pretty long to figure out where the specs were stored, and how to extract configuration information from Docker, so I certainly could be doing this in a sub-optimal way. Let me know!

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -167,7 +167,7 @@ func libcontainerConfigToContainerSpec(config *libcontainerConfigs.Config, mi *i
 	}
 	spec.Cpu.Mask = utils.FixCpuMask(config.Cgroups.CpusetCpus, mi.NumCores)
 
-	spec.HasNetwork = true
+	spec.HasNetwork = len(config.Networks) > 0
 	spec.HasDiskIo = true
 
 	return spec


AnanyaKumar:
I authored this commit.

AnanyaKumar:
Another question: should I add test cases for the code? I could use a fake for the dockerContainerHandler and examine the returned spec.

I've tested the code on the GUI (the network tab does not show up when a docker container is run with --net=host). What's the cAdvisor policy for testing small changes?

container/docker/handler.go
```diff
@@ -167,7 +167,8 @@ func libcontainerConfigToContainerSpec(config *libcontainerConfigs.Config, mi *i
 	}
 	spec.Cpu.Mask = utils.FixCpuMask(config.Cgroups.CpusetCpus, mi.NumCores)
 
-	spec.HasNetwork = true
+	ctnr, _ := self.client.InspectContainer(self.id)
+	spec.HasNetwork = ctnr.HostConfig.NetworkMode == "bridge"
```
vmarmol: I wonder whether we can get this from the libcontainer config?

vmarmol:
@AnanyaKumar sorry, didn't mean to make that a challenge :D I just meant that it would be a small change. What you're doing is along the lines of what I had in mind.

vmarmol:
@AnanyaKumar if there is a simple-ish way to add a test for this then I would certainly encourage it. That is pretty hard today unfortunately...

AnanyaKumar:
I checked libcontainer config (https://github.com/docker/libcontainer/blob/master/configs/config.go), but it didn't seem like it was keeping track of the HostConfig or the NetworkMode. I also logged libcontainer.config.Networks and libcontainer.config.SystemProperties and they did not seem to contain information about the NetworkMode. However, I might have missed something!

However, I realized that there's an issue with my code. The code I inserted has nothing to do with libcontainerConfig and so it should not be in the function "libcontainerConfigToContainerSpec". I'll extract it to GetSpec and send a revised PR.

AnanyaKumar:
I'm using Docker.runconfig's method to check if the network mode is a bridge - I think this is better because the string representation of "bridge" could potentially change.

container/docker/handler.go
```diff
@@ -184,6 +182,10 @@ func (self *dockerContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	}
 
 	spec := libcontainerConfigToContainerSpec(libcontainerConfig, mi)
+
+	ctnr, _ := self.client.InspectContainer(self.id)
```
vmarmol: We should check the error like we do above.

container/docker/handler.go
```diff
@@ -184,6 +182,10 @@ func (self *dockerContainerHandler) GetSpec() (info.ContainerSpec, error) {
...
+	spec.HasNetwork = runconfig.NetworkMode(ctnr.HostConfig.NetworkMode).IsBridge()
```
vmarmol: Nice!

Another possible option is to check if the libcontainer config has an empty [Network](https://github.com/docker/libcontainer/blob/master/configs/config.go#L99). This would save us the Docker inspect call in every `GetSpec()` which would be nice. If that's possible, let's do that :)

vmarmol: We can also check the [type](https://github.com/docker/libcontainer/blob/master/configs/network.go#L9) on the Network config.

AnanyaKumar:
Checking the size of the Network attribute in libContainerConfig was a good idea! I logged the output of libContainerConfig when net=host was set to true and when it wasn't, and checked for the differences in output, but I seem to have missed the Network attribute :(

vmarmol:
No worries :) Change LGTM, but it looks like you included a few other commits. A rebase gone wrong maybe?

AnanyaKumar:
I'll push a fix - thanks @rjnagal for helping me out!

AnanyaKumar:
Ok, I've removed the extra commits, and squashed my commits!

rjnagal:
I signed it

vmarmol:
LGTM, thanks @AnanyaKumar!

AnanyaKumar:
@rjnagal Can/should this PR be merged in? Just pinging you so that we don't forget about the PR :)

rjnagal:
Thanks for the reminder! I thought Victor merged this already.

PR #765, opened 2015-06-10
Set Content-Type on static files.

### LINKED ISSUE #763: CSS Broken in Newest Chrome
I *think* that we're not sending the CSS files with the proper MIME type and the newest Chrome is being more strict.

screenshot 2015-06-09 at 10 53 29 am

vmarmol:
Fixes #763

/cc @rjnagal


rjnagal:
LGTM

PR #767, opened 2015-06-10
Remove stats from cache when container is destroyed

vmarmol:
/cc @rjnagal


vmarmol:
Verified this fixes the leak! Will cut a 0.15.1 with this once we merge.

rjnagal:
wow, this is embarrassing :)

LGTM

PR #769, opened 2015-06-10
Create package cloudinfo

AnanyaKumar:
Started implementing the cloudinfo package. This is my first time writing production Go code, so I'm sending a short (50 line) diff for feedback. Thanks!

### DIFF
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -115,17 +115,17 @@ type NetInfo struct {
 type CloudProvider string
 
 const (
-	GCE              CloudProvider = "GCE"
-	AWS                            = "AWS"
-	BAREMETAL                      = "BAREMETAL"
-	UNKNOWN_PROVIDER               = "UNKNOWN"
+	GCE            CloudProvider = "GCE"
+	AWS                          = "AWS"
+	Baremetal                    = "Baremetal"
+	UnkownProvider               = "Unknown"
 )
 
 type InstanceType string
 
 const (
-	NO_INSTANCE      InstanceType = "NONE"
-	UNKNOWN_INSTANCE              = "UNKNOWN"
+	NoInstance      InstanceType = "None"
+	UnknownInstance              = "Uknown"
 )
 
 type MachineInfo struct {
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -112,6 +112,22 @@ type NetInfo struct {
 	Mtu int64 `json:"mtu"`
 }
 
+type CloudProvider string
+
+const (
+	GCE              CloudProvider = "GCE"
+	AWS                            = "AWS"
+	BAREMETAL                      = "BAREMETAL"
+	UNKNOWN_PROVIDER               = "UNKNOWN"
+)
+
+type InstanceType string
+
+const (
+	NO_INSTANCE      InstanceType = "NONE"
+	UNKNOWN_INSTANCE              = "UNKNOWN"
+)
+
 type MachineInfo struct {
 	// The number of cores in this machine.
 	NumCores int `json:"num_cores"`
@@ -143,6 +159,12 @@ type MachineInfo struct {
 	// Machine Topology
 	// Describes cpu/memory layout and hierarchy.
 	Topology []Node `json:"topology"`
+
+	// Cloud provider the machine belongs to.
+	CloudProvider CloudProvider `json:"cloud_provider"`
+
+	// Type of cloud instance (e.g. GCE standard) the machine is.
+	InstanceType InstanceType `json:"instance_type"`
 }
 
 type VersionInfo struct {
--- a/utils/cloudinfo/cloudinfo.go
+++ b/utils/cloudinfo/cloudinfo.go
@@ -1,0 +1,87 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Get information about the cloud provider (if any) cAdvisor is running on.
+
+package cloudinfo
+
+import (
+	info "github.com/google/cadvisor/info/v1"
+)
+
+type CloudInfo interface {
+	GetCloudProvider() info.CloudProvider
+	GetInstanceType() info.InstanceType
+}
+
+type realCloudInfo struct {
+	cloudProvider info.CloudProvider
+	instanceType  info.InstanceType
+}
+
+func NewRealCloudInfo() CloudInfo {
+	cloudProvider := detectCloudProvider()
+	instanceType := detectInstanceType(cloudProvider)
+	return &realCloudInfo{
+		cloudProvider: cloudProvider,
+		instanceType:  instanceType,
+	}
+}
+
+func (self *realCloudInfo) GetCloudProvider() info.CloudProvider {
+	return self.cloudProvider
+}
+
+func (self *realCloudInfo) GetInstanceType() info.InstanceType {
+	return self.instanceType
+}
+
+func detectCloudProvider() info.CloudProvider {
+	switch {
+	case inGCE():
+		return info.GCE
+	case inAWS():
+		return info.AWS
+	case inBareMetal():
+		return info.BAREMETAL
+	}
+	return info.UNKNOWN_PROVIDER
+}
+
+func detectInstanceType(cloudProvider info.CloudProvider) info.InstanceType {
+	switch cloudProvider {
+	case info.GCE:
+		return getGceInstanceType()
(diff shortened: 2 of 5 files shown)


AnanyaKumar:
I authored these commits.

vishh:
ok to test

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+)
+
+type CloudProvider int
+
+const (
+	GCE CloudProvider = iota
+	AWS
+	BAREMETAL
+	OTHER
```
vishh: Could this be "unknown" instead?

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,56 @@
...
+	case inAWS():
+		return AWS
+	case inBareMetal():
+		return BAREMETAL
+	}
+	return OTHER
+}
+
+func inGCE() bool {
+	_, err := http.Get("http://metadata.google.internal/" +
```
vishh: nit: The line wrap isn't necessary.

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,56 @@
...
+type CloudProvider int
+
+const (
+	GCE CloudProvider = iota
+	AWS
+	BAREMETAL
+	OTHER
+)
+
+func GetCloudProvider() CloudProvider {
```
vishh: Consider creating an interface for this method. It will be useful for unit testing.

vishh:
@AnanyaKumar: This PR seems to be mostly fine.

utils/cloudinfo/cloudinfo.go
AnanyaKumar: Done.

utils/cloudinfo/cloudinfo.go
AnanyaKumar: Removed.

utils/cloudinfo/cloudinfo.go
AnanyaKumar: Added.

AnanyaKumar:
Thanks for your feedback @vishh! I've made the changes, and I'll send a PR with more changes later on :)

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,96 @@
...
+	GetInstanceType(CloudProvider) (string, error)
+}
+
+type realCloudInfo struct{}
+
+func NewRealCloudInfo() (CloudInfo, error) {
+	return &realCloudInfo{}, nil
+}
+
+func (self *realCloudInfo) GetCloudProvider() (CloudProvider, error) {
```
rjnagal: I'd suggest we split this method into two:

a private detect() method that detects which cloudprovider the node is running on. This can be called once in NewRealCloudInfo() and we can save the result as this isn't going to change across the machine's lifetime.

The public part of the interface should just return the detected type rather than calling detection logic on every invocation. We can also cache instance type as that's mostly the same detect call.

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,96 @@
...
+		return GCE, nil
+	case inAWS():
+		return AWS, nil
+	case inBareMetal():
+		return BAREMETAL, nil
+	}
+	return UNKNOWN, nil
+}
+
+func (self *realCloudInfo) GetInstanceType(cp CloudProvider) (string, error) {
```
rjnagal: If we save cloudprovider and instance type internally, we don't need to accept cloudprovider as a parameter.

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,96 @@
...
+	switch cp {
+	case GCE:
+		return getGceInstanceType()
+	case AWS:
+		return getAwsInstanceType()
+	}
+	return NOINSTANCE, nil
+}
+
+func inGCE() bool {
```
rjnagal: Its likely we'd would expand cloudprovider utils to return more than just instance type. I'd suggest moving cloudprovider logic to its own file for supported clouds (gce for now). The default/unimplemented ones can stay here.

utils/cloudinfo/cloudinfo.go
AnanyaKumar: Great idea! Changed the code.

utils/cloudinfo/cloudinfo.go
AnanyaKumar: True!

utils/cloudinfo/cloudinfo.go
AnanyaKumar: That's true - changed!

AnanyaKumar:
The feature has been implemented. Couple of notes:
1. I have tested this on my local machine and it works as expected. I'm in the process of testing it in GCE (I'm installing cAdvisor on a GCE instance right now)
2. Errors were, for the most part, removed. If there was an error in getting the cloud provider, then the cloud provider will be set to "UNKNOWN", so the errors are contained in the return value.
3. I've changed the type of the constants to strings, so that they're more readable in the API output.

Let me know if you have any feedback!

info/v1/machine.go
```diff
@@ -14,6 +14,10 @@
 
 package v1
 
+import (
+	"github.com/google/cadvisor/utils/cloudinfo"
+)
```
rjnagal: I'd suggest moving the datastructures reported in the API  (CloudProvider, InstanceType) here. That way, consumers of the api don't have to include utils/cloudinfo.

rjnagal:
minor code movement.
LGTM, otherwise.

AnanyaKumar:
I didn't detect the instance type correctly in GCE in the previous commit. I've corrected this, and tested on an n1-standard-1 instance (it outputs the correct instance type).

info/v1/machine.go
AnanyaKumar: Sorry, I didn't understand why the data structures should go into machine.go Why shouldn't consumers of the API have to include utils/cloudinfo? Aren't the methods that retrieve information about the cloud provider in utils/cloudinfo anyway? Thanks!

AnanyaKumar:
@rjnagal Does this look fine?

rjnagal:
I think this is fine.
Can you squash the commits?

AnanyaKumar:
@rjnagal Awesome, thanks for your feedback. I've squashed the commits!

info/v1/machine.go
```diff
@@ -112,6 +112,22 @@ type NetInfo struct {
 	Mtu int64 `json:"mtu"`
 }
 
+type CloudProvider string
+
+const (
+	GCE              CloudProvider = "GCE"
+	AWS                            = "AWS"
+	BAREMETAL                      = "BAREMETAL"
```
vmarmol: nit: Go style is camelcase, this is fine for the above, but for the others it should be Baremetal and UnknownProvider

info/v1/machine.go
```diff
@@ -112,6 +112,22 @@ type NetInfo struct {
...
+	UNKNOWN_PROVIDER               = "UNKNOWN"
+)
+
+type InstanceType string
+
+const (
+	NO_INSTANCE      InstanceType = "NONE"
```
vmarmol: nit: Same here

info/v1/machine.go
```diff
@@ -143,6 +159,12 @@ type MachineInfo struct {
 	// Machine Topology
 	// Describes cpu/memory layout and hierarchy.
 	Topology []Node `json:"topology"`
+
+	// Cloud provider the machine belongs to.
+	CloudProvider CloudProvider `json:"cloud_provider"`
+
+	// Type of cloud instance (e.g. GCE standard) the machine is.
+	InstanceType InstanceType `json:"instance_type"`
```
vmarmol: nit: omitempty that way it is not included when it is not used. Same for CloudProvider:

```
`json:"instance_type,omitempty"`
```

vmarmol: Although I guess we have Unknown which will make this never happen :P ignore...

utils/cloudinfo/cloudinfo.go
```diff
@@ -0,0 +1,87 @@
...
+	GetCloudProvider() info.CloudProvider
+	GetInstanceType() info.InstanceType
+}
+
+type realCloudInfo struct {
+	cloudProvider info.CloudProvider
+	instanceType  info.InstanceType
+}
+
+func NewRealCloudInfo() CloudInfo {
```
vmarmol: nit: Just `NewCloudInfo()`

utils/cloudinfo/gce.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"io"
+	"net/http"
+	"strings"
+
+	info "github.com/google/cadvisor/info/v1"
+)
+
+func inGCE() bool {
+	_, err := http.Get("http://metadata.google.internal/computeMetadata/v1/instance/machine-type")
```
vmarmol: I recommend using: https://godoc.org/google.golang.org/cloud/compute/metadata

It's pretty simple code, but useful to use a library by the authors/maintainers :)

AnanyaKumar:
@vmarmol Sorry I missed your comments until you spoke to me IRL a couple of days ago. I've pushed a diff based on the corrections to the code that you suggested. I'll respond faster if you have any follow ups!

vmarmol:
LGTM, thanks @AnanyaKumar!

PR #775, opened 2015-06-19
Add config file for custom metrics and pull config file into a suitable structure

anushree-n:


anushree-n:
I signed it.

collector/collector_manager_test.go
```diff
@@ -14,6 +14,7 @@
 
 package collector
 
+/*
```
vmarmol: Why comment out?

collector/collector.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
vmarmol: I'd recommend having this in `collector/nginx` so we can do: `nginx.NewCollector()`

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+
+package collector
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"time"
+)
+
+type Collector struct {
```
vmarmol: Rename to NginxCollector and leave the other as CollectorInterface

collector/collector.go
```diff
@@ -0,0 +1,37 @@
...
+
+package collector
+
+import (
+	"time"
+
+	"github.com/google/cadvisor/info/v2"
+)
+
+func NewNginxCollector() (*Collector, error) {
```
rjnagal: I think we don't need Nginx specific collector - just make name as an arg to NewCollector(). This should be the regular config-based collector generator. nginx just happens to be one of the configs.

collector/set_config.go
rjnagal: This is supposed to be a generic collector constructed from a config.

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+	name               string
+	configFile         Config
+	nextCollectionTime time.Time
+	err                error
+}
```
rjnagal: Document the fields. Add comments about what each field means.
eg. Is err the latest collection error or only set when we fail to parse the config?

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+
+type Config struct {
+	Endpoint      string         `json:"endpoint"`
+	MetricsConfig []metricConfig `json:"metricsConfig"`
+}
+
+type metricConfig struct {
+	Name             string `json:"name"`
+	MetricType       string `json:"metricType"`
```
rjnagal: define enum for metric types - so the users have a way to know what we support.

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+	Units            string `json:"units"`
+	PollingFrequency string `json:"pollingFrequency"`
```
rjnagal: comments? polling frequency in seconds?

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func SetCollectorConfig(collector *Collector, file string) error {
+	configFile, err := ioutil.ReadFile(file)
+	if err != nil {
+		collector.err = err
+	} else {
+		var configInJSON Config
+
+		err1 := json.Unmarshal(configFile, &configInJSON)
```
rjnagal: you can reuse err here, instead of err1.

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+		if err1 != nil {
+			collector.err = err1
```
rjnagal: return directly from here and avoid the else part.

Also prepend some more information to error message - like printing out the config read.

collector/set_config.go
```diff
@@ -0,0 +1,60 @@
...
+		} else {
+			collector.err = nil
+			collector.configFile = configInJSON
+		}
+	}
+
+	return collector.err
```
rjnagal: How about passing the config to constructor and failing construction if the config doesn't parse? Would it be cleaner than storing the error state and holding on to a defunct collector?

anushree-n:
Thank you @vmarmol  and @rjnagal ! I will make the necessary modifications.

collector/collector_manager_test.go
anushree-n: There was a name collision between the Collector struct and the Collector interface(temporarily renamed as CollectorInterface). Once we finalize on the names for these, we can make the necessary changes in all places(eg manager.go uses Collector as an interface) and put back the test.

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"time"
+
+	"github.com/google/cadvisor/info/v2"
+)
+
+type Collector struct {
```
vmarmol: I'd call this `GenericCollector` or `GeneralCollector` and leave the interface as `Collector` :)

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+	//name of the collector
+	name string
+
+	//holds information extracted from the config file for a collector
+	configFile Config
+
+	//time at which metrics need to be collected next
+	nextCollectionTime time.Time
```
vmarmol: It looks like we can remove this one as you suggested.

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+
+	//error (if any) during metrics collection
+	err error
```
vmarmol: You probably don't need to keep track of the last error, it will just be returned by the collector.

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+type Config struct {
+	//the endpoint to hit to scrape metrics
+	Endpoint string `json:"endpoint"`
+
+	//holds information about different metrics that can be collected
+	MetricsConfig []metricConfig `json:"metricsConfig"`
+}
+
+// metricConfig holds information extracted from the config file about a metric
+type metricConfig struct {
```
vmarmol: This should probably be public too.

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+	//the name of the metric
+	Name string `json:"name"`
+
+	//enum type for the metric type
+	MetricType MetricType `json:"metricType"`
+
+	//data type of the metric (eg: integer, string)
```
vmarmol: Units are more like: bytes, seconds, nanoseconds per second, etc.

collector/collector.go
```diff
@@ -0,0 +1,100 @@
...
+	Units string `json:"units"`
+
+	//the frequency at which the metric should be collected (in seconds)
+	PollingFrequency string `json:"pollingFrequency"`
```
vmarmol: We should represent this as a `time.Duration` or an integer and say it is in seconds.

vmarmol:
Closing in favor of #787

PR #776, opened 2015-06-20
Add RegisterAllCollectors in collector_manager.go

anushree-n:
- Add a config file for custom metrics
- Read config file into a suitable structure
- Add a test for extraction of config file (set_config_test)
- Add the function 'RegisterAllCollectors' in collector_manager.go
- Add a test for collector_manager in collector_manager_test

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -1,0 +1,23 @@
+package collector
+
+import (
+	"time"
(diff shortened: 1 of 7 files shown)


anushree-n:
I signed it.

Issue #688, opened 2015-05-06
Set docker container name and meta data as prometheus labels


discordianfish:
cadvisor supports exposing metrics for prometheus consumption. Right now it only sets the container ID as a label but it would be useful to also set the container name and potential meta data[1] as labels.

1) https://docs.docker.com/userguide/labels-custom-metadata/

MrMMorris:
yes please! I need to be able to differentiate containers as I don't name them and docker labels seems ot be the best way to do it :smile:

timpoultney:
+1 here.

jimmidyson:
Think docker container name & labels are now in the prometheus exposed metrics. Can this be closed?

discordianfish:
Awesome, then it can be closed!

Colstuwjx:
I have some docker envs which need to inject as prom labels, how to achieve that?
It seems that we need a more flexible way to determine prom labels...

Colstuwjx:
I have been checked the [source code](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L660), and found that cadvisor will set some default labels, and envs are already exist.

BTW,  I also found that there is a `ContainerSpec` struct to determine the container metadata, but the env will be [whitelist](https://github.com/google/cadvisor/blob/359ef5b1008bb0497cbcb843829356782ab8875d/info/v2/container.go#L81)? How to setup these configs? 

I cannot find any documentation about this, and I suggest `docker inspect` result should have a simple  way to inject as part of `ContainerSpect`.

Issue #546, opened 2015-02-27
Grouping by App (marathon) 


samek:
Hi,
now that i have cadvisor up and running through marathon I'm getting data into the influxdb. 

The problem that I'm facing is that I don't know which app is under docker name.

Let me try to explain: 
When marathon starts a docker it gives him a name for example 
![screen shot 2015-02-27 at 15 46 58](https://cloud.githubusercontent.com/assets/125693/6414464/e739a142-be97-11e4-9f09-0d5f2677f4b6.png)

for cadvisor it's named mesos-3deefa59-6981-4069-8c74-911aead8b396

As you can see in the screenshot I have 2 nginx images also started and they have completely different names. 
In the influxdb I cannot differentiate/group by apps since they are presented as mesos name. 

Now, Marathon passes some environment variables into the started dockers which could be used to group it. 

"MARATHON_APP_VERSION=2015-02-27T13:38:50.135Z",
"HOST=10.0.0.193",
"MESOS_TASK_ID=nginx2.f726a4c5-be85-11e4-82c2-56847afe9799",
"PORT=31005",
"PORTS=31005",
"PORT_80=31005",
"MARATHON_APP_ID=/nginx2",
"PORT0=31005",
"MESOS_SANDBOX=/mnt/mesos/sandbox",
"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
"NGINX_VERSION=1.7.10-1~wheezy"

those filelds are located in the docker dir config.json and should be accessible by cadvisor. 

Question is would it be wise to also sent MARATHON_APP_ID into the influxdb in order to group stats by app ?

Or How would this be approached correctly ?

vmarmol:
Today we have a concept of "namespaces" for container names and those come with a list of aliases used to address the container. Docker containers are under the "docker" namespace and its aliases are the Docker ID and name. This feature may be served by the "marathon" namespace and have the app ID as one of the aliases. This we will then push to the InfluxDB backend.

We probably don't have the bandwidth to tackle this anytime soon, but we will gladly take PRs towards that goal if you'd like :)

rjnagal:
If we did a generic change of exporting aliases to storage backend,
wouldn't it resolve this problem too?

vmarmol:
Yes, you need to do both (I *think* we export the aliases today, if not we should). Here the aliases they're referring to are not a Docker container name or ID, is that correct @samek?

rjnagal:
[...]

PR #780, opened 2015-06-23
Export image name and env variables as prometheus labels

Marmelatze:
I had the same issue as described in #546. The names from containers created with Mesos/Marathon are not telling something about the app this container is running. With this PR any env variable can be exported as prometheus label. I also added the current image a container is running for monitoring the performance of some images.

I hope this use case isn't to specific.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -93,6 +93,7 @@ func newDockerContainerHandler(
 	usesAufsDriver bool,
 	cgroupSubsystems *containerLibcontainer.CgroupSubsystems,
 	inHostNamespace bool,
+	exposedMetadata []string,
 ) (container.ContainerHandler, error) {
 	// Create the cgroup paths.
 	cgroupPaths := make(map[string]string, len(cgroupSubsystems.MountPoints))
@@ -140,6 +141,18 @@ func newDockerContainerHandler(
 	handler.labels = ctnr.Config.Labels
 	handler.image = ctnr.Config.Image
 	handler.networkMode = ctnr.HostConfig.NetworkMode
+
+	// split env vars to get metadata map.
+	if len(exposedMetadata) > 0 {
+		for _, envVar := range ctnr.Config.Env {
+			splits := strings.SplitN(envVar, "=", 2)
+			for _, exposedVar := range exposedMetadata {
+				if splits[0] == exposedVar {
+					handler.labels[strings.ToLower(exposedVar)] = splits[1]
+				}
+			}
+		}
+	}
 
 	return handler, nil
 }
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -493,6 +493,11 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 		baseLabelValues := []string{id, name, image}[:len(baseLabels)]
 
+		for labelKey, labelValue := range container.Spec.Labels {
+			baseLabels = append(baseLabels, labelKey)
+			baseLabelValues = append(baseLabelValues, labelValue)
+		}
+
 		// Container spec
 		desc := prometheus.NewDesc("container_start_time_seconds", "Start time of the container since unix epoch in seconds.", baseLabels, nil)
 		ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.CreationTime.Unix()), baseLabelValues...)
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -42,6 +42,8 @@ var DockerNamespace = "docker"
 var dockerRootDir = flag.String("docker_root", "/var/lib/docker", "Absolute path to the Docker state root directory (default: /var/lib/docker)")
 var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path to the Docker run directory (default: /var/run/docker)")
 
+var dockerMetadataEnvs = flag.String("docker_metadata_env", "", "Comma seperated list with names of env variables, which will be exported as metadata (default: empty)")
+
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
 func DockerStateDir() string {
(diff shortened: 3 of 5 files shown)


vmarmol:
ok to test

vmarmol:
@Marmelatze this looks good for the most part. I know image has been requested for a while. Env variables I'm not so sure we want to do however...what is the intended usecase? Just afraid that we're exposing data that is tricky to use and easy to abuse.

Marmelatze:
I have the same use case as the guy in #546. I have some Mesos/Marathon hosts and Mesos assigns some random id as container name (I expect Mesos/kubernetes and Mesos/Swarm to do the same).

In my dashboard (cadvisor on every host -> prometheus -> grafana) I can see that a container uses 100% CPU but I only can see the host its running on and the name (mesos-1234-5678...). To determine what application this container is running I have to SSH into that host and inspect the container. There I can see the mesos task-id, the marathon app name and version.

The reverse way is also time consuming: I have more than one instance of this app running and I want to know how these are performing (CPU, RAM, Network). So I have to check Marathon or Mesos to get a list of hosts, that app is deployed to. Than I have to ssh into these. List all containers and look which could be my app. Inspect these to be sure. Go back to the dashboard and filter for the names.

I know some sensitive data, like passwords and access tokens, are stored in env variables, so I made this optional and the exposed variables configurable. 

For Mesos/Marathon it would be:
`--docker_metadata_env=MESOS_TASK_ID,MARATHON_APP_ID,MARATHON_APP_VERSION`

container/docker/handler.go
```diff
@@ -71,6 +71,10 @@ type dockerContainerHandler struct {
 
 	// Metadata labels associated with the container.
 	labels map[string]string
+
+	image string
```
vmarmol: nit: add comments for these

info/v1/container.go
```diff
@@ -72,6 +72,10 @@ type ContainerReference struct {
 	// Namespace under which the aliases of a container are unique.
 	// An example of a namespace is "docker" for Docker containers.
 	Namespace string `json:"namespace,omitempty"`
+
+	Image string `json:"image,omitempty"`
```
vmarmol: These should not be in the reference but rather in the ContainerSpec above (and Metadata should be merged with labels I would think)

Marmelatze:
Thanks for the feedback. With all container labels exported as prometheus labels it would also fix #688

vmarmol:
LGTM, thanks @Marmelatze! Can you rebase and repush to kick the CI?

MrMMorris:
@Marmelatze thanks so much for taking this on. It is basically a requirement when using Prometheus with Consul if I want to populate graphs dynamically. Now I just have to LABEL my images and It will Just Work™ :grinning:

Marmelatze:
Rebased to current master and tests seem to pass

gtmtech:
+1 I have exactly the same usecase (mesos, classifying containers by appname for aggregation and averaging reasons across a cluster). Hopefully this can be merged ASAP

jimmidyson:
Would really like to see this merged, although only want the image name personally.

rjnagal:
@Marmelatze Let's split this up. I think image name makes a lot of sense. We can merge that quickly.

Do the containers in Mesos use labels - something more structured than env variables and intended for exporting? We can pick up runtime image labels too for identifying the apps.

jimmidyson:
Image name added in #869.

Marmelatze:
OK thanks, but unfortunately mesos doesn't seem set labels or something else. Just the env variables.

jzaefferer:
@Marmelatze could you rebase this to remove the changes already introduced by #869? Unless you've found an alternative solution by now; if so, could you share?

@rjnagal was the answer to your question satisfying? We've got the same usecase as @Marmelatze and would like to see this addressed.

MikeMichel:
+1 for rebase

koboltmarky:
+1 for rebase

jzaefferer:
(bad typo in previous comment, I *couldn't* figure it out)

Marmelatze:
Rebased to current master and removed the image name parts.

jzaefferer:
The commit message still mentions the image, but that should be trivial to fix when landing this.

MikeMichel:
@Marmelatze awesome, works perfect.

thx

metrics/prometheus.go
```diff
@@ -493,6 +493,11 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 		baseLabelValues := []string{id, name, image}[:len(baseLabels)]
 
+		for labelKey, labelValue := range container.Spec.Labels {
+			baseLabels = append(baseLabels, labelKey)
```
jimmidyson: `labelKey` will need to be sanitized as Prometheus labels need to match [this regexp](https://github.com/prometheus/client_golang/blob/8c899cd8208a366b6899d397e4bc3ced7ab073e3/prometheus/desc.go#L19).

You can copy the sanitize method from [here](https://github.com/prometheus/prometheus/blob/b964da4b75a589a6ded82becf9c456d1ac79b25d/util/strutil/strconv.go#L110-L114) & please add a test for this.

vishh:
@Marmelatze: Can you rebase?

dqminh:
@vishh @Marmelatze  im also interested to see this and ability to specify container labels as tags. Let me know if you want some help with this ( rebase / take over etc. )

deedubs:
Is anyone running this yet?

MikeMichel:
yes https://hub.docker.com/r/mikemichel/cadvisor/tags/

```
docker run --name cadvisor --volume=/:/rootfs:ro --volume=/var/run:/var/run:rw --volume=/sys:/sys
:ro --volume=/mnt/docker/:/var/lib/docker:ro --publish=8080:8080 --detach=true   mikemichel/cadvisor:metaenvs --docker_metadata_env=MESOS_TASK_ID,MARATHO
N_APP_ID,MARATHON_APP_VERSION
```

deedubs:
@rjnaga feels like the security concerns are minimized by only exporting whitelisted ENV vars no?

scalp42:
@Marmelatze any chance ? :rotating_light:

dqminh:
@vishh @rjnagal @jimmidyson  i like to use this patch. Should i take over and do a rebase on behalf of @Marmelatze ?

vishh:
Yeah. Feel free to work on top of existing patches.

jimmidyson:
Fixed by #780.

r0fls:
>Fixed by #780.

Well hey now, that's a circular reference! 😄  Do you mean #1023 ?

PR #783, opened 2015-06-25
Fix mirrored network stats

vmarmol:
/cc @rjnagal


rjnagal:
LGTM.

Can you look at test failure?

vmarmol:
@rjnagal needed to change the command a bit in the unit test. Fixed. Thanks for the review!

PR #788, opened 2015-06-26
Add a self-built canary cAdvisor.

vmarmol:
Much credit to @afein 

/cc @rjnagal

### DIFF
--- a/deploy/canary/Dockerfile
+++ b/deploy/canary/Dockerfile
@@ -1,0 +1,11 @@
+FROM golang:latest
+MAINTAINER vmarmol@google.com
+
+
+RUN apt-get install -y git
+RUN git clone https://github.com/google/cadvisor.git /go/src/github.com/google/cadvisor
+RUN go get github.com/tools/godep
+RUN cd /go/src/github.com/google/cadvisor && godep go build .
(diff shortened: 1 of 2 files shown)


vmarmol:
The goal is to have an Automated Build with an up-to-date cAdvisor. This is not the official release since it is ~600MB in size, but it'll allow people to test the latest features.

vishh:
Nice idea! :+1: LGTM

vmarmol:
It was @afein's idea :D

deploy/canary/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:latest
```
rjnagal: why did we change the tag to beta and have the dir deploy/canary?

rjnagal:
small naming nit, LGTM

deploy/canary/Dockerfile
vmarmol: Moved canary -> beta for builds we make manually. The automated build will be the canary. The thought is that users can try a fixed version without us having to make a release. We can probably replace a lot of the "try vmarmol/cadvisor" with this.

PR #793, opened 2015-07-01
Use watch instead of sh.

vmarmol:
Use of sh was making the test flaky and harder to read.

@AnanyaKumar @anushree-n could one of you take a look when you have time? This fixes the integration test.

### DIFF
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -249,7 +249,7 @@ func TestDockerContainerNetworkStats(t *testing.T) {
 	defer fm.Cleanup()
 
 	// Wait for the container to show up.
-	containerId := fm.Docker().RunBusybox("sh", "-c", "wget www.google.com && ping www.google.com")
+	containerId := fm.Docker().RunBusybox("watch", "-n1", "wget", "https://www.google.com/")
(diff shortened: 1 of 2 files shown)


vmarmol:
Failure is due to a different, but known flake. *sigh* I'd say to ignore it for now, but we should still get this PR in.

vmarmol:
We're now skipping the known flake. PATL @anushree-n

We don't have to wait for Travis as it runs the same things Jenkins already did.

anushree-n:
Looks good! Thanks!

AnanyaKumar:
Sorry I didn't comment on this - I was trying to figure the code out, and understand what the test was doing, but looks like Anushree got to it first! 

What happens if the first wget is executed after the asserts in lines 264 to 270? Would the test fail because the busy box container hasn't yet sent a packet? (Sorry I probably missed something)

vmarmol:
@AnanyaKumar yes it will most likely miss a packet and the test will fail. It is a flaky test unfortunately...

In practice it doesn't happen often. We need to fix most of the tests in this fixture for cases like this.

AnanyaKumar:
Thanks for the clarification - besides that it looked fine to me :)

PR #787, opened 2015-06-26
Collector Config

anushree-n:
Create a config : Includes endpoint and metrics metadata
Extract config information
Create nginx collector using the config

PR #802, opened 2015-07-06
Collection of custom metrics

anushree-n:
Based on #787 
Adds Implementation of Collect() method of Collector interface

### DIFF
--- a/collector/generic_collector.go
+++ b/collector/generic_collector.go
@@ -18,7 +18,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"regexp"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/cadvisor/info/v1"
@@ -92,6 +95,70 @@ func (collector *GenericCollector) Name() string {
 
 //Returns collected metrics and the next collection time of the collector
 func (collector *GenericCollector) Collect() (time.Time, []v1.Metric, error) {
-	//TO BE IMPLEMENTED
-	return time.Now(), nil, nil
+	minNextColTime := collector.configFile.MetricsConfig[0].PollingFrequency
+	for _, metricConfig := range collector.configFile.MetricsConfig {
+		if metricConfig.PollingFrequency < minNextColTime {
+			minNextColTime = metricConfig.PollingFrequency
+		}
+	}
+	currentTime := time.Now()
+	nextCollectionTime := currentTime.Add(time.Duration(minNextColTime * time.Second))
+
+	uri := collector.configFile.Endpoint
+	response, err := http.Get(uri)
+	if err != nil {
+		return nextCollectionTime, nil, err
+	}
+
+	defer response.Body.Close()
+
+	pageContent, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nextCollectionTime, nil, err
+	}
+
+	metrics := make([]v1.Metric, len(collector.configFile.MetricsConfig))
+	var errorSlice []error
+
+	for ind, metricConfig := range collector.configFile.MetricsConfig {
+		regex, err := regexp.Compile(metricConfig.Regex)
+		if err != nil {
+			return nextCollectionTime, nil, err
+		}
+
+		matchString := regex.FindStringSubmatch(string(pageContent))
+		if matchString != nil {
+			if metricConfig.Units == "float" {
+				regVal, err := strconv.ParseFloat(strings.TrimSpace(matchString[1]), 64)
+				if err != nil {
+					errorSlice = append(errorSlice, err)
+				}
+				metrics[ind].FloatPoints = []v1.FloatPoint{
+					{Value: regVal, Timestamp: currentTime},
+				}
+			} else if metricConfig.Units == "integer" || metricConfig.Units == "int" {
+				regVal, err := strconv.ParseInt(strings.TrimSpace(matchString[1]), 10, 64)
+				if err != nil {
+					errorSlice = append(errorSlice, err)
+				}
+				metrics[ind].IntPoints = []v1.IntPoint{
+					{Value: regVal, Timestamp: currentTime},
+				}
+
+			} else {
+				errorSlice = append(errorSlice, fmt.Errorf("Unexpected value of 'units' for metric '%v' in config ", metricConfig.Name))
+			}
+		} else {
+			errorSlice = append(errorSlice, fmt.Errorf("No match found for regexp: %v for metric '%v' in config", metricConfig.Regex, metricConfig.Name))
+		}
+
+		metrics[ind].Name = metricConfig.Name
+		if metricConfig.MetricType == "gauge" {
+			metrics[ind].Type = v1.MetricGauge
+		} else if metricConfig.MetricType == "counter" {
+			metrics[ind].Type = v1.MetricCumulative
+		}
+	}
+
+	return nextCollectionTime, metrics, compileErrors(errorSlice)
 }
--- a/collector/generic_collector_test.go
+++ b/collector/generic_collector_test.go
@@ -15,10 +15,14 @@
 package collector
 
 import (
+	"fmt"
 	"io/ioutil"
+	"net/http"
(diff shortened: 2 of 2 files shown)


anushree-n:
I signed it.

vmarmol:
Lets rebase this off of master since we merged the other PR

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+				regVal, _ := strconv.ParseInt(strings.Trim(matchString[1], " "), 10, 64)
+				metricValue := make([]v2.IntPoint, 1)
+				metricValue[0].Value = regVal
+				metricValue[0].Timestamp = currentTime
+				metrics[ind].IntPoints = metricValue
+				metrics[ind].FloatPoints = nil
+			}
+		} else {
+			return nextCollectionTime, nil, errors.New("No match found for metric regexp: " + metricConfig.Regex)
+		}
```
anushree-n: This can be skipped if we want to continue collection of other valid metrics.

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+		metrics[ind].Name = metricConfig.Name
+		if metricConfig.MetricType == "gauge" {
+			metrics[ind].Type = v2.MetricGauge
+		} else if metricConfig.MetricType == "counter" {
+			metrics[ind].Type = v2.MetricCumulative
+		}
+	}
+
+	return nextCollectionTime, metrics, nil
 }
```
anushree-n: Do we need to re-calculate nextCollectionTime here?

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+
+		metrics[ind].Name = metricConfig.Name
+		if metricConfig.MetricType == "gauge" {
+			metrics[ind].Type = v2.MetricGauge
+		} else if metricConfig.MetricType == "counter" {
+			metrics[ind].Type = v2.MetricCumulative
+		}
+	}
+
+	return nextCollectionTime, metrics, nil
```
anushree-n: Do we need to re-calculate nextCollectionTime here?

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
 
 //Returns collected metrics and the next collection time of the collector
 func (collector *GenericCollector) Collect() (time.Time, []v2.Metric, error) {
```
vmarmol: There are a few things we do per collection that we should be able to do once at initialization:
- get minPollingFrequency
- Compile regexps

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
-	//TO BE IMPLEMENTED
-	return time.Now(), nil, nil
+	minNextColTime := collector.configFile.MetricsConfig[0].PollingFrequency
```
vmarmol: nit: minPollingFrequency?

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+	currentTime := time.Now()
+	nextCollectionTime := currentTime.Add(time.Duration(minNextColTime * time.Second))
+
+	uri := collector.configFile.Endpoint
+	response, err := http.Get(uri)
+	if err != nil {
+		return nextCollectionTime, nil, err
+	}
+
+	pageContent, err := ioutil.ReadAll(response.Body)
```
vmarmol: We need to close the body:

```
defer response.Body.Close()
```

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+
+	metrics := make([]v2.Metric, len(collector.configFile.MetricsConfig))
+
+	for ind, metricConfig := range collector.configFile.MetricsConfig {
+		regex, err := regexp.Compile(metricConfig.Regex)
+		if err != nil {
+			return nextCollectionTime, nil, err
+		}
+
+		matchString := regex.FindStringSubmatch(string(pageContent[:]))
```
vmarmol: Do we need the extra slice copy?

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+		if matchString != nil {
+			if strings.Contains(matchString[1], ".") {
```
vmarmol: We should check the length of the match too

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+				regVal, _ := strconv.ParseFloat(strings.Trim(matchString[1], " "), 64)
```
vmarmol: Can use: `strings.TrimSpace()` here and below

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+				metricValue := make([]v2.FloatPoint, 1)
+				metricValue[0].Value = regVal
+				metricValue[0].Timestamp = currentTime
+				metrics[ind].FloatPoints = metricValue
+				metrics[ind].IntPoints = nil
```
vmarmol: We shouldn't need to do this, it's the default

collector/generic_collector.go
```diff
@@ -57,6 +62,63 @@ func (collector *GenericCollector) Name() string {
...
+		}
+
+		matchString := regex.FindStringSubmatch(string(pageContent[:]))
+		if matchString != nil {
+			if strings.Contains(matchString[1], ".") {
+				regVal, _ := strconv.ParseFloat(strings.Trim(matchString[1], " "), 64)
+				metricValue := make([]v2.FloatPoint, 1)
+				metricValue[0].Value = regVal
+				metricValue[0].Timestamp = currentTime
+				metrics[ind].FloatPoints = metricValue
```
vmarmol: Maybe simpler:

```
metrics[ind].FloatPoints = []v2.FloatPoint{
  {
    Value: regval,
    Timestamp: currentTime,
  }
}
```

collector/generic_collector.go
vmarmol: We should also check the return of the conversion. Here and below

collector/generic_collector.go
vmarmol: We should probably keep a slice of errors that we then compile together. We try to get all the metrics we can and only return errors at the end. We can use:

https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/util/errors/errors.go#L31

collector/generic_collector.go
vmarmol: What you have today is correct I think. We'll be closer to collecting every specified interval.

collector/generic_collector.go
anushree-n: Used this approach in #810

collector/generic_collector.go
```diff
@@ -57,6 +62,68 @@ func (collector *GenericCollector) Name() string {
...
+
+	for ind, metricConfig := range collector.configFile.MetricsConfig {
+		regex, err := regexp.Compile(metricConfig.Regex)
+		if err != nil {
+			return nextCollectionTime, nil, err
+		}
+
+		matchString := regex.FindStringSubmatch(string(pageContent))
+		if matchString != nil {
+			if strings.Contains(matchString[1], ".") {
```
rjnagal: Can't we rely on the config to know if we are expecting floats or ints?

vmarmol: +1

anushree-n: My concern was the way in which the metric would be expressed : int, integer, int32, number ... 
But yes, I think this is a better approach. We can use enum for metric units if necessary

rjnagal:
@anushree-n Can you rebase and squash?

anushree-n:
Done, @rjnagal  :)

rjnagal:
LGTM

PR #803, opened 2015-07-07
Lower threshold for "unlimited".

### LINKED ISSUE #796: cadvisor with no swap shows 8.00 EB memory limit /swap limit
If you don't have a swap file (as is the case on CoreOS by default),
8.00 EB is shown as the memory limit and swap limit.  

Clearly that's not right, I wish it was though! :-)

vmarmol:
Seems like some kernels report a value a bit lower than MaxInt64.

Fixes #796 

/cc @rjnagal

### DIFF
--- a/pages/containers.go
+++ b/pages/containers.go
@@ -18,7 +18,6 @@ package pages
 import (
 	"fmt"
 	"html/template"
-	"math"
 	"net/http"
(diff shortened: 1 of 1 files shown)


rjnagal:
LGTM

PR #810, opened 2015-07-09
Modify Collector structure

anushree-n:
@rjnagal , @vmarmol  This is a rough draft of how we could change the Collector structure

The newly added fields 'minPollingFrequency' and 'regexps' in collectorInfo field in Collector structure are necessary for collection of metrics.
We can avoid calculation of these during actual metric collection to prevent repeated calculation during every metric calculation.

### DIFF
--- a/collector/generic_collector.go
+++ b/collector/generic_collector.go
@@ -16,7 +16,9 @@ package collector
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
+	"regexp"
 	"time"
 
 	"github.com/google/cadvisor/info/v2"
@@ -28,6 +30,17 @@ type GenericCollector struct {
 
 	//holds information extracted from the config file for a collector
 	configFile Config
+
+	//holds information necessary to extract metrics
+	info *collectorInfo
+}
+
+type collectorInfo struct {
+	//minimum polling frequency among all metrics
+	minPollingFrequency time.Duration
+
+	//regular expresssions for all metrics
+	regexps []*regexp.Regexp
 }
 
 //Returns a new collector using the information extracted from the configfile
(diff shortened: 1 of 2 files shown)


collector/generic_collector.go
```diff
@@ -45,8 +57,20 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
...
+	minPollFrequency := configInJSON.MetricsConfig[0].PollingFrequency
+	regexprs := make([]*regexp.Regexp, len(configInJSON.MetricsConfig))
+
+	for ind, metricConfig := range configInJSON.MetricsConfig {
+		if metricConfig.PollingFrequency < minPollFrequency {
+			minPollFrequency = metricConfig.PollingFrequency
+		}
+
+		//regexprs[ind] is set to nil if metricConfig.Regex cannot be parsed
+		regexprs[ind], _ = regexp.Compile(metricConfig.Regex)
```
rjnagal: Why not fail collector creation if regex can't be compiled?

rjnagal:
ok to test

rjnagal:
small nit on not keeping errored collectors around. looks okay otherwise.

collector/generic_collector.go
```diff
@@ -45,8 +57,20 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 
 	//TODO : Add checks for validity of config file (eg : Accurate JSON fields)
 
+	minPollFrequency := configInJSON.MetricsConfig[0].PollingFrequency
```
vmarmol: We should check that there is at least one (we have the TODO above, but we should at least check that so we don't crash)

collector/generic_collector.go
vmarmol: +1 to checking the error

collector/generic_collector.go
```diff
@@ -45,8 +57,20 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
...
+			minPollFrequency = metricConfig.PollingFrequency
+		}
+
+		//regexprs[ind] is set to nil if metricConfig.Regex cannot be parsed
+		regexprs[ind], _ = regexp.Compile(metricConfig.Regex)
+	}
+
 	return &GenericCollector{
-		name: collectorName, configFile: configInJSON,
+		name: collectorName, configFile: configInJSON, info: &collectorInfo{minPollingFrequency: minPollFrequency, regexps: regexprs},
```
vmarmol: Can we space it out a bit to make it easier to read? :)

```Go
return &GenericCollector{
    name: collectorName,
    configFile: configInJSON,
    info: &collectorInfo{
        minPollingFrequency: minPollFrequency,
        regexps: regexprs
    },
}
```

collector/generic_collector.go
anushree-n: The idea was to allow collection of other valid metrics by creating the constructor even if a few regexps errored. Can we can remove the entire metric whose regex is invalid and proceed with creation of a collector if there is atleast one valid metric?

rjnagal: My thinking is that if regex compile fails, then we should just error out and not create the collector. If we do partial collections, it might be hard for the user to notice that something was actually wrong with supplied regex. Failing to collect a metric is okay and we can continue processing, but invalid regex is probably something user should be reported back to user with a failure.

vmarmol: My thought (and I think Rohit's too), is that here we're able to tell the user that their config is wrong and allow them to fix it.

You bring up a good point about trying to fail with reduced functionality (but in that case, we should at least return the error). Only concern is that if some metrics show up and some don't the user would think that something was broken in cAdvisor (or that the metric is empty), whereas a clear yes no to the metric would possibly make them think the config was wrong.

We can go either way :) but I think this makes it clear that we should be doing very good reporting of errors on metric collection so users know what's happening. We should export some stats and possibly a page.

vmarmol: And it looks like we were on the same page @rjnagal :) lol

anushree-n: I see the point It might be  unclear to the user how only a few metrics have been collected. Thanks @rjnagal @vmarmol . Will fix it.

rjnagal:
LGTM

PR #811, opened 2015-07-10
Update gomock dependency

afein:
Gomock has been removed from code.google.com and is now hosted under github.

This was blocking the godep restore command from executing, leading to all sorts of issues when attempting to use cadvisor as a dependency in other projects.

Note: go get github.com/golang/mock/gomock does not get the correct version of gomock, as internal imports still point to code.google.com. 
This Godep change was generated by manually cloning gomock to GOPATH and then using godep save ./...

\cc @rjnagal


rjnagal:
LGTM, thanks for fixing it !

rjnagal:
ok to test

vmarmol:
Are we also changing the code to depend on the new location?

rjnagal:
@vmarmol yes, mockfs.go

afein:
yes! there was only one usage:
https://github.com/google/cadvisor/pull/811/files#diff-b99eec7556cdc85a8fbee378f4131125L21

rjnagal:
this change didn't delete the gomock from code.google.com though.

PR #812, opened 2015-07-10
Removed the gomock files from the old location

afein:
sorry for that! forgot about the old gomock files
\cc @rjnagal @vmarmol


vmarmol:
ok to test

vmarmol:
Thanks :)

PR #815, opened 2015-07-13
Cache custom metrics

anushree-n:
v1.ContainerStats structure is being used to cache stats. 
To include CustomMetrics [ ]v2.Metric (originally defined in v2/metric.go), metric.go was moved from v2 to v1.
Should v2.ContainerStats be used instead to cache and retrieve stats?

### DIFF
--- a/collector/collector_manager.go
+++ b/collector/collector_manager.go
@@ -22,11 +22,10 @@ import (
 	"github.com/google/cadvisor/info/v1"
 )
 
-type collectorManager struct {
-	collectors []*collectorData
+type GenericCollectorManager struct {
+	Collectors         []*collectorData
+	NextCollectionTime time.Time
 }
-
-var _ CollectorManager = &collectorManager{}
 
 type collectorData struct {
 	collector          Collector
@@ -35,26 +34,27 @@ type collectorData struct {
 
 // Returns a new CollectorManager that is thread-compatible.
 func NewCollectorManager() (CollectorManager, error) {
-	return &collectorManager{
-		collectors: []*collectorData{},
+	return &GenericCollectorManager{
+		Collectors:         []*collectorData{},
+		NextCollectionTime: time.Now(),
 	}, nil
 }
 
-func (cm *collectorManager) RegisterCollector(collector Collector) error {
-	cm.collectors = append(cm.collectors, &collectorData{
+func (cm *GenericCollectorManager) RegisterCollector(collector Collector) error {
+	cm.Collectors = append(cm.Collectors, &collectorData{
 		collector:          collector,
 		nextCollectionTime: time.Now(),
 	})
 	return nil
 }
 
-func (cm *collectorManager) Collect() (time.Time, []v1.Metric, error) {
+func (cm *GenericCollectorManager) Collect() (time.Time, []v1.Metric, error) {
 	var errors []error
 
 	// Collect from all collectors that are ready.
 	var next time.Time
 	var metrics []v1.Metric
-	for _, c := range cm.collectors {
+	for _, c := range cm.Collectors {
 		if c.nextCollectionTime.Before(time.Now()) {
 			nextCollection, newMetrics, err := c.collector.Collect()
 			if err != nil {
@@ -69,7 +69,7 @@ func (cm *collectorManager) Collect() (time.Time, []v1.Metric, error) {
 			next = c.nextCollectionTime
 		}
 	}
-
+	cm.NextCollectionTime = next
 	return next, metrics, compileErrors(errors)
 }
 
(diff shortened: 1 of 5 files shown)


info/v1/container.go
```diff
@@ -418,7 +418,8 @@ type ContainerStats struct {
 	Filesystem []FsStats `json:"filesystem,omitempty"`
 
 	// Task load stats
-	TaskStats LoadStats `json:"task_stats,omitempty"`
+	TaskStats     LoadStats `json:"task_stats,omitempty"`
+	CustomMetrics []Metric  `json:"custom_metrics,omitempty"`
```
vmarmol: nit: add a comment and whitespace

manager/container.go
```diff
@@ -397,27 +399,44 @@ func (c *containerData) updateLoad(newLoad uint64) {
...
+	if metricsErr != nil {
+		// Ignore errors if the container is dead.
+		if !c.handler.Exists() {
+			return nextCollectionTime, nil
+		}
+
+		// Metrics may be partially populated, push those before we return an error.
+		metricsErr = fmt.Errorf("%v, continuing to push metrics", metricsErr)
+	}
+	if metrics == nil {
```
vmarmol: This would stop getting all regular stats if we don't have any custom stats. Any reason we're integration collection here instead of the callsite above? (the one we commented out)

anushree-n: This condition should not have been here! Will remove it.

The intent was to  enable collection of metrics within 'housekeepingTick' so that the housekeeping time includes the time taken for metric collection.

rjnagal:
Can you address the nit about adding comments and rebase?

manager/container.go
```diff
@@ -342,10 +343,10 @@ func (c *containerData) housekeeping() {
...
-		nextCollectionTime, _, err := c.collectorManager.Collect()
-		if err != nil && c.allowErrorLogging() {
-			glog.Warningf("[%s] Collection failed: %v", c.info.Name, err)
-		}
+		// nextCollectionTime, _, err := c.collectorManager.Collect()
+		// if err != nil && c.allowErrorLogging() {
+		//	glog.Warningf("[%s] Collection failed: %v", c.info.Name, err)
+		//}
 
 		// Next housekeeping is the first of the stats or the custom collector's housekeeping.
```
anushree-n: Do we need to retain the metric collection here, @rjnagal , @vmarmol ? Only issue is that the metric collection time will not be included in housekeeping time in this case. 

However, the current approach of pushing metric collection to updateStats seems more convoluted.

manager/container.go
```diff
@@ -304,7 +305,7 @@ func (c *containerData) housekeeping() {
 		default:
 			// Perform housekeeping.
 			start := time.Now()
-			c.housekeepingTick()
+			nextCollectionTime = c.housekeepingTick()
```
rjnagal: I think we don't need to change the housekeeping time. Supporting collection of custom metrics at a rate faster than regular housekeeping will be problematic. 

How about making UpdateCustomStats() equivalent to UpdateStats(), and calling both of these from tick. We can save the next time for customMetric collection in containerData.

manager/container.go
rjnagal: Given that we are now saving all stats in one place, it doesn't make sense to collect metrics at this level.
Either doing it in UpdateStats() or calling UpdateCustomStats() from housekeepingTick() should be fine.

collector/collector_manager.go
```diff
@@ -22,11 +22,12 @@ import (
...
 
-type collectorManager struct {
-	collectors []*collectorData
+type GenericCollectorManager struct {
+	Collectors         []*collectorData
+	NextCollectionTime time.Time
 }
 
-var _ CollectorManager = &collectorManager{}
+//var _ CollectorManager = &collectorManager{}
```
rjnagal: remove the commented out line? Where was this being used?

manager/container.go
```diff
@@ -342,17 +342,12 @@ func (c *containerData) housekeeping() {
...
 		// TODO(vmarmol): Export metrics.
 		// Run custom collectors.
-		nextCollectionTime, _, err := c.collectorManager.Collect()
-		if err != nil && c.allowErrorLogging() {
-			glog.Warningf("[%s] Collection failed: %v", c.info.Name, err)
-		}
+		// nextCollectionTime, _, err := c.collectorManager.Collect()
+		// if err != nil && c.allowErrorLogging() {
+		//	glog.Warningf("[%s] Collection failed: %v", c.info.Name, err)
+		//}
```
rjnagal: more commented out code

rjnagal:
LGTM

PR #818, opened 2015-07-16
Modification in Collect() method of GenericCollector

anushree-n:
Calculation of minimum polling frequency and check for validity of regular expressions of all metrics of a collector has been done during initialization of the collector.

### DIFF
--- a/collector/generic_collector.go
+++ b/collector/generic_collector.go
@@ -66,10 +66,16 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 	}
 
 	minPollFrequency := configInJSON.MetricsConfig[0].PollingFrequency
+
+	//set minPollFrequency to housekeepingInterval if config returns minpollFrequency=0
+	if minPollFrequency == 0 {
+		minPollFrequency = 1 * time.Second
+	}
+
 	regexprs := make([]*regexp.Regexp, len(configInJSON.MetricsConfig))
 
 	for ind, metricConfig := range configInJSON.MetricsConfig {
-		if metricConfig.PollingFrequency < minPollFrequency {
+		if metricConfig.PollingFrequency < minPollFrequency && metricConfig.PollingFrequency != 0 {
 			minPollFrequency = metricConfig.PollingFrequency
 		}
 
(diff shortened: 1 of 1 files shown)


collector/generic_collector.go
```diff
@@ -95,14 +95,8 @@ func (collector *GenericCollector) Name() string {
...
 func (collector *GenericCollector) Collect() (time.Time, []v1.Metric, error) {
-	minNextColTime := collector.configFile.MetricsConfig[0].PollingFrequency
-	for _, metricConfig := range collector.configFile.MetricsConfig {
-		if metricConfig.PollingFrequency < minNextColTime {
-			minNextColTime = metricConfig.PollingFrequency
-		}
-	}
 	currentTime := time.Now()
-	nextCollectionTime := currentTime.Add(time.Duration(minNextColTime * time.Second))
+	nextCollectionTime := currentTime.Add(time.Duration(collector.info.minPollingFrequency * time.Second))
```
rjnagal: In line 68: It looks like we don't check if the user provided a polling frequency for any metric config. We need to set in a default if json returns zero and ignore any metricConfig.PollingFrequency which is zero.

anushree-n: Added the change. :)

collector/generic_collector.go
```diff
@@ -78,6 +78,10 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 			return nil, fmt.Errorf("Invalid regexp %v for metric %v", metricConfig.Regex, metricConfig.Name)
 		}
 	}
+	//set minPollFrequency to maxHousekeepingInterval if config returns minpollFrequency=0
+	if minPollFrequency == 0 {
+		minPollFrequency = 60 * time.Second
```
rjnagal: Do we want the default to be one minute? The default housekeeping is 1s.

collector/generic_collector.go
```diff
@@ -78,6 +78,10 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 			return nil, fmt.Errorf("Invalid regexp %v for metric %v", metricConfig.Regex, metricConfig.Name)
 		}
 	}
+	//set minPollFrequency to maxHousekeepingInterval if config returns minpollFrequency=0
+	if minPollFrequency == 0 {
```
rjnagal: I think we should move this logic up to replace line 68. Start with the default and then compare in the loop (#72) and lower it if we find a user-defined value.

rjnagal:
LGTM

PR #819, opened 2015-07-17
Add 50th and 95th percentiles

mvdan:
These are some of the stats that will later be needed for kubelet to expose via an endpoint.

### DIFF
--- a/summary/percentiles.go
+++ b/summary/percentiles.go
@@ -34,14 +34,17 @@ func (a uint64Slice) Len() int           { return len(a) }
 func (a uint64Slice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 func (a uint64Slice) Less(i, j int) bool { return a[i] < a[j] }
 
-// Get 90th percentile of the provided samples. Round to integer.
-func (self uint64Slice) Get90Percentile() uint64 {
+// Get percentile of the provided samples. Round to integer.
+func (self uint64Slice) GetPercentile(d float64) uint64 {
+	if d < 0.0 || d > 1.0 {
+		return 0
+	}
 	count := self.Len()
 	if count == 0 {
 		return 0
 	}
 	sort.Sort(self)
-	n := float64(0.9 * (float64(count) + 1))
+	n := float64(d * (float64(count) + 1))
 	idx, frac := math.Modf(n)
 	index := int(idx)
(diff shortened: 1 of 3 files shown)


summary/percentiles.go
```diff
@@ -88,26 +91,31 @@ func (self *resource) Add(p info.Percentiles) {
...
+		Max:        val,
+		Fifty:      val,
+		Ninety:     val,
+		NinetyFive: val,
 	}
 	self.Add(sample)
 }
 
 // Get max, average, and 90p from existing samples.
-func (self *resource) GetPercentile() info.Percentiles {
```
rjnagal: optional nit: Can we rename to GetAllPercentiles() instead. It makes searching/reading names easier if they don't differ in just the plural.

rjnagal:
LGTM with a small nit.

I think this changes is good for cAdvisor in general.
I haven't yet found a good concise way to put these numbers up on cAdvisor UI (plain table is ugly).
Take a stab at it if you get some time, @mvdan

summary/percentiles.go
mvdan: Sure.

mvdan:
Initially these were going to be used by kubelet, but in the end we're doing them in heapster with @afein. If this is in general useful, like you say, it can of course be merged. If not, I'm fine with closing it too.

I can have a look at the UI part, yeah. Should that be on this PR?

mvdan:
nit fixed. Also, note the TODO added. I was just confused by that code and the 90p comment so was hoping someone could shed some light on that.

rjnagal:
The UI comment was an aside, it can be done separately.

We currently get 90p for an hour by using 90p values from last 60 mins sample. If we do the same logic for 95p and 50p, we'll burn a it more memory. I think for now, we only collect up to 60 mins samples, so its not that bad. If we did a day worth of samples per container, then it would become noticeable.

mvdan:
So should I just remove the TODO on the samples line and this can be merged?

mvdan:
@rjnagal Removed the TODO

mvdan:
CC @vishh

mvdan:
I've improved the `GetPercentile(float64)` tests and exported the type.

vmarmol:
ok to test

vmarmol:
LGTM

PR #821, opened 2015-07-18
Reduce the amount of bad Docker.inspects when using docker_only flag in combination with Systemd

basvdlei:
When running with the `docker_only` flag, the manager will keep trying to find a handler for the "ignored" raw containers during the housekeeping (every minute.) Each time a container is passed to the docker driver it will send an inspect request to the Docker daemon, resulting in loads of log messages:

```log
Jul 18 19:18:45 hactar docker[1024]: time="2015-07-18T19:18:45+02:00" level=info msg="+job container_inspect(rsyslog.service)"
Jul 18 19:18:45 hactar docker[1024]: no such id: rsyslog.service
Jul 18 19:18:45 hactar docker[1024]: time="2015-07-18T19:18:45+02:00" level=info msg="-job container_inspect(rsyslog.service) = ERR (1)"
Jul 18 19:18:45 hactar docker[1024]: time="2015-07-18T19:18:45+02:00" level=error msg="Handler for GET /containers/{name:.*}/json returned error: no such id: rsyslog.service"
Jul 18 19:18:45 hactar docker[1024]: time="2015-07-18T19:18:45+02:00" level=error msg="HTTP Error: statusCode=404 no such id: rsyslog.service"
```

This PR adds an extra check in the CanHandleAndAccept method of the Docker driver to filter out non-Docker containers when using Systemd before doing a Docker inspect.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -44,7 +44,7 @@ var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path t
 
 // Regexp that identifies docker cgroups, containers started with
 // --cgroup-parent have another prefix than 'docker'
-var dockerCgroupRegexp = regexp.MustCompile(`[A-z]+-([a-z0-9]+)\.scope`)
+var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -131,14 +131,20 @@ func ContainerNameToDockerId(name string) string {
 	return id
 }
 
+func isContainerName(name string) bool {
+	if UseSystemd() {
+		return dockerCgroupRegexp.MatchString(path.Base(name))
+	}
+	return true
+}
+
 // Docker handles all containers under /docker
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
 	canAccept := true
 
-	// When using Systemd all docker containers have a .scope suffix
-	if UseSystemd() && !strings.HasSuffix(path.Base(name), ".scope") {
-		return false, canAccept, nil
+	if !isContainerName(name) {
+		return false, canAccept, fmt.Errorf("invalid container name")
 	}
 
 	// Check if the container is known to docker and it is active.
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -131,16 +131,6 @@ func ContainerNameToDockerId(name string) string {
 	return id
 }
 
-// Returns a full container name for the specified Docker ID.
-func FullContainerName(dockerId string) string {
-	// Add the full container name.
-	if UseSystemd() {
-		return path.Join("/system.slice", fmt.Sprintf("docker-%s.scope", dockerId))
-	} else {
-		return path.Join("/docker", dockerId)
-	}
-}
-
 // Docker handles all containers under /docker
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -42,6 +42,10 @@ var DockerNamespace = "docker"
 var dockerRootDir = flag.String("docker_root", "/var/lib/docker", "Absolute path to the Docker state root directory (default: /var/lib/docker)")
 var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path to the Docker run directory (default: /var/run/docker)")
 
+// Regexp that identifies docker cgroups, containers started with
+// --cgroup-parent have another prefix than 'docker'
+var dockerCgroupRegexp = regexp.MustCompile(`[A-z]+-([a-z0-9]+)\.scope`)
+
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
 func DockerStateDir() string {
@@ -119,8 +123,9 @@ func ContainerNameToDockerId(name string) string {
 
 	// Turn systemd cgroup name into Docker ID.
 	if UseSystemd() {
-		id = strings.TrimPrefix(id, "docker-")
-		id = strings.TrimSuffix(id, ".scope")
+		if matches := dockerCgroupRegexp.FindStringSubmatch(id); matches != nil {
+			id = matches[1]
+		}
 	}
 
 	return id
@@ -140,6 +145,12 @@ func FullContainerName(dockerId string) string {
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
 	canAccept := true
+
+	// When using Systemd all docker containers have a .scope suffix
+	if UseSystemd() && !strings.HasSuffix(path.Base(name), ".scope") {
+		return false, canAccept, nil
+	}
+
 	// Check if the container is known to docker and it is active.
 	id := ContainerNameToDockerId(name)
 
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -294,32 +294,8 @@ func (self *dockerContainerHandler) GetStats() (*info.ContainerStats, error) {
 }
 
 func (self *dockerContainerHandler) ListContainers(listType container.ListType) ([]info.ContainerReference, error) {
-	if self.name != "/docker" {
-		return []info.ContainerReference{}, nil
-	}
-	opt := docker.ListContainersOptions{
-		All: true,
-	}
-	containers, err := self.client.ListContainers(opt)
-	if err != nil {
-		return nil, err
-	}
-
-	ret := make([]info.ContainerReference, 0, len(containers)+1)
-	for _, c := range containers {
-		if !strings.HasPrefix(c.Status, "Up ") {
-			continue
-		}
-
-		ref := info.ContainerReference{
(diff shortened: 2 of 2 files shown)


container/docker/factory.go
```diff
@@ -139,6 +139,12 @@ func FullContainerName(dockerId string) string {
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
 	canAccept := true
+
+	// When using SystemD all docker containers have a docker- prefix
+	if UseSystemd() && !strings.HasPrefix(path.Base(name), "docker-") {
+		return false, canAccept, nil
```
rjnagal: I think this might not be true if a docker container is started with --cgroup_parent flag. parent cgroup name will replace docker in that case.

basvdlei: You are correct. I've not used that feature before :) So I just did a simple test and the prefix `docker-` is indeed replaced with the parent name. But it also didn't show up as a docker container in a build of master though. Probably because the docker factory is already relying on the `docker-` prefix when Systemd is detected :(

basvdlei:
I'm pretty sure containers are started as a `.scope` unit. This is good enough for filtering on my systemd based machines (Debian and CoreOS), where I don't have any other scope units. I've also changed the ContainerNameToDockerId to handle containers started with `--cgroup-parent`.

basvdlei:
Hi, is there something I can do to get this in?

whosthatknocking:
Just wondering, it looks like the scope of this PR is,
  - enabling support for `--cgroup-parent`
  - filter accept to only `.scope`

Perhaps these should be,
  - Two PRs?
  - To minimize any performance impact should the `regexp.MustCompile` be moved outside the func

vishh:
ok to test

vishh:
LGTM

container/docker/factory.go
```diff
@@ -118,8 +118,12 @@ func ContainerNameToDockerId(name string) string {
 
 	// Turn systemd cgroup name into Docker ID.
 	if UseSystemd() {
-		id = strings.TrimPrefix(id, "docker-")
-		id = strings.TrimSuffix(id, ".scope")
+		// Using a regexp because container started with
+		// --cgroup-parent have another prefix than 'docker'
+		re := regexp.MustCompile(`[A-z]+-([a-z0-9]+)\.scope`)
```
jimmidyson: As @whosthatknocking said, please move this to a var outside of the func.

basvdlei:
Moved the regexp out of the function as requested. Squashed and rebased so the automated builds won't fail (on changed dependencies.)

jimmidyson:
How should the non-docker cgroup parent be handled in `FullContainerName`?

basvdlei:
The first thing that comes to mind is to ask Docker for the cgroup parent name. It seems to be part of the inspect result. The current `FullContainerName` implementation however doesn't have a reference to the Docker client.

When scanning the codebase, `FullContainerName` is only used in `dockerContainerHandler.ListContainers`. So adding an extra argument that passes a Docker client reference shouldn't have that big of an impact.

But I don't like the amount of extra API call to Docker this will generate, for a case most users won't ever need. 

We could first try to test if `docker-%s.scope` cgroup exists before bugging Docker about to cgroup parent name?

container/docker/factory.go
```diff
@@ -42,6 +42,10 @@ var DockerNamespace = "docker"
 var dockerRootDir = flag.String("docker_root", "/var/lib/docker", "Absolute path to the Docker state root directory (default: /var/lib/docker)")
 var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path to the Docker run directory (default: /var/run/docker)")
 
+// Regexp that identifies docker cgroups, containers started with
+// --cgroup-parent have another prefix than 'docker'
+var dockerCgroupRegexp = regexp.MustCompile(`[A-z]+-([a-z0-9]+)\.scope`)
```
jimmidyson: This regex can probably be locked to `([a-z0-9]{64})\.scope$`, including anchoring to end. I wonder if we could use that to get the parent cgroup too? Something like `(.*)-([a-z0-9]{64})\.scope$`?

whosthatknocking: `(.+)-([a-z0-9]{64})\.scope$`?

container/docker/factory.go
```diff
@@ -140,6 +145,12 @@ func FullContainerName(dockerId string) string {
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
 	canAccept := true
+
+	// When using Systemd all docker containers have a .scope suffix
+	if UseSystemd() && !strings.HasSuffix(path.Base(name), ".scope") {
```
whosthatknocking: Feels appropriate to follow existing format and express the check under `isContainerName` or similar function?

container/docker/factory.go
jimmidyson: AFAIK format is `<cgroupParent>-<containerId>.scope`. Container ID is 64 characters (hence the `{64}` so from this regex we should be able to extract the cgroup parent (normally `docker`) & the container ID. I can't imagine any conflict with other scope units so this should be pretty safe.

jimmidyson:
@basvdlei Are you still working on this PR? If not do you mind if I pick this up & swap over the regex? Would really like to get this in - nice optimisation.

basvdlei:
Haven't spend much time on it, and when I did, I've mostly been banging my head on finding a clean solution for `FullContainerName`. The more I look at it, the more it seems like dead code. I'll take a look at your PR in a moment.

I like whosthatknocking's regex `(.+)` for the parent name, since it's never nil. AFAIK docker does not place any restrictions on the name, as far as it results in a valid unit name.

container/docker/factory.go
jimmidyson: Ah I missed @whosthatknocking's `+` as opposed to my `*` - much better idea. Will update #896 with that.

container/docker/handler.go
```diff
@@ -294,32 +294,8 @@ func (self *dockerContainerHandler) GetStats() (*info.ContainerStats, error) {
 }
 
 func (self *dockerContainerHandler) ListContainers(listType container.ListType) ([]info.ContainerReference, error) {
-	if self.name != "/docker" {
```
jimmidyson: How does removing this affect non-systemd Docker installs?

basvdlei: In all the tests I've done, it doesn't. I have tested this on Ubuntu Precise (upstart), Debian Jessie/Sid (sysv init) and Debian Jessie (systemd).

When the manager is started it registers the `/` container and starts searching for containers using `detectSubcontainers("/")`. Since `/` is technically a raw container, all Docker containers are discovered using their entries in the cgroup mount. 

I have not found any place where the `dockerContainerHandler` with the name `/docker` is created. Because it doesn't exist, this **if** will never be **false**

jimmidyson: Coincentally I've just run cadvisor on Ubnutu 14.04 & it does actually use `/docker`...

```
condescending_bhabha (/docker/f5a39e9c98788da89c050ca45d266ce0ddc31bd0a7e3ecbc618e166c17ade17d)
```

basvdlei: Yes, docker will create a cgroup named `docker`, and is found as subcontainer of `/`. But `CanHandleAndAccept` will be **false** when the manager tries to get a `dockerContainerHandler` for it. Because the inspect will fail: https://github.com/google/cadvisor/blob/master/container/docker/factory.go#L147

You can also see this happening in the Docker daemon logs:
```log
ERRO[437132] Handler for GET /containers/{name:.*}/json returned error: no such id: docker 
ERRO[437132] HTTP Error                                    err=no such id: docker statusCode=404
```
Only when the container `/docker/f5a39e9c98788da89c050ca45d266ce0ddc31bd0a7e3ecbc618e166c17ade17d` is found it will actually create a handler.

jimmidyson: The docker ID is extracted from the container name - https://github.com/google/cadvisor/blob/499737d7ad3f754a63036e9c4d0d5f77f2889fed/container/docker/factory.go#L118. It is that ID that is used in the docker inspect.

Again, this is tested & working properly on Ubuntu 14.04.

basvdlei: Maybe we are saying the same thing. Or are you saying this code is still needed? If so, do you know where the `dockerContainerHandler` named *only* `/docker` is created? Because only then will `ListContainers` actually do something.

In my tests, with this function stubbed, everything also still works on a non-systemd machine and also see Docker containers with the `/docker` cgroup eg.

```
clever_wright (/docker/64f26ce9d60219c5d730f31abeabc39c47fe7aa282343c3a3a8c820a80db5b70)
```

jimmidyson: omg you're totally right - sorry i misread the condition somehow. Doh!

Yes I agree - redundant code to be removed.

container/docker/factory.go
```diff
@@ -119,27 +123,24 @@ func ContainerNameToDockerId(name string) string {
...
+		if matches := dockerCgroupRegexp.FindStringSubmatch(id); matches != nil {
+			id = matches[1]
+		}
 	}
 
 	return id
 }
 
-// Returns a full container name for the specified Docker ID.
-func FullContainerName(dockerId string) string {
```
jimmidyson: Why is this deleted? Isn't this used by the web pages?

jimmidyson: OK it's redundant too. I like this PR - lots deleted :)

jimmidyson:
@vishh LGTM but need to retest - not sure if i'm in the admin list in Jenkins to request that with a comment here?

container/docker/factory.go
jimmidyson: Agreed. Also using the regexp rather than `HasSuffix` to check the container name would make more sense.

jimmidyson:
LGTM - thanks!

PR #823, opened 2015-07-20
Change logic for determining minimum polling frequency.

rjnagal:
Since polling is tied to housekeeping, minimum supported polling
frequency is 1s.

Users can specify polling frequency higher than 1s. The polling loop
will be called at the minimum frequency specified in config as long as
its higher than the minimum supported frequency.

### DIFF
--- a/collector/generic_collector.go
+++ b/collector/generic_collector.go
@@ -65,24 +65,27 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 		return nil, fmt.Errorf("No metrics provided in config")
 	}
 
-	minPollFrequency := configInJSON.MetricsConfig[0].PollingFrequency
-
-	//set minPollFrequency to housekeepingInterval if config returns minpollFrequency=0
-	if minPollFrequency == 0 {
-		minPollFrequency = 1 * time.Second
-	}
-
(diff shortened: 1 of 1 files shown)


rjnagal:
@anushree-n  I think this is the model we want to go with. WDYT?

collector/generic_collector.go
```diff
@@ -85,6 +82,12 @@ func NewCollector(collectorName string, configfile string) (*GenericCollector, e
 		}
 	}
 
+	// Minimum supported polling frequency is 1s.
+	minSupportedFrequency := 1 * time.Second
```
vmarmol: In theory it should be the value of minHousekeeping flag :)

rjnagal: yes, but importing HousekeepingInterval adds a cyclic dependency between manager and collector. I'll clean that up next.

anushree-n:
I believe the earlier model took the minimum of next stats collection and next custom metrics collection as the next housekeeping time.

If the minimum supported frequency is 1s, yes, this looks good! :)

rjnagal:
yeah, the intention is to not go below the housekeeping interval.

rjnagal:
self-merging ....

PR #825, opened 2015-07-21
Moved max_housekeeping and allow_dynamic_housekeeping flags to cadvisor.go

piosz:


vmarmol:
We should make this change in the Kubernetes configuration rather than the default for cAdvisor I think.

piosz:
The changed is required to change the value of max_housekeeping_interval in Kubernetes. PTAL

vmarmol:
LGTM, thanks @piosz!

vmarmol:
ok to test

piosz:
Thanks for review!

PR #826, opened 2015-07-21
Added simple Makefile

piosz:


vmarmol:
I've been resisting adding a Makefile since not having one makes it clear that we're "just Go". This is the second request for a Makefile, is this something people expect and prefer?

piosz:
I definitely prefer typing ```make test``` instead of ```godep go test ./... -test.short```, but this is just my private feeling;)

vmarmol:
LGTM, will wait for @rjnagal's LGTM

rjnagal:
I worry that having a makefile would encourage people to make regular "godep go ..." not work. We'll also force go purists to install make :)

These are minor complaints. I won't be too sad if this got in. My build/test alias are still shorter than make test :)

mvdan:
Argument against this makefile in particular - running `deps`, `clean` and `build` in `test` before the actual `godep go test` will make the whole thing noticeably slower than just running `godep go test` directly.

Why are those dependencies for `test` anyway?

I generally agree with @rjnagal here - a Makefile to share targets as if they were aliases amongst people is fine, as long as running the simple go/godep commands will still be straightforward :)

piosz:
I won't argue;)

PR #827, opened 2015-07-21
Add logic to read custom metric config files from container root.

rjnagal:
Docker does not provide the rootfs path through docker inspect or statefile
and the path is dependent on the storage driver being used.

Instead of enumerating the storage drivers, we pick a pid from the container
and get the config from /proc/pid/root. Although a bit expensive, this method
works for non-docker containers too.

### DIFF
--- a/collector/collector_manager.go
+++ b/collector/collector_manager.go
@@ -22,6 +22,8 @@ import (
 	"github.com/google/cadvisor/info/v1"
 )
 
+const metricLabelPrefix = "io.cadvisor.metric."
+
 type GenericCollectorManager struct {
 	Collectors         []*collectorData
 	NextCollectionTime time.Time
@@ -38,6 +40,17 @@ func NewCollectorManager() (CollectorManager, error) {
 		Collectors:         []*collectorData{},
 		NextCollectionTime: time.Now(),
 	}, nil
+}
+
+func GetCollectorConfigs(labels map[string]string) map[string]string {
+	configs := map[string]string{}
+	for k, v := range labels {
+		if strings.HasPrefix(k, metricLabelPrefix) {
+			name := strings.TrimPrefix(k, metricLabelPrefix)
+			configs[name] = v
+		}
+	}
+	return configs
 }
 
 func (cm *GenericCollectorManager) RegisterCollector(collector Collector) error {
--- a/manager/container.go
+++ b/manager/container.go
@@ -17,8 +17,10 @@ package manager
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"math"
 	"os/exec"
+	"path"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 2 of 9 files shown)


rjnagal:
@anushree-n this should make the configs visible to cadvisor
The label for config should have the prefix "io.cadvisor.metric.<configname>"

collector/collector_manager.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/google/cadvisor/info/v1"
 )
 
+const metricLabelPrefix = "io.cadvisor.metric."
```
vmarmol: Are these supposed to be `io.cadvisor.metric.<what goes here?>`

manager/manager.go
```diff
@@ -689,6 +689,29 @@ func (m *manager) GetProcessList(containerName string, options v2.RequestOptions
...
+			return err
+		}
+		glog.V(3).Infof("Got config from %q: %q", v, configFile)
+
+		newCollector, err := collector.NewCollector(k, configFile)
+		if err != nil {
+			glog.Infof("failed to create collector for container %q, config %q: %v", cont.info.Name, k, err)
+			return err
+		}
+		err = cm.RegisterCollector(newCollector)
```
anushree-n: Just a small query : Can 
err = cont.cm.RegisterCollector(newCollector)

be used instead of passing cm as an argument ?

anushree-n:
This looks good, @rjnagal  :) Thank you.

manager/manager.go
```diff
@@ -689,6 +689,29 @@ func (m *manager) GetProcessList(containerName string, options v2.RequestOptions
 	return ps, nil
 }
 
+func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *containerData, cm collector.CollectorManager) error {
+	for k, v := range collectorConfigs {
+		configFile, err := cont.ReadFile(v, m.inHostNamespace)
+		if err != nil {
+			glog.Infof("failed to read config file %q for config %q, container %q: %v", v, k, cont.info.Name, err)
```
vmarmol: Remove the glog's and merge it into the error. We already log the error in the calling function

manager/container.go
```diff
@@ -136,11 +138,32 @@ func (c *containerData) getCgroupPath(cgroups string) (string, error) {
...
+// Takes in a path relative to container root.
+func (c *containerData) ReadFile(filepath string, inHostNamespace bool) ([]byte, error) {
+	pids, err := c.getContainerPids(inHostNamespace)
+	if err != nil {
+		return nil, err
+	}
+	// TODO(rjnagal): Optimize by just reading container's cgroup.proc file when in host namespace.
+	rootfs := "/"
+	if !inHostNamespace {
+		rootfs = "/rootfs"
```
vmarmol: We should eventually make the `/rootfs` a flag, we're relying on it in many places. For another PR :)

vmarmol:
Minor comments, looks good overall.

collector/collector_manager.go
rjnagal: An identifying name  -eg . io.cadvisor.metric.nginx. This allows for multiple config to be specified.
If there are any processing errors, the name is what users will see in their error logs - "failed to parse config nginx"

vmarmol: Thanks! I'm guessing we derive no meaning from the name, just an identifier?

manager/container.go
rjnagal: WDYT about moving all flags and consts to a const package?
Right now, some of the files have circular dependencies. We can have one single file for all tunables.

vmarmol: SGTM

collector/collector_manager.go
rjnagal: yes, just an identifier.

manager/manager.go
rjnagal: ah, that's right. Fixed.

manager/manager.go
rjnagal: done.

rjnagal:
Fixed the comments and re-pushed.

manager/manager.go
```diff
@@ -700,17 +722,26 @@ func (m *manager) createContainer(containerName string) error {
...
 
+	// Add collectors
+	labels := handler.GetContainerLabels()
+	collectorConfigs := collector.GetCollectorConfigs(labels)
+	err = m.registerCollectors(collectorConfigs, cont)
+	if err != nil {
+		glog.Infof("failed to register collectors for %q: %v", containerName, err)
+		return err
+	}
+
```
anushree-n: Do we need to add the following after line 780 ?
if collectorConfigs != nil {                                     //err != nil check for m.registerCollectors is already present
        contSpec.HasCustomMetrics = true
}

vmarmol:
LGTM, merging. Feel free to address @anushree-n's point in another PR :)

PR #829, opened 2015-07-22
UI changes to display custom metrics

anushree-n:
The function startCustomMetrics is called once to setup the custom graphs depending on the number of custom metrics.
The function drawCustomMetrics is called periodically to draw all the graphs of custom metrics.

### DIFF
--- a/pages/containers.go
+++ b/pages/containers.go
@@ -231,20 +231,21 @@ func serveContainersPage(m manager.Manager, w http.ResponseWriter, u *url.URL) e
 	}
 
 	data := &pageData{
-		DisplayName:        displayName,
-		ContainerName:      escapeContainerName(cont.Name),
-		ParentContainers:   parentContainers,
-		Subcontainers:      subcontainerLinks,
-		Spec:               cont.Spec,
-		Stats:              cont.Stats,
-		MachineInfo:        machineInfo,
-		IsRoot:             cont.Name == "/",
-		ResourcesAvailable: cont.Spec.HasCpu || cont.Spec.HasMemory || cont.Spec.HasNetwork || cont.Spec.HasFilesystem,
-		CpuAvailable:       cont.Spec.HasCpu,
-		MemoryAvailable:    cont.Spec.HasMemory,
-		NetworkAvailable:   cont.Spec.HasNetwork,
-		FsAvailable:        cont.Spec.HasFilesystem,
-		Root:               rootDir,
+		DisplayName:            displayName,
+		ContainerName:          escapeContainerName(cont.Name),
+		ParentContainers:       parentContainers,
+		Subcontainers:          subcontainerLinks,
+		Spec:                   cont.Spec,
+		Stats:                  cont.Stats,
+		MachineInfo:            machineInfo,
+		IsRoot:                 cont.Name == "/",
+		ResourcesAvailable:     cont.Spec.HasCpu || cont.Spec.HasMemory || cont.Spec.HasNetwork || cont.Spec.HasFilesystem,
+		CpuAvailable:           cont.Spec.HasCpu,
+		MemoryAvailable:        cont.Spec.HasMemory,
+		NetworkAvailable:       cont.Spec.HasNetwork,
+		FsAvailable:            cont.Spec.HasFilesystem,
+		CustomMetricsAvailable: cont.Spec.HasCustomMetrics,
+		Root: rootDir,
 	}
 	err = pageTemplate.Execute(w, data)
 	if err != nil {
--- a/pages/containers_html.go
+++ b/pages/containers_html.go
@@ -221,6 +221,16 @@ const containersHtmlTemplate = `
           </div>
         </div>
 	{{end}}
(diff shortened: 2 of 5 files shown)


pages/containers.go
```diff
@@ -231,20 +231,21 @@ func serveContainersPage(m manager.Manager, w http.ResponseWriter, u *url.URL) e
...
-		Root:               rootDir,
+		DisplayName:            displayName,
+		ContainerName:          escapeContainerName(cont.Name),
+		ParentContainers:       parentContainers,
+		Subcontainers:          subcontainerLinks,
+		Spec:                   cont.Spec,
+		Stats:                  cont.Stats,
+		MachineInfo:            machineInfo,
+		IsRoot:                 cont.Name == "/",
+		ResourcesAvailable:     cont.Spec.HasCpu || cont.Spec.HasMemory || cont.Spec.HasNetwork || cont.Spec.HasFilesystem || cont.Spec.HasCustomMetrics,
```
anushree-n: Addition of customMetrics to ResourcesAvailable is probably not necessary, as cpu, memory, network, fs are the standard metrics which need to be present.

pages/static/containers_js.go
```diff
@@ -468,6 +468,30 @@ function drawFileSystemUsage(machineInfo, stats) {
 	}
 }
 
+// Draw the graph for Custom Metrics.
```
anushree-n: Assumption: PollingFrequency is same for all custom metrics which leads to custom_metrics.length being same in all stats corresponding to different timestamps.

pages/containers.go
rjnagal: ResourcesAvailable is used to draw the spec table. We can skip CustomMetrics as the details won't show up there.

pages/containers_html.go
```diff
@@ -221,6 +221,16 @@ const containersHtmlTemplate = `
           </div>
         </div>
 	{{end}}
+        {{if .CustomMetricsAvailable}}
```
rjnagal: We should make this its own h3 heading as Application Metrics.

pages/static/containers_js.go
```diff
@@ -468,6 +468,30 @@ function drawFileSystemUsage(machineInfo, stats) {
...
+			var elements = [];
+                	elements.push(stats.stats[i].timestamp);
+			if (cur.int_points.length != 0) {
+				elements.push(cur.int_points[0])
+			} else if (cur.float_points.length !=0) {
+				elements.push(cur.float_points[0])
+			}
+                	data.push(elements);
+        	}
+        	drawLineChart(titles, data, elementId+"-"+cur.name, "Value")
```
rjnagal: Why not just show cur.name ? Do we also get units in the config?

rjnagal:
Let's hold on the UI changes a bit. I want to change our storage structure first.

The way we are storing stats is a bit inefficient.
The current structure stores name, type, label and int/float array for each stat point  (because we started to append metrics directly to containerStats).

I think we need to do two things:
1. take out name, type, label and push them only once in spec.
2. only store one struct {intval, floatval} per metric in container stats. We can add a map from metric name to this new struct. If we dont' want to waste storing two values for each metric, we can make two custom metrics map - one for int and one for float.

The other option is to not append custom metrics under ContainerStats, but that will need another timeStore.

I am going to start with separating out the spec first. let me know what you think @anushree-n @vmarmol

anushree-n:
Append metrics under ContainerStats
     - Already in place
     - Logically , custom metrics follows cpu, memory stats
     - If polling frequency of custom metrics is higher than that of regular stats, containerStats can be uneven.

Different timeStore
     -Extra overhead
     - Handling of data(to display in API and UI) may be easier

vmarmol:
@anushree-n is right about the different periods. Long term we want to move everything to metrics which would solve the timing problem and be the most efficient storage mechanism. That will probably take some time though :) @rjnagal can't we coalesce the name,labels into the same metric? The way it is exposed today it should handle multiple points with the same metadata. Maybe we need an intermediate structure as an output to the Collects.

pages/static/containers_js.go
```diff
@@ -696,12 +705,67 @@ function refreshStats() {
...
+	if(metricsInfo.length == 0) {
+		return;
+	}
+
+	for (var containerName in metricsInfo) {
+		var container = metricsInfo[containerName];
+		var metricIndex = -1 ;
+		for (var metricName in container) {
+			metricIndex++;
+			var titles = ["Time", containerInfo.spec.custom_metrics[metricIndex].units];
```
rjnagal: The order of stats and spec doesn't match. In my tests, the units gets mixed up.
Maps in go don't have a fixed order. I'd suggest changing the loop to iterate over the spec and locate spec.custom_metrics[index].name in metrics map returned by appmetrics.

anushree-n:
@rjnagal Let me know if any changes need to be done. Thanks!

rjnagal:
This looks great.

LGTM.

Squash and lets commit.

anushree-n:
Squashed. :)

PR #830, opened 2015-07-22
Enable custom metrics when collector configuration files are found

anushree-n:


manager/manager.go
```diff
@@ -782,6 +782,10 @@ func (m *manager) createContainer(containerName string) error {
 		return err
 	}
 
+	if collectorConfigs != nil {
+		contSpec.HasCustomMetrics = true
+	}
```
rjnagal: This doesn't seem to work for the api as we end up calling UpdateSpec() and overwriting this.

anushree-n: Oh, right! The spec gets re-written. Thanks for poiniting it out! I shall close the PR.

PR #831, opened 2015-07-23
Two small fixes to custom metric collection.

rjnagal:
- a typo in minPollingFrequency multiplies it with time.Second twice.
- Updating custom metrics is unnecessarily called for all containers.


rjnagal:
@anushree-n @vmarmol simple bugfixes to get custom metrics going.

vmarmol:
LGTM

PR #833, opened 2015-07-23
Change storage of custom metrics

anushree-n:
@rjnagal @vmarmol 

Would this be something we are interested in ?
We are using the same storage structure in a slightly different way.
Each metric will have  metric name, type, labels, a slice of intPoints(timestamp+value), a slice of floatPoints. Either intpoints or floatPoints will be empty for each metric. Collect function only updates intPoints or floatPoints.

### DIFF
--- a/collector/generic_collector.go
+++ b/collector/generic_collector.go
@@ -36,6 +36,8 @@ type GenericCollector struct {
 
 	//holds information necessary to extract metrics
 	info *collectorInfo
+
+	metrics []v1.Metric
 }
 
 type collectorInfo struct {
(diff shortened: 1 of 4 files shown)


collector/generic_collector.go
```diff
@@ -36,6 +36,8 @@ type GenericCollector struct {
 
 	//holds information necessary to extract metrics
 	info *collectorInfo
+
+	metrics []v1.Metric
```
anushree-n: There is redundancy here as we store metrics as part of the Collector struct as well as return it in the Collect() method.

anushree-n:
Some issues need to be addressed here though 
- The maximum size of IntPoints/FloatPoints that metrics can hold.
- How collector manager can gather metrics since Collect() method would return historic data too.

rjnagal:
We did bits of these in #838 

Closing this one. Lets do the remaining part - handling counter, rate - as a new PR.

Issue #834, opened 2015-07-24
Container memory is off


KevinPike:
It appears that a docker container process table is reporting RSS and virtual memory as KiB when it should be MiB.

Here's an example:
<img width="812" alt="screenshot 2015-07-24 08 15 22" src="https://cloud.githubusercontent.com/assets/1403622/8877851/a1909354-31dc-11e5-8e81-2e2d841f89d1.png">

<img width="1076" alt="screenshot 2015-07-24 08 15 34" src="https://cloud.githubusercontent.com/assets/1403622/8877848/98490a88-31dc-11e5-8851-5f31c90ee3d9.png">

The memory reported by the process table appears in KiB while the memory reported by the graph is MiB.

rjnagal:
Thanks for reporting @KevinPike It's missing a conversion. Sending out a PR now.

KevinPike:
Thanks @rjnagal for quickly addressing it. I think, with my very limited cAdvisor knowledge, that the PR will resolve this.

dashpole:
closed via #835

PR #835, opened 2015-07-24
Fix converion of rss and vsz in ps output.

rjnagal:


rjnagal:
fixes #834

vmarmol:
LGTM, thanks for the quick fix @rjnagal!

PR #838, opened 2015-07-25
Add custom metrics to spec.

rjnagal:
Remove spec-related fields from stat.
We can simplify the stats a bit further by handling Int and Float better.
But this was big enough change already.
Verified v1 and v2 spec/stats/appmetrics APIs.

### DIFF
--- a/info/v1/metric.go
+++ b/info/v1/metric.go
@@ -32,38 +32,21 @@ const (
 	MetricDelta = "delta"
 )
 
-// An exported metric.
-type Metric struct {
+// Spec for custom metric.
+type MetricSpec struct {
 	// The name of the metric.
 	Name string `json:"name"`
 
 	// Type of the metric.
 	Type MetricType `json:"type"`
-
-	// Metadata associated with this metric.
-	Labels map[string]string
-
-	// Value of the metric. Only one of these values will be
-	// available according to the output type of the metric.
-	// If no values are available, there are no data points.
-	IntPoints   []IntPoint   `json:"int_points,omitempty"`
-	FloatPoints []FloatPoint `json:"float_points,omitempty"`
 }
 
-// An integer metric data point.
-type IntPoint struct {
+// An exported metric.
+type MetricVal struct {
 	// Time at which the metric was queried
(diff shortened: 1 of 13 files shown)


rjnagal:
@anushree-n @vmarmol Still a bunch of cleanup to do, but I think this is a more condensed format for stats and would be easier to handle.

info/v1/metric.go
```diff
@@ -32,38 +32,21 @@ const (
...
+// Spec for custom metric.
+type MetricSpec struct {
 	// The name of the metric.
 	Name string `json:"name"`
 
 	// Type of the metric.
 	Type MetricType `json:"type"`
-
-	// Metadata associated with this metric.
-	Labels map[string]string
```
vmarmol: The labels should go with the values?

vmarmol:
@anushree-n and I spoke on Friday and we thought that we may want to break the custom metrics into their own completely (not rely on Stats). This is more in line with where we want to go long-term with everything being a metric. I'm OK with putting them in Stats as we're doing here in the short-term if we think it'll make things simpler. From some of the issues with storage, it may not.

info/v1/metric.go
rjnagal: I didn't know what the labels for each metric were supposed to do? For now, I just removed it.

vmarmol: We spoke a bit offline, but labels allow us to identify different parameters of the same metric. For example, if I am exporting disk size as the "disk_size" metric and I have 2 disk: sda1 and sda2 I can export a label "disk_id=sda1" to specify these metrics are for that disk

rjnagal: I see. Wouldn't we have naming conflicts if we have two stats series that are both called disk_usage and disambiguation depends on the label. I would lean towards pushing that kind of info in the name, in a structured way.

I think this will be relevant when we implement a prometheus reader. Let's revisit it then.

rjnagal:
Chatted with Victor. He has no big issues with the change as such. Merging it and following up with other PRs to resolve issues around namespacing and timestamp alignment.

PR #839, opened 2015-07-30
Add units and data type to spec for custom metrics.

rjnagal:
Renamed units to data_type in collector config.
Use units for actual display units eg MB/s, "number of connections".


rjnagal:
@anushree-n This invalidates existing scripts because:
- we renamed units to data_type.
- dropped supporting both int and integer, just using int.

jnagal:nginx has the updated config.

vmarmol:
LGTM

PR #841, opened 2015-08-01
Add an option to not display some custom metrics on UI.

rjnagal:
These metrics are still accessible from the API.

### DIFF
--- a/collector/config.go
+++ b/collector/config.go
@@ -29,22 +29,26 @@ type Config struct {
 
 // metricConfig holds information extracted from the config file about a metric
 type MetricConfig struct {
-	//the name of the metric
+	// the name of the metric
 	Name string `json:"name"`
 
-	//enum type for the metric type
+	// enum type for the metric type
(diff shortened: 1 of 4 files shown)


vmarmol:
OK with the change but wondering if this is something we can do on the UI instead. Seems like adding API support is a lot. What metrics do we think this would apply for?

rjnagal:
@vmarmol With prometheus, there are tons of metrics that are reported and putting them up in UI isn't that useful.

I would like to condense UI output in three ways:
- Add a way to remove a metric from UI (like we do with pgfaults for regular metrics).
- Add a tag to group multiple metrics on to the same graph (eg read and write connections for nginx)
- Add a drop down to change metrics in a graph (eg make all labels for a metric grouped as drop-down in prometheus)

We should also probably have a cap of graphs we show on a page to keep it responsive.
Let's discuss ways to do it without API input.

rjnagal:
Need to think of a different way to do it. Closing this one for now.

PR #844, opened 2015-08-07
issue: high cpu usage

jhjeong-kr:
housekeeping function is invoked 1sec(default) even if system time changed greatly.
let's suppose that cadvisor is running on a vm and the vm is about to be suspended for 1 day,
and, after 1 day, vm is resumed.
At the moment, cadvisor runs at full speed, because nexthousekeeping is current housekeeping + 1sec.
Therefore, cadvisor collects stats 1day * 24hours * 60minutes * 60seconds times without sleep.
In my systems, cadvisor occupies 100% cpu time almost for 3 hours.
soluton:
If cadvisor notices time change(allowedHouseKeepingGap or higher), cadvisor has to adjust nexthousekeeping.
I implemented this patch with allowedHouseKeepingGap constant.

Signed-off-by: Jin-Hwan Jeong <jhjeong.kr@gmail.com>

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -416,6 +416,8 @@ func (c *containerData) housekeeping() {
 		// Schedule the next housekeeping. Sleep until that time.
 		if time.Now().Before(next) {
 			time.Sleep(next.Sub(time.Now()))
+		} else {
+			next = time.Now()
 		}
 		lastHousekeeping = next
 	}


jhjeong-kr:
I signed it!

2015년 8월 7일 (금) 오후 3:41, cadvisorJenkinsBot <notifications@github.com>님이 작성:

vmarmol:
ok to test

manager/container.go
```diff
@@ -63,6 +63,7 @@ type containerData struct {
 	loadAvg                  float64 // smoothed load average seen so far.
 	housekeepingInterval     time.Duration
 	maxHousekeepingInterval  time.Duration
+	allowedHouseKeepingGap   time.Duration
```
vmarmol: Looks like we never set a default for this one?

vmarmol:
Actually, what if in housekeeping we just set next to now when next is not "before" now? Should keep this from happening without needed a configurable jump.

Nice find btw :)

jhjeong-kr:
@vmarmol Yes, Your suggestion is the simplest way to stop the issue, but I want to distinguish between two cases, 1) system time change(by suspend/resume), 2) actually host machine is very busy(really really busy), so I tried to fix nexthousekeeping() function.
If you want, I can update PR.

manager/container.go
jhjeong-kr: It was dropped while testing via cli argument.
At the moment, it would be better to set as a constant.
I set 1 minute as a default in housekeeping().

vmarmol:
Hey @jhjeong-kr I would prefer we make the change. I'm not sure we need to distinguish between the cases unless you strongly disagree.

jhjeong-kr:
@vmarmol OK. May I update the source as you recommend? If you prefer a simpler way, I can update this PR within a couple of hours.

vmarmol:
Updating the source SGTM.

jhjeong-kr:
I updated the source code.

vmarmol:
LGTM, thanks @jhjeong-kr!

Issue #843, opened 2015-08-06
No processes listed on centos systems except in /


aecolley:
The ps on Centos6 (procps version 3.2.8) formats the CGROUP column differently from the ps used on Ubuntu (procps-ng version 3.3.9). It has no subsystem number (so no ":" before the subsystem name) and entries are separated by ";" instead of ",".

The cgroupPathRegExp in manager/container.go should be changed to reflect it. I suggest ".*?\\w:(/[^,;]*)[,;].*".

aecolley:
Mangled by markdown. Let me try one more time: `".*?\\w:(/[^,;]*)[,;].*"`

rjnagal:
Thanks for reporting @aecolley. The regex looks okay. Feel free to send a CL, or I'll do it soonish :)
We'll cut a new release within a week

rjnagal:
verified on centos6.

PR #846, opened 2015-08-08
Fix cgroup name parsing logic in ps output for centos6.

rjnagal:
Centos 6 uses a different ps lib resulting in different output.


rjnagal:
Fixes #843

vmarmol:
LGTM

PR #849, opened 2015-08-10
Adding STDOUT new storage

jimonreal:
Createed the feature proposed on issue 847
There are changes that could be made on how to write to STDOUT.

It needs besides the storage_driver=io, the storage_driver_host=$HOSTNAME that way you may associate the logs, perhaps this could be added in the same code, but it's done this way to make it flexible to the user.

PR #850, opened 2015-08-11
Stdout storage

jimonreal:
Resolving branch and commit issues made. This PR is originally from PR #849

### DIFF
--- a/storage/stdout/stdout.go
+++ b/storage/stdout/stdout.go
@@ -17,6 +17,7 @@ package stdout
 import (
 	"bytes"
 	"fmt"
+
 	info "github.com/google/cadvisor/info/v1"
 )
 
@@ -84,11 +85,9 @@ func (driver *stdoutStorage) AddStats(ref info.ContainerReference, stats *info.C
 		return nil
 	}
 
-	var containerName string
+	containerName := ref.Name
 	if len(ref.Aliases) > 0 {
 		containerName = ref.Aliases[0]
-	} else {
-		containerName = ref.Name
 	}
 
 	var buffer bytes.Buffer
--- a/storage/stdout/stdout.go
+++ b/storage/stdout/stdout.go
@@ -1,0 +1,117 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stdout
+
+import (
+	"bytes"
+	"fmt"
+	info "github.com/google/cadvisor/info/v1"
+)
+
+type stdoutStorage struct {
+	Namespace string
+}
+
+const (
+	colCpuCumulativeUsage string = "cpu_cumulative_usage"
+	// Memory Usage
+	colMemoryUsage string = "memory_usage"
+	// Working set size
+	colMemoryWorkingSet string = "memory_working_set"
+	// Cumulative count of bytes received.
+	colRxBytes string = "rx_bytes"
+	// Cumulative count of receive errors encountered.
+	colRxErrors string = "rx_errors"
+	// Cumulative count of bytes transmitted.
+	colTxBytes string = "tx_bytes"
+	// Cumulative count of transmit errors encountered.
+	colTxErrors string = "tx_errors"
+	// Filesystem summary
+	colFsSummary = "fs_summary"
+	// Filesystem limit.
+	colFsLimit = "fs_limit"
+	// Filesystem usage.
+	colFsUsage = "fs_usage"
+)
+
+func (driver *stdoutStorage) containerStatsToValues(stats *info.ContainerStats) (series map[string]uint64) {
+	series = make(map[string]uint64)
+
+	// Cumulative Cpu Usage
+	series[colCpuCumulativeUsage] = stats.Cpu.Usage.Total
+
+	// Memory Usage
+	series[colMemoryUsage] = stats.Memory.Usage
+
+	// Working set size
+	series[colMemoryWorkingSet] = stats.Memory.WorkingSet
+
+	// Network stats.
+	series[colRxBytes] = stats.Network.RxBytes
+	series[colRxErrors] = stats.Network.RxErrors
+	series[colTxBytes] = stats.Network.TxBytes
+	series[colTxErrors] = stats.Network.TxErrors
+
+	return series
+}
+
+func (driver *stdoutStorage) containerFsStatsToValues(series *map[string]uint64, stats *info.ContainerStats) {
+	for _, fsStat := range stats.Filesystem {
+		// Summary stats.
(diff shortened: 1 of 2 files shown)


storage/stdout/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+		(*series)[fsStat.Device+"."+colFsUsage] = fsStat.Usage
+	}
+}
+
+func (driver *stdoutStorage) AddStats(ref info.ContainerReference, stats *info.ContainerStats) error {
+	if stats == nil {
+		return nil
+	}
+
+	var containerName string
```
mvdan: It's a bit more idiomatic to do:

```
containerName := ref.Name
if len(ref.Aliases) > 0 {
    containerName = ref.Aliases[0]
}

File: storage/stdout/stdout.go [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+		containerName = ref.Name
+	}
+
+	var buffer bytes.Buffer
+	buffer.WriteString(fmt.Sprintf("cName=%s host=%s", containerName, driver.Namespace))
+
+	series := driver.containerStatsToValues(stats)
+	driver.containerFsStatsToValues(&series, stats)
+	for key, value := range series {
+		buffer.WriteString(fmt.Sprintf(" %s=%v", key, value))
```
mvdan: If this data can be very long - i.e. if this line can be very long - consider splitting the output into multiple lines.

This has several advantages:

* Better readability
* No need to use a buffer, just write directly line by line to stdout
* Writing very long chunks of text to stdout in one go is not a very good idea on most systems

storage/stdout/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type stdoutStorage struct {
+	Namespace string
+}
+
+const (
+	colCpuCumulativeUsage string = "cpu_cumulative_usage"
+	// Memory Usage
+	colMemoryUsage string = "memory_usage"
```
mvdan: Do you need to specify that these consts are strings? (probably not)

storage/stdout/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stdout
+
+import (
+	"bytes"
+	"fmt"
+	info "github.com/google/cadvisor/info/v1"
```
mvdan: We generally group and separate imports, so add an empty line before the `info` import since it's not a standard library import.

storage/stdout/stdout.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (driver *stdoutStorage) containerFsStatsToValues(series *map[string]uint64, stats *info.ContainerStats) {
+	for _, fsStat := range stats.Filesystem {
+		// Summary stats.
+		(*series)[colFsSummary+"."+colFsLimit] += fsStat.Limit
+		(*series)[colFsSummary+"."+colFsUsage] += fsStat.Usage
+
+		// Per device stats.
+		(*series)[fsStat.Device+"."+colFsLimit] = fsStat.Limit
+		(*series)[fsStat.Device+"."+colFsUsage] = fsStat.Usage
```
mvdan: This block of code is a bit weird. As a start, you could do the string concatenation only once.

Is the map dereference necessary? I thought Go took care of this, although I'm right now unsure about maps.

storage/stdout/stdout.go
jimonreal: I got your point. That would be for someone who would want to read it right? but if you think in a logging system like logstash, fluentd, sumologic, splunk, etc. This is better, and you do not have to add extra characters for new line. The other thing on outputing all together is that you will not get mixed up data from other nodes, making this system mix all up.

Let me know what you think about it

storage/stdout/stdout.go
jimonreal: probably not

storage/stdout/stdout.go
jimonreal: I took it from statsd code, let me check

storage/stdout/stdout.go
jimonreal: I'll change it, thanks for the feedback

storage/stdout/stdout.go
mvdan: So you're saying that if you do line by line instead of writing it all at once, different goroutines might work at the same time and thus their lines might get mixed?

If you are worried about that, you could do the writes to stdout with a worker and a channel, something like `chan []string`, acting as a queue of chunks of lines.

If there is still a strong reason why a single line is preferred over multiple lines, then sticking with the current code sounds fine to me.

jimonreal: More than code rutines working at the same time, is the logging system collector that will not know which line correspond to which.
For a logging system this is nothing, and its easier to parse strings without the new line in the same log.

If you think that I could use chan []string but to improve performance, but still write everything in one line, then that sounds good.

storage/stdout/stdout.go
jimonreal: If not constant, what could it be? Just right the names down in the arrays on the methods? Or other type may I use?

storage/stdout/stdout.go
jimonreal: What concat? They are not identically if you see one is Summary the other is Device

mvdan: Whoops, you're right. Read too fast.

storage/stdout/stdout.go
mvdan: Constants don't need to be typed. You can just do `foo = "bar"` instead of `foo string = "bar"``

mvdan: Clarification - you might need a constant to be typed for some reason, but I highly doubt this is the case.

storage/stdout/stdout.go
mvdan: If the intention is to write only one line per `AddStats()` call, then this code is fine. If it ever becomes slow due to the buffers getting big, we can look into ways of improving it.

storage/stdout/stdout.go
```diff
@@ -0,0 +1,116 @@
...
+	var buffer bytes.Buffer
+	buffer.WriteString(fmt.Sprintf("cName=%s host=%s", containerName, driver.Namespace))
+
+	series := driver.containerStatsToValues(stats)
+	driver.containerFsStatsToValues(&series, stats)
+	for key, value := range series {
+		buffer.WriteString(fmt.Sprintf(" %s=%v", key, value))
+	}
+
+	fmt.Println(buffer.String())
```
mvdan: You should probably return `Println`'s error:

```go
_, err := fmt.Println(buffer.String())
return err
```

mvdan:
Besides the constants and error comments, LGTM.

@cadvisorJenkinsBot ok to test

CC @vmarmol

mvdan:
It appears I'm not an admin. Could someone trigger the tests?

storage/stdout/stdout.go
jimonreal: Ok. Thanks for the help, everyday learning something new :)

storage/stdout/stdout.go
jimonreal: So, should I just removed the constant declaration? and keep them as global vars?

mvdan: Just remove the `string` word from all of these const declarations. It should still build and work the same.

jimonreal:
Done ;)

mvdan:
LGTM

rjnagal:
Thanks @jimonreal and @mvdan for multiple iterations.

LGTM

jimonreal:
Thank you guys for the help! @vishh @mvdan @rjnagal

PR #851, opened 2015-08-12
Add Prometheus collector

anushree-n:


anushree-n:
Will reopen once #852 is finalized.

PR #852, opened 2015-08-13
Modify generic collector

anushree-n:
These are the changes that need to be done before we add a prometheus collector.

### DIFF
--- a/pages/static/containers_js.go
+++ b/pages/static/containers_js.go
@@ -740,31 +740,31 @@ function drawCustomMetrics(elementId, containerInfo, metricsInfo) {
 	if(metricsInfo.length == 0) {
 		return;
 	}
-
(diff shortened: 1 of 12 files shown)


pages/static/containers_js.go
```diff
@@ -740,31 +740,31 @@ function drawCustomMetrics(elementId, containerInfo, metricsInfo) {
...
+                                	        value = metric[attribute];
+                                        	elements.push(value);
+                                	}
+                    	       		if (elements.length<2) {
+                        	                elements.push(0);
+                                	}
+                                	data.push(elements);
+                        	}
+  		                drawLineChart(titles, data, elementId+"-"+metricName, metricUnits);
+			}
```
anushree-n: This function needs to be re-written when we have different labels for one metric.

rjnagal:
LGTM

PR #853, opened 2015-08-13
Add Prometheus Collector

anushree-n:


manager/manager.go
```diff
@@ -705,15 +705,28 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
...
-		newCollector, err := collector.NewCollector(k, configFile)
-		if err != nil {
-			glog.Infof("failed to create collector for container %q, config %q: %v", cont.info.Name, k, err)
-			return err
-		}
-		err = cont.collectorManager.RegisterCollector(newCollector)
-		if err != nil {
-			glog.Infof("failed to register collector for container %q, config %q: %v", cont.info.Name, k, err)
-			return err
+		if strings.HasPrefix(k, "prometheus") || strings.HasPrefix(k, "Prometheus") {
```
rjnagal: Let's submit this PR. But we should change the way we detect that its a prometheus config.
We can add a "type" field to the config which can default to REST and can be overridden to "prometheus". That way we don't end up overloading labels.

rjnagal:
looks okay... Can you please squash the commits?

anushree-n:
@rjnagal , Squashed.

rjnagal:
LGTM

PR #855, opened 2015-08-14
Adding Hostname from the server and not as a parameter for stdout

jimonreal:
Cleaning the stdout code a bit, making the hostname not a parameter from stdin but using the one from storagedriver.
Cleaning storagedriver to remove duplicate code lines for the hostname.

### DIFF
--- a/storage/stdout/stdout.go
+++ b/storage/stdout/stdout.go
@@ -108,7 +108,7 @@ func (driver *stdoutStorage) Close() error {
 	return nil
 }
 
-func New(hostname, namespace string) (*stdoutStorage, error) {
+func New(hostname string) (*stdoutStorage, error) {
 	stdoutStorage := &stdoutStorage{
 		Hostname: hostname,
 	}
--- a/storage/stdout/stdout.go
+++ b/storage/stdout/stdout.go
@@ -22,7 +22,7 @@ import (
 )
 
 type stdoutStorage struct {
-	Namespace string
+	Hostname string
 }
 
 const (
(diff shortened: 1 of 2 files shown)


rjnagal:
ok to test

rjnagal:
go vet error:

+go vet github.com/google/cadvisor/...
+godep go test -v -race -test.short github.com/google/cadvisor/...
# github.com/google/cadvisor
./storagedriver.go:91: not enough arguments in call to stdout.New

storage/stdout/stdout.go
```diff
@@ -108,9 +108,9 @@ func (driver *stdoutStorage) Close() error {
 	return nil
 }
 
-func New(namespace string) (*stdoutStorage, error) {
+func New(hostname, namespace string) (*stdoutStorage, error) {
```
rjnagal: namespace not required anymore.

mvdan:
LGTM

mvdan:
Ping @rjnagal

rjnagal:
When cadvisor is running in a container, os.hostname() is not going to return the hostname of the physical host. Is that acceptable for this use-case? It will still be different for each cadvisor instance.

jimonreal:
That will not work. Then I would have to get the HOST hostname from someplace else

vishh:
@jimonreal: You can override the default hostname that docker sets. So this PR should work even with docker.

timstclair:
This PR has been labeled inactive due to lack of recent activity. If it remains in this state for more than 2 weeks, it will be closed. If it is still being developed, please provide a status update.

timstclair:
Closing as **inactive**. Please reopen if development resumes.

PR #860, opened 2015-08-20
Always to a live lookup of version info instead of caching.

pwittrock:
This PR changes how VersionInfo is served
old: lookup/create and cache the version info at manager creation time
new: lookup/create the version info each time it is requested

This is necessary to correctly report the VersionInfo in the event that the docker daemon is not running at the time the manager is created.  This case may be encountered when cAdvisor is started in the same program that initializes the docker daemon / container bridge.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -161,8 +161,7 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 	if err != nil {
 		return nil, err
 	}
-	newManager.versionInfo = *versionInfo
-	glog.Infof("Version: %+v", newManager.versionInfo)
+	glog.Infof("Version: %+v", *versionInfo)
 
 	newManager.eventHandler = events.NewEventManager(parseEventsStoragePolicy())
 	return newManager, nil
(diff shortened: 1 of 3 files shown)


vishh:
ok to test

manager/manager.go
```diff
@@ -658,7 +656,9 @@ func (m *manager) GetMachineInfo() (*info.MachineInfo, error) {
 }
 
 func (m *manager) GetVersionInfo() (*info.VersionInfo, error) {
-	return &m.versionInfo, nil
+	// TODO: Consider caching this and periodically updating.  The VersionInfo may change if
+	// the docker daemon is started after the cAdvisor client is created
```
rjnagal: nit: Add to TODO - Might also be helpful to cache, so we can return last known docker version if docker is down at the time of query.

rjnagal:
LGTM.

Thanks for fixing it, @pwittrock !

vishh:
LGTM

manager/manager.go
pwittrock: Done.

pwittrock:
Thanks everyone.  PTAL. Feedback addressed.

vishh:
retest this please

vishh:
@cadvisorJenkinsBot: retest this please

pwittrock:
Recent push was to squash the commit for the PR review comments

rjnagal:
LGTM

PR #861, opened 2015-08-20
Add instantaneous cpu metrics

mvdan:
They will be useful in kubelet, since we want to expose instantaneous data in cores instead of accumulated data in seconds.

CC @vishh @vmarmol @rjnagal

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 	"path"
 	"strconv"
+	"time"
 
 	"github.com/golang/glog"
 	info "github.com/google/cadvisor/info/v1"
@@ -449,8 +450,63 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 	}
 }
 
+func instCpuStats(last, cur *info.ContainerStats) (*v2.CpuInstStats, error) {
+	if last == nil {
+		return nil, nil
+	}
+	if !cur.Timestamp.After(last.Timestamp) {
+		return nil, fmt.Errorf("container stats move backwards in time")
+	}
+	if len(last.Cpu.Usage.PerCpu) != len(cur.Cpu.Usage.PerCpu) {
+		return nil, fmt.Errorf("different number of cpus")
+	}
+	timeDelta := cur.Timestamp.Sub(last.Timestamp)
+	if timeDelta <= 100*time.Millisecond {
+		return nil, fmt.Errorf("time delta unexpectedly small")
+	}
+	// Nanoseconds to gain precision and avoid having zero seconds if the
+	// difference between the timestamps is just under a second
+	timeDeltaNs := uint64(timeDelta.Nanoseconds())
+	convertToRate := func(lastValue, curValue uint64) (uint64, error) {
+		if curValue < lastValue {
+			return 0, fmt.Errorf("cumulative stats decrease")
+		}
+		valueDelta := curValue - lastValue
+		return (valueDelta * 1e9) / timeDeltaNs, nil
+	}
+	total, err := convertToRate(last.Cpu.Usage.Total, cur.Cpu.Usage.Total)
+	if err != nil {
+		return nil, err
+	}
+	percpu := make([]uint64, len(last.Cpu.Usage.PerCpu))
+	for i := range percpu {
+		var err error
+		percpu[i], err = convertToRate(last.Cpu.Usage.PerCpu[i], cur.Cpu.Usage.PerCpu[i])
+		if err != nil {
+			return nil, err
+		}
+	}
+	user, err := convertToRate(last.Cpu.Usage.User, cur.Cpu.Usage.User)
+	if err != nil {
+		return nil, err
+	}
+	system, err := convertToRate(last.Cpu.Usage.System, cur.Cpu.Usage.System)
+	if err != nil {
+		return nil, err
+	}
+	return &v2.CpuInstStats{
+		Usage: v2.CpuInstUsage{
+			Total:  total,
+			PerCpu: percpu,
+			User:   user,
+			System: system,
+		},
+	}, nil
+}
+
 func convertStats(cont *info.ContainerInfo) []v2.ContainerStats {
-	stats := []v2.ContainerStats{}
+	stats := make([]v2.ContainerStats, 0, len(cont.Stats))
+	var last *info.ContainerStats
 	for _, val := range cont.Stats {
 		stat := v2.ContainerStats{
 			Timestamp:        val.Timestamp,
@@ -463,6 +519,13 @@ func convertStats(cont *info.ContainerInfo) []v2.ContainerStats {
 		}
 		if stat.HasCpu {
 			stat.Cpu = val.Cpu
+			cpuInst, err := instCpuStats(last, val)
+			if err != nil {
+				glog.Warningf("Could not get instant cpu stats: %v", err)
+			} else {
+				stat.CpuInst = cpuInst
+			}
+			last = val
 		}
 		if stat.HasMemory {
 			stat.Memory = val.Memory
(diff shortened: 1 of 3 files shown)


mvdan:
Doing the PR with only the data structure first so that if anyone is strongly against this, it can be said before I actually implement it.

vishh:
ok to test

info/v2/container.go
```diff
@@ -86,8 +86,11 @@ type ContainerStats struct {
 	// The time of this stat point.
 	Timestamp time.Time `json:"timestamp"`
 	// CPU statistics
-	HasCpu bool        `json:"has_cpu"`
-	Cpu    v1.CpuStats `json:"cpu,omitempty"`
+	HasCpu bool `json:"has_cpu"`
+	// In nanoseconds (aggregated)
+	Cpu v1.CpuStats `json:"cpu,omitempty"`
+	// In nanocores (instantaneous)
```
vishh: why not milli or micro cores?

mvdan: Since aggregated is in nanoseconds, I thought I'd stay consistent and not lose any precision.

vishh:
To summarize the issue, kubelet needs to expose cpu (and possibly network, diskio) usage in units that are similar to how users configure limits, which is in milli cores. One option we had is to perform caching and deriving instantaneous values in kubelet. But that seemed wasteful because cAdvisor can easily (cheaply maybe) calculate instantaneous data and expose it. 
Our dynamic housekeeping might end up shifting the units for instantaneous usage over time though.

info/v2/container.go
vishh: Also the unit is cores per second right?

mvdan: Yes, going to fix that.

mvdan:
@vmarmol @rjnagal is this a terrible idea?

info/v2/container.go
```diff
@@ -86,8 +86,11 @@ type ContainerStats struct {
 	// The time of this stat point.
 	Timestamp time.Time `json:"timestamp"`
 	// CPU statistics
-	HasCpu bool        `json:"has_cpu"`
-	Cpu    v1.CpuStats `json:"cpu,omitempty"`
+	HasCpu bool `json:"has_cpu"`
+	// In nanoseconds (aggregated)
+	Cpu v1.CpuStats `json:"cpu,omitempty"`
+	// In nanocores per second (instantaneous)
+	CpuInst v1.CpuStats `json:"cpu_inst,omitempty"`
```
rjnagal: I think we shouldn't overload CpuStats to work for both. Dropping units in the base definition is confusing.
Let's split them and we can evolve them separately. 

Overall supporting instant metrics in cAdvisor SG.

mvdan: Fair enough. If I split them, should the new definition be in v1 or v2?

Thanks for your input!

rjnagal: Kubelet uses v1 today. If you are not planning to update kubelet to move to v2, you can put it in v1. Else, move it to v2.

mvdan: I would have assumed that v1 was frozen by now. Otherwise, shouldn't changes in v1 be in v2 too?

Would moving kubelet to v2 be an issue?

vishh: Let's do it in v2. Kubelet can switch to using v2. We should consider doing
this for other instantaneous metrics as well.

mvdan:
@rjnagal how does this look? Doing a unit test for `instCpuStats()` now.

mvdan:
Finished including extensive unit tests. Doing integration tests now.

mvdan:
@rjnagal done. I did not do the integration tests because the v2 client is lacking, and I did not add the `DockerContainer` method to the v2 client because the `v2.ContainerInfo` type is missing. This PR would get rather large if I did all of that, and I don't know what you want to do with `v2.ContainerInfo` either.

api/versions.go
```diff
@@ -449,8 +449,62 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+	if !cur.Timestamp.After(last.Timestamp) {
+		return nil, fmt.Errorf("container stats move backwards in time")
+	}
+	if len(last.Cpu.Usage.PerCpu) != len(cur.Cpu.Usage.PerCpu) {
+		return nil, fmt.Errorf("different number of cpus")
+	}
+	// Nanoseconds to gain precision and avoid having zero seconds if the
+	// difference between the timestamps is just under a second
+	timeDelta := uint64(cur.Timestamp.Sub(last.Timestamp).Nanoseconds())
+	if timeDelta <= 0 {
```
rjnagal: We should probably refuse if the delta is less than 100ms or so.

api/versions.go
```diff
@@ -449,8 +449,62 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+		return nil, fmt.Errorf("refusing to use a non-positive time delta")
+	}
+	convert := func(lastValue, curValue uint64) (uint64, error) {
```
rjnagal: s/convert/convertToRate/ ?

info/v2/container.go
```diff
@@ -204,3 +207,27 @@ type NetworkStats struct {
...
+
+// Instantaneous CPU stats
+type CpuInstStats struct {
+	Usage CpuInstUsage `json:"usage"`
+}
+
+// CPU usage time statistics.
+type CpuInstUsage struct {
+	// Total CPU usage.
+	// Units: nanocores per second
```
rjnagal: Are we using nanocores/s for precision? Will we convert these back to millicores/s in kubelet?

api/versions.go
mvdan: Why is that? Is it never supposed to be that small?

api/versions.go
mvdan: Sure.

info/v2/container.go
mvdan: I used nanocores/s for both precision and consistency with the other metrics, which are also nano.

Kubelet will expose milli, so yes, it will divide the numbers. Although it's not converting "back", since I obtain the instantaneous metrics from accumulated nanoseconds, not milliseconds.

mvdan:
@rjnagal I can't just switch kubelet to using cadvisor's info/v2 because some types are missing. Saw your TODO, so I copied whatever types were necessary and weren't in v2 already.

mvdan:
Note that I left `v1.FsInfo` instead of using the modified `FsInfo` in v2. Should I change that?

mvdan:
Actually, more stuff is missing. Would you move it to v1 and keep the kubelet as-is, or keep it in v2 and make all the modifications in cadvisor and kubelet for it to work?

api/versions.go
rjnagal: Our timestamps are one seconds apart. If we took samples too close, it might exaggerate a small spike to a large number. Best to ignore tiny bursts or idle cycles.

info/v2/container.go
rjnagal: ok, nanocores/s is fine.

api/versions.go
mvdan: OK.

rjnagal:
Are you planning to use the /stats api in kubelet or add a different one for instantaneous stats? I thought you were already change kubelet to have its own native data-structures. If you are going to end up doing translation, I'd prefer to do that with v2. If you are re-using cadvisor structs, then do the minimal work to get you going.

I'd be fine with both.

mvdan:
The /stats api in kubelet will be removed, I'm ading `/experimental/v1/nodeMetrics` which translates from the cadvisor stats into its own api types.

Okay, I'll continue with v2. Just wanted to ask since I wasn't expecting v2 to be this lacking.

mvdan:
@rjnagal comments addressed. Removed the v2 stuff, I'll do that on another PR once I have everything working in kubelet.

rjnagal:
LGTM

Thanks for fixing the v2 stuff too!

PR #864, opened 2015-08-21
Fix network info for docker containers running with --net=host.

rjnagal:
Docker still reports a loop device for these. Need to check for
more than one device to mark network as available.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -167,7 +167,16 @@ func libcontainerConfigToContainerSpec(config *libcontainerConfigs.Config, mi *i
 	}
 	spec.Cpu.Mask = utils.FixCpuMask(config.Cgroups.CpusetCpus, mi.NumCores)
 
-	spec.HasNetwork = len(config.Networks) > 0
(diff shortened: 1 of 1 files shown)


rjnagal:
@vmarmol This fixes empty network display for kube containers.

container/docker/handler.go
```diff
@@ -167,7 +167,9 @@ func libcontainerConfigToContainerSpec(config *libcontainerConfigs.Config, mi *i
 	}
 	spec.Cpu.Mask = utils.FixCpuMask(config.Cgroups.CpusetCpus, mi.NumCores)
 
-	spec.HasNetwork = len(config.Networks) > 0
+	// Docker reports a loop device for containers with --net=host. Ignore
+	// those too.
+	spec.HasNetwork = len(config.Networks) > 1
```
vmarmol: WDYT if we count the non-loopback devices? If in the future this changes we don't have to worry about it.

rjnagal: Sure, changed to counting non-loopback devices.

vmarmol:
LGTM, thanks @rjnagal!

PR #865, opened 2015-08-25
Get network stats by switching network namespace on newer Docker versions.

### LINKED ISSUE #822: Zero stats for Docker containers network/blkio
On Fedora 22 with docker-1.7.0-6.git74e7a7a.fc22.x86_64, all Docker containers report zero network traffic & blkio. Is there something that needs to be configured on the host to enable this reporting?

jimmidyson:
Fixes #822

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -15,14 +15,22 @@
 package libcontainer
 
 import (
+	"bufio"
 	"fmt"
+	"net"
+	"os/exec"
 	"path"
+	"regexp"
+	"runtime"
+	"strings"
 	"time"
 
 	"github.com/docker/libcontainer"
 	"github.com/docker/libcontainer/cgroups"
+	"github.com/golang/glog"
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/utils/sysinfo"
+	"github.com/vishvananda/netns"
 )
 
 type CgroupSubsystems struct {
@@ -74,7 +82,7 @@ var supportedSubsystems map[string]struct{} = map[string]struct{}{
 }
 
 // Get cgroup and networking stats of the specified container
-func GetStats(cgroupManager cgroups.Manager, networkInterfaces []string) (*info.ContainerStats, error) {
+func GetStats(cgroupManager cgroups.Manager, networkInterfaces []string, pid int) (*info.ContainerStats, error) {
 	cgroupStats, err := cgroupManager.GetStats()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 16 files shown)


vishh:
ok to test

container/libcontainer/helpers.go
```diff
@@ -93,14 +99,101 @@ func GetStats(cgroupManager cgroups.Manager, networkInterfaces []string) (*info.
...
+// Borrowed from libnetwork Stats - https://github.com/docker/libnetwork/blob/master/sandbox/interface_linux.go
+// In older kernels (like the one in Centos 6.6 distro) sysctl does not have netns support. Therefore
+// we cannot gather the statistics from /sys/class/net/<dev>/statistics/<counter> files. Per-netns stats
+// are naturally found in /proc/net/dev in kernels which support netns (ifconfig relyes on that).
+const (
+	netStatsFile = "/proc/net/dev"
+	base         = "[  ]*%s:([  ]+[0-9]+){16}"
+)
+
+func scanInterfaceStats(data, ifName string) (info.InterfaceStats, error) {
```
vishh: How about parsing the entire file once, and index it based on the Interface name?

vishh:
Thanks for this PR @jimmidyson :100:

vishh:
The e2e tests for network stats are failing.

jimmidyson:
@vishh Looking back looks like the e2e tests have failed on most PRs recently... although I would have expected this PR to fix those failures :)

I have just tested it manually against coreos beta (the reported failure) & it works fine. I can't work out how to run the integration tests manually against a local VM - seems like the integration tests are coupled to GCE... not very friendly IMO. Any way to run these locally against a VM?

vishh:
The integration tests are run against a few linux distros. Here are the
instruction for running it against your VM:
https://github.com/google/cadvisor/blob/master/docs/integration_testing.md#integration-testing-cadvisor

jimmidyson:
Right:

"Today We only support remote execution in Google Compute Engine since that is where we run our continuous builds."

Working on making it a bit more flexible - would you be open to that? I don't have a GCE account to run integration tests myself.

jimmidyson:
@vishh I've just run integration tests against coreos beta on GCE & it worked fine. Any chance you could retest?

container/libcontainer/helpers.go
jimmidyson: Changed to that - good suggestion, thanks.

vishh:
Works for me! Thanks @jimmidyson!! LGTM.

jimmidyson:
Thanks @vishh.

Don't suppose there's any chance we could get a release cut with this fix to update deps in kubernetes & openshift?

vishh:
We can update cadvisor deps right away in kubernetes. I am not sure when
the next Kubernetes release will happen though.

jimmidyson:
Ah OK so no need to tag cadvisor first? Just use a non tagged commit in kubernetes godeps? As long as it's in next release whenever that is I'm happy.

vishh:
I think tagging makes sense! Created a tag '0.16.0.1'.

jimmidyson:
Great thanks!

jimmidyson:
@vishh One thing I have noticed in further (post-merge ;)) testing is when running in a Docker container cadvisor needs to be in the hosts `pid` namespace (see https://docs.docker.com/reference/run/#pid-settings-pid) to be able to switch network namespaces based on pid of the docker containers. I don't see this as a problem but something that needs to be documented. Shall I do that?

vishh:
Ah! Thats true. It cannot access the procfs of host. Do we have to
guarantee backwards compat with old kernels? Figuring out veth interfaces
and getting stats via sysfs seems to be way cheaper than switching
namespaces every second for every container.

jimmidyson:
No we don't - this is how docker works, switching namespaces. Switching namespaces is cheap, no?

jimmidyson:
This is the approach of libnetwork & how docker itself describes how to do high performance metrics collection - see https://docs.docker.com/articles/runmetrics/#tips-for-high-performance-metric-collection.

vishh:
Hmm. If we were to be network setup agnostic, I agree that entering the
namespace might be the best strategy.

jimmidyson:
I think we can use this technique for all versions of docker & get rid of all the veth handling code.

I'm not sure what you mean by network setup agnostic? Aren't we? Shouldn't we be?

vishh:
Yes. I agree. What I meant by network agnostic was not having to know if
the underlying networking mechanism is veth or ipvlan, etc.

PR #867, opened 2015-08-26
Add empty healthz test_utils to fix go build ./...

mvdan:
Without any .go file that doesn't end in _test, it fails:

no buildable Go source files in [...]/integration/tests/healthz

An empty source file works around this.


vishh:
ok to test

mvdan:
all green :)

PR #868, opened 2015-08-27
Use network namespace for all network metrics, including host

jimmidyson:


jimmidyson:
@vishh As discussed in #865, this now switches namespace for any processes, including the cadvisor process itself to get network metrics.

vishh:
ok to test

rjnagal:
I think there's a lighter approach than switching the network namespace.

If you can get a pid for a process running in the docker container, we can read stats using the root view of that pid:

eg. from inside a container:

$ docker exec -ti  3ae8d35423f8 /bin/sh
/ # cat /sys/class/net/eth0/statistics/tx_bytes 
313707668
# cat /sys/class/net/eth0/ifindex 
3824

From outside, using a pid (we already get this in cadvisor)
# cat /proc/24342/root/sys/class/net/eth0/ifindex 
3824
# cat /proc/24342/root/sys/class/net/
eth0/ lo/   
# /proc/24342/root/sys/class/net/eth0/statistics/tx_bytes 
313708886

WDYT?

jimmidyson:
@rjnagal I was following the recommendation from Docker itself - https://docs.docker.com/articles/runmetrics/#tips-for-high-performance-metric-collection. This is also the approach of libnetwork which is where most of the code is borrowed from.

I have put together a benchmark for this vs straight proc fs without setns - results show that the overhead of switching ns is negligible:

```
# go test -bench=. -benchmem -benchtime=30s
testing: warning: no tests to run
PASS
BenchmarkSetNs    100000            464509 ns/op           14876 B/op        163 allocs/op
BenchmarkSysFs    100000            459523 ns/op           14735 B/op        157 allocs/op
ok      github.com/jimmidyson/benchmark-setns   122.377s
```

Gisted the benchmark code & results at https://gist.github.com/jimmidyson/81448529975bc5466bb4.

jimmidyson:
Closing in favour of @rjnagal's approach in #870.

PR #869, opened 2015-08-28
Add image name to container spec & prometheus metrics

### LINKED ISSUE #848: Add Image Name to the metrics collected
It would be really helpful to have the imageName of the running container, that way you may associate all containers of the same image running on the same host, and on different hosts.

jimmidyson:
Fixes #848 

As discussed in https://github.com/google/cadvisor/pull/780#issuecomment-134657301, this just exports the image name as this has most immediate value. We can discuss docker labels separately perhaps?

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -74,6 +74,9 @@ type dockerContainerHandler struct {
 
 	// The container PID used to switch namespaces as required
 	pid int
+
+	// Image name used for this container.
+	image string
 }
 
 func newDockerContainerHandler(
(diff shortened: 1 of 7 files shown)


rjnagal:
LGTM

Thanks @jimmidyson We already expose labels. The only remaining discussion is around env variables, which I am still not convinced is safe to do.

jimmidyson:
Thanks!

Agree on env vars - dangerous considering they're used for generated passwords, secret data, etc.

Issue #634, opened 2015-04-07
Feature Request: Support ELK


catrixs:
ELK stack is heavily used, but cAdvisor only support influxdb. So we want to implement a log stash forward drive base on Lumberjack Protocol.

Is it a good idea?

vishh:
Is ELK built to handle metrics data?

feiyang21687:
@vishh Yah. We use ELK to collect container metrics data in production environment.

ELK represents ```ElasticSearch + LogStash + Kibana```. We think cAdvisor could be able to forward metrics data to LogStash, and then we could use ElasticSearch and Kibana to create dashboard, and like this one:

![2015-03-28 12 00 27](https://cloud.githubusercontent.com/assets/1303976/7037378/78bb1898-ddd4-11e4-8b2d-58b493b41533.png)

wangjingbomail:
It is a good idea

vishh:
Ok! Interesting! Are you interested in posting a PR to add elasticsearch as
a backend?

thaJeztah:
Nice looking screenshot! Definitely interesting

feiyang21687:
@vishh I'm working on it.

vishh:
Great!

ternel:
Would be great :)

bastnic:
:+1:

feiyang21687:
https://github.com/google/cadvisor/pull/693

We just submit a pull request which is very simple. Just implements a redis driver which use ```lpush``` to push metrics json to redis, then logstash could forward them to ElasticSearch.

Next step, we want implement a native ElasticSearch driver.

rjnagal:
awesome! thanks :+1:

kangman:
:+1:

jason2055141:
Hi @feiyang21687 How did you get that? What plugin are you using to collect data's or logs in containers? Can you please advice? Thanks.

feiyang21687:
Hi @jason2055141 We use `cAdvisor` to collect stats data of container, use `scribe` to collect logs of container, use --volume to store the logs of container on local host, use ELK and Graphite to visualize the data.

jason2055141:
@feiyang21687 
Thanks men. I will try that scribe to collects logs from container. Thanks again.

kerwin:
+1

trompx:
An elasticsearch storage driver would definitly be a great add ! instead of having to store data to redis first then forward to logstash to save it on elasticsearch.

[...]

PR #875, opened 2015-09-08
:beer: Add support for elasticsearch storage driver

hacpai:
An elasticsearch storage driver would definitly be a great add.
https://github.com/google/cadvisor/issues/634

I am interested in posting a PR to add elastic search as a backend with @feiyang21687 help.

cAdvisor was able to store metrics data in ElasticSearch and I use Kibana to create dashboard, and like this one:


rjnagal:
LGTM

wish ELK dependency had more fine grained packages for us to import :sad:

gregbkr:
Hello, thank you for adding the ELK support to cadvisor, that is a great news! ^^
This command work, and I can see the data in kibana (in index: cadvisor):
  docker run \
  --volume=/:/rootfs:ro \
  --volume=/var/run:/var/run:rw \
  --volume=/sys:/sys:ro \
  --publish=8080:8080 \
  --detach=true \
  --name=cadvisor \
  google/cadvisor:latest \
  -storage_driver="elasticsearch" -alsologtostderr=true -storage_driver_es_host="http://elasticsearch_ip:9200"

By any chance, anyone tried to make a dashboard of these data? I am stuck with the field container_Name, which container some "/" and is analysed (and cut). Usually I use the field container_Name.raw to escape but it is not available. Anyone could help, or share some dashboards?

Thanks a lot! and good weekend!
Greg.
visu_cadvisor

PR #876, opened 2015-09-08
use relative links on docker pages

basvdlei:
The links on the docker pages do not work when running under a different context. Links to static content is one level too high (`../../static/`) and container links are defined absolute `http:/xxx/docker/...`.

This PR makes the `/docker` pages behave the same way as `/containers`.


vishh:
ok to test

PR #882, opened 2015-09-11
Add flag -k to journalctl show only kernel messages

lborguetti:
Hi guys,

I have a problem when my application sends many logs for journald, the command "journalctl -f" generates a high consumption of CPU. In some cases killing the journald.

I added the flag -k in journalctl to show only the kernel logs.

Do you see any problem in this change?

Thanks


lborguetti:
I signed it!

vishh:
ok to test

PR #888, opened 2015-09-19
Validate systemd cgroup name as docker id

whosthatknocking:
cAdvisor (0.16.0.2) when started with default runtime options on system with systemd service manager calls Docker API for systemd service files.

```
Sep 18 20:06:12 docker: time="2015-09-18T20:06:12.299962820-07:00" level=info msg="GET /containers/rsyslog.service/json"
Sep 18 20:06:12 docker: time="2015-09-18T20:06:12.300056485-07:00" level=error msg="Handler for GET /containers/{name:.*}/json returned error: no such id: rsyslog.service"
Sep 18 20:06:12 docker: time="2015-09-18T20:06:12.300074752-07:00" level=error msg="HTTP Error" err="no such id: rsyslog.service" statusCode=404
```

The purpose of this PR is to remove unnecessary docker API calls by validating container names against currently defined container name patterns, `docker-` prefix and `.scope` suffix.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -136,10 +136,25 @@ func FullContainerName(dockerId string) string {
 	}
 }
 
+// Returns true if container name matches pattern
+func isContainerName(name string) bool {
+	id := path.Base(name)
+	if UseSystemd() {
+		return strings.HasPrefix(id, "docker-") && strings.HasSuffix(id, ".scope")
+	} else {
+		return true
+	}
+}
+
 // Docker handles all containers under /docker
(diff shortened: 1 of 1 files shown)


vishh:
ok to test

jimmidyson:
This is trying to achieve the same as #821. #821 also handles the case where a container is started with a different cgroup parent other than docker so prefer that PR over this.

jimmidyson:
@whosthatknocking Happy for me to close this & get #821 merged once it's fixed up the cgroup parent stuff?

PR #892, opened 2015-09-23
New version 0.17.0

rjnagal:


rjnagal:
@vishh

rjnagal:
Will add the tag once this is merged.

vishh:
LGTM. Instead of a tag, can we create a release branch on HEAD to manage patch releases?

rjnagal:
We never patch branches anyway. It seems like an overkill. I'd expect packaging managers to keep branches of their official releases.

PR #895, opened 2015-09-26
Fix docker containers relative links

jimmidyson:
A few missed links when running cadvisor under a different path, behind nginx or in a kubelet.


vishh:
add to whitelist

vishh:
LGTM

jimmidyson:
Self-merging after @vishh's LGTM

PR #896, opened 2015-09-26
Systemd docker container improvements, including customized cgroup parent

jimmidyson:
Includes commit from #821 & extends to include cgroup parent in full container name. #821 can be closed in preference to this.

/cc @basvdlei @whosthatknocking

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -44,7 +44,7 @@ var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path t
 
 // Regexp that identifies docker cgroups, containers started with
 // --cgroup-parent have another prefix than 'docker'
-var dockerCgroupRegexp = regexp.MustCompile(`[A-z]+-([a-z0-9]+)\.scope`)
+var dockerCgroupRegexp = regexp.MustCompile(`(.+)-([a-z0-9]{64})\.scope$`)
 
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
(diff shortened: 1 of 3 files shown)


container/docker/handler.go
```diff
@@ -312,7 +314,7 @@ func (self *dockerContainerHandler) ListContainers(listType container.ListType)
 		}
 
 		ref := info.ContainerReference{
-			Name:      FullContainerName(c.ID),
+			Name:      FullContainerName(c.ID, self.cgroupParent),
```
basvdlei: Is this not setting the cgroup parent the same for all containers? Since this for is looping over all docker containers (self.client is a reference to the Docker client), but self.cgroupParent is only for the current container.

container/docker/factory.go
```diff
@@ -42,6 +42,10 @@ var DockerNamespace = "docker"
 var dockerRootDir = flag.String("docker_root", "/var/lib/docker", "Absolute path to the Docker state root directory (default: /var/lib/docker)")
 var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path to the Docker run directory (default: /var/run/docker)")
 
+// Regexp that identifies docker cgroups, containers started with
+// --cgroup-parent have another prefix than 'docker'
+var dockerCgroupRegexp = regexp.MustCompile(`(.+)-([a-z0-9]{64})\.scope$`)
```
vishh: Did you test this on a systemd distro?

vishh:
@jimmidyson: Doesn't `--cgroup-parent` affect non-systemd systems as well? I don't see this PR addressing that.

vishh:
@cadvisorJenkinsBot: add to whitelist

container/docker/factory.go
jimmidyson: Yes. Fedora 22 with Docker 1.8.2

jimmidyson:
@vishh: `--cgroup-parent` does apply to non-systemd docker, correct, but I can't think of a simple way to identify that from the cgroup container name. This optimisation will only help with systemd, which is acceptable to me given the wider adoption of systemd now.

container/docker/factory.go
whosthatknocking: RH 7.1 with Docker 1.8.2

container/docker/handler.go
jimmidyson: This is actually redundant as this is only used for non-systemd Dockers where `self.cgroupParent` is empty.

jimmidyson:
Closing - #821 sorted this all out :)

PR #906, opened 2015-10-06
Registration of custom function to map container name to prometheus labels

jimmidyson:
This helps with the prometheus metrics in Kubernetes where the container name has info on namespace, pod name & container name. Disabled by default, but useful when embedded inside another app, like the kubelet.

/cc @vishh


vishh:
LGTM

jimmidyson:
Thanks again!

PR #907, opened 2015-10-06
Control cpu usage when getting disk usage.

### LINKED ISSUE #898: Increase the period of getting container filesystem/network stats
Frequent checking could cause high cpu usage, as reported by a kubernetes user in https://github.com/kubernetes/kubernetes/issues/10451#issuecomment-143255549

edit: The current housekeeping period is 1s

/cc @vishh

vishh:
1. Run `du` less often
2. Set `ulimit -t` just in case.

While I was at it, I also refactor docker storage path handling a bit. A lot of cleanup is still pending :(

I intend to add support for `overlay` and `device mapper` storage drivers in a subsequent PR.

Fixes #898 (I guess)

cc @yujuhong

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -30,7 +30,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -62,12 +61,7 @@ type RealFsInfo struct {
 	labels map[string]string
 }
 
-type Context struct {
-	// docker root directory.
-	DockerRoot string
-}
-
-func NewFsInfo(context Context) (FsInfo, error) {
+func NewFsInfo() (FsInfo, error) {
 	mounts, err := mount.GetMounts()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


fs/fs.go
```diff
@@ -282,7 +259,7 @@ func (self *RealFsInfo) GetDirFsDevice(dir string) (*DeviceInfo, error) {
 }
 
 func (self *RealFsInfo) GetDirUsage(dir string) (uint64, error) {
-	out, err := exec.Command("du", "-s", dir).CombinedOutput()
+	out, err := exec.Command("/bin/sh", "-c", "ulimit -t 1; du -s", dir).CombinedOutput()
```
jimmidyson: Isn't it the larger docker containers that will take longer? Setting the cpu time this low will likely mean no fs stats for large containers, which reduces the usefulness of this check: finding which containers are using up your disk.

jimmidyson: Perhaps `nice`ing it would be better?

vishh: Setting `nice` values is a good idea.

validate/validate.go
```diff
@@ -19,16 +19,18 @@ package validate
...
 	"fmt"
-	"github.com/google/cadvisor/manager"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"path"
 	"strings"
 
+	"github.com/google/cadvisor/manager"
+
```
yujuhong: remove the extra line

fs/types.go
```diff
@@ -63,4 +63,6 @@ type FsInfo interface {
 
 	// Returns the mountpoint associated with a particular device.
 	GetMountpointForDevice(device string) (string, error)
+
+	AddLabels(context Context)
```
yujuhong: Needs a one line comment to be consistent with all the functions above

fs/fs.go
yujuhong: +1. I don't know if it's useful to set a hard 1-second limit as @jimmidyson mentioned.

fs/fs.go
```diff
@@ -282,7 +259,7 @@ func (self *RealFsInfo) GetDirFsDevice(dir string) (*DeviceInfo, error) {
 }
 
 func (self *RealFsInfo) GetDirUsage(dir string) (uint64, error) {
-	out, err := exec.Command("du", "-s", dir).CombinedOutput()
+	out, err := exec.Command("/bin/sh", "-c", "nice -n 19 du -s", dir).CombinedOutput()
```
jimmidyson: Probably don't need to wrap it in `/bin/sh` now as it's not a binary, not a shell builtin. This will also help with the args - `dir` is outside of the `-c` param to `sh` right now & probably won't be used at all in this call.

I'd go with `out, err := exec.Command("nice", "-n", "19", "du", "-s", dir).CombinedOutput()`

Unless I'm missing it, could you add a test for this too? Should be pretty easy to set up a file with known file in an otherwise empty dir & run `GetDirUsage` on it to check.

fs/types.go
vishh: Done

validate/validate.go
vishh: Fixed

fs/fs.go
vishh: Done. I added an integration test.

vishh:
I'm having lots of difficulty in debugging integration test failures. To move forward with this PR, I added an unit test as @jimmidyson suggested. I will add the integration tests in a subsequent PR.

vishh:
Looks like the refactoring is breaking cadvisor. While I debug this, I opened #909 to handle the primary issue this PR is intended to solve.

vishh:
I'm closing this for now.

PR #910, opened 2015-10-08
Compute fs usage for docker containers less often.

vishh:
`du` can cause a lot of cpu usage.

#898

### DIFF
--- a/container/docker/fsHandler.go
+++ b/container/docker/fsHandler.go
@@ -1,0 +1,109 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Handler for Docker containers.
+package docker
+
+import (
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/cadvisor/fs"
+)
+
+type fsHandler interface {
+	start()
+	usage() uint64
+	stop()
+}
+
+type realFsHandler struct {
+	sync.RWMutex
+	lastUpdate  time.Time
+	usageBytes  uint64
+	period      time.Duration
+	storageDirs []string
+	fsInfo      fs.FsInfo
+	// Tells the container to stop.
+	stopChan chan struct{}
+}
+
+const longDu = time.Second
+
+var _ fsHandler = &realFsHandler{}
+
+func newFsHandler(period time.Duration, storageDirs []string, fsInfo fs.FsInfo) fsHandler {
+	return &realFsHandler{
+		lastUpdate:  time.Time{},
+		usageBytes:  0,
+		period:      period,
+		storageDirs: storageDirs,
+		fsInfo:      fsInfo,
+		stopChan:    make(chan struct{}, 1),
+	}
+}
+
+func (fh *realFsHandler) needsUpdate() bool {
+	return time.Now().After(fh.lastUpdate.Add(fh.period))
+}
+
+func (fh *realFsHandler) update() error {
+	var usage uint64
+	for _, dir := range fh.storageDirs {
+		// TODO(Vishh): Add support for external mounts.
+		dirUsage, err := fh.fsInfo.GetDirUsage(dir)
+		if err != nil {
+			return err
+		}
+		usage += dirUsage
+	}
+	fh.Lock()
+	defer fh.Unlock()
+	fh.lastUpdate = time.Now()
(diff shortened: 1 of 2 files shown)


vishh:
cc @yujuhong

jimmidyson:
1m frequency sounds good to me, but might be a problem for containers that have written a lot of data - see https://github.com/kubernetes/kubernetes/issues/10451#issuecomment-143327536.

Should there be a check to see if `du` is already running so it doesn't overlap with a previous run?

BTW, this idea has got me thinking about a new way to collect metrics in cadvisor, each check individually scheduleable & feeding data back over a channel to central collector.

vishh:
Individual container housekeeping is synchronous. There will not be multiple stat aggregations in flight for a given container.
The memory cache is being used for synchronization as of now.

jimmidyson:
So does that mean that the length of time for `du` to run will impact on cpu/memory stats collection? Is that a problem?

yujuhong:
> 1m frequency sounds good to me, but might be a problem for containers that have written a lot of data - see kubernetes/kubernetes#10451 (comment).

The actual call would be blocked until finished, so it's not worse than today...

> BTW, this idea has got me thinking about a new way to collect metrics in cadvisor, each check individually scheduleable & feeding data back over a channel to central collector.

I am fine with the idea, but would like to make sure cadvisor doesn't consume too much resources.

> So does that mean that the length of time for du to run will impact on cpu/memory stats collection? Is that a problem?

This is not a new problem, but I am also curious. @vishh, would this affect heapster?

vishh:
@jimmidyson @yujuhong: Thanks for the review. I updated the fs handler to run in a separate go thread. This should prevent default housekeeping from being blocked by `du`.

vishh:
@cadvisorJenkinsBot: test this please

vishh:
Let's ignore travis since [jenkins](http://104.154.52.74/job/cadvisor-e2e/1863/) has already run the unit tests.

container/docker/fsHandler.go
```diff
@@ -0,0 +1,113 @@
...
+	fh.lastUpdate = time.Now()
+	fh.usageBytes = usage
+	return nil
+}
+
+func (fh *realFsHandler) trackUsage() {
+	for {
+		start := time.Now()
+		select {
+		case <-fh.stopChan:
```
yujuhong: nit (optional): since here it only checks for a stop signal, you can close the channel at stop(). The code here can then be simplified to check whether a channel is closed.

container/docker/fsHandler.go
```diff
@@ -0,0 +1,113 @@
...
+	start()
+	usage() uint64
+	stop()
+}
+
+type realFsHandler struct {
+	sync.RWMutex
+	lastUpdate  time.Time
+	usageBytes  uint64
+	frequency   time.Duration
```
yujuhong: nit: frequency seems to suggest "rate". perhaps switching to "period" or "interval"?

container/docker/fsHandler.go
```diff
@@ -0,0 +1,113 @@
...
+	storageDirs []string
+	fsInfo      fs.FsInfo
+	// Tells the container to stop.
+	stopChan chan bool
```
yujuhong: nit (optional): use `struct{}` instead.

container/docker/fsHandler.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
```
yujuhong: s/2014/2015

container/docker/fsHandler.go
vishh: Done.

container/docker/fsHandler.go
vishh: Done

container/docker/fsHandler.go
vishh: Done

container/docker/fsHandler.go
vishh: Done

vishh:
@yujuhong: comments addressed. PTAL

vishh:
@yujuhong: Got it. Updated.

yujuhong:
lgtm

vishh:
Thanks for the review! Merging this PR.

container/docker/fsHandler.go
```diff
@@ -0,0 +1,109 @@
...
+			return
+		}
+		if err := fh.update(); err != nil {
+			glog.V(2).Infof("failed to collect filesystem stats - %v", err)
+		}
+		duration := time.Since(start)
+		if duration > longDu {
+			glog.V(3).Infof("`du` on following dirs took %v: %v", duration, fh.storageDirs)
+		}
+		next := start.Add(fh.period)
```
jimmidyson: Wouldn't this be better off using a select on `time.After` & stop channel?

jimmidyson: I'll send in a PR later

container/docker/fsHandler.go
```diff
@@ -0,0 +1,109 @@
...
+		time.Sleep(next.Sub(time.Now()))
+	}
+}
+
+func (fh *realFsHandler) start() {
+	go fh.trackUsage()
+}
+
+func (fh *realFsHandler) stop() {
```
jimmidyson: I'm missing where `stop` is actually called?

PR #916, opened 2015-10-13
Switch to normal static web resources & use go-rice for embedding

### LINKED ISSUE #879: minified/obfuscated resources in "pages/static"; privacy breach
Files in [pages/static](../tree/master/pages/static) contain embedded JavaScript and CSS resources in minified obfuscated form. This is a bad practice just like committing pre-build binaries.
Those files should be composed on build time.
Minified resources are same as pre-compiled binaries -- un-readable, non-modifiable and non-distributable in [Debian](http://debian.org).

Lately it's been [demonstrated](https://zyan.scripts.mit.edu/blog/backdooring-js/) how minified JavaScript can become an attack vector (i.e. security breach).

It is hard to check whether minified resources were tampered with.

Finally there is a privacy breach as resources bundled to `pages/static` have references to external web sites such as _google.com_ and _ajax.googleapis.com_.

Please replace minified resources with original uncompressed ones (please make sure that there are no external references), compose static `.go` resources on build-time and avoid committing minified junk to source tree.

Thanks.

jimmidyson:
Fixes #879


jimmidyson:
@mvdan @vishh @onlyjob Thoughts? Much prefer this way of working, easier to work with the web assets. Does require an internet connection though to be able to use Google jsapi & chart API - Google's ToS say so...

onlyjob:
I did not test this approach but from the brief look at 9f52cad it appears to be much nicer indeed. The only minor concern is that there are 5 new go(.rice) dependencies in _Godeps_ but that's not an issue.

As for charts I'm quite happy that source-less files are no longer bundled.
ToS compliance is quite important as well but privacy breach (i.e. unconditional load of 3rd party JS) is still there. Maybe some other charting library like D3 could be a potential replacement? Or something from [this list](http://www.sitepoint.com/15-best-javascript-charting-libraries/) as long as free and open source alternative is chosen. Google JSAPI and Charts are not free software so I hope they can be replaced eventually...

jimmidyson:
@vishh Any comments?

vishh:
@jimmidyson: This change LGTM as long as it doesn't change existing functionality.

timstclair:
Is it possible to exclude directories or files in Godeps? I don't like that we're pulling in that `doge.jpg` (why is there a dependency on the examples directory anyway?).

http/handlers.go
```diff
@@ -53,36 +53,32 @@ func RegisterHandlers(mux httpMux.Mux, containerManager manager.Manager, httpAut
...
 		secrets := auth.HtpasswdFileProvider(httpAuthFile)
-		authenticator := auth.NewBasicAuthenticator(httpAuthRealm, secrets)
-		mux.HandleFunc(static.StaticResource, authenticator.Wrap(staticHandler))
-		if err := pages.RegisterHandlersBasic(mux, containerManager, authenticator); err != nil {
-			return fmt.Errorf("failed to register pages auth handlers: %s", err)
-		}
-		authenticated = true
+		authenticator = auth.NewBasicAuthenticator(httpAuthRealm, secrets)
 	}
 	if httpAuthFile == "" && httpDigestFile != "" {
```
timstclair: (nit) prefer: `} else if httpDigestFile != "" {`

pages/static/static.go
```diff
@@ -17,43 +17,15 @@
...
-	// Set Content-Type if we were able to detect it.
-	contentType := mime.TypeByExtension(path.Ext(resource))
-	if contentType != "" {
-		w.Header().Set("Content-Type", contentType)
-	}
-
-	_, err := w.Write([]byte(content))
-	return err
+func FileServer() http.Handler {
+	box := rice.MustFindBox("")
```
timstclair: I think it would be better to use `FindBox` here and return the error.

timstclair:
This was superseded by https://github.com/google/cadvisor/pull/1217 and https://github.com/google/cadvisor/pull/1245. If you think we should be using go-rice instead, let's open a new discussion.

PR #923, opened 2015-10-16
Drop C code for FS sizes

jimmidyson:


jimmidyson:
@mvdan @vishh

vishh:
Why this change?

jimmidyson:
Just to tidy up the C code as it's provided by the std lib directly.

vishh:
LGTM

PR #926, opened 2015-10-19
Support devicemapper storage for docker images dir

### LINKED ISSUE #920: No disk capacity/usage is reported for Docker using the thinpool lvm devicemapper storage driver
Based on comments I see in the code, this is not implemented yet:
https://github.com/google/cadvisor/blob/master/fs/fs.go#L108
https://github.com/google/cadvisor/blob/master/fs/fs.go#L116

However, it's really important for Kubernetes image garbage collection to work on systems that typically use the thinpool lvm storage driver like RHEL or Fedora.

jimmidyson:
Fixes #920

This currently only implements docker storage dir usage & not container FS usage.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -81,11 +81,17 @@ func RootDir() string {
 	return *dockerRootDir
 }
 
+type storageDriver string
+
+const (
+	devicemapperStorageDriver storageDriver = "devicemapper"
+	aufsStorageDriver         storageDriver = "aufs"
+)
+
 type dockerFactory struct {
 	machineInfoFactory info.MachineInfoFactory
 
-	// Whether docker is running with AUFS storage driver.
-	usesAufsDriver bool
+	storageDriver storageDriver
 
 	client *docker.Client
 
@@ -110,7 +116,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		name,
 		self.machineInfoFactory,
 		self.fsInfo,
-		self.usesAufsDriver,
+		self.storageDriver,
 		&self.cgroupSubsystems,
 		inHostNamespace,
 	)
@@ -184,6 +190,10 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
 
 // Register root container before running this function!
 func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
+	if UseSystemd() {
+		glog.Infof("System is using systemd")
+	}
+
 	client, err := Client()
 	if err != nil {
 		return fmt.Errorf("unable to communicate with docker daemon: %v", err)
@@ -207,32 +217,16 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 	}
 
 	// Check that the libcontainer execdriver is used.
-	information, err := client.Info()
+	information, err := DockerInfo()
 	if err != nil {
 		return fmt.Errorf("failed to detect Docker info: %v", err)
 	}
-	usesNativeDriver := false
-	for _, val := range *information {
-		if strings.Contains(val, "ExecutionDriver=") && strings.Contains(val, "native") {
-			usesNativeDriver = true
-			break
-		}
-	}
-	if !usesNativeDriver {
+	execDriver, ok := information["ExecutionDriver"]
+	if !ok || !strings.HasPrefix(execDriver, "native") {
 		return fmt.Errorf("docker found, but not using native exec driver")
 	}
 
-	usesAufsDriver := false
-	for _, val := range *information {
-		if strings.Contains(val, "Driver=") && strings.Contains(val, "aufs") {
-			usesAufsDriver = true
-			break
-		}
-	}
-
-	if UseSystemd() {
-		glog.Infof("System is using systemd")
-	}
+	sd, _ := information["Driver"]
 
 	cgroupSubsystems, err := libcontainer.GetCgroupSubsystems()
 	if err != nil {
@@ -243,7 +237,7 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 	f := &dockerFactory{
 		machineInfoFactory: factory,
 		client:             client,
-		usesAufsDriver:     usesAufsDriver,
+		storageDriver:      storageDriver(sd),
 		cgroupSubsystems:   cgroupSubsystems,
 		fsInfo:             fsInfo,
 	}
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -62,9 +62,9 @@ type dockerContainerHandler struct {
 	// Manager of this container's cgroups.
 	cgroupManager cgroups.Manager
 
-	usesAufsDriver bool
-	fsInfo         fs.FsInfo
-	storageDirs    []string
+	storageDriver storageDriver
+	fsInfo        fs.FsInfo
+	storageDirs   []string
 
 	// Time at which this container was created.
 	creationTime time.Time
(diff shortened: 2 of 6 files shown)


jimmidyson:
@vishh Noticed that there's not a CentOS vm in integration tests to test this on. Any chance of adding one in? All the current integration test platforms use aufs. Well, unsure about coreos tbh.

vishh:
I think CoreOS uses btrfs. I want to switch to a newer version of CoreOS with overlay support. 
I will add a CentOS node today. Do we need to test against fedora as well?

container/docker/factory.go
```diff
@@ -207,32 +217,16 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
...
-		if strings.Contains(val, "Driver=") && strings.Contains(val, "aufs") {
-			usesAufsDriver = true
-			break
-		}
-	}
-
-	if UseSystemd() {
-		glog.Infof("System is using systemd")
-	}
+	sd, _ := information["Driver"]
```
vishh: Just curious: Has docker every changed the value for this key in the past? 
Previously the code was performing `string.Contains` and now we use the value as-is. Hence the concern.

container/docker/handler.go
```diff
@@ -127,14 +127,15 @@ func newDockerContainerHandler(
...
+		storageDriver:      storageDriver,
 		fsInfo:             fsInfo,
 		rootFs:             rootFs,
 		storageDirs:        storageDirs,
 		fsHandler:          newFsHandler(time.Minute, storageDirs, fsInfo),
 	}
 
-	if usesAufsDriver {
+	switch storageDriver {
+	case "aufs":
```
vishh: nit: `case aufsStorageDriver`

container/docker/handler.go
```diff
@@ -228,9 +229,8 @@ func (self *dockerContainerHandler) GetSpec() (info.ContainerSpec, error) {
 
 	spec := libcontainerConfigToContainerSpec(libcontainerConfig, mi)
 	spec.CreationTime = self.creationTime
-	if self.usesAufsDriver {
-		spec.HasFilesystem = true
-	}
+	// For now only enable for aufs filesystems
+	spec.HasFilesystem = self.storageDriver == "aufs"
```
vishh: nit: `self.storageDriver == aufsStorageDriver`

fs/fs.go
```diff
@@ -295,3 +325,83 @@ func getVfsStats(path string) (uint64, uint64, uint64, error) {
...
+	err := json.Unmarshal([]byte(driverStatus), &config)
+	if err != nil {
+		return "", 0, 0, err
+	}
+	poolName := dockerStatusValue(config, "Pool Name")
+	if len(poolName) == 0 {
+		return "", 0, 0, fmt.Errorf("Could not get dm pool name")
+	}
+
+	poolName = path.Join("/dev/mapper", poolName)
```
vishh: Question: Should this be relative to `/rootfs`?

fs/fs.go
```diff
@@ -295,3 +325,83 @@ func getVfsStats(path string) (uint64, uint64, uint64, error) {
...
+
+	dmTable, err := exec.Command("dmsetup", "table", poolName).Output()
```
vishh: Is `dmsetup` available on the docker base image that cAdvisor is using?

fs/fs.go
```diff
@@ -295,3 +325,83 @@ func getVfsStats(path string) (uint64, uint64, uint64, error) {
...
+		"%d %d %s %d:%d %d:%d %d %d %d %s",
+		&bkt, &bkt, &bkts, &bkt, &bkt, &major, &minor, &bkt, &bkt, &bkt, &bkts)
+	if err != nil {
+		return "", 0, 0, err
+	}
+	return poolName, major, minor, nil
+}
+
+func getDMStats(poolName string) (uint64, uint64, uint64, error) {
+	dmTable, err := exec.Command("dmsetup", "table", poolName).Output()
```
vishh: Why run this command again?

vishh: Nvm. I misread the code.

fs/fs.go
```diff
@@ -295,3 +325,83 @@ func getVfsStats(path string) (uint64, uint64, uint64, error) {
...
+	}
+
+	var (
+		dataBlkSize, bkt uint64
+		bkts             string
+	)
+
+	_, err = fmt.Fscanf(bytes.NewReader(dmTable),
+		"%d %d %s %d:%d %d:%d %d %d %d %s",
+		&bkt, &bkt, &bkts, &bkt, &bkt, &bkt, &bkt, &dataBlkSize, &bkt, &bkt, &bkts)
```
vishh: Should we cache the value of `dataBlkSize` from the previous call to `dmsetup table <poolName>` instead?

manager/manager.go
```diff
@@ -1188,19 +1193,13 @@ func (m *manager) DockerInfo() (DockerStatus, error) {
 			out.NumContainers = n
 		}
 	}
-	// cut, trim, cut - Example format:
-	// DriverStatus=[["Root Dir","/var/lib/docker/aufs"],["Backing Filesystem","extfs"],["Dirperm1 Supported","false"]]
 	if val, ok := info["DriverStatus"]; ok {
+		var driverStatus [][]string
```
vishh: Thanks for the cleanup :exclamation:

container/docker/factory.go
jimmidyson: The free form options are a bit scary but there is no other way AFAIK to get this info out of Docker. The use of `Contains` was previously unnecessary - the keys/values _should_ never have changed... I'd prefer to list the options more explicitly rather than using `Contains` anyway if need be, but this should work fine.

fs/fs.go
jimmidyson: No... damn that's killed this approach really then. Or can we add in dmsetup to the cadvisor image?

fs/fs.go
jimmidyson: Yep

container/docker/factory.go
vishh: Acknowledged.

fs/fs.go
vishh: We can add it. 
I wonder if switching to alpine linux base image is necessary here. There is an [alpine linux package]( https://pkgs.alpinelinux.org/package/main/armhf/device-mapper) for this purpose

jimmidyson:
@vishh If you could get a fedora test env up as well that would be awesome.

jimmidyson:
I love alpine so happy to make the switch. Gives us more options in future as well I think.

vishh:
@jimmidyson: I have added a CentOS and RHEL v7.1 instances added. Are you targeting this PR for kubernetes v1.1?

jimmidyson:
@vishh Yes I'm targeting this for 1.1 & OpenShift 3.1

jimmidyson:
@vishh I've switched to alpine & added devicemapper tools to both release & canary Dockerfiles

ncdc:
cc @smarterclayton

deploy/Dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM progrium/busybox
-MAINTAINER dengnan@google.com vmarmol@google.com vishnuk@google.com
+FROM alpine:3.2
```
vishh: Did you get a chance to test this change?

AFAIK our integration tests do not use the docker images as of now.

vishh:
LGTM. Feel free to self-merge once you test the docker image changes.

jimmidyson:
Tested the Docker image, works great.

jimmidyson:
@ncdc @smarterclayton Any comments before I merge?

ncdc:
I won't be able to review until tomorrow morning at the earliest.

jimmidyson:
@ncdc Any chance of a review? Would like to merge ASAP & get a PR in to k8s to update cadvisor dep.

ncdc:
@jimmidyson sorry, I probably can't get to it this morning :disappointed: If it works for you, I have no objections. I'm assuming that, worst case, this isn't any *worse* than the current code in master, right? :smile:

jimmidyson:
@ncdc Of course it works - for me at least ;)

jimmidyson:
Merging...

Issue #929, opened 2015-10-20
Unmarshalling error when `AdditionalGroups` field in libcontainer config.go is out of sync with Docker


pmorie:
I discovered during implementation of supplemental group support in Kubernetes that `AdditionalGroups` field has diverged between Docker and cadvisor:

* [version in cadvisor](https://github.com/google/cadvisor/blob/master/Godeps/_workspace/src/github.com/docker/libcontainer/configs/config.go#L87)
* [version in Docker](https://github.com/docker/docker/blob/master/vendor/src/github.com/opencontainers/runc/libcontainer/configs/config.go#L148)

When cadvisor tries to read the config laid down by a Docker that vendors a different version of the config, an unmarshalling error occurs.

This hasn't been an issue so far in Kubernetes because we haven't used the `AdditionalGroups` field in the docker API yet.  How has this kind of change been handled in the past?  It seems like cadvisor has to be able to parse whatever Docker might lay down.

jimmidyson:
@pmorie Do you know what version of Docker the type of `AdditionalGroups` was changed in? Wondering if we can just update cadvisor's config to string & forget about handling old format.

jimmidyson:
Fixed in #930 by bumping libcontainer again. I wouldn't worry about the old format as it didn't look like it worked properly anyway.

PR #930, opened 2015-10-20
Docker client update including reuse

jimmidyson:


jimmidyson:
Also fixes #929

vishh:
Other than the outstanding comment, this PR LGTM.

Thanks a lot @jimmidyson for this PR! :)

jimmidyson:
Switched to `sync.Once` as suggested by @vishh. This will not try to recreate client if it ever fails but I think that's OK.

vishh:
My bad. I forgot about a change that I have been meaning to do. 
As of now cAdvisor requires docker daemon to be up before it starts up. I want to change this behavior soon, because it makes kubelet internal architecture complex. 
I'm OK with this PR for now, and altering the current behavior should be easy in the future thanks to this PR!

vishh:
Ignore my previous comment. Since client creation does not require the daemon to be up, as long as client creation doesn't fail, we should be fine. Sorry about the noise!

vishh:
LGTM.

dchen1107:
LGTM

yujuhong:
`sync.Once` looks good. Thanks for fixing this!

dchen1107:
@jimmidyson Are you going to cut a release with this one or you want to include devicemapper fix too?

jimmidyson:
@dchen1107 I want the devicemapper enhancement too if you're OK to wait?

PR #933, opened 2015-10-21
there shows empty data for network metrics in k8s cluster, the root cause is that ,the network type blocks the metrics data

jekyang:
if the network mode is --net=container:name, cadvisor will consider it  not having network, then in the case of k8s, network metrics from pod containers will be lost.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -21,7 +21,7 @@ import (
 	"path"
 	"strings"
 	"time"
-
+	"regexp"
 	"github.com/docker/libcontainer/cgroups"
 	cgroup_fs "github.com/docker/libcontainer/cgroups/fs"
 	libcontainerConfigs "github.com/docker/libcontainer/configs"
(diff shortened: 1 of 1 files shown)


jimmidyson:
This is on purpose. In Kubernetes it is the network of the infra container (`POD`) in the pod that handles all networking so it should only be reported by that container, otherwise metrics are duplicated.

jekyang:
I just saw that kubelet  will export the metrics by api /stats/container, /stats/pod/containers /stats/<namespace>/<pod name>/<uid>/<container name>, but the data from kubelet&cadvisor is also empty even i get it manually during debugging.  by "docker inspect" the pod container, there's empty value in HostConfig.NetworkMode, so there's something should be fixed by kubelet?

jimmidyson:
Can you share the `docker inspect` output? An empty `NetworkMode` seems odd.

vishh:
Kubelet metrics APIs are being developed. Once that's done, network metrics will be exposed as part of POD metrics.
As of now, the data that cadvisor exposes per container is assumed to be local to a container. 
You should see an `infra` container for every pod in k8s that cadvisor does expose. That container should have network metrics.

jimmidyson:
@jekyang Can I close this? Are you happy with the discussion & explanation above?

jimmidyson:
Closing - feel free to reopen if you think this is still a problem.

PR #934, opened 2015-10-21
Fix imported package names to not use mixedCaps or under_scores

mqliang:
All package names should be lower case, no under_scores or mixedCaps. The same rule applies to renaming imported packages. (https://blog.golang.org/package-names).

This PR helps to fix this.


jimmidyson:
:+1: but needs a rebase.

PR #937, opened 2015-10-22
Add make release target

jimmidyson:
@vishh This means you can just do `make release` & tag will be created & binaries uploaded. Will need to create changelog manually for now, but thinking how we automate that with something like https://github.com/skywinder/github-changelog-generator in future.


jimmidyson:
Flake...

jimmidyson:
No code changes so self-merging.

vishh:
LGTM

vishh:
We usually publish a few checksums for the binary. Some users have explicitly asked for it. Can you update the 0.19.1 release with the checksums?

PR #942, opened 2015-10-27
Drop regexp for net stats parsing

jimmidyson:
Reported in kubernetes/kubernetes#16296

/cc @vishh @yujuhong

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


container/libcontainer/helpers.go
```diff
@@ -166,22 +167,16 @@ func scanInterfaceStats(netStatsFile string) ([]info.InterfaceStats, error) {
...
 
-			_, err := fmt.Sscanf(line, "%s %d %d %d %d %d %d %d %d %d %d %d %d %d %d %d %d",
-				&i.Name, &i.RxBytes, &i.RxPackets, &i.RxErrors, &i.RxDropped, &bkt, &bkt, &bkt,
-				&bkt, &i.TxBytes, &i.TxPackets, &i.TxErrors, &i.TxDropped, &bkt, &bkt, &bkt, &bkt)
+		i := info.InterfaceStats{}
 
-			if err != nil {
-				return stats, fmt.Errorf("failure opening %s: %v", netStatsFile, err)
-			}
+		numRead, _ := fmt.Sscanf(line, netstatsLine,
```
yujuhong: Do we not care about the error anymore?...

jimmidyson: I don't think so. we've handled the file read error so we have a string that we're just trying to match.

vishh: Wouldn't Sscanf return an error if the expected number of elements, as specified in the `format`, is not present?

jimmidyson:
Switched to using the error rather than the number of fields read.

jimmidyson:
retest this please

jimmidyson:
retest this please

jimmidyson:
retest this please

jimmidyson:
Tested myself & works fine against coreos beta...

yujuhong:
lgtm. Thanks for the fix!

jimmidyson:
Merging after LGTM.

PR #957, opened 2015-11-09
Include log usage for aufs driver.

vishh:
Signed-off-by: Vishnu Kannan <vishnuk@google.com>

cc @yujuhong


vishh:
@jimmidyson: PTAL.

jimmidyson:
After recent discussion, happy with this approach.

One thing: could you enable the storage auditing regardless of the storage driver & only add the aufs dir if aufs is being used? Right now this only works for aufs storage driver but we could still track log usage for any storage drivers.

vishh:
@jimmidyson: Makes sense. Updated.

jimmidyson:
LGTM

vishh:
I appreciate the thorough review @jimmidyson!

jimmidyson:
No worries at all!

Issue #947, opened 2015-11-03
Handle non-init systemd process


vishh:
If systemd is not running as the init daemon, docker does not use it for container cgroup management. However, cAdvisor on the other hand, assumes docker is using systemd even if it is not the init daemon.

There seems to be use cases for running systemd as a regular non-init process :)

cAdvisor needs to handle this scenario.

andyzheng0831:
cc/ @andyzheng0831

andyzheng0831:
It will be great if we can fix this issue soon and make it merged in next k8s release post 1.1

dyema:
Regardless of whether systemd is init or not, I think it's desirable for cAdvisor to support the configuration where systemd is not managing cgroups for docker containers.

andyzheng0831:
@vishh May I ask about the status of bug? As we already cut k8s 1.1.1 release, would it possible for us to speed up fixing this issue, so that this will unblock our effort ASAP.

vishh:
#960 is out for review. Once that is merged, I can cherry-pick that into
K8s v1.1 branch.

dashpole:
closing via #960

PR #960, opened 2015-11-13
Add a `--nosystemd` flag 

vishh:
Setting this flag to true, forces cAdvisor to not assume systemd to be the cgroups owner for docker containers.

#947

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -46,6 +46,8 @@ var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path t
 // --cgroup-parent have another prefix than 'docker'
 var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
+var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
+
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
 func DockerStateDir() string {
@@ -53,13 +55,14 @@ func DockerStateDir() string {
 }
 
 // Whether the system is using Systemd.
-var useSystemd bool
+var useSystemd = false
 var check = sync.Once{}
 
 func UseSystemd() bool {
 	check.Do(func() {
-		useSystemd = false
-
+		if *noSystemd {
+			return
+		}
 		// Check for system.slice in systemd and cpu cgroup.
 		for _, cgroupType := range []string{"name=systemd", "cpu"} {
 			mnt, err := cgroups.FindCgroupMountpoint(cgroupType)


container/docker/factory.go
```diff
@@ -24,7 +24,6 @@ import (
 	"sync"
 
 	"github.com/docker/libcontainer/cgroups"
-	"github.com/fsouza/go-dockerclient"
```
jimmidyson: Doesn't this cause compilation failure for you?

vishh: Yeah. Fixed it now.

jimmidyson:
So the use case for this is systems where systemd is running but docker is not run via systemd?

container/docker/factory.go
```diff
@@ -46,6 +46,8 @@ var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path t
 // --cgroup-parent have another prefix than 'docker'
 var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
+var noSystemd = flag.Bool("nosystemd", false, "Don't assume systemd is being used by docker.")
```
jimmidyson: How about something like `Explicitly disable  systemd support for Docker cgroups`?

container/docker/factory.go
```diff
@@ -59,17 +61,18 @@ var check = sync.Once{}
 func UseSystemd() bool {
 	check.Do(func() {
 		useSystemd = false
```
jimmidyson: Isn't this a noop consisting when we run this?

container/docker/factory.go
```diff
@@ -59,17 +61,18 @@ var check = sync.Once{}
...
-		for _, cgroupType := range []string{"name=systemd", "cpu"} {
-			mnt, err := cgroups.FindCgroupMountpoint(cgroupType)
-			if err == nil {
-				// systemd presence does not mean systemd controls cgroups.
-				// If system.slice cgroup exists, then systemd is taking control.
-				// This breaks if user creates system.slice manually :)
-				if utils.FileExists(path.Join(mnt, "system.slice")) {
-					useSystemd = true
-					break
+		if !*noSystemd {
```
jimmidyson: Can we just return here rather than inventing the whole next block?

container/docker/factory.go
vishh: Done

container/docker/factory.go
vishh: Done

container/docker/factory.go
vishh: I don't get your comment? Why is it a no-op?

vishh:
@cadvisorJenkinsBot: retest this please

vishh:
@jimmidyson: PTAL.

jimmidyson:
Thanks @vishh. LGTM

PR #964, opened 2015-11-20
ClockSpeed not defined in /proc/cpuinfo for AArch64, so return 0.

yghannam:
Signed-off-by: Yazen Ghannam <yazen.ghannam@linaro.org>


jimmidyson:
I wonder if we'd be better using build constraints for stuff like this? Tiny performance improvement, but possibly worth exploring in this case.

yghannam:
CLA signed through Linaro.

@jimmidyson 
This fixes a runtime error so I'm not sure that a build constraint will be easy to do in this case. I think there may be some other parts that expect machineInfo.CpuFrequency to be defined.

jimmidyson:
@yghannam Understood. LGTM. Thanks very much!

jimmidyson:
ok to test

jimmidyson:
retest this please

jimmidyson:
Merging. Thanks @yghannam!

PR #966, opened 2015-11-21
[Godeps] changed docker/libcontainer dependency to runc/libcontainer

afein:
This PR changes the github.com/docker/libcontainer dependency to github.com/opencontainers/runc/libcontainer

This change was necessary in order to build CAdvisor on an ARM architecture, as the old libcontainer dependency enforces the x86_64 architecture.
For future reference, an ARM-based Docker image with CAdvisor compiled on a Raspberry Pi is available at afein/rpi-cadvisor

Steps taken to generate this set of changes on a clean GOPATH:
- godep restore
- rm -rf Godeps
- go get github.com/opencontainers/runc
- point all libcontainer imports to opencontainers/runc/libcontainer
- godep save ./...


Godeps/Godeps.json
```diff
@@ -29,32 +29,31 @@
...
+		{
+			"ImportPath": "github.com/coreos/go-systemd/util",
+			"Comment": "v4",
+			"Rev": "b4a58d95188dd092ae20072bac14cece0e67c388"
 		},
 		{
 			"ImportPath": "github.com/docker/docker/pkg/mount",
-			"Comment": "v1.8.3-1-g2732df1",
-			"Rev": "2732df1dfb35f797a90e163afffe23564600407b"
+			"Comment": "v1.4.1-4831-g0f5c9d3",
```
jimmidyson: Why is the docker package being downgraded?

jimmidyson:
Thank you for sending this PR in. Definitely needs to be done some time.

This seems to be changing more than just the runc dependencies. Can you confirm that those other dependency changes are required? Most importantly the downgrade of the docker packages is concerning.

Do you know what impact this has on kubernetes, if any? Has kubernetes migrated to runc? I guess we need to keep that aligned, or doesn't it matter?

Does this have any impact on docker versions we support? I'm not sure what our compatibility policy is but runc is not used on older dockers. Will this cause problems?

afein:
It appears that the docker downgrade to v1.4.1-4831-g0f5c9d3 was performed because of runc's dependency on that version. I will edit this PR to disregard the downgrade, even though it seems that the imported packages are not affected. The other package upgrades were, again, performed due to runc's dependencies on these versions.

However, from a quick look, I can see that Kubernetes has not migrated to runc yet. Therefore, introducing this set of changes to CAdvisor at this point would require Kubernetes to migrate to runc during their next godep update of CAdvisor. I am not aware of the current release plans of both projects so this PR could be delayed until Kubernetes is ready to migrate.

As far as I'm aware, there should be no compatibility concerns with docker, as the previous version that was being linked with CAdvisor (v1.8.3) already uses runc/libcontainer internally.

Please let me know if you feel any other changes are needed to this PR, aside from retaining the docker version.

afein:
Updated this PR to include the latest available tag for docker/docker, v1.9.1

jimmidyson:
@vishh Are you happy with these godeps bumps? I can't think of any problems but don't want to bump stuff it it's going to make it difficult/impossible to bump kubernetes' dependencies on next cadvisor release.

jimmidyson:
@yujuhong Any concerns about the Docker dependency bump, especially with kubernetes in mind?

jimmidyson:
ok to test

jimmidyson:
retest this please

Godeps/Godeps.json
```diff
@@ -8,7 +8,7 @@
 		{
 			"ImportPath": "bitbucket.org/ww/goautoneg",
 			"Comment": "null-5",
-			"Rev": "75cd24fc2f2c2a2088577d12123ddee5f54e0675"
+			"Rev": "'75cd24fc2f2c2a2088577d12123ddee5f54e0675'"
```
jimmidyson: This is doing the same bizarre single quotes for me locally too. Any idea why? I assume it's not going to cause any problems going forward?

jimmidyson:
LGTM. Thanks @afein!

PR #969, opened 2015-11-22
add support for cgroup  CPUQuota  and CPUPeriod prometheus limit

f0:
This adds support for the CPUQuota cgroup limit.

This actually does not work , i miss something, any hints?

### DIFF
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -207,6 +207,16 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 		if utils.FileExists(cpuRoot) {
 			spec.HasCpu = true
 			spec.Cpu.Limit = readUInt64(cpuRoot, "cpu.shares")
+			spec.Cpu.Period = readUInt64(cpuRoot, "cpu.cfs_period_us")
+			quota := readString(cpuRoot, "cpu.cfs_quota_us")
+
+			if quota != "-1" {
+				val, err := strconv.ParseUint(quota, 10, 64)
+				if err != nil {
+					glog.Errorf("raw driver: Failed to parse CPUQuota from %q: %s", path.Join(cpuRoot, "cpu.cfs_quota_us"), err)
+				}
+				spec.Cpu.Quota = val
+			}
 		}
 	}
 
--- a/info/v2/container.go
+++ b/info/v2/container.go
@@ -36,6 +36,10 @@ type CpuSpec struct {
 	// Cpu affinity mask.
 	// TODO(rjnagal): Add a library to convert mask string to set of cpu bitmask.
 	Mask string `json:"mask,omitempty"`
+	// CPUQuota Default is disabled
+	Quota uint64 `json:"quota,omitempty"`
+	// Period is the CPU reference time in ns e.g the quota is compared aginst this.
+	Period uint64 `json:"period,omitempty"`
 }
 
 type MemorySpec struct {
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -542,10 +542,16 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.CreationTime.Unix()), baseLabelValues...)
 
 		if container.Spec.HasCpu {
+			desc = prometheus.NewDesc("container_spec_cpu_period", "CPU period of the container.", baseLabels, nil)
+			ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.Cpu.Period), baseLabelValues...)
+			if container.Spec.Cpu.Quota != 0 {
+				desc = prometheus.NewDesc("container_spec_cpu_quota", "CPU quota of the container.", baseLabels, nil)
+				ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.Cpu.Quota), baseLabelValues...)
+			}
 			desc := prometheus.NewDesc("container_spec_cpu_shares", "CPU share of the container.", baseLabels, nil)
 			ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.Cpu.Limit), baseLabelValues...)
-		}
 
+		}
 		if container.Spec.HasMemory {
 			desc := prometheus.NewDesc("container_spec_memory_limit_bytes", "Memory limit for the container.", baseLabels, nil)
 			ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, specMemoryValue(container.Spec.Memory.Limit), baseLabelValues...)
--- a/metrics/prometheus_test.go
+++ b/metrics/prometheus_test.go
@@ -55,7 +55,13 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 				Aliases: []string{"testcontaineralias"},
 			},
 			Spec: info.ContainerSpec{
-				Image:        "test",
+				Image:  "test",
+				HasCpu: true,
+				Cpu: info.CpuSpec{
+					Limit:  1000,
+					Period: 10,
+					Quota:  10000,
+				},
 				CreationTime: time.Unix(1257894000, 0),
 				Labels: map[string]string{
 					"foo.label": "bar",
--- a/metrics/testdata/prometheus_metrics
+++ b/metrics/testdata/prometheus_metrics
@@ -116,6 +116,15 @@ container_network_transmit_packets_total{foo_env="prod",foo_label="bar",id="test
 # HELP container_scrape_error 1 if there was an error while getting container metrics, 0 otherwise
 # TYPE container_scrape_error gauge
 container_scrape_error 0
+# HELP container_spec_cpu_period CPU period of the container.
+# TYPE container_spec_cpu_period gauge
+container_spec_cpu_period{foo_env="prod",foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 10
+# HELP container_spec_cpu_quota CPU quota of the container.
+# TYPE container_spec_cpu_quota gauge
+container_spec_cpu_quota{foo_env="prod",foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 10000
+# HELP container_spec_cpu_shares CPU share of the container.
+# TYPE container_spec_cpu_shares gauge
+container_spec_cpu_shares{foo_env="prod",foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1000
 # HELP container_start_time_seconds Start time of the container since unix epoch in seconds.
 # TYPE container_start_time_seconds gauge
 container_start_time_seconds{foo_env="prod",foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1.257894e+09
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -23,6 +23,8 @@ type CpuSpec struct {
 	Limit    uint64 `json:"limit"`
 	MaxLimit uint64 `json:"max_limit"`
 	Mask     string `json:"mask,omitempty"`
+	Quota    uint64 `json:"quota,omitempty"`
+	Period   uint64 `json:"period,omitempty"`
 }
 
 type MemorySpec struct {


jimmidyson:
Thanks for this. What doesn't work? No metric or no values reported in the metric?

f0:
cadvisor builds sucessfully, but in the prometheus output the quota is not exported , but is present in the specific cgroup

jimmidyson:
Can you share the whole `/metrics` endpoint output please? Perhaps in a gist.

f0:
sure https://gist.github.com/f0/1938007f1d9054f57c56

jimmidyson:
And you're sure you built & replaced cadvisor binary with the newly built binary? Everything looks fine to me but the output.

f0:
hm, yes i am sure but i double check ...

f0:
hm yes its the correct binary, but its not shown

jimmidyson:
OK I'll build with your PR locally tomorrow & see if I can see what's going on.

f0:
@jimmidyson   thx

jimmidyson:
This works perfectly fine for me locally... How are you building? Just run `make` & you should get a binary in the root of the project dir to use.

jimmidyson:
Well when I say it works perfectly I mean I get output in the `/metrics` endpoint.

But in logs (running with `--logtostderr`):

    E1123 09:15:09.954624 11428 handler.go:149] raw driver: Failed to parse int "-1" from file "/sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us": strconv.ParseUint: parsing "-1": invalid syntax

Need a test for this as well if possible.

f0:
@jimmidyson yep this happens when no Quota is set, so it need to be coverd.... i run build/build.sh

f0:
@jimmidyson  hm is there a specific go version needed to build ? atm i have 1.4.3

jimmidyson:
1.4 should be fine. Do you get errors when building with 1.4?

f0:
@jimmidyson yes

fk@linux-3zvu:~/work/src/f0/cadvisor> make
>> formatting code
>> building binaries
 >   cadvisor
# github.com/google/cadvisor
usage: 6l [options] main.6
  -1	use alternate profiling code
  -8	assume 64-bit addresses
  -B info
    	define ELF NT_GNU_BUILD_ID note
  -C	check Go calls to C code
  -D addr
    	data address
  -E sym
    	entry symbol
  -I interp
    	set ELF interp
  -L dir
    	add dir to library path
  -H head
    	header type
  -K	add stack underflow checks
  -O	print pc-line tables
  -Q	debug byte-register code gen
  -R rnd
    	address rounding
  -S	check type signatures
  -T addr
    	text address
  -V	print version and exit
  -W	disassemble input
  -X name value
    	define string data
  -Z	clear stack frame on entry
  -a	disassemble output
  -c	dump call graph
  -d	disable dynamic executable
  -extld ld
    	linker to run in external mode
  -extldflags ldflags
    	flags for external linker
  -f	ignore version mismatch
  -g	disable go package data checks
  -installsuffix suffix
    	pkg directory suffix
  -k sym
    	set field tracking symbol
  -linkmode mode
    	set link mode (internal, external, auto)
  -n	dump symbol table
  -o outfile
    	set output file
  -r dir1:dir2:...
    	set ELF dynamic linker search path
  -race
    	enable race detector
  -s	disable symbol table
  -shared
    	generate shared object (implies -linkmode external)
  -tmpdir dir
    	leave temporary files in this directory
  -u	reject unsafe packages
  -v	print link trace
  -w	disable DWARF generation
godep: go exit status 2
Makefile:32: recipe for target 'build' failed
make: *** [build] Error 1

jimmidyson:
Just pushed a fix so you can build on 1.4 as well (`ldflags` was 1.5 stylee). Rebase against master & you should be able to build on 1.4 with `make`.

f0:
ok now i can build, and i got these errors, but why , the typh has a field Quota

'''
fk@linux-3zvu:~/work/src/f0/cadvisor> make
>> formatting code
>> building binaries
 >   cadvisor
>> running tests
# f0/cadvisor/container/raw
container/raw/handler.go:203: spec.Cpu.Quota undefined (type v1.CpuSpec has no field or method Quota)
# f0/cadvisor/metrics
metrics/prometheus.go:523: container.Spec.Cpu.Quota undefined (type v1.CpuSpec has no field or method Quota)
```

**Comment by f0** [comment]
@jimmidyson   hm any Idea why spec.Cpu.Quota is not present, its in the files...

**Comment by f0** [comment]
hm ok, was a problem in my buildenv,  but the Quota still not work

**Comment by f0** [comment]
updated with check for -1

**Comment by f0** [comment]
@jimmidyson  hm please reopen

**Comment by f0** [comment]
@jimmidyson   ok have fixed some things, now i get these errors while building, but why, the type has this field? And it does not work in my environment....

```
 running tests
 github.com/f0/cadvisor/container/raw
container/raw/handler.go:206: spec.Cpu.Quota undefined (type v1.CpuSpec has no field or method Quota)
 github.com/f0/cadvisor/metrics
metrics/prometheus.go:523: container.Spec.Cpu.Quota undefined (type v1.CpuSpec has no field or method Quota)
```

**Comment by jimmidyson** [comment]
Discussed on IRC: cadvisor needs to be cloned to `$GOPATH/src/github.com/google/cadvisor` & a remote set to your fork, rather than cloning to `$GOPATH/src/github.com/f0/cadvisor`.

**Comment by f0** [comment]
@jimmidyson  after fixing gopath and some other orrors it finaly workt THX

**Comment by f0** [comment]
also adding support for the cpu period setting (with this we finally can calculate the percentual cpu usage )

**Comment by f0** [comment]
@jimmidyson  when will this be merged?

**Comment by jimmidyson** [comment]
So this works for you?

**Comment by jimmidyson** [comment]
Would you be able to add a test to `cadvisor/metrics/prometheus_test.go` please?

**Comment by f0** [comment]
yes works for me, and yes i try to add a test

**Comment by jimmidyson** [comment]
Thank you. Works for me too btw.

File: container/raw/handler.go [hunk 1]
```diff
@@ -200,6 +200,17 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 		if utils.FileExists(cpuRoot) {
 			spec.HasCpu = true
 			spec.Cpu.Limit = readInt64(cpuRoot, "cpu.shares")
+			spec.Cpu.Period = readInt64(cpuRoot, "cpu.cfs_period_us")
+			quota := readString(cpuRoot, "cpu.cfs_quota_us")
```
jimmidyson: Why not use the `readInt64` method as above?

f0: because if the quota is notr set, -1 is returned and internally readInt64 uses an uint and this breaks with -1

metrics/prometheus.go
```diff
@@ -519,6 +519,10 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		if container.Spec.HasCpu {
 			desc := prometheus.NewDesc("container_spec_cpu_shares", "CPU share of the container.", baseLabels, nil)
 			ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.Cpu.Limit), baseLabelValues...)
+			desc = prometheus.NewDesc("container_spec_cpu_quota", "CPU quota of the container.", baseLabels, nil)
+			ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, float64(container.Spec.Cpu.Quota), baseLabelValues...)
```
jimmidyson: What value will this have if quota is unset (-1)? I assume zero - is zero really a value we want to send through?

container/raw/handler.go
jimmidyson: I didn't check that - so really `readInt64` should really be `readUInt64`...

metrics/prometheus.go
f0: hm if the quota is not set (-1) i do not set container.Spec.Cpu.Quota and the its empthy and is discarded  by omitifempthy , if i understand his correct

container/raw/handler.go
f0: yes this would be much cleaner

metrics/prometheus.go
jimmidyson: What `omitifempty`? You're sending through a metric to Prometheus here with value `0` if `container.Spec.Cpu.Quota` is unset.

jimmidyson: Ah you mean the `omitempty` tag on the struct field? That's only for JSON serialization of the specifie struct, not sending metrics to Prometheus.

f0: hm ok, what does go assume if a value is not set nil? i only set container.Spec.Cpu.Quota to a value when the read does not return -1

jimmidyson: https://golang.org/ref/spec#The_zero_value

f0: ok in this case 0 would be returned, hm is there a way to discard the output completly?

jimmidyson: Is 0 a valid value for CPU quota - I'm not sure tbh.

f0:
```
cpu.cfs_quota_us
specifies the total amount of time in microseconds (µs, represented here as "us") for which all tasks in a cgroup can run during one period (as defined by cpu.cfs_period_us). As soon as tasks in a cgroup use up all the time specified by the quota, they are throttled for the remainder of the time specified by the period and not allowed to run until the next period. If tasks in a cgroup should be able to access a single CPU for 0.2 seconds out of every 1 second, set cpu.cfs_quota_us to 200000 and cpu.cfs_period_us to 1000000. Note that the quota and period parameters operate on a CPU basis. To allow a process to fully utilize two CPUs, for example, set cpu.cfs_quota_us to 200000 and cpu.cfs_period_us to 100000.
Setting the value in cpu.cfs_quota_us to -1 indicates that the cgroup does not adhere to any CPU time restrictions. This is also the default value for every cgroup (except the root cgroup).
''''

i think its a valid value

**Comment by f0** [comment]
rebased

**Comment by f0** [comment]
@jimmidyson  ok i  do not found any documentation about the valid values, i tested it, 0 is not allowed, the minimal value is 1s (in ns ) in cpu.cfs_quota_us or 1% with e.g systemd settings, so i am not sure whats the best way, maybe completely disable the output when its set to -1 ?

**Comment by jimmidyson** [comment]
That sounds good - if -1 then don't send the metric.

**Comment by f0** [comment]
@jimmidyson  i am not sure how to add a test in cadvisor/metrics/prometheus_test.go, can you give me a hint?

**Comment by f0** [comment]
@jimmidyson  ping

**Comment by jimmidyson** [comment]
Sorry. You'll need to add the stats somewhere in https://github.com/google/cadvisor/blob/master/metrics/prometheus_test.go#L61-L151 & update https://github.com/google/cadvisor/blob/master/metrics/testdata/prometheus_metrics accordingly. It's a bit manual, but worthwhile.

File: info/v2/container.go [hunk 3]
```diff
@@ -36,6 +36,10 @@ type CpuSpec struct {
 	// Cpu affinity mask.
 	// TODO(rjnagal): Add a library to convert mask string to set of cpu bitmask.
 	Mask string `json:"mask,omitempty"`
+	// CPUQuota Defazult is disabled
```
vishh: nit: s/Defazult/default

vishh: What is the value when it is disabled?

container/raw/handler.go
```diff
@@ -200,6 +200,16 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 		if utils.FileExists(cpuRoot) {
 			spec.HasCpu = true
 			spec.Cpu.Limit = readUInt64(cpuRoot, "cpu.shares")
```
vishh: Why add this metric only to the raw driver?
Is it possible to extract this metrics even for docker containers?

vishh:
add to whitelist

info/v2/container.go
f0: @vishh its -1 in the cgroup file

jimmidyson:
retest this please

f0:
@jimmidyson  hm i do not get how to name the testdata in  https://github.com/google/cadvisor/blob/master/metrics/testdata/prometheus_metrics

f0:
rebased , commits squashed and the first half of testing

f0:
@jimmidyson  hm i do something wrong.... the testing does not really work, i have no idea why

metrics/testdata/prometheus_metrics
```diff
@@ -13,6 +13,9 @@ container_cpu_usage_seconds_total{cpu="cpu03",id="testcontainer",image="test",na
 # HELP container_cpu_user_seconds_total Cumulative user cpu time consumed in seconds.
 # TYPE container_cpu_user_seconds_total counter
 container_cpu_user_seconds_total{id="testcontainer",image="test",name="testcontaineralias"} 6e-09
+# container_spec_cpu_quota HELP CPU quota of the container
```
jimmidyson: Check the format of the lines - `HELP` should be at the start of the comment. You can run tests locally btw with `make test`.

f0:
@jimmidyson 

hm i dont get it now i have this error....
=== RUN TestPrometheusCollector
--- FAIL: TestPrometheusCollector (0.02s)
        prometheus_test.go:189: want # HELP container_spec_cpu_quota CPU quota of the container, got # HELP container_fs_io_current Number of I/Os currently in progress

jimmidyson:
The order of the output is important. Think metrics are ordered alphabetically so you'll need to put the new metric in the right place in the test file.

f0:
@jimmidyson   ah ok so then the ne test should be after 
container_start_time_seconds  and before container_tasks_state, but this also gives me the above error(with another error text)

jimmidyson:
Should be before `container_start_time_seconds`.

f0:
@jimmidyson   also does not work

f0:
ok, iow i got it !    rebased and test added

metrics/prometheus_test.go
```diff
@@ -55,7 +55,11 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 				Aliases: []string{"testcontaineralias"},
 			},
 			Spec: info.ContainerSpec{
-				Image:        "test",
+				Image: "test",
```
jimmidyson: You need to add in `HasCpu: true,` in here so that the CPU specs are output as metrics. You'll probably find then that there are a couple of other metrics which haven't been tested for that you'll need to insert into `metrics/testdata/prometheus_metrics` to get this to pass.

vishh:
Ping @f0!

f0:
@vishh  @jimmidyson  sorry have not so much spare time atm, try to get this in the next days

jimmidyson:
If you don't manage to get to it by Monday I can pull it into a separate PR & fix up the tests.

f0:
@jimmidyson  feel free, i have no chance to get this until monday, i am 90% offline at the weekend , sorry.

f0:
@jimmidyson @vishh finally done (and now i understand the test thing ....)

container/raw/handler.go
jimmidyson: @f0 Any thoughts on this?

f0: @jimmidyson  hm not really , why not but i think this should be done in a different PR

f0:
@jimmidyson  @vishh   any outstanding issues with this PR?

jimmidyson:
no LGTM

PR #971, opened 2015-11-23
Fix for go 1.4 build compat

jimmidyson:


jimmidyson:
Self-merging as this is only `ldflags` tidy, no functional change.

PR #974, opened 2015-11-24
Godeps: Update glog version.

yifan-gu:
I tried to pull #966 and then update the kubernetes' Godep to use runc/libcontainer, found that kubernetes' glog is newer there. 

So would like to make this update in cadvisor to make it easier to bump godeps in kubernetes (otherwise will have to update a bunch of packages by hand instead of `rm -fr Godeps/, godep save ./...`

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -72,7 +72,7 @@
 		},
 		{
 			"ImportPath": "github.com/golang/glog",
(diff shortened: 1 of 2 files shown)


jimmidyson:
ok to test

jimmidyson:
Thanks @yifan-gu! Merging.

yifan-gu:
Thank you @jimmidyson

PR #976, opened 2015-11-24
Fix Devicemapper output parsing

jimmidyson:
See kubernetes/kubernetes#17725 for more details.

/cc @vishh

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -359,8 +359,8 @@ func dockerDMDevice(driverStatus string) (string, uint, uint, uint, error) {
 	)
 
 	_, err = fmt.Fscanf(bytes.NewReader(dmTable),
-		"%d %d %s %d:%d %d:%d %d %d %d %s",
-		&bkt, &bkt, &bkts, &bkt, &bkt, &major, &minor, &dataBlkSize, &bkt, &bkt, &bkts)
+		"%d %d %s %d:%d %d:%d %d",
+		&bkt, &bkt, &bkts, &bkt, &bkt, &major, &minor, &dataBlkSize)
 	if err != nil {
 		return "", 0, 0, 0, err
 	}
@@ -379,8 +379,8 @@ func getDMStats(poolName string, dataBlkSize uint) (uint64, uint64, uint64, erro
 	)
 
 	_, err = fmt.Fscanf(bytes.NewReader(dmStatus),
-		"%d %d %s %d %d/%d %d/%d %s %s %s %s",
-		&bkt, &bkt, &bkts, &bkt, &bkt, &bkt, &used, &total, &bkts, &bkts, &bkts, &bkts)
+		"%d %d %s %d %d/%d %d/%d",
+		&bkt, &bkt, &bkts, &bkt, &bkt, &bkt, &used, &total)
 	if err != nil {
 		return 0, 0, 0, err
 	}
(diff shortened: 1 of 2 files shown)


yujuhong:
Might be useful to add a simple unit test. Would help the reviewers too :)

jimmidyson:
@yujuhong Now you're making me work hard... you're right of course. Will add some tomorrow.

fs/fs.go
```diff
@@ -359,8 +359,8 @@ func dockerDMDevice(driverStatus string) (string, uint, uint, uint, error) {
 	)
 
 	_, err = fmt.Fscanf(bytes.NewReader(dmTable),
-		"%d %d %s %d:%d %d:%d %d %d %d %s",
-		&bkt, &bkt, &bkts, &bkt, &bkt, &major, &minor, &dataBlkSize, &bkt, &bkt, &bkts)
+		"%d %d %s %d:%d %d:%d %d",
```
timstclair: Is there any kernel documentation or paths you can link to here to make this easier to verify, and easier to understand/modify in the future?

Same below.

jimmidyson:
Refactored to use `strings.Fields` instead of `fmt.Fscanf` after benchmarking the two:

```
BenchmarkFstatf-8         300000              5029 ns/op             240 B/op         10 allocs/op
BenchmarkFields-8        1000000              1390 ns/op             304 B/op          3 allocs/op
```

Added tests as requested, including a case with the output value in the original report of this problem (kubernetes/kubernetes#17725).

Link to DM kernel docs added.

jimmidyson:
retest this please

jimmidyson:
retest this please

jimmidyson:
@yujuhong @timstclair Review would be appreciated so we can get this fix into kubernetes ASAP.

yujuhong:
Thanks for the PR. It looks good overall and the test shows the expected output.

Just a question: I couldn't find the exact output fields of `dmsetup status <pool>`. Is it not clearly documented?

jimmidyson:
> Is it not clearly documented?

 Not that I've found. I worked out the line format & confirmed with @rhvgoyal.

yujuhong:
Ah I see. LGTM.

jimmidyson:
Thanks @yujuhong. Merging.

rhvgoyal:
Here is the link to "dmsetup status <pool>" command output description.

https://github.com/torvalds/linux/blob/master/Documentation/device-mapper/thin-provisioning.txt#L273

Following fields should be visible.

    <transaction id> <used metadata blocks>/<total metadata blocks>
    <used data blocks>/<total data blocks> <held metadata root>
    [no_]discard_passdown ro|rw

yujuhong:
@rhvgoyal, I saw the description but the order seemed different than the output @jimmidyson provided in the test.

jimmidyson:
@yujuhong You mean at https://github.com/google/cadvisor/pull/976/files#diff-88a0b03cd7789db9964ce3da2ff9fd37R110 & following?

    0 409534464 thin-pool 64085 3705/4161600 88106/3199488 - rw no_discard_passdown queue_if_no_space -

From @rhvgoyal description above, this is `88106` used data blocks out of total `3199488` data blocks as the test specifies?

jimmidyson:
Well now I actually read it, it does look different. The output I used in the tests was from my local box & from a bug report in Kubernetes. I guess there's some prefix that's different somehow but it is correct :)

PR #978, opened 2015-11-25
Regexp tidy up

jimmidyson:
Cleaned up some areas where regexps were either messy or unnecessary.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -172,9 +172,12 @@ func (self *dockerFactory) DebugInfo() map[string][]string {
 	return map[string][]string{}
 }
 
+var (
+	version_regexp_string = `(\d+)\.(\d+)\.(\d+)`
+	version_re            = regexp.MustCompile(version_regexp_string)
+)
+
 func parseDockerVersion(full_version_string string) ([]int, error) {
-	version_regexp_string := "(\\d+)\\.(\\d+)\\.(\\d+)"
-	version_re := regexp.MustCompile(version_regexp_string)
 	matches := version_re.FindAllStringSubmatch(full_version_string, -1)
 	if len(matches) != 1 {
 		return nil, fmt.Errorf("version string \"%v\" doesn't match expected regular expression: \"%v\"", full_version_string, version_regexp_string)
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -20,7 +20,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -202,8 +201,6 @@ func scanTcpStats(tcpStatsFile string) (info.TcpStat, error) {
 		return stats, fmt.Errorf("failure opening %s: %v", tcpStatsFile, err)
 	}
 
-	tcpStatLineRE, _ := regexp.Compile("[0-9:].*")
-
 	tcpStateMap := map[string]uint64{
 		"01": 0, //ESTABLISHED
 		"02": 0, //SYN_SENT
@@ -223,18 +220,24 @@ func scanTcpStats(tcpStatsFile string) (info.TcpStat, error) {
 
 	scanner.Split(bufio.ScanLines)
 
+	// Discard header line
+	b := scanner.Scan()
+	if !b {
+		return stats, scanner.Err()
+	}
+
 	for scanner.Scan() {
-
 		line := scanner.Text()
-		//skip header
-		matched := tcpStatLineRE.MatchString(line)
 
-		if matched {
-			state := strings.Fields(line)
-			//#file header tcp state is the 4 filed:
-			//sl local_address rem_address st tx_queue rx_queue tr tm->when retrnsmt  uid timeout inode
-			tcpStateMap[state[3]]++
+		state := strings.Fields(line)
+		// TCP state is the 4th field.
+		// Format: sl local_address rem_address st tx_queue rx_queue tr tm->when retrnsmt  uid timeout inode
+		tcpState := state[3]
+		_, ok := tcpStateMap[tcpState]
+		if !ok {
+			return stats, fmt.Errorf("invalid TCP stats line: %v", line)
 		}
+		tcpStateMap[tcpState]++
 	}
 
 	stats = info.TcpStat{
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -35,8 +35,6 @@ import (
 	"github.com/golang/glog"
 )
 
-var partitionRegex = regexp.MustCompile("^(:?(:?s|xv)d[a-z]+\\d*|dm-\\d+)$")
-
 const (
 	LabelSystemRoot   = "root"
 	LabelDockerImages = "docker-images"
@@ -223,6 +221,8 @@ func (self *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, er
 	}
 	return filesystems, nil
 }
+
+var partitionRegex = regexp.MustCompile(`^(?:(?:s|xv)d[a-z]+\d*|dm-\d+)$`)
 
 func getDiskStatsMap(diskStatsFile string) (map[string]DiskStats, error) {
 	diskStatsMap := make(map[string]DiskStats)
--- a/metrics/prometheus_test.go
+++ b/metrics/prometheus_test.go
@@ -152,6 +152,11 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 	}, nil
 }
 
+var (
+	includeRe = regexp.MustCompile(`^(?:(?:# HELP |# TYPE)container_|cadvisor_version_info\{)`)
+	ignoreRe  = regexp.MustCompile(`^container_last_seen\{`)
+)
+
 func TestPrometheusCollector(t *testing.T) {
 	prometheus.MustRegister(NewPrometheusCollector(testSubcontainersInfoProvider{}, nil))
 
@@ -171,8 +176,6 @@ func TestPrometheusCollector(t *testing.T) {
 	// (https://github.com/prometheus/client_golang/issues/58), we simply compare
 	// verbatim text-format metrics outputs, but ignore certain metric lines
 	// whose value depends on the current time or local circumstances.
-	includeRe := regexp.MustCompile("^(?:(?:# HELP |# TYPE)container_|cadvisor_version_info{)")
-	ignoreRe := regexp.MustCompile("^container_last_seen{")
 	for i, want := range wantLines {
 		if !includeRe.MatchString(want) || ignoreRe.MatchString(want) {
 			continue
--- a/utils/machine/machine.go
+++ b/utils/machine/machine.go
@@ -34,12 +34,17 @@ import (
 
 // The utils/machine package contains functions that extract machine-level specs.
 
-var cpuRegExp = regexp.MustCompile("processor\\t*: +([0-9]+)")
-var coreRegExp = regexp.MustCompile("core id\\t*: +([0-9]+)")
-var nodeRegExp = regexp.MustCompile("physical id\\t*: +([0-9]+)")
-var CpuClockSpeedMHz = regexp.MustCompile("cpu MHz\\t*: +([0-9]+.[0-9]+)")
-var memoryCapacityRegexp = regexp.MustCompile("MemTotal: *([0-9]+) kB")
-var swapCapacityRegexp = regexp.MustCompile("SwapTotal: *([0-9]+) kB")
+var (
+	cpuRegExp  = regexp.MustCompile(`^processor\s*:\s*([0-9]+)$`)
+	coreRegExp = regexp.MustCompile(`^core id\s*:\s*([0-9]+)$`)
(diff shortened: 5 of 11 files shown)


container/docker/factory.go
```diff
@@ -172,9 +172,12 @@ func (self *dockerFactory) DebugInfo() map[string][]string {
 	return map[string][]string{}
 }
 
+var (
+	version_regexp_string = "(\\d+)\\.(\\d+)\\.(\\d+)"
+	version_re            = regexp.MustCompile(version_regexp_string)
```
timstclair: What's the point of moving this out of the function? Personally I find it more readable if I can see the regexp right where it's used (and I don't think this should be used outside that function).

container/docker/factory.go
```diff
@@ -172,9 +172,12 @@ func (self *dockerFactory) DebugInfo() map[string][]string {
 	return map[string][]string{}
 }
 
+var (
+	version_regexp_string = "(\\d+)\\.(\\d+)\\.(\\d+)"
```
timstclair: Maybe change this to a raw string so you don't need to escape the backslash?

container/docker/factory.go
jimmidyson: Putting it in the function means the RE gets wastefully compiled every time the function is called. In this case probably not called more than once, but normally best to move it outside so trying to keep it consistent.

container/docker/factory.go
jimmidyson: Good catch,

container/libcontainer/helpers.go
```diff
@@ -224,16 +221,14 @@ func scanTcpStats(tcpStatsFile string) (info.TcpStat, error) {
...
-		matched := tcpStatLineRE.MatchString(line)
-
-		if matched {
-			state := strings.Fields(line)
-			//#file header tcp state is the 4 filed:
-			//sl local_address rem_address st tx_queue rx_queue tr tm->when retrnsmt  uid timeout inode
-			tcpStateMap[state[3]]++
+
+		state := strings.Fields(line)
+		//#file header tcp state is the 4th field:
```
timstclair: (nit) space after //

timstclair: Also this comment reads a little weird. Maybe rephrase to:
```
// TCP state is the 4th field.
// Format: sl local_address rem_address st tx_queue rx_queue tr tm->when retrnsmt  uid timeout inode
```

container/libcontainer/helpers.go
```diff
@@ -224,16 +221,14 @@ func scanTcpStats(tcpStatsFile string) (info.TcpStat, error) {
...
+		//sl local_address rem_address st tx_queue rx_queue tr tm->when retrnsmt  uid timeout inode
+		tcpState := state[3]
+		if _, ok := tcpStateMap[tcpState]; ok {
```
timstclair: If this is not OK it means you encountered an unexpected state. I think it's worth logging an error or warning if that happens (although that could generate a LOT of log spam if the file is corrupted, so it might be good to rate limit to a single log entry per call)

fs/fs.go
```diff
@@ -35,7 +35,7 @@ import (
 	"github.com/golang/glog"
 )
 
-var partitionRegex = regexp.MustCompile("^(:?(:?s|xv)d[a-z]+\\d*|dm-\\d+)$")
+var partitionRegex = regexp.MustCompile(`^(:?(:?s|xv)d[a-z]+\d*|dm-\d+)$`)
```
timstclair: Does this regex optionally start with `::`, or is that supposed to be a non-capturing group? If the latter, it's backwards:
`(?:re)`

Also would prefer to move this closer to where it's used.

container/docker/factory.go
timstclair: Ah, makes sense. Maybe move the declaration to just above the function in that case? Also, consider making the string const.

metrics/prometheus_test.go
```diff
@@ -152,6 +152,11 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 	}, nil
 }
 
+var (
+	includeRe = regexp.MustCompile("^(?:(?:# HELP |# TYPE)container_|cadvisor_version_info{)")
```
timstclair: Does this compile correctly? I would expect that `{` needs to be escaped.

utils/machine/machine.go
```diff
@@ -34,12 +34,18 @@ import (
...
 // The utils/machine package contains functions that extract machine-level specs.
 
-var cpuRegExp = regexp.MustCompile("processor\\t*: +([0-9]+)")
-var coreRegExp = regexp.MustCompile("core id\\t*: +([0-9]+)")
-var nodeRegExp = regexp.MustCompile("physical id\\t*: +([0-9]+)")
-var CpuClockSpeedMHz = regexp.MustCompile("cpu MHz\\t*: +([0-9]+.[0-9]+)")
-var memoryCapacityRegexp = regexp.MustCompile("MemTotal: *([0-9]+) kB")
-var swapCapacityRegexp = regexp.MustCompile("SwapTotal: *([0-9]+) kB")
+var (
+	cpuRegExp            = regexp.MustCompile("processor\\t*: +([0-9]+)")
```
timstclair: (nit) raw strings

timstclair: I wonder if ` +` should be `\s+`?

timstclair: Actually, same with `\t` (use `\s`)? Might be more robust.

Same below for all these.

utils/machine/machine.go
```diff
@@ -34,12 +34,18 @@ import (
...
+	coreRegExp           = regexp.MustCompile("core id\\t*: +([0-9]+)")
+	nodeRegExp           = regexp.MustCompile("physical id\\t*: +([0-9]+)")
+	CpuClockSpeedMHz     = regexp.MustCompile("cpu MHz\\t*: +([0-9]+.[0-9]+)")
```
timstclair: escape decimal? `\.`

utils/machine/machine.go
```diff
@@ -34,12 +34,18 @@ import (
...
+	memoryCapacityRegexp = regexp.MustCompile("MemTotal: *([0-9]+) kB")
+	swapCapacityRegexp   = regexp.MustCompile("SwapTotal: *([0-9]+) kB")
+
+	cpuClockSpeedMHz = regexp.MustCompile("clock\\t*: +([0-9]+.[0-9]+)MHz")
```
timstclair: here too: `\.`

timstclair: Unused? Did you mean to remove the lines where this was used below? Also, consider differentiating the variable name more from `CpuClockSpeedMHz`

utils/oomparser/oomparser.go
```diff
@@ -29,12 +29,11 @@ import (
...
-var containerRegexp *regexp.Regexp = regexp.MustCompile(
-	`Task in (.*) killed as a result of limit of (.*)`)
-var lastLineRegexp *regexp.Regexp = regexp.MustCompile(
-	`(^[A-Z]{1}[a-z]{2} .*[0-9]{1,2} [0-9]{1,2}:[0-9]{2}:[0-9]{2}) .* Killed process ([0-9]+) \(([0-9A-Za-z_]+)\)`)
-var firstLineRegexp *regexp.Regexp = regexp.MustCompile(
-	`invoked oom-killer:`)
+var (
+	containerRegexp *regexp.Regexp = regexp.MustCompile(`Task in (.*) killed as a result of limit of (.*)`)
+	lastLineRegexp  *regexp.Regexp = regexp.MustCompile(`(^[A-Z]{1}[a-z]{2} .*[0-9]{1,2} [0-9]{1,2}:[0-9]{2}:[0-9]{2}) .* Killed process ([0-9]+) \(([0-9A-Za-z_]+)\)`)
+	firstLineRegexp *regexp.Regexp = regexp.MustCompile(`invoked oom-killer:`)
```
timstclair: Why are these pointers?

utils/oomparser/oomparser.go
```diff
@@ -29,12 +29,11 @@ import (
...
 
-var containerRegexp *regexp.Regexp = regexp.MustCompile(
-	`Task in (.*) killed as a result of limit of (.*)`)
-var lastLineRegexp *regexp.Regexp = regexp.MustCompile(
-	`(^[A-Z]{1}[a-z]{2} .*[0-9]{1,2} [0-9]{1,2}:[0-9]{2}:[0-9]{2}) .* Killed process ([0-9]+) \(([0-9A-Za-z_]+)\)`)
-var firstLineRegexp *regexp.Regexp = regexp.MustCompile(
-	`invoked oom-killer:`)
+var (
+	containerRegexp *regexp.Regexp = regexp.MustCompile(`Task in (.*) killed as a result of limit of (.*)`)
+	lastLineRegexp  *regexp.Regexp = regexp.MustCompile(`(^[A-Z]{1}[a-z]{2} .*[0-9]{1,2} [0-9]{1,2}:[0-9]{2}:[0-9]{2}) .* Killed process ([0-9]+) \(([0-9A-Za-z_]+)\)`)
```
timstclair: `{1}` unnecessary

(optional) `[0-9A-Za-z_]` can be replaced with `\w`, although I'm not sure that's more readable.

jimmidyson:
@timstclair Thanks for a thorough code review - this is making it a much more worthwhile endeavour than it would have otherwise been :)

utils/sysinfo/sysinfo.go
```diff
@@ -65,18 +65,14 @@ func GetBlockDeviceInfo(sysfs sysfs.SysFs) (map[string]info.DiskInfo, error) {
...
-		} else {
-			matches := schedulerRegExp.FindSubmatch([]byte(sched))
-			if len(matches) < 2 {
-				sched = "none"
-			} else {
-				sched = string(matches[1])
+		disk_info.Scheduler = "none"
+		blkSched, err := sysfs.GetBlockDeviceScheduler(name)
+		if err == nil {
+			matches := schedulerRegExp.FindSubmatch([]byte(blkSched))
```
timstclair: This regex should use a raw string too.

container/libcontainer/helpers.go
jimmidyson: Probably best to return an error - can you trust stats from a corrupt file?

timstclair: Yeah, makes sense.

utils/machine/machine.go
jimmidyson: Seems as though it was always unused... removing.

jimmidyson: Ah hang on messed up removing stuff below. Will add it back in.

timstclair: Actually, maybe just combine this with `CpuClockSpeedMHz`? Something like:
```
`(?:cpu MHz|clock)\t*: +([0-9]+\.[0-9]+)(?:MHz)?`
```

timstclair:
> @timstclair Thanks for a thorough code review - this is making it a much more worthwhile endeavour than it would have otherwise been :)

No problem! Thanks for the cleanup!

utils/machine/machine.go
jimmidyson: Great minds...

jimmidyson:
All comments addressed. Good review!

jimmidyson:
As an observation there is a lot we could do to lower resource utilisation. Regexps are hungry & need to be used judiciously. They're overused here where we could probably do some cheaper string parsing.

jimmidyson:
Integration tests pass, so it must be good right? :stuck_out_tongue_winking_eye:

utils/machine/machine.go
```diff
@@ -157,6 +160,7 @@ func GetTopology(sysFs sysfs.SysFs, cpuinfo string) ([]info.Node, int, error) {
 				lastNode = -1
 			}
 			lastThread = thread
+			continue
```
timstclair: This whole loop could use some cleanup. I think it would be much more efficient and readable to switch on `strings.TrimSpace(line[:strings.Index(line, ":")-1])`

That's sort of a separate change, so up to you where / whether you want to address it.

timstclair:
LGTM (please squash)

> Integration tests pass, so it must be good right?

Of course! I always like to have some simple unit tests for regexes, but I won't force you to add them if you aren't up for it...

> As an observation there is a lot we could do to lower resource utilisation. Regexps are hungry & need to be used judiciously. They're overused here where we could probably do some cheaper string parsing.

Agreed. Although I don't think it's worth spending too much time on this without first doing some profiling to identify the places it really matters.

utils/sysinfo/sysinfo.go
jimmidyson: Done.

jimmidyson:
Yeah we need to do more profiling of cadvisor in general. There have been quite a few performance issues when embedded in the kubelet & we need to identify areas for improvement. Regexes were a recurrent theme in those perf problems, hence me looking at them here. I think with the work we've done recently in cadvisor this is improved significantly, but further work needs to be done to keep cadvisor's footprint as low as possible.

utils/machine/machine.go
jimmidyson: Agreed. Not going to do it here, luckily AFAICT this is only called once on start of cadvisor so efficiency isn't paramount. Would like to improve it though, for vanity more than anything else :-b

jimmidyson:
retest this please

jimmidyson:
Self-merging after LGTM.

PR #985, opened 2015-11-27
Cleanup code from gometalinter output

jimmidyson:
More to do but a quick delete of a couple of hundred lines of unused code isn't too bad for now.


jimmidyson:
Self-merging as integration test cluster is broken & this is just unused code.

PR #986, opened 2015-11-29
Fix machine ID file path when running in container

### LINKED ISSUE #955: machineIdFilePath not found in containerized Kubelet
In containerized kubelet on startup I get an error because it's looking for /etc/machine-id which it can't find.  Should machineIdFilePath be looking at /rootfs/etc/machine-id if it doesn't find /etc/machine-id?

jimmidyson:
Fixes #955 

This is a quick fix, we need to do a full clean-up sometime of filepaths when running in a container rather than passing `inHostNamespace` everywhere, but this is good enough for now IMO.

### DIFF
--- a/manager/machine.go
+++ b/manager/machine.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"flag"
 	"io/ioutil"
+	"path/filepath"
 	"strings"
 	"syscall"
 
@@ -50,8 +51,13 @@ func getInfoFromFiles(filePaths string) string {
 	return ""
 }
 
-func getMachineInfo(sysFs sysfs.SysFs, fsInfo fs.FsInfo) (*info.MachineInfo, error) {
-	cpuinfo, err := ioutil.ReadFile("/proc/cpuinfo")
+func getMachineInfo(sysFs sysfs.SysFs, fsInfo fs.FsInfo, inHostNamespace bool) (*info.MachineInfo, error) {
+	rootFs := "/"
+	if !inHostNamespace {
+		rootFs = "/rootfs"
+	}
+
+	cpuinfo, err := ioutil.ReadFile(filepath.Join(rootFs, "/proc/cpuinfo"))
 	clockSpeed, err := machine.GetClockSpeed(cpuinfo)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


vishh:
LGTM.

jimmidyson:
@vishh The integration test environment is screwed & needs fixing. I've tried to contact you before to fix but no response. I have no access to integration environment or I would have looks - perhaps you could sort me out with access? Also give me access to cadvisor docker hub builds?

vishh:
@jimmidyson: I was on vacation for the last two weeks. I will take a look at the e2e infra today. I will also add you to the admin list on the jenkins server.

As for dockerhub, I'm not sure if I can grant access to non-googlers since it is a Google owned organization.
Moving to a new org is a possibility, but I suspect users will still try to access `google/cadvisor`.

jimmidyson:
Hope you had a great, well-deserved, break.

You should be able to add me as a collaborator on the Docker hub repository for just cadvisor - think that's now possible.

vishh:
I have marked you as an `Admin` on jenkins. Try logging in again.

jimmidyson:
Perfect thanks @vishh - will have a nose around & see if it's anything obvious I can fix.

jimmidyson:
Seems to be a problem with coreos beta only. Investigating.

manager/machine.go
```diff
@@ -97,9 +103,9 @@ func getMachineInfo(sysFs sysfs.SysFs, fsInfo fs.FsInfo) (*info.MachineInfo, err
 		DiskMap:        diskMap,
 		NetworkDevices: netDevices,
 		Topology:       topology,
-		MachineID:      getInfoFromFiles(*machineIdFilePath),
+		MachineID:      getInfoFromFiles(path.Join(rootFs, *machineIdFilePath)),
 		SystemUUID:     systemUUID,
-		BootID:         getInfoFromFiles(*bootIdFilePath),
+		BootID:         getInfoFromFiles(path.Join(rootFs, *bootIdFilePath)),
```
carmark: How about using `path/filepath` instead of `path`?  `path` is intended for forward slash-separated paths only (such as those used in URLs), while `path/filepath` manipulates paths across different operating systems, although the cadvisor only support linux system.

jimmidyson: Good catch!

jimmidyson:
retest this please

jimmidyson:
retest this please

jimmidyson:
Merging after tests now pass.

PR #987, opened 2015-11-30
re-order the import package

carmark:
Make sure the import package in each file as the follow order:
* `golang` self package
* `cadvisor` package
* third-party package

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -28,11 +28,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/glog"
 	"github.com/google/cadvisor/events"
 	httpmux "github.com/google/cadvisor/http/mux"
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/manager"
+
+	"github.com/golang/glog"
 )
 
 const (
--- a/api/versions.go
+++ b/api/versions.go
@@ -21,10 +21,11 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/golang/glog"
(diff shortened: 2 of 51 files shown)


jimmidyson:
Thanks for the PR. Did you do this manually or via a tool like `goimports`? If via a tool then can we also get the check added to the build scripts to verify subsequent PRs don't mess up ordering?

carmark:
@jimmidyson 

I read the code, and modify the import packages manually at the same time.  The tool `goimports` seems not working for this change.

jimmidyson:
Why do you feel strongly about this package order? What different order does `goimports` produce? My issue with manually doing this is keeping it ordered going forward. Need tooling to support.

carmark:
This order make the packages imported very clear, the tool `goimports` will only separate the golang packages and other packages into two parts.  However, the order of this patch will have another parts.

It is truly that there is no any tool to support this, and I do not very strongly feel this patch is very necessary, but [kubernetes](https://github.com/kubernetes/kubernetes) uses this kind of order.

jimmidyson:
OK that makes sense.

Going to have to keep a close eye on import order in future.

Thanks!

jimmidyson:
ok to test

jimmidyson:
Merging as the integration testing environment is broken, but this change only affects import order & unit tests pass.

jimmidyson:
Thanks @carmark.

carmark:
Thanks @jimmidyson.  I will also try to keep a close eye on the new import packages.

Issue #989, opened 2015-11-30
cadvisor fails to unmarshal container config with Docker 1.8.3


jimmidyson:
Not sure if it happens with other versions prior to Docker 1.9, but since the merge of #966 to update runc container config fails to unmarshal properly. The problem is with this commit opencontainers/runc@c6e406af243fab0c9636539c1cb5f4d60fe0787f, turning certain fields e.g. https://github.com/opencontainers/runc/commit/c6e406af243fab0c9636539c1cb5f4d60fe0787f#diff-6318b8edda382c370345a4cb18e70b95R70 from string to slice.

@afein Would you be able to try to fix up?

afein:
I will not have the time to actually dig into this in detail any time soon.

We can revert #966 and I can resubmit the Godep bump PR with the exact version of runc that corresponds to Docker 1.8.3. Would that be an adequate fix?

jimmidyson:
Understand about time constraints. Thanks for replying.

I don't want to revert it as it will break with Docker >= 1.9 so need to come up with a way to handle it all, probably a custom Unmarshal function or something. I'll see if I can come up with something soon-ish.

carmark:
@jimmidyson 

I have a simple fix, it works well after basic test.  I will submit a PR later when I do more test.

jimmidyson:
@carmark awesome! Feel free to submit a PR now so we can help review, test, etc if you want.

PR #990, opened 2015-12-01
fix unmarshal container config failure with Docker 1.8.3

carmark:
fix https://github.com/google/cadvisor/issues/989

Test with docker 1.8.3, it works well.


jimmidyson:
Looks good but needs a couple of tests if possible with configs from < 1.8, 1.8.3 & >=1.9 just to check we haven't broken anything.

carmark:
@jimmidyson 

Yeah, I will do more tests to make sure there is no any regression.

Thanks

container/libcontainer/compatibility.go
```diff
@@ -159,8 +195,22 @@ func ReadConfig(dockerRoot, dockerRun, containerID string) (*configs.Config, err
 
 		var state libcontainer.State
 		err = json.Unmarshal(out, &state)
-		if err != nil {
-			return nil, err
+		if err != nil && strings.Contains(err.Error(), "json: cannot unmarshal") {
```
jimmidyson: Should we check the type of this is a `json.UnmarshalTypeError` instead of checking the string contents?

container/libcontainer/compatibility.go
```diff
@@ -159,8 +195,22 @@ func ReadConfig(dockerRoot, dockerRun, containerID string) (*configs.Config, err
...
+			var oldstate oldState
```
jimmidyson: Can we take this out into a separate function? Better readability & easier to test.

jimmidyson: I mean this whole convert old Config to new block.

container/libcontainer/compatibility.go
```diff
@@ -234,6 +284,75 @@ func ReadConfig(dockerRoot, dockerRun, containerID string) (*configs.Config, err
...
 
+func getThrottleFromString(str string) *configs.ThrottleDevice {
+	fields := strings.Split(str, " ")
+	if len(fields) < 2 {
+		rate, _ := strconv.ParseUint(fields[0], 10, 64)
+		return &configs.ThrottleDevice{Rate: rate}
+	}
+	var (
+		rstr  string = fields[1]
+		major int64  = 8 // by default, use /dev/sda
```
jimmidyson: Can you link to where this is documented?

jimmidyson:
I'm wondering if we could just drop the throttled devices for now as we don't use that config AFAIK so processing for no need. I'm happy to do that & put that in a comment unless you can think of a reason why not to do that.

carmark:
@jimmidyson 

When I implemented this, I also want to drop these throttled devices, but I am afraid I missed something.  So I will remove those codes and put the new changes into a new function.

Thanks,

carmark:
@jimmidyson 

I updated the changes, and tested it with docker 1.8.2, 1.8.3 and 1.9.1.  All of them works well.  Could you please review them again?

container/libcontainer/compatibility.go
```diff
@@ -234,6 +276,58 @@ func ReadConfig(dockerRoot, dockerRun, containerID string) (*configs.Config, err
...
+		HugetlbLimit:      old.HugetlbLimit,
+		Slice:             old.Slice,
+		OomKillDisable:    old.OomKillDisable,
+		MemorySwappiness:  old.MemorySwappiness,
+		NetPrioIfpriomap:  old.NetPrioIfpriomap,
+		NetClsClassid:     old.NetClsClassid,
+	}
+
+	// For now, we do not need these ThrottleDevices
+	newCgroup.BlkioThrottleReadBpsDevice = []*configs.ThrottleDevice{}
```
jimmidyson: These are all redundant assignments aren't they?

container/libcontainer/compatibility.go
```diff
@@ -158,9 +192,17 @@ func ReadConfig(dockerRoot, dockerRun, containerID string) (*configs.Config, err
 		}
 
 		var state libcontainer.State
-		err = json.Unmarshal(out, &state)
-		if err != nil {
-			return nil, err
+		if err = json.Unmarshal(out, &state); err != nil {
+			if _, ok := err.(*json.UnmarshalTypeError); ok {
```
jimmidyson: Not sure this would be a pointer type?

jimmidyson:
Minor nits. Really need some tests now just to confirm it's fixed & compatible.

jimmidyson:
ok to test

container/libcontainer/compatibility.go
carmark: Yeah, I checked most of codes which used `json.UnmarshalTypeError`, it should be pointer type.

carmark:
@jimmidyson 

>Minor nits. Really need some tests now just to confirm it's fixed & compatible.

I also think so, it will be better to have more integration test cases for different docker version.

jimmidyson:
We have integration tests for multiple versions right now, this is why most PRs are broken now...

Could you please add some unit tests & then we can merge?

Thanks very much for this - very much appreciated.

carmark:
Sure, I will try to add some unit tests on this.

jimmidyson:
Thank you!

container/libcontainer/compatibility.go
```diff
@@ -144,6 +144,40 @@ type preAPINetwork struct {
 	TxQueueLen int `json:"txqueuelen,omitempty"`
 }
 
+type oldCgroup struct {
```
vishh: nit: old doesn't really mean anything once we face this issue again. 
Shall we rename this to v1Cgroup or something similar to that? Let's also add a few comments describing the versions of docker this version is compatible with.

container/libcontainer/compatibility.go
vishh: +1. I'd expect higher layers to be resilient to `nil` values.

vishh:
Thanks for the fix @carmark !

container/libcontainer/compatibility.go
carmark: Yeah, I will rename it and add some comments in next commit with the unit test.

carmark:
@vishh  @jimmidyson 

Made some changes based on your comments and add the basic unit tests.  I am not sure why the `default` test fails.

jimmidyson:
Flaky test

jimmidyson:
retest this please

jimmidyson:
retest this please

jimmidyson:
LGTM

jimmidyson:
Thanks @carmark!

PR #992, opened 2015-12-01
Make docker build more consistent across branches/tags

jimmidyson:
This enables a simple `make docker-build` to build Docker images which will be tagged with the currently checked out tag or branch, with the exception of master branch which will be tagged as canary.

`make docker-push` will push the created image.

This will enable us to run CI builds for cadvisor in a consistent fashion for both master & tags.


jimmidyson:
@vishh Comments appreciated.

deploy/canary/Dockerfile
```diff
@@ -1,10 +0,0 @@
-FROM golang:latest
```
vishh: Why delete this file? This Dockerfile makes it possible to build images using just `docker build ...`?
Once we get the jenkins automatic builder to work reliably using makefiles, I'd be comfortable with removing this Dockerfile.

jimmidyson: The `Dockerfile` has moved in this PR to the top level dir & you can use just `make docker-build` to build the image.

vishh: With this change will I be able to run `docker build -t ..` without requiring `make` and `golang`?

jimmidyson: No - that was my original intention but was a bit tricky to achieve while still running on alpine due to glibc linking issues for runc. I'll take another look & see what I can do.

vishh: Ok. My assumption is that this `canary` Dockerfile is still functional.

jimmidyson: What do you mean? At the end of this I'd like to see a single Dockerfile that we can use Docker hub to build on commit to master (`canary` tag) & any tag (tagged with appropriate tag).

jimmidyson:
With the new Dockerfile in the root of the repo you can build with `docker build ...` & nothing more.

I'd like to move to this approach & use Docker hub builds with builds from regex tags so we can ensure a Docker image is built for every tag automatically when we create a release tag.

vishh:
Can Dockerhub support docker builds using Makefiles?
To build with just `docker build ....`, I need access to source code and go
tools from within the image right?

jimmidyson:
With the latest Dockerfile it is just a case of docker build, no make required. Try it out!

deploy/canary/Dockerfile
pwittrock: Yeah would be great to get rid of the canary.  If dockerhub automated builds can do a periodic build from master/head as well, then we can stop building it from jenkins.

pwittrock:
I just tried it out, and it is pretty slick.  Even works on my mac.

I got a couple of these types of warnings:
link: warning: option -X github.com/google/cadvisor/version.Version 0.20.1 may not work in future releases; use -X github.com/google/cadvisor/version.Version=0.20.1

Are these expected? / Is it possible to make them go away?

jimmidyson:
Those warnings are due to the fact that cadvisor has to support with go 1.4 & 1.5. The only way to get rid of these warnings is to change the linker flags in build scripts in a way that would be incompatible with 1.4. I don't mind making go 1.5 a requirement personally.

vishh:
+1 from my side for requiring 1.5

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM alpine:3.2
+MAINTAINER dengnan@google.com vmarmol@google.com vishnuk@google.com jimmidyson@gmail.com
+
+WORKDIR /gopath/src/github.com/google/cadvisor
+COPY    . /gopath/src/github.com/google/cadvisor
+
+RUN apk add --update --repository=http://dl-4.alpinelinux.org/alpine/edge/community/ -t build-deps go sqlite-dev btrfs-progs-dev bash linux-headers make git gcc g++ && \
```
pwittrock: Since this is run after the copy, docker won't cache any of these commands.  Does it make sense to split out the commands that do not depend on the src code and run them first?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    apk add --update device-mapper && \
+    export GOPATH=/gopath && \
```
pwittrock: Why use the docker ENV directives for these so that they would be available to any images that inherit from this one?

Dockerfile
jimmidyson: This would end up with a bigger final image as the build dependencies will still be present in lower layers.

Dockerfile
jimmidyson: Because these env vars aren't needed after the image build & all go build tools are cleaned up anyway. Not expecting any images to extend this one & build go stuff.

Dockerfile
pwittrock: Interesting.  I hadn't considered some of the implications of the way docker handles layers.  Does this imply that cleaning up data from previous layers does not reduce the aggregate image size.  How important is image size for cadvisor?  Is there a threshold we want to keep it under?  FWIW, the git src looks like it consumes ~30MB, while the image "virtual size" is ~65MB, so copying the source and building cadvisor within the Dockerfile will increase the image size by 30MB.  (not sure if this is considered a lot or trivial)

```
$ docker exec -ti e58b09c05e64 du -h -d 1 /gopath/
29.7M   /gopath/src
9.8M    /gopath/bin
764.0K  /gopath/pkg
40.3M   /gopath/
```

```
$ docker images
REPOSITORY                                        TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
pwittrock/cadvisor                                latest              763321a5339a        About an hour ago   62.56 MB
```

Dockerfile
pwittrock: Yeah if none of the tools are present, it doesn't help much.  I was thinking in the future we could even support running tests through Docker to make cross platform development easier, but this isn't a goal at this point.

Dockerfile
pwittrock: Image sizes for canary and 0.19.3 for reference (760MB and 35MB respectively)
```
bash-3.2$ docker images | grep cadvisor
google/cadvisor                                canary               ef54655a0485        17 minutes ago      760.5 MB
google/cadvisor                                0.19.3               548619cdac12        4 weeks ago         34.99 MB
```

deploy/canary/Dockerfile
jimmidyson: Docker hub can do builds on any push to any branch/tag & tag the docker image appropriately.

Dockerfile
jimmidyson: So this change makes the prod image larger as it still includes source in the final image - no way to delete that unfortunately as it's in a lower layer. If we want to keep small image size without source then perhaps we should go with an external build via jenkins/circleci? Pretty easy to set up.

Dockerfile
jimmidyson: I'd expect most tests to go through a second docker image & use the cadvisor API rather than extending this image.

pwittrock:
fyi, new runs of the Jenkins GCE e2e tests should export the build log (+inline in the browser)

Dockerfile
pwittrock: yeah, the docker build isn't optimized to support small image sizes.  I really don't have enough context on how the cadvisor image is used to know if the increase in size is an issue for folks using it.  I like your approach because it provides a consistent way of building cadvisor, and having docker hub build the image sounds like it could reduce maintenance cost.

timstclair:
This PR has been labeled inactive due to lack of recent activity. If it remains in this state for more than 2 weeks, it will be closed. If it is still being developed, please provide a status update.

PR #993, opened 2015-12-01
Make flaky network test less flaky

jimmidyson:


integration/tests/api/docker_test.go
```diff
@@ -253,6 +253,9 @@ func TestDockerContainerNetworkStats(t *testing.T) {
 	containerId := fm.Docker().RunBusybox("watch", "-n1", "wget", "https://www.google.com/")
 	waitForContainer(containerId, fm)
 
+	// Add a little sleep to let the request complete a few times
```
vishh: What request? `docker run`? `waitForContainer` is supposed to ensure that the container is running right?
Are you trying to give some time for cadvisor to gather metrics?

jimmidyson: The `wget` req on line 253 (https://github.com/google/cadvisor/pull/993/files#diff-6333e565f1256ee66767acf24186c52cR253). That's what should be causing network traffic. The problem I think that's happening is there's a race between container is up check via `waitForContainer` & first `wget` requests.

jimmidyson:
Closing - this doesn't seem to have helped at all. Wondering now if the test is actually valid?

vishh:
That test needs some more love and care :(

PR #996, opened 2015-12-03
Fix case statements dealing with storageDriver

jonboulle:
cf0adcc817e422e86966a24a1dc758369f3c076f introduced two switch
statements to facilitate the addition of the `overlayStorageDriver`;
unfortunately neither of them conform to the Go switch semantic, which
does not fallthrough unless explicitly requested. In one case this was
innocuous (because a `break` was effectively the same as a no-op) but in
the other it would cause the `HasFilesystem` bool to not be set
appropriately in the case of `aufsStorageDriver` being used.

IMHO it's also more idiomatic to perform the default behaviour in the
default case rather than pre-setting and overriding it.


vishh:
ok to test

vishh:
Thanks for the fix @jonboulle!

jimmidyson:
thanks @jonboulle - merging.

PR #999, opened 2015-12-03
Modified links to the right lines

Mistermatt007:
I should have done it in the previous PR


jimmidyson:
Even better ;)

jimmidyson:
We'll have to remember to keep docs up to date if the source changes.

Mistermatt007:
The doc I edited needs to be more complete IMO

PR #1002, opened 2015-12-04
Fix usage of housekeeping_interval flag

timstclair:
Defer calculation of `loadDecay`. Flags must be parsed before they can
be read, and therefore cannot be reliably be read at package init time.


jimmidyson:
ok to test

jimmidyson:
LGTM

jimmidyson:
Thanks @timstclair.

PR #1006, opened 2015-12-04
Fix goroutine leak in docker fs handler logic.

### LINKED ISSUE #994: Container filesystem check doesn't handle container deletes
Continually logs:

    failed to collect filesystem stats - du command failed on /var/lib/docker/containers/568ec44c421500145adddfe250d9ee24512a030e880aff21318361ed1f9b5a33 with output du: cannot access ‘/var/lib/docker/containers/568ec44c421500145adddfe250d9ee24512a030e880aff21318361ed1f9b5a33’: No such file or directory

For containers that are started & then deleted.

@vishh This one for you I think.

vishh:
Fixes #994

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -78,4 +78,7 @@ type ContainerHandler interface {
 
 	// Returns whether the container still exists.
 	Exists() bool
+
+	// Cleanup frees up any resources being held like fds or go routines, etc.
+	Cleanup()
 }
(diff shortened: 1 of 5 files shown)


vishh:
I'm starting to work on an e2e test that will attempt to catch goroutine leaks. I tested this PR manually.

dchen1107:
LGTM

PR #1008, opened 2015-12-05
Integration test to track goroutine leaks.

vishh:
This PR only captures the docker side logic.
We need to add another test that captures cadvisor as a whole.

### DIFF
--- a/integration/framework/framework.go
+++ b/integration/framework/framework.go
@@ -18,7 +18,11 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os/exec"
+	"regexp"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -125,6 +129,7 @@ type CadvisorActions interface {
 	// Returns a cAdvisor client to the machine being tested.
 	Client() *client.Client
 	ClientV2() *v2.Client
+	GetGoRoutines() (int, error)
 }
 
 type realFramework struct {
@@ -161,6 +166,28 @@ func (self HostnameInfo) FullHostname() string {
 
 func (self *realFramework) T() *testing.T {
 	return self.t
+}
+
+func (self *realFramework) GetGoRoutines() (int, error) {
+
+	pprofURL := fmt.Sprintf("%s/debug/pprof", self.Hostname().FullHostname())
+	resp, err := http.Get(pprofURL)
+	if err != nil {
+		return -1, err
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return -1, err
+	}
+	r := regexp.MustCompile("right>([0-9]+)<td")
+	for _, line := range strings.Split(string(body), "\n") {
+		// <tr><td align=right>42<td><a href="goroutine?debug=1">goroutine</a>
+		if strings.Contains(line, "goroutine") {
+			return strconv.Atoi(r.FindStringSubmatch(line)[1])
+		}
+	}
+	return -1, fmt.Errorf("failed to find number of goroutines from '/debug/pprof' output - %q", string(body))
 }
 
 func (self *realFramework) Hostname() HostnameInfo {
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -270,3 +270,22 @@ func TestDockerContainerNetworkStats(t *testing.T) {
 	assert.NotEqual(stat.Network.RxBytes, stat.Network.TxBytes, "Network tx and rx bytes should not be equal")
 	assert.NotEqual(stat.Network.RxPackets, stat.Network.TxPackets, "Network tx and rx packets should not be equal")
 }
+
+// Tests goroutine leaks in docker handler.
+func TestGoRoutineSanity(t *testing.T) {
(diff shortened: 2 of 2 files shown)


vishh:
cc @dchen1107 @yujuhong

integration/tests/api/docker_test.go
```diff
@@ -270,3 +270,22 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+	fm := framework.New(t)
+
+	assert := assert.New(t)
+	initialGoRoutines, err := fm.Cadvisor().GetGoRoutines()
+	assert.NoError(err)
+	containerId1 := fm.Docker().RunPause()
+	containerId2 := fm.Docker().RunPause()
+	waitForContainer(containerId1, fm)
+	waitForContainer(containerId2, fm)
+	fm.Cleanup()
```
jimmidyson: Shouldn't this be deferred straight after `framework.New()` to make sure we don't leave any containers lying around if it bombs out earlier?

integration/framework/framework.go
```diff
@@ -163,6 +168,28 @@ func (self *realFramework) T() *testing.T {
...
+		return -1, err
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return -1, err
+	}
+	r := regexp.MustCompile("right>([0-9]+)<td")
+	for _, line := range strings.Split(string(body), "\n") {
+		// <tr><td align=right>42<td><a href="goroutine?debug=1">goroutine</a>
```
jimmidyson: This seems really fragile. Is this the only way to get this info?

yujuhong: You can get this from Prometheus endpoint /metrics as well.

```
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 235
```

integration/tests/api/docker_test.go
```diff
@@ -270,3 +270,22 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+	// Adding the sleep here to let cAdvisor stabilize internally.
+	time.Sleep(5 * time.Second)
```
yujuhong: Using an arbitrary sleep time seems brittle. Is there any other way to confirm that cadvisor has stabilized?

integration/tests/api/docker_test.go
```diff
@@ -270,3 +270,22 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+	goRoutines, err := fm.Cadvisor().GetGoRoutines()
+	assert.NoError(err)
+	assert.True(goRoutines <= initialGoRoutines, "expected initial go routines to match current count.")
```
yujuhong: Are we sure that there wouldn't be any short-lived goroutines causing the test to fail, e.g., goroutines serving http requests, etc? I am also not sure whether garbage collector may spawn goroutines, but I think it's possible. 

How about we start more containers in this test so that we can allow some margin for fluctuations in #goroutines.

vishh: I tried this test a few times and it works. The number of threads do change a bit before it stabilizes. Hence the delay.

Irrespective of how many containers we start, the purpose of the test is to check that the initial and final number of goroutines match.

integration/tests/api/docker_test.go
vishh: Not that I can think of? Any suggestions?

integration/tests/api/docker_test.go
vishh: That's how it usually works. In this test though, I want to explicitly delete the containers.

integration/framework/framework.go
vishh: Golang is yet to formalize a format for pprof data. I don't expect golang to change this format anytime soon. 
This is a test. As long as the data is valid, do we really care about how we get the data?

integration/tests/api/docker_test.go
yujuhong: I don't have any good idea, but I'll throw something out. How about letting the test check periodically for a longer timeout until #goroutines <= initial#goroutines is true? This way we end the test quickly when cadvisor responds faster, but wait for a longer timeout if cadvisor is slow for other reasons.

vishh: Sure.

integration/tests/api/docker_test.go
yujuhong: Is there any undesirable side-effect of calling the Cleanup() function twice? If not, it might still be worth adding `defer fm.Cleanup()` to handle misbehaved tests

vishh: AFAIK Cleanup() is idempotent. It will attempt to cleanup and fail, which
is probably OK. I really don't think that this is an issue with this test
though.

jimmidyson:
How about a Prometheus metric to hold metrics for number of specific goroutines? As goroutine starts, increment gauge; decrement on stop. That seems like a more foolproof way to do this?

yujuhong:
> How about a Prometheus metric to hold metrics for number of specific goroutines? As goroutine starts, increment gauge; decrement on stop. That seems like a more foolproof way to do this?

That sounds like a good idea, and will probably be more useful for debugging goroutine leak in a regular cluster.

timstclair:
Ping. What's the status of this?

PR #1010, opened 2015-12-08
Added possibility to not ignore NumStats field in ContainerInfoRequest

piosz:


piosz:
cc @dchen1107 @fgrzadkowski @jszczepkowski @mwielgus @vishh

jimmidyson:
Can you describe the use case please?

jimmidyson:
Shouldn't we just respect max results if it's specified?

timstclair:
> Shouldn't we just respect max results if it's specified?

+1 - I think we should just remove the logic that ignores NumStats when a time range is specified.

piosz:
Yes, that would be easier but we have already defined semantic of the request (see [container.go](https://github.com/google/cadvisor/blob/master/info/v1/container.go#L96)) and this would be a breaking change. @rjnagal @vishh @vmarmol any thoughts?

Since we are downsampling the stats to exactly one in [Heapster](https://github.com/kubernetes/heapster/blob/master/sources/datasource/kubelet.go#L69) we can do it on cadvisor side and reduce the network traffic.

vishh:
In a given time range, if NumStats is specified, which samples should be returned? Should it be the oldest or the most recent?

piosz:
The most recent one.

vishh:
That change SGTM. We can perhaps send out an announcement on
`google-containers@` regarding this API change.

jimmidyson:
Definitely most recent samples.

I'm happy with the breaking change. cadvisor is still only a 0.x release so API stability not guaranteed :stuck_out_tongue_closed_eyes:

piosz:
cadvisorJenkinsBot writing comments... sounds interesting ;)

piosz:
Changed the code to make a breaking change.

jimmidyson:
@cadvisorJenkinsBot has become sentient!

vishh:
ok to test

rjnagal:
That was me reply through mail and expecting github to do the right thing. I can be easily mistaken for a bot though :(

vishh:
add to whitelist

piosz:
@rjnagal @vishh could you please grant LGTM to the PR?

The integration test is broken:
```
Error 0: failed to make remote testing directory: command "gcloud" ["compute" "ssh" "--zone" "us-central1-f" "cadvisor-ubuntu-trusty" "--" "mkdir" "-p" "/tmp/cadvisor-5471"] failed with error: exit status 1 and output: "ERROR: (gcloud.compute.ssh) The required property [project] is not currently set.\nYou may set it for your current workspace by running:\n\n  $ gcloud config set project VALUE\n\nor it can be set temporarily by the environment variable [CLOUDSDK_CORE_PROJECT]\n"
Build step 'Execute shell' marked build as failure
```

jimmidyson:
That's normally a transient error - let's try retesting...

jimmidyson:
retest this please

piosz:
thanks @jimmidyson

jimmidyson:
I haven't looked (lazy!) but are we defaulting max stats to -1 somewhere?

piosz:
According to the documentation the default is 60 (see https://github.com/google/cadvisor/blob/master/info/v1/container.go#L95).

jimmidyson:
The failure is unrelated... Jenkins seems to get like this sometimes.

jimmidyson:
retest this please

jimmidyson:
Why 60? Assume that means 1 minute of stats but seems a strange default when housekeeping time is configurable.

But we can discuss that in a different PR. IMO -1 should be the default if unspecified.

piosz:
create an issue #1011 for this purpose

jimmidyson:
Anyway LGTM

jimmidyson:
Unit tests run fine for me locally so happy to merge.

piosz:
Thanks @jimmidyson. What is the release process of cadvisor?

jimmidyson:
Are you looking to get this into kubernetes ASAP? I have a PR to bump cadvisor so I can tag it here & include this bump in that PR if you'd like?

piosz:
Yes, please. Thank you very much in advance.

vishh:
Piotr, can you send out an announcement about this API change to
google-containers@ ?

PR #1013, opened 2015-12-10
Fix cgroupPathRegExp to match path after the first colon after devices

Snorch:
If in getCgroupPath in cgroups we have some other hierarchies after
"devices" using ".*" will cause matching wrong string as a container path
so we need to prefer fewer matches here with ".*?".

e.g.
If cgroups is 14:devices,name=named_cgroup:/test.slice,1:name=systemd:/user.slice/user-1000.slice/session-1.scope
match[1] will be /user.slice/user-1000.slice/session-1.scope
but not /test.slice as intended.

These fixes the commit:
4cbd91c761 Make getCgroupPath work in case of named or multi- hierarchies

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -43,7 +43,7 @@ import (
 // Housekeeping interval.
 var HousekeepingInterval = flag.Duration("housekeeping_interval", 1*time.Second, "Interval between container housekeepings")
 
-var cgroupPathRegExp = regexp.MustCompile(`.*devices.*:(.*?)[,;$].*`)
+var cgroupPathRegExp = regexp.MustCompile(`devices[^:]*:(.*?)[,;$]`)
 
 type containerInfo struct {
 	info.ContainerReference


vishh:
ok to test

vishh:
@Snorch: Wonder why those cgroups are co-mounted. Is there a specific reason?

manager/container.go
```diff
@@ -43,7 +43,7 @@ import (
 // Housekeeping interval.
 var HousekeepingInterval = flag.Duration("housekeeping_interval", 1*time.Second, "Interval between container housekeepings")
 
-var cgroupPathRegExp = regexp.MustCompile(`.*devices.*:(.*?)[,;$].*`)
+var cgroupPathRegExp = regexp.MustCompile(`.*devices.*?:(.*?)[,;$].*`)
```
timstclair: I think using a negated character set is prefered to lazy matching. Also, the `.*` on either end isn't doing anything except slowing this down.

timstclair:
I wasn't able to reproduce your result: https://gist.github.com/timstclair/5a1abd7fdee2ba93e337
The original regexp yielded the expected "/test.slice". What version of go were you using?

Snorch:
Sorry my example were not working, here is an example with real cgroups string I got on my node:
https://play.golang.org/p/NrmYySNVIB

manager/container.go
Snorch: Thanks for your comment, I will update pull request.

timstclair: Can you remove the `.*` at the beginning and end too?

Snorch: Ok, updated.

timstclair:
@k8s-bot ok to test

LGTM

PR #1017, opened 2015-12-11
Retry cadvisor e2e tests to compensate for test flakyness

pwittrock:


jimmidyson:
@pwittrock:

> The following files are not properly formatted:
integration/runner/runner.go integration/runner/runner.go

jimmidyson:
@vishh I do hope that cadvisor CI isn't moving to a non-public Jenkins... makes it hard for non-Google maintainers to help with contributions.

pwittrock:
@jimmidyson wow that was an amazingly fast turn around.  ran `go vet` and `go fmt` on runner.go.

jimmidyson:
@pwittrock Aim to please :-b

Can no longer see why Jenkins build is failing as it's been moved to internal google jenkins it seems... Will have to defer to @vishh for that sadly.

jimmidyson:
retest this please

pwittrock:
@jimmidyson SGMT.  fwiw the test failures were because the internal jenkins is not fully configured, and it seemed to hide the external jenkins test results.  Will sync up with vishh w.r.t to your comment about the challenges with internal jenkins PR builders.

pwittrock:
retest this please

vishh:
@pwittrock: I wonder if it is possible to delete previous posts of k8s-bot whenever it posts a new update? In its current state its too spammy :(

integration/runner/runner.go
```diff
@@ -117,11 +118,17 @@ func PushAndRunTests(host, testDir string) error {
 	// Run the tests.
 	glog.Infof("Running integration tests targeting %q...", host)
 	err = RunCommand("godep", "go", "test", "github.com/google/cadvisor/integration/tests/...", "--host", host, "--port", portStr)
```
vishh: nit: To avoid having two commands that are expected to be identical, can we increase the retry count to be `6` instead and run the command only inside the loop or cache the command in a separate var?

vishh:
Just a nit. Otherwise LGTM. Feel free to self-merge @pwittrock. Thanks for the fix!! Much appreciated.

pwittrock:
I can manually delete them, or I can disable them entirely.  The spammyness on this PR is in part due to me triggering the build a bunch while I was working on the internal cadvisor Jenkins PR builder.  I believe the kubernetes/kubernetes project has the same behavior of leaving the PR builder comments around instead of deleting them:  https://github.com/kubernetes/kubernetes/pull/17808

integration/runner/runner.go
pwittrock: Done.

jimmidyson:
Link to build log doesn't work btw...

pwittrock:
Thanks for the heads up about the build log.  This is a known problem caused by an issue in the google-storage-plugin not recognizing valid gce scopes.  I am working with our build team on a work around but e.t.a. for a fix is still a couple days.  I have tried to restrict the Experimental PR builder to my own PRs only through the admin + whitelist to minimize annoyance meanwhile.

Issue #1016, opened 2015-12-11
Introduce direct API for application metrics


jszczepkowski:
Currently, cAdvisor read configuration of application metrics from containers labels and from config file mount in container root directory. In kubernets, cAdvisor is a part of kubelet. To configure custom metrics, kubelet need to create and mount a dummy configuration file. This is a bit hacky solution.

We should consider exposing an internal API for configuring custom metrics for containers. Such API could be used by kubelet.

jszczepkowski:
CC @vishh

ncdc:
cc @DirectXMan12 

@jszczepkowski @vishh please let us know if this is an area where we could help out

vishh:
@ncdc we can use some help here. Current configuration model for custom metrics is meant for Stavanger cadvisor and doesn't make sense in the kubelet case.

DirectXMan12:
+1

Also, at the very least, I think it would make sense to allow just dropping the config currently stored in the file into a container label.  For example:
```
$ docker run ... -l io.cadvisor.metric.prometheus="{\"endpoint\": \"http://localhost:8080/metrics\"}" mycontainer
```

I can write up something like this, if you'd like

jszczepkowski:
> Also, at the very least, I think it would make sense to allow just dropping the config currently stored in the file into a container label.

+1

I really like this idea.

DirectXMan12:
It should be fairly quick to implement.  I'll see what I can whip up.

vishh:
To be clear, for the kubelet use case, what we need is an internal API that
can communicate to the cadvisor core what custom metrics are supposed to be
collected.
The labels based API is cute, but using labels for storing what is
`Annotations` makes me a bit unhappy.
Ideally the config should be part of the Docker images.

DirectXMan12:
> To be clear, for the kubelet use case, what we need is an internal API that can communicate to the cadvisor core what custom metrics are supposed to be collected.

Sure.  The labels-based idea is a quick nice-to-have.

> Ideally the config should be part of the Docker images.

Why?  It seems like it might not be uncommon to have a metrics endpoint that exposes a lot of metrics, but you might only want to use a few.  It seems silly to have to entirely rebuild your Docker image to change which metrics are exposed, when you're not really making a change that affects the rest of the image itself.  It also seems odd to leak cAdvisor details into the filesystem of the container itself.

vishh:
I guess it depends on the use case. Image authors will be aware of what
[...]

PR #1018, opened 2015-12-11
Support Loading Metrics Configs from Labels

DirectXMan12:
Previously, the custom metrics configuration was loaded
from a file in the container, by specifiying a file path
as the value for a label of the form "io.cadvisor.metrics.xyz".

Now, you can also just specify the configuration JSON directly
in label, by using a label with the name
"io.cadvisor.metric-config.xyz".

Part of #1016

### DIFF
--- a/collector/collector_manager.go
+++ b/collector/collector_manager.go
@@ -22,7 +22,10 @@ import (
 	"github.com/google/cadvisor/info/v1"
 )
 
-const metricLabelPrefix = "io.cadvisor.metric."
+const (
+	metricLabelPrefix   = "io.cadvisor.metric."
+	metricInLabelPrefix = "io.cadvisor.metric-config."
+)
 
 type GenericCollectorManager struct {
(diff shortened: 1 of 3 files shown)


vishh:
Which configuration will take precedence? File based or label based?

DirectXMan12:
@vishh what, specifically, do you mean?  Are you asking what happens if you have two configurations for the same endpoint, one file-based and one direct-label-based?  In that case, it would be exactly as if you had two file-based labels pointing to the same endpoint.  The collectors with file-based-configs get registered before collectors with direct-label-based-configs, but AFAICT the way it would work would be like the combination of the two, since all collectors get processed and add their metrics to the set.

vishh:
Embedding custom metrics configuration as part of the image makes it
possible to share the config.
If the primary use case is to query specific custom metrics from cAdvisor,
that use case is already supported.

PR #1023, opened 2015-12-16
Carry #780: Export env variables as prometheus labels

dqminh:
This rebase the patch in #780 and add a few more fixes:
- fix regex in test to do the correct thing
- sanitize all label name before setting them as prometheus label

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -49,6 +49,8 @@ var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
 var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
 
+var dockerMetadataEnvs = flag.String("docker_metadata_env", "", "Comma seperated list with names of env variables, which will be exported as metadata (default: empty)")
+
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
 func DockerStateDir() string {
@@ -117,6 +119,9 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 	if err != nil {
 		return
 	}
+
+	exposedMetadata := strings.Split(*dockerMetadataEnvs, ",")
+
 	handler, err = newDockerContainerHandler(
 		client,
 		name,
@@ -125,6 +130,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		self.storageDriver,
 		&self.cgroupSubsystems,
 		inHostNamespace,
+		exposedMetadata,
 	)
 	return
 }
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -49,7 +49,7 @@ var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
 var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
 
-var dockerMetadataEnvs = flag.String("docker_metadata_env", "", "Comma seperated list with names of env variables, which will be exported as metadata (default: empty)")
+var dockerEnvWhitelist = flag.String("docker_env_metadata_whitelist", "", "a comma-separated list of environment variable keys that needs to be collected for docker containers")
 
 // TODO(vmarmol): Export run dir too for newer Dockers.
 // Directory holding Docker container state information.
@@ -120,7 +120,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		return
 	}
 
-	exposedMetadata := strings.Split(*dockerMetadataEnvs, ",")
+	metadataEnvs := strings.Split(*dockerEnvWhitelist, ",")
 
 	handler, err = newDockerContainerHandler(
 		client,
@@ -130,7 +130,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		self.storageDriver,
 		&self.cgroupSubsystems,
 		inHostNamespace,
-		exposedMetadata,
+		metadataEnvs,
 	)
 	return
 }
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -66,8 +66,9 @@ type dockerContainerHandler struct {
 	// Time at which this container was created.
 	creationTime time.Time
 
-	// Metadata labels associated with the container.
+	// Metadata associated with the container.
 	labels map[string]string
+	envs   map[string]string
 
 	// The container PID used to switch namespaces as required
 	pid int
@@ -93,7 +94,7 @@ func newDockerContainerHandler(
 	storageDriver storageDriver,
 	cgroupSubsystems *containerlibcontainer.CgroupSubsystems,
 	inHostNamespace bool,
-	exposedMetadata []string,
+	metadataEnvs []string,
 ) (container.ContainerHandler, error) {
 	// Create the cgroup paths.
 	cgroupPaths := make(map[string]string, len(cgroupSubsystems.MountPoints))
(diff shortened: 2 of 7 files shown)


dqminh:
@vishh  i've checked our CLAs and i think we are ok here. Let me know what you think.

container/docker/handler.go
```diff
@@ -157,6 +158,18 @@ func newDockerContainerHandler(
 	handler.image = ctnr.Config.Image
 	handler.networkMode = ctnr.HostConfig.NetworkMode
 
+	// split env vars to get metadata map.
+	if len(exposedMetadata) > 0 {
+		for _, envVar := range ctnr.Config.Env {
+			splits := strings.SplitN(envVar, "=", 2)
+			for _, exposedVar := range exposedMetadata {
+				if splits[0] == exposedVar {
+					handler.labels[strings.ToLower(exposedVar)] = splits[1]
```
vishh: Should we keep env as a separate field?

dqminh: hmm in the end they all come back as metrics labels so i dont really think they are necessary.
Labels makes a lot of sense for metrics, env much less so. We are just stuck with env here because some orchestration platform doesnt support writing labels to docker container.

dqminh:
ping @vishh

container/docker/handler.go
vishh: Sure. The data generated here can be exposed in many different ways - REST API, prometheus endpoint, storage sinks, etc. I'd prefer keeping env and labels separate at this layer. 
If the prometheus endpoint chooses to combine labels and env, that's fine by me.

dqminh: agreed. Updated it to add `Envs` as a source of metadata to `ContainerSpec`. PTAL

container/docker/factory.go
```diff
@@ -49,6 +49,8 @@ var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
 var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
 
+var dockerMetadataEnvs = flag.String("docker_metadata_env", "", "Comma seperated list with names of env variables, which will be exported as metadata (default: empty)")
```
vishh: This feels more like labels. Why not expose this additional metadata as labels instead?

vishh: Infact why not add it to docker containers directly? Is this needed to provide backwards compat with old docker versions?

dqminh: main reason was noted here: https://github.com/google/cadvisor/pull/78

tldr: the orchestrator doesnt understand docker labels, and set their metadata as env variable. Until that's fixed, we need this to make sense of those containers.

vishh: Ok. Why add a flag to cadvisor?

dqminh: this is to select the env that we want to publish as metrics labels ( I made a similar flag for logging in docker https://github.com/docker/docker/pull/16961 )

For example, a mesos container has the following configs:
```
"Env": [
            "GOMAXPROCS=64",
            "MARATHON_APP_VERSION=2015-11-18T22:55:53.442Z",
            "MARATHON_APP_DOCKER_IMAGE=docker.cfops.it/elssftp:2015.10.17",
            "MESOS_TASK_ID=pdx_elsent_svc_elssftp.3f90dfce-9231-11e5-81c8-2c600c83f52f",
            "MARATHON_APP_ID=/pdx/elsent/svc/elssftp",
            "MESOS_SANDBOX=/mnt/mesos/sandbox",
            "MESOS_CONTAINER_NAME=mesos-d311c627-c678-4651-9345-3b46a02c7d99-S115.9961fba3-3680-47d4-ba5a-2b724aa46515",
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
            "container=docker"
],
"Labels": {}, // sad :(
```

we only interest in MARATHON_APP_ID,MESOS_TASK_ID,MESOS_CONTAINER_NAME as the container identifiers.

A flag is reasonably useful for use as we only use 1 orchestrator i.e. mesos, so all containers will have the same set of env keys. This is the chosen approach in the original PR.

I can imagine an alternative approach would be to let the administrator defined their desired env keys as a well known label, for example:

```
// this will not be set as metrics in prometheus
"Labels": { "cadvisor.metadata_env": "MARATHON_APP_ID,MESOS_TASK_ID,MESOS_CONTAINER_NAME" }
```

metrics/prometheus.go
```diff
@@ -571,3 +581,11 @@ func specMemoryValue(v uint64) float64 {
 	}
 	return float64(v)
 }
+
+var invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
+
+// sanitizeLabelName replaces anything that doesn't match
+// client_label.LabelNameRE with an underscore.
+func sanitizeLabelName(name string) string {
```
vishh: Why is this needed? Does it make sense to enable it by default?

container/docker/factory.go
vishh: Thanks for explaining @dqminh. The flag name doesn't seem to convey the fact that it is a *filter*. Can you fix the naming and description to make it clear that it is a filter?

info/v1/container.go
```diff
@@ -43,8 +43,9 @@ type ContainerSpec struct {
 	// Time at which the container was created.
 	CreationTime time.Time `json:"creation_time,omitempty"`
 
-	// Metadata labels associated with this container.
+	// Metadata associated with this container.
 	Labels map[string]string `json:"labels,omitempty"`
+	Envs   map[string]string `json:"envs,omitempty"`
```
vishh: We need to add this to API v2 as well.

metrics/prometheus.go
dqminh:
```
Label names may contain ASCII letters, numbers, as well as underscores.
```

From: http://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
I *think* this makes sense as the default. The other option is to ignore the invalid labels.

jimmidyson: We should sanitize rather than drop. This is normal practice for prometheus metrics.

container/docker/factory.go
dqminh: SGTM. What do you think of
```
-allowed-docker-env-as-metadata: a comma-separated list of allowed environment variable names for docker container that should have their name values set as the container's metadata
```

vishh: How about `--docker-env-metadata-whitelist: a comma-separated list of environment variable keys that needs to be collected for docker containers`?
And BTW will this default to all env variables? To me that feels like a good default.

dqminh:
```
--docker-env-metadata-whitelist: a comma-separated list of environment variable keys that needs to be collected for docker containers
```
:+1: 

Hmm, i thinkcollecting all env variables as the default is debatable.

In term of metrics collection, many env variables are strictly for runtime usage and doesn't make much sense in term of metrics which is why the whitelist approach ( i.e. disabled by default ) feels right. I think enabling all env variables by default will likely bite some users in their foot if they are not careful.

However, if we are talking about collecting information of container for exposing to others i.e. cadvisor is the only place to discover containers' information on the system then collecting all env by default makes sense.

dqminh:
@vishh added references to API v2 and changed the flag.

I'm keeping the env whitelist i.e. not collecting all env variables by default. PTAL.

container/docker/factory.go
vishh: I do agree with you. Env variables sometimes contain auth information. Disabled by default SGTM.
If we were to support collecting all env variables, then we might have to support wildcards in the flag though.

container/docker/factory.go
```diff
@@ -49,6 +49,8 @@ var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
 
 var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
 
+var dockerMetadataEnvs = flag.String("docker_env_metadata_whitelist", "", "a comma-separated list of environment variable keys that needs to be collected for docker containers")
```
vishh: nit: `dockerEnvWhitelist` would be a more readable name for this variable.

info/v2/container.go
```diff
@@ -64,8 +64,9 @@ type ContainerSpec struct {
 	// An example of a namespace is "docker" for Docker containers.
 	Namespace string `json:"namespace,omitempty"`
 
-	// Metadata labels associated with this container.
```
vishh: nit: Update the description. The existing description is valid for `Labels` field. Let's add a new one for `Envs` and mention that only whitelisted env vars are exposed, to avoid users filing issues that this field is always empty.

vishh:
Just a couple of nits. Otherwise LGTM! Thanks @dqminh

dqminh:
@vishh updated

vishh:
LGTM

vishh:
can you squash your commits?

dqminh:
@vishh squashed the changes to envs into 1 commits.

vishh:
LGTM

vishh:
@k8s-bot test this please

vishh:
@k8s-bot test this please

vishh:
Merging this since e2e has passed.

PR #1032, opened 2016-01-04
Only retry whitelisted test failure…

pwittrock:
… flakes instead of retrying on any failure.

We should explicitly be retrying on certain conditions and not blindly doing so.

### DIFF
--- a/integration/runner/runner.go
+++ b/integration/runner/runner.go
@@ -121,7 +121,6 @@ func PushAndRunTests(host, testDir string) error {
 
 	// Run the tests in a retry loop.
 	glog.Infof("Running integration tests targeting %q...", host)
-
 	for i := 0; i <= *testRetryCount; i++ {
 		// Check if this is a retry
 		if i > 0 {
@@ -141,7 +140,6 @@ func PushAndRunTests(host, testDir string) error {
 			break
 		}
 	}
-
 	if err != nil {
 		err = fmt.Errorf("error on host %s: %v", host, err)
 	}
--- a/integration/runner/runner.go
+++ b/integration/runner/runner.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"flag"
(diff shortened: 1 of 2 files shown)


jimmidyson:
LGTM! Self-merge when travis passes.

Nice work with the build logs link btw. Do you think we can disable the `default` jenkins build now?

pwittrock:
Thanks Jimmi.  :)

I'd love to disable the default Jenkins build now, especially because it seems to be in a bad state.  I'll disable it but keep the job around in case we decide we need it again for whatever reason.

Seems like the regex isn't properly catching the message, I am going to try to fix it and will comment back when it is resolved.

jimmidyson:
@pwittrock Wonder if we can do this in a slightly smarter way? Perhaps extracting regexes into separate file to match multiple flaky tests in future rather than changing code?

pwittrock:
I can separate the configuration from code by moving to a file.  Was hoping the list would remain small enough and change infrequently enough that the indirection would not be justified, but I suppose that is a bit naive.  :P  How does a newline separated list of regexes in a new file called integration/runner/retrywhitelist.txt sound?  I am thinking we probably want the file to be flag configurable?  Whitelist file location / naming suggestions?  Other considerations?

pwittrock:
PTAL.  Extracted the regexp to a file.

pwittrock:
Since the recent changes require the whitelist file to be set by flag to enable retries, they are not enabled for the PR builds

pwittrock:
@jimmidyson Since you added comments after the lgtm / self-merge, would you like to re-review (instead of leaving the lgtm)?  Let me know if you prefer the changes or had something else in mind.  Cheers.

integration/runner/runner.go
```diff
@@ -195,6 +205,32 @@ func Run() error {
 	return nil
 }
 
+// initWhiltelist initializes the whitelist of test failures that can be retried.
+func initWhitelist() {
```
jimmidyson: `initRetryWhitelist`?

jimmidyson:
LGTM after function renamed. Thanks!

jimmidyson:
LGTM - waiting on travis (being slow today). Feel free to self-merge if someone else doesn't get their first.

pwittrock:
Is it normal to take 5+ hours for travis, or do we need to kick something?

jimmidyson:
Travis has these blips every now & then. I'd rather use circleci tbh where I never see these kinds of issues.

PR #1033, opened 2016-01-05
Add the support for kafka in cAdvisor's storage, including output of container id and labels

arichardet:
Steps to run and configure with custom broker and topic. Default broker is `localhost:9092`. Default topic is `stats`.
`sudo ./cadvisor -storage_driver=kafka -storage_driver_kafka_broker_list=localhost:9092 -storage_driver_kafka_topic=stats`

This will push json encoded container statistics onto the designated topic. Includes additional output of container id and labels, added to the v1 container spec.

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -70,6 +70,9 @@ type ContainerSpec struct {
 
 // Container reference contains enough information to uniquely identify a container
 type ContainerReference struct {
+	// The container id
+	Id string `json:"id,omitempty"`
+
 	// The absolute name of the container. This is unique on the machine.
 	Name string `json:"name"`
 
@@ -80,6 +83,8 @@ type ContainerReference struct {
 	// Namespace under which the aliases of a container are unique.
 	// An example of a namespace is "docker" for Docker containers.
 	Namespace string `json:"namespace,omitempty"`
+
+	Labels map[string]string `json:"labels,omitempty"`
 }
 
 // Sorts by container name.
--- a/storage/kafka/kafka.go
+++ b/storage/kafka/kafka.go
@@ -1,0 +1,114 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kafka
+
+import (
+	"encoding/json"
+	"flag"
+	"os"
+	"strings"
+	"time"
+
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/storage"
+	"github.com/google/cadvisor/utils/container"
+
+	kafka "github.com/Shopify/sarama"
+	"github.com/golang/glog"
+)
+
+func init() {
+	storage.RegisterStorageDriver("kafka", new)
+}
+
+var (
+	brokers = flag.String("storage_driver_kafka_broker_list", "localhost:9092", "kafka broker(s) csv")
+	topic   = flag.String("storage_driver_kafka_topic", "stats", "kafka topic")
+)
+
+type kafkaStorage struct {
+	producer    kafka.AsyncProducer
+	topic       string
+	machineName string
+}
+
+type detailSpec struct {
+	Timestamp       time.Time            `json:"timestamp"`
+	MachineName     string               `json:"machine_name,omitempty"`
+	ContainerName   string               `json:"container_Name,omitempty"`
+	ContainerID     string               `json:"container_Id,omitempty"`
+	ContainerLabels map[string]string    `json:"container_labels,omitempty"`
+	ContainerStats  *info.ContainerStats `json:"container_stats,omitempty"`
+}
+
+func (driver *kafkaStorage) infoToDetailSpec(ref info.ContainerReference, stats *info.ContainerStats) *detailSpec {
+	timestamp := time.Now()
+	containerID := ref.Id
+	containerLabels := ref.Labels
+	containerName := container.GetPreferredName(ref)
+
+	detail := &detailSpec{
+		Timestamp:       timestamp,
+		MachineName:     driver.machineName,
+		ContainerName:   containerName,
+		ContainerID:     containerID,
+		ContainerLabels: containerLabels,
+		ContainerStats:  stats,
+	}
+	return detail
+}
+
+func (driver *kafkaStorage) AddStats(ref info.ContainerReference, stats *info.ContainerStats) error {
+	detail := driver.infoToDetailSpec(ref, stats)
+	b, err := json.Marshal(detail)
+
+	driver.producer.Input() <- &kafka.ProducerMessage{
+		Topic: driver.topic,
+		Value: kafka.StringEncoder(b),
+	}
+
+	return err
+}
+
+func (self *kafkaStorage) Close() error {
+	return self.producer.Close()
+}
+
+func new() (storage.StorageDriver, error) {
+	machineName, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+	return newStorage(machineName)
+}
+
+func newStorage(machineName string) (storage.StorageDriver, error) {
+	config := kafka.NewConfig()
+	config.Producer.RequiredAcks = kafka.WaitForAll
+
+	brokerList := strings.Split(*brokers, ",")
+	glog.V(4).Infof("Kafka brokers:%q", brokers)
+
+	producer, err := kafka.NewAsyncProducer(brokerList, config)
+	if err != nil {
+		return nil, err
+	}
+	ret := &kafkaStorage{
+		producer:    producer,
+		topic:       *topic,
+		machineName: machineName,
+	}
+	return ret, nil
+}
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -15,6 +15,11 @@
 			"Rev": "f92a08f515e1bf718e995076a37c2f534b1deb08"
 		},
 		{
+			"ImportPath": "github.com/Shopify/sarama",
+			"Comment": "v1.7.0",
+			"Rev": "87ec8d76e89cc002ff5673ee7598ae1db2f32424"
+		},
+		{
 			"ImportPath": "github.com/Sirupsen/logrus",
 			"Comment": "v0.7.3-2-g26709e2",
 			"Rev": "26709e2714106fb8ad40b773b711ebce25b78914"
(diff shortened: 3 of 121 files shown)


arichardet:
I signed it!

info/v1/container.go
```diff
@@ -64,10 +64,16 @@ type ContainerSpec struct {
...
 	Image string `json:"image,omitempty"`
+
+	// The ID of the container
+	ContainerId string `json:"container_id,omitempty"`
 }
 
 // Container reference contains enough information to uniquely identify a container
 type ContainerReference struct {
+	// The container id
+	Id string `json:"id"`
```
timstclair: Why is this required for a reference, but optional for the spec? I think these should be consistent, or at least the Reference should have weaker requirements than the spec.

storage/common_flags.go
```diff
@@ -26,3 +26,5 @@ var ArgDbName = flag.String("storage_driver_db", "cadvisor", "database name")
 var ArgDbTable = flag.String("storage_driver_table", "stats", "table name")
 var ArgDbIsSecure = flag.Bool("storage_driver_secure", false, "use secure connection with database")
 var ArgDbBufferDuration = flag.Duration("storage_driver_buffer_duration", 60*time.Second, "Writes in the storage driver will be buffered for this duration, and committed to the non memory backends as a single transaction")
+var ArgKafkaBrokerList = flag.String("storage_driver_kafka_broker_list", "localhost:9092", "kafka broker(s) csv")
+var ArgKafkaTopic = flag.String("storage_driver_kafka_topic", "stats", "kafka topic")
```
timstclair: I don't think these are common flags. They should be in the kafka package (e.g. like [bigquery](https://github.com/google/cadvisor/blob/master/storage/bigquery/client/client.go#L28))

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
timstclair: 2016 :)

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	MachineName     string               `json:"machine_name,omitempty"`
+	ContainerName   string               `json:"container_Name,omitempty"`
+	ContainerId     string               `json:"container_Id,omitempty"`
+	ContainerLabels map[string]string    `json:"container_labels,omitempty"`
+	ContainerStats  *info.ContainerStats `json:"container_stats,omitempty"`
+}
+
+func (driver *kafkaStorage) containerStatsAndDefaultValues(ref info.ContainerReference, stats *info.ContainerStats) *detailSpec {
+	timestamp := stats.Timestamp.UnixNano() / 1E3
+	var containerName string
```
timstclair: (nit) move this down to where it's used (just before the `if`)

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+type detailSpec struct {
+	Timestamp       int64                `json:"timestamp"`
+	MachineName     string               `json:"machine_name,omitempty"`
+	ContainerName   string               `json:"container_Name,omitempty"`
+	ContainerId     string               `json:"container_Id,omitempty"`
+	ContainerLabels map[string]string    `json:"container_labels,omitempty"`
+	ContainerStats  *info.ContainerStats `json:"container_stats,omitempty"`
+}
+
+func (driver *kafkaStorage) containerStatsAndDefaultValues(ref info.ContainerReference, stats *info.ContainerStats) *detailSpec {
```
timstclair: This method name confusing to me. Consider changing it to something like `createDetailSpec`, `getDetailSpec`, or `infoToDetailSpec`.

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	storage.RegisterStorageDriver("kafka", new)
+}
+
+type kafkaStorage struct {
+	producer    kafka.SyncProducer
+	topic       string
+	machineName string
+}
+
+type detailSpec struct {
```
timstclair: This is basically the exported API, right? Would it be useful to put this in its own package & export it, so that go kafka consumers can import the type?

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	Timestamp       int64                `json:"timestamp"`
```
timstclair: Document the format of the timestamp. Can you use time.Time instead?

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	MachineName     string               `json:"machine_name,omitempty"`
+	ContainerName   string               `json:"container_Name,omitempty"`
+	ContainerId     string               `json:"container_Id,omitempty"`
+	ContainerLabels map[string]string    `json:"container_labels,omitempty"`
+	ContainerStats  *info.ContainerStats `json:"container_stats,omitempty"`
+}
+
+func (driver *kafkaStorage) containerStatsAndDefaultValues(ref info.ContainerReference, stats *info.ContainerStats) *detailSpec {
+	timestamp := stats.Timestamp.UnixNano() / 1E3
```
timstclair: Microseconds is a bit of an unusual unit - is that a Kafka thing?

Also, for clarity, prefer:
```
timestamp := stats.Timestamp.UnixNano() / time.Microsecond
```

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	var containerName string
+	containerId := ref.Id
+	containerLabels := ref.Labels
+
+	if len(ref.Aliases) > 0 {
+		containerName = ref.Aliases[0]
+	} else {
+		containerName = ref.Name
+	}
```
timstclair: Can you move this block to a shared utility function? Something like `GetPreferredName`. It's non-obvious to me that in practice `Name` is usually the ID, and `Aliases[0]` is a more accurate representation of the container "name".

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+		ContainerId:     containerId,
+		ContainerLabels: containerLabels,
+		ContainerStats:  stats,
+	}
+	return detail
+}
+
+func (driver *kafkaStorage) AddStats(ref info.ContainerReference, stats *info.ContainerStats) error {
+	detail := driver.containerStatsAndDefaultValues(ref, stats)
+	b, _ := json.Marshal(detail)
```
timstclair: Is there a reason to ignore the error here? I think at the very least you should log it.

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+		Topic: driver.topic,
+		Value: kafka.StringEncoder(b),
+	}
+
+	_, _, err := driver.producer.SendMessage(msg)
+	if err != nil {
+		fmt.Println("kafka storage error:", err)
+	}
+
+	if stats == nil {
```
timstclair: This if is unnecessary, as you return `nil` below.

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (driver *kafkaStorage) AddStats(ref info.ContainerReference, stats *info.ContainerStats) error {
+	detail := driver.containerStatsAndDefaultValues(ref, stats)
+	b, _ := json.Marshal(detail)
+	msg := &kafka.ProducerMessage{
+		Topic: driver.topic,
+		Value: kafka.StringEncoder(b),
+	}
+
+	_, _, err := driver.producer.SendMessage(msg)
```
timstclair: `AddStats(..)` should not block. Can you just use the `kafka.AsyncProducer` instead?

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return nil
+}
+
+func (self *kafkaStorage) Close() error {
+	return self.producer.Close()
+}
+
+func new() (storage.StorageDriver, error) {
```
timstclair: I'm surprised that go allows this, since `new` is a built-in. Since I'm confused as to what function `new` actually refers to in this file, I'd prefer to call this something like `newKafka` or `newDriver`

timstclair: or `newStorage` :) You can just inline newStorage in this method.

storage/kafka/kafka.go
```diff
@@ -0,0 +1,128 @@
...
+func newStorage(machineName,
+	brokers,
+	topic string,
+) (storage.StorageDriver, error) {
+	config := kafka.NewConfig()
+	config.Producer.RequiredAcks = kafka.WaitForAll
+	config.Producer.Retry.Max = 10
+
+	brokerList := strings.Split(brokers, ",")
+	fmt.Println("Kafka brokers:", strings.Join(brokerList, ", "))
```
timstclair: This should be a log statement (probably info level 4).

timstclair: Just log brokers, rather than joining on the same character you split on.

carmark:
could you please re-order the import packages as the follow order:
* golang self package
* cadvisor package
* third-party package

timstclair:
Did you run `godep save`? I think you're missing the sarama deps.

timstclair:
@k8s-bot ok to test

storage/kafka/kafka.go
arichardet: I was aiming to be consistent with the convention set by [this refactor](https://github.com/google/cadvisor/commit/b0ce5720b03b009c4bc48a57857c0737cb2ae07f).

timstclair: I see, that should probably be fixed. But, consistency trumps, so I'm ok with what you have for now. Let me know when you're ready for me to take another look.

arichardet:
@timstclair I've addressed your PR feedback.

arichardet:
Travis-CI is not liking the Kafka.go header going from 2015 -> 2016

timstclair:
Oh yeah, that needs to be fixed.

timstclair:
It looks like you somehow picked up a bunch of extraneous commits. Can you try rebasing?

arichardet:
@timstclair I tried rebasing and it this is what I got :-1: I can close this PR and resubmit as a fresh one that is cleanly in one commit if thats ok.

timstclair:
It looks like your commits for this PR are interspersed with merged commits, in which case you might need to manually repair it. I'd suggest trying `git rebase -i HEAD~35` and moving all your commits so they are grouped together at the bottom. If that doesn't fix it, try rebasing again using the instructions here: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/development.md#keeping-your-development-fork-in-sync

arichardet:
@timstclair I was able to clean it up.

timstclair:
Great, but unfortunately github won't show the diff now because of all the godeps changes. Could you please move the godeps changes to a separate commit, so that I can review the other files separately? Thanks!

arichardet:
@timstclair I've separated them.

arichardet:
@timstclair Is there anything else you need me to do?

arichardet:
@timstclair should close this and resubmit a clean PR? Thanks

timstclair:
Sorry, I was out of office last week. I'll take a final look at this later today.

timstclair:
Just a few more comments. Thanks for your patience.

arichardet:
@timstclair comments addressed. Thanks

timstclair:
LGTM, just 1 small comment

arichardet:
@timstclair comment addressed. Thanks

timstclair:
LGTM. Thanks!

timstclair:
@k8s-bot ok to test

PR #921, opened 2015-10-16
Start Manager v2 API

timstclair:
Reorganize some manager V2 APIs & add a V2 ContainerInfo (and method for retrieving it).

This PR looks a lot bigger than it actually is. Most of the changes are just moves. New code includes:

- V2 ContainerInfo API
- V2 GetContainerInfo
- manager_v2.go (and corresponding test) to separate V2 APIs from V1 APIs. The V2 APIs are exported by the  V2() method on `Manager`, but the `v2Manager` is intentionally not exported.
- test for ConvertStats (which caught a bug: `Labels` were not converted)

This change is a prerequisite for https://github.com/kubernetes/kubernetes/issues/12483

PR #1035, opened 2016-01-05
Expose container's write layer usage separately, with API refactoring

vishh:
1. Update filesystem stats APIs.
2. Add BaseUsage to containers which includes only the rootfs usage, if available.

cc @pwittrock @timstclair


vishh:
This PR has not been tested yet.

api/versions.go
```diff
@@ -63,6 +64,7 @@ func getApiVersions() []ApiVersion {
 	v1_2 := newVersion1_2(v1_1)
 	v1_3 := newVersion1_3(v1_2)
 	v2_0 := newVersion2_0()
+	v2_1 := newVersion2_1()
 
 	return []ApiVersion{v1_0, v1_1, v1_2, v1_3, v2_0}
```
pwittrock: Does this need to be updated as well?

container/docker/fsHandler.go
```diff
@@ -32,11 +32,13 @@ type fsHandler interface {
...
-	usageBytes  uint64
-	period      time.Duration
-	storageDirs []string
-	fsInfo      fs.FsInfo
+	lastUpdate     time.Time
+	usageBytes     uint64
+	baseUsageBytes uint64
+	period         time.Duration
+	writable       string
+	other          string
```
pwittrock: what is other?

api/versions.go
```diff
@@ -549,6 +601,87 @@ func convertStats(cont *info.ContainerInfo) []v2.ContainerStats {
 	return stats
 }
 
+func convertStatsToV2(cont *info.ContainerInfo) []v2.ContainerStats {
```
timstclair: Can we separate request handling from conversion? (At least separate files). E.g. something like https://github.com/google/cadvisor/pull/921/files#diff-aac76d5d60dea772214f3f48e0b21705 (might make more sense in a separate PR).

Longer term, I think we should switch to a model closer to the k8s API, with an unversioned internal model & versioned client APIs.

info/v2/machine.go
```diff
@@ -86,3 +88,125 @@ func GetAttributes(mi *v1.MachineInfo, vi *v1.VersionInfo) Attributes {
...
+// MachineStats contains usage statistics for the entire machine.
+type MachineStats struct {
+	// The time of this stat point.
+	Timestamp time.Time `json:"timestamp"`
+	// In nanoseconds (aggregated)
+	Cpu *v1.CpuStats `json:"cpu,omitempty"`
+	// In nanocores per second (instantaneous)
+	CpuInst *CpuInstStats `json:"cpu_inst,omitempty"`
+	// Disk IO statistics
+	DiskIo *v1.DiskIoStats `json:"diskio,omitempty"`
```
timstclair: Aren't most of the DiskIo stats duplicated in DiskStats? Can this field be removed?

info/v2/machine.go
```diff
@@ -86,3 +88,125 @@ func GetAttributes(mi *v1.MachineInfo, vi *v1.VersionInfo) Attributes {
...
+	ReadsMerged *uint64 `json:"reads_merged,omitempty"`
+
+	// Number of sectors read
+	// This is the total number of sectors read successfully.
+	SectorsRead *uint64 `json:"sectors_read,omitempty"`
+
+	// Number of milliseconds spent reading
+	// This is the total number of milliseconds spent by all reads (as
+	// measured from __make_request() to end_that_request_last()).
+	ReadTime *uint64 `json:"read_time,omitempty"`
```
timstclair: Consider calling this `ReadMillis`, to make the unit explicit. Same below.

info/v2/container.go
```diff
@@ -114,6 +114,28 @@ type ContainerStats struct {
...
+type ContainerStats struct {
+	// The time of this stat point.
+	Timestamp time.Time `json:"timestamp"`
+	// CPU statistics
+	// In nanoseconds (aggregated)
+	Cpu *v1.CpuStats `json:"cpu,omitempty"`
+	// In nanocores per second (instantaneous)
+	CpuInst *CpuInstStats `json:"cpu_inst,omitempty"`
+	// Disk IO statistics
+	DiskIo *v1.DiskIoStats `json:"diskio,omitempty"`
```
timstclair: Do we still need per-device statistics per-container?

api/versions.go
```diff
@@ -451,6 +453,56 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+	case machineStatsApi:
+		glog.V(4).Infof("Api - MachineStats(%v)", request)
+		cont, err := m.GetRequestedContainersInfo("/", opt)
+		if err != nil {
+			return err
+		}
+		return writeResult(convertMachineStats(cont), w)
+	case statsApi:
+		name := getContainerName(request)
+		glog.V(4).Infof("Api - Stats: Looking for stats for container %q, options %+v", name, opt)
```
timstclair: (nit) make the 2 log messages more consistent

info/v2/machine.go
pwittrock: Why do we use time.Time in MachineStats, and uint64 here?

api/versions.go
```diff
@@ -451,6 +453,56 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+		conts, err := m.GetRequestedContainersInfo(name, opt)
+		if err != nil {
+			return err
+		}
+		contStats := make(map[string][]v2.ContainerStats, 0)
```
timstclair: (nit) s/`0`/`len(conts)`/

info/v2/machine.go
```diff
@@ -86,3 +88,125 @@ func GetAttributes(mi *v1.MachineInfo, vi *v1.VersionInfo) Attributes {
...
+	// DiskStats for this device.
+	DiskStats `json:"inline"`
+}
+
+// DiskStats contains per partition usage information.
+// This information is only available at the machine level.
+type DiskStats struct {
+	// Number of reads completed
+	// This is the total number of reads completed successfully.
+	ReadsCompleted *uint64 `json:"reads_completed,omitempty"`
```
pwittrock: Why are we using pointers to ints instead of just ints?

info/v2/machine.go
timstclair: time.Time is for a timestamp, not a duration. This could be time.Duration though. I agree that overloading Time in the field name is a little confusing.

pwittrock: maybe: ReadDuration *time.Duration

api/versions.go
vishh: yes. good catch.

api/versions.go
vishh: Sure. I don't like this model either. It is very cumbersome. We should scope out the changes required for refactoring the existing API framework.

container/docker/fsHandler.go
vishh: :) Other includes logs, container metadata, etc. I add better documentation.

info/v2/container.go
vishh: It is useful for detecting containers that thrash the disk.

info/v2/machine.go
vishh: No. DiskStats is not scoped at the cgroup. It is at the node level.

info/v2/machine.go
vishh: With `uint64` a value of `0` will be treated as unset. With a pointer, we can use `nil` to express that the value is unset or unknown.

info/v2/machine.go
vishh: Sure. That will be possible here, but existing APIs will still expose them as uint64.

vishh:
So I assume the API reworking is OK. I will address the comments, test and then update the PR.

Thanks for the review.

info/v2/machine.go
timstclair: Write, but isn't this also at the node level?

info/v2/machine.go
pwittrock: leaving a uint64 makes sense to me then.  Prefer changing the name to ReadDuration or ReadDurationMillis

info/v2/machine.go
vishh: They are not really the same. But we can start with dropping DiskIO for now.

api/versions.go
vishh: Fixed

info/v2/machine.go
vishh: I changed it to duration.

vishh:
@timstclair @pwittrock: updated PR and added unit and integration tests. PTAL.

info/v2/machine.go
```diff
@@ -86,3 +88,97 @@ func GetAttributes(mi *v1.MachineInfo, vi *v1.VersionInfo) Attributes {
...
+	ReadsMerged *uint64 `json:"reads_merged,omitempty"`
+
+	// Number of sectors read
+	// This is the total number of sectors read successfully.
+	SectorsRead *uint64 `json:"sectors_read,omitempty"`
+
+	// Time spent reading
+	// This is the total number of milliseconds spent by all reads (as
+	// measured from __make_request() to end_that_request_last()).
+	ReadDuration time.Duration `json:"read_duration,omitempty"`
```
timstclair: time.Duration is just a alias of int64, so these should still be pointer types.

info/v2/container.go
```diff
@@ -85,7 +85,7 @@ type ContainerSpec struct {
 	Image string `json:"image,omitempty"`
 }
 
-type ContainerStats struct {
+type DeprecatedContainerStats struct {
```
timstclair: While I like this change, I wonder if it would be better to split it into a separate PR? Both because it seems a little unrelated to the disk stats, and we could make sure the disk changes are not blocked while dealing with merge conflicts etc. from the API change. If you think it's easier to keep it in this PR, I won't stop you...

vishh: I'm fine with splitting it up, but I won't have time to work on it until
Thursday. So we can either merge this in if it's a blocker out wait for two
more days..

timstclair: As mentioned offline, go with whatever you think will be easiest in terms of unblocking the critical parts. If you think you can get the critical part merged faster by splitting this up, go with that. Otherwise, I'm fine with leaving it as is.

vishh:
@timstclair: Splitting up the PR would be a bit time consuming. I just completed rebasing and it took quite some time. I'd much rather spend my energy improving the existing API machinery.

info/v2/conversions.go
```diff
@@ -0,0 +1,319 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
timstclair: (nit) can you rename this file to conversion.go so that git picks up the diff correctly?

timstclair:
LGTM

vishh:
Merged the files @timstclair

vishh:
@timstclair: Tests are passing. Merging.

PR #1036, opened 2016-01-05
Expose all memory stats contained in memory.stat via API and /metrics

### LINKED ISSUE #975: Export RSS size in /metrics
The metric exported as container_memory_usage_bytes includes page cache. It would be nice if cadvisor also exported the RSS size, which is often more interesting.

guoshimin:
fixes #975

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -306,6 +306,9 @@ type MemoryStats struct {
 	// Units: Bytes.
 	Usage uint64 `json:"usage"`
 
+	// Stats
+	Stats map[string]uint64 `json:"stats"`
+
 	// The amount of working set memory, this includes recently accessed memory,
 	// dirty memory, and kernel memory. Working set is <= "usage".
 	// Units: Bytes.
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -306,7 +306,9 @@ type MemoryStats struct {
 	// Units: Bytes.
 	Usage uint64 `json:"usage"`
 
-	// Stats
+	// Various statistics reported in memory.stat in cgroupfs. See the section
+	// "5.2 stat file" of https://www.kernel.org/doc/Documentation/cgroups/memory.txt
+	// for details.
 	Stats map[string]uint64 `json:"stats"`
 
 	// The amount of working set memory, this includes recently accessed memory,
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -127,6 +127,21 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					return values
 				},
 			}, {
+				name:        "container_memory_stats",
+				help:        "Container memory stats",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"type"},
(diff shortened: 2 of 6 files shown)


metrics/prometheus.go
```diff
@@ -127,6 +127,21 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					return values
 				},
 			}, {
+				name:        "container_memory_stats",
+				help:        `Various memory statistics. See the section "5.2 stat file" of https://www.kernel.org/doc/Documentation/cgroups/memory.txt for details`,
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"stat_name"},
```
jimmidyson: nit: can we just make this `name`?

jimmidyson:
One nit, but otherwise LGTM.

jimmidyson:
ok to test

jimmidyson:
@pwittrock How do I trigger Jenkins builds now?

info/v1/container.go
```diff
@@ -306,6 +306,11 @@ type MemoryStats struct {
 	// Units: Bytes.
 	Usage uint64 `json:"usage"`
 
+	// Various statistics reported in memory.stat in cgroupfs. See the section
```
vishh: We already include some of the stats from `memory.stat`. For example `WorkingSet` is derived using data from `memory.stat`.
Can we instead expose specific fields? Explicit fields will help with understanding of the API as well.

info/v1/container.go
```diff
@@ -306,6 +306,11 @@ type MemoryStats struct {
...
+	// "5.2 stat file" of https://www.kernel.org/doc/Documentation/cgroups/memory.txt
+	// for details.
+	Stats map[string]uint64 `json:"stats"`
```
vishh: We will have to update `v2` API as well.

info/v1/container.go
guoshimin: All this is just plumbing to get to the information in memory.stat, which is officially documented at https://www.kernel.org/doc/Documentation/cgroups/memory.txt. I don't find it tasteful to manually enumerate the entries here and duplicate the official documentation.

As a compromise, I can limit the scope of the PR to only expose RSS, which is the original request in #975.

info/v1/container.go
guoshimin: The v2 API references v1: https://github.com/google/cadvisor/blob/e4262b91b17cc132a35a023865f9ef97a048aa21/info/v2/container.go#L102

info/v1/container.go
vishh: Yeah. Let's limit to RSS for now. We get quite a few questions/issues around the data exposed by cAdvisor. Honestly, all the data that cAdvisor exposes is documented as part of kernel APIs.

info/v1/container.go
vishh: My bad. :+1:

info/v1/container.go
dqminh: I would like to also see at least `cache` exposed.

guoshimin:
Opened another PR #1055. Closing this one.

PR #1041, opened 2016-01-07
Make Go 1.5 a requirement

jimmidyson:
As discussed in https://github.com/google/cadvisor/pull/992#issuecomment-165908173 this makes 1.5 a requirement of using the build tooling (`make`) & disables the 1.4 build for travis (`vet` now doesn't support 1.4 any more so this works around this issue too).

/cc @vishh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 sudo: false
 go:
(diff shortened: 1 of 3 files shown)


vishh:
LGTM

Issue #606, opened 2015-03-21
Separate API object from info objects


rjnagal:
Move public API objects from info/v1 and info/v2 to api.

Use reflective conversions to fill in API objects from internal structure.

vishh:
@rjnagal: Any progress on this? New contributors are confused because of this split.

iwankgb:
Closing very old issue.

PR #1045, opened 2016-01-09
Add V2 ContainerInfo API

timstclair:
Changes include:

- Add V2 ContainerInfo struct
- Move v1/v2 conversion functions to a reusable package (accounts for most lines changed)
- Add V2 ContainerInfo accessor method to Manager

This is a trimmed down version of https://github.com/google/cadvisor/pull/921. Notably, I removed the changes to add a V2 manager interface, because it is not required by https://github.com/kubernetes/kubernetes/issues/12483.

This PR is required for https://github.com/kubernetes/kubernetes/issues/12483


timstclair:
@vishh - This will conflict with your changes in https://github.com/google/cadvisor/pull/1035, how would you like to handle it?

vishh:
@timstclair: My PR is not yet complete. So you win :) Reviewing this now.

info/conversion/conversion.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/info/v2"
+)
+
+// Get V2 container spec from v1 container info.
```
vishh: Should these conversion methods reside in the "v2" package since they are specific for v2?

info/v2/container.go
```diff
@@ -52,6 +52,14 @@ type MemorySpec struct {
 	SwapLimit uint64 `json:"swap_limit,omitempty"`
 }
 
+type ContainerInfo struct {
+	// Describes the container.
+	Spec ContainerSpec `json:"spec,omitempty"`
```
vishh: Given that we are essentially breaking the existing API, which I think is ok, shall we get rid of some of the fields in `Spec`? I want to get rid of all the `Has*` fields in place of pointer values in Stats.

vishh:
Overall code LGTM. Just a couple of comments about the API and the structure.

info/conversion/conversion.go
timstclair: Good call, done.

info/v2/container.go
timstclair: What do you mean by "essentially breaking the existing API"? I'm trying to change the minimum in this PR since it's blocking the stats API implementation, and also since I think you've already done the necessary work for this in https://github.com/google/cadvisor/pull/1035? Is there a reason not to wait for https://github.com/google/cadvisor/pull/1035 to remove the pointer fields?

vishh: #1035 doesn't change ContainerStats. This PR is changing the wire format for the `/stats` API right?

timstclair: No, I wasn't planning on it. This just adds a new method (`GetContainerInfoV2`) which will be called from kubelet for serving the new API at `/stats/summary`. I do not intend to make any changes to the other `/stats/...` endpoints.

vishh: Ah got it.. Sorry I got confused. I really wish we separate the internal API from the wire format soon.

vishh:
LGTM

vishh:
Tests are failing. Not sure why.

timstclair:
The tests don't like 2016: https://github.com/google/cadvisor/pull/1047 needs to be submitted.

info/v2/container.go
timstclair: > I really wish we separate the internal API from the wire format soon.

Agreed. I think [this issue](https://github.com/google/cadvisor/issues/606) is asking for that? I'll try to prioritize it after the 1.2 freeze.

timstclair:
Squashed.

timstclair:
@k8s-bot test this

timstclair:
I forgot to update the imports after moving the conversion files to v2. Fixed.

PR #1047, opened 2016-01-12
Fix check_boilerplate.sh for 2016

timstclair:
Copied the kubernetes boilerplate checker from
https://github.com/kubernetes/kubernetes/tree/master/hack/boilerplate,
and updated templates for cAdvisor.


jimmidyson:
LGTM

Just wait for Travis to pass & self-merge if someone doesn't get there first.

PR #1051, opened 2016-01-14
Fix FS usage goroutine leaks

jimmidyson:
Reported in kubernetes/kubernetes#19633

/cc @vishh @timothysc

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -746,6 +746,18 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 
 // Create a container.
 func (m *manager) createContainer(containerName string) error {
+	m.containersLock.Lock()
+	defer m.containersLock.Unlock()
+
+	namespacedName := namespacedContainerName{
+		Name: containerName,
+	}
+
+	// Check that the container didn't already exist.
+	if _, ok := m.containers[namespacedName]; ok {
+		return nil
+	}
+
 	handler, accept, err := container.NewContainerHandler(containerName, m.inHostNamespace)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


timothysc:
Awesome, thx @jimmidyson for the quick fix.

vishh:
@jimmidyson: Can you describe the issue you are trying to fix with this PR?

manager/manager.go
```diff
@@ -746,6 +746,9 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 
 // Create a container.
 func (m *manager) createContainer(containerName string) error {
+	m.containersLock.Lock()
```
yujuhong: Now that you moved out the creation of goroutines from the initialization of container handlers, is it necessary to guard the entire function with the lock?

timothysc:
@vishh the linked issue ( kubernetes/kubernetes#19633) contains the complete details.  

cadvisor can leak goroutines.

manager/manager.go
jimmidyson: The guard is really between `createContainer` & `destroyContainer`.

jimmidyson:
@vishh Goroutines were leaked as detailed in kubernetes/kubernetes#19633. @timothysc found this was due to tracking filesystem usage. The prior version was starting tracking filesystem usage even if the container had already been seen - this was not being cleaned up. This PR ensures that the track usage goroutine is only started once.

manager/manager.go
```diff
@@ -775,35 +778,24 @@ func (m *manager) createContainer(containerName string) error {
...
-		defer m.containersLock.Unlock()
-
-		namespacedName := namespacedContainerName{
-			Name: containerName,
-		}
-
-		// Check that the container didn't already exist.
-		_, ok := m.containers[namespacedName]
-		if ok {
-			return true
```
vishh: As an alternative, if we were to invoke `handler.Cleanup()` here, would that also fix the issue?

manager/manager.go
```diff
@@ -746,6 +746,9 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 
 // Create a container.
 func (m *manager) createContainer(containerName string) error {
+	m.containersLock.Lock()
+	defer m.containersLock.Unlock()
+
 	handler, accept, err := container.NewContainerHandler(containerName, m.inHostNamespace)
```
vishh: If we were to check if the container exists before invoking this Factory method, wouldn't that be better?

manager/manager.go
jimmidyson: `handler.Cleanup()` should only be called when a container is removed. Not sure why you would do that here?

manager/manager.go
jimmidyson: Yeah it would.

jimmidyson: Fixed.

manager/manager.go
vishh: The problem IIUC is that of not invoking Cleanup() before letting go GC the container object. 
I intended to say that if `alreadyExists` is `true`, then we can invoke Cleanup() right away.
But I prefer not creating the object if it already exists and invoking Cleanup() iif we return error instead of starting housekeeping. 
That would let us not add the new `Start` method.

vishh:
@jimmidyson: That was quick :) LGTM. I don't like the additional `Start` and `Cleanup` to be frank. I will try to find alternative methods to avoid them. For now, let's merge this PR.

jimmidyson:
@vishh I agree this was just yet another quick fix... Really need to review cadvisor's structure so we can tidy things up like this.

PR #1054, opened 2016-01-14
Bump dev version to 0.21.0

jimmidyson:


jimmidyson:
Self-merging - only a version change.

PR #1055, opened 2016-01-15
Expose rss and cache via API and /metrics

### LINKED ISSUE #975: Export RSS size in /metrics

guoshimin:
Fixes #975. This PR supersedes #1036.


vishh:
LGTM

vishh:
ok to test

vishh:
@guoshimin: Kindly fix the test issues.

Issue #1053, opened 2016-01-14
Introduce jitter on housekeeping intervals.


timothysc:
In profiling the kubelet and mapping the data, we see cpu spikes on housekeeping intervals and cadvisor appears to be the culprit of the spikes.  

Is there any reason why we couldn't introduce jitter on per container data collection? 

```
  0.18s 0.033% 0.033%    428.55s 78.33%  github.com/google/cadvisor/manager.(*containerData).housekeeping
         0     0% 0.033%    426.76s 78.01%  github.com/google/cadvisor/manager.(*containerData).housekeepingTick
     0.10s 0.018% 0.051%    426.76s 78.01%  github.com/google/cadvisor/manager.(*containerData).updateStats
         0     0% 0.051%    425.49s 77.77%  github.com/google/cadvisor/container/libcontainer.GetStats
     0.08s 0.015% 0.066%    415.93s 76.03%  github.com/google/cadvisor/container/raw.(*rawContainerHandler).GetStats
     0.01s 0.0018% 0.068%    413.70s 75.62%  github.com/google/cadvisor/container/libcontainer.networkStatsFromProc
     0.23s 0.042%  0.11%    413.63s 75.61%  github.com/google/cadvisor/container/libcontainer.scanInterfaceStats
```

/cc @vishh @jimmidyson @ncdc @mattf @jeremyeder

jimmidyson:
Sounds like a good idea to me.

mwielgus:
Ok, if the jitter is not too big (lets say less 1 - 2 sec)

jimmidyson:
Closing - thanks @timothysc.

PR #1057, opened 2016-01-15
Add jitter to housekeeping interval

timothysc:
per https://github.com/google/cadvisor/issues/1053 adding jitter to housekeeping intervals similar to what is done internally in kubernetes.  

@jimmidyson @vishh

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
+	"math/rand"
 	"os/exec"
 	"path"
 	"regexp"
@@ -76,6 +77,17 @@ type containerData struct {
 
 	// Runs custom metric collectors.
 	collectorManager collector.CollectorManager
+}
+
+// jitter returns a time.Duration between duration and duration + maxFactor * duration,
+// to allow clients to avoid converging on periodic behavior.  If maxFactor is 0.0, a
+// suggested default value will be chosen.
+func jitter(duration time.Duration, maxFactor float64) time.Duration {
+	if maxFactor <= 0.0 {
+		maxFactor = 1.0
+	}
+	wait := duration + time.Duration(rand.Float64()*maxFactor*float64(duration))
+	return wait
 }
 
 func (c *containerData) Start() error {
@@ -356,7 +368,7 @@ func (self *containerData) nextHousekeeping(lastHousekeeping time.Time) time.Tim
 		}
 	}
 
-	return lastHousekeeping.Add(self.housekeepingInterval)
+	return lastHousekeeping.Add(jitter(self.housekeepingInterval, 1.0))
 }
 
 // TODO(vmarmol): Implement stats collecting as a custom collector.


manager/container.go
```diff
@@ -306,8 +322,8 @@ func newContainerData(containerName string, memoryCache *memory.InMemoryCache, h
 	cont := &containerData{
 		handler:                  handler,
 		memoryCache:              memoryCache,
-		housekeepingInterval:     *HousekeepingInterval,
-		maxHousekeepingInterval:  maxHousekeepingInterval,
+		housekeepingInterval:     jitter(*HousekeepingInterval, JitterFactor),
+		maxHousekeepingInterval:  jitter(maxHousekeepingInterval, JitterFactor),
```
vishh: We should cap the housekeeping at `maxHousekeepingInternal`. I'd recommend not changing this.

manager/container.go
```diff
@@ -306,8 +322,8 @@ func newContainerData(containerName string, memoryCache *memory.InMemoryCache, h
 	cont := &containerData{
 		handler:                  handler,
 		memoryCache:              memoryCache,
-		housekeepingInterval:     *HousekeepingInterval,
-		maxHousekeepingInterval:  maxHousekeepingInterval,
+		housekeepingInterval:     jitter(*HousekeepingInterval, JitterFactor),
```
vishh: So the range of values will be [*housekeepingInterval, 3*HousekeepingInterval)?
That will mean for a 10 second housekeeping, it will be [10, 30), which is not acceptable.
Isn't the goal to spread housekeeping evenly. So wouldn't a factor of `1` be enough?

manager/container.go
```diff
@@ -356,7 +372,7 @@ func (self *containerData) nextHousekeeping(lastHousekeeping time.Time) time.Tim
 		}
 	}
 
-	return lastHousekeeping.Add(self.housekeepingInterval)
+	return lastHousekeeping.Add(jitter(self.housekeepingInterval, JitterFactor))
```
vishh: Isn't it enough to introduce jitter here alone and ceil it at maxHousekeepingInterval?

manager/container.go
jimmidyson: Agreed. I assume the point of this is to spread the collection across the housekeeping interval? I don't think this should affect the actual housekeeping interval but instead just add a random sleep (e.g. with `timer.After`) in to the housekeeping loop?

The aim must be to collect from all containers in the effective period between housekeeping runs?

manager/container.go
timothysc: sgtm

manager/container.go
timothysc: 1 is sufficient.

manager/container.go
timothysc: I think your initial starting point should also have jitter in the case of initial burst of containers.

vishh: Makes sense. Wrap `cont.Start()` inside of go thread then?

On Fri, Jan 15, 2016 at 1:30 PM, Timothy St. Clair <notifications@github.com

manager/container.go
jimmidyson: Doesn't that have the potential of skipping a housekeeping period as @vishh detailed above?

manager/container.go
jimmidyson: Why? If we're jittering in housekeeping then this seems to be enough to spread load across the housekeeping period.

manager/container.go
timothysc: k, will converge on 1 till we have more data.

manager/container.go
timothysc: I'm ok with just having it here on the 1st iteration, we can revise if needed.

timothysc:
PTAL comments addressed. 
For simplicity I'm keeping to the easiest mod and if we need to iterate, then I will follow up after we run another performance evaluation.

manager/container.go
```diff
@@ -40,6 +41,10 @@ import (
 	"github.com/golang/glog"
 )
 
+const (
+	JitterFactor = 1.0
```
jimmidyson: This is a bit redundant isn't it? Remove for now, add back in later if different from constant 1?

manager/container.go
```diff
@@ -78,6 +83,17 @@ type containerData struct {
 	collectorManager collector.CollectorManager
 }
 
+// jitter returns a time.Duration between duration and duration + maxFactor * duration,
+// to allow clients to avoid converging on periodic behavior.  If maxFactor is 0.0, a
+// suggested default value will be chosen.
+func jitter(duration time.Duration, maxFactor float64) time.Duration {
```
jimmidyson: Drop maxFactor? Always set to 1 in this PR.

timothysc: I think we should keep the function parameter at minimum, and if we want to remove the constant and simply default the argument that's fine.  But at this point we're splitting hairs.

vishh: Making it variable might invalidate assumptions made over cAdvisor APIs.
For example, heapster's internal logic might break if the housekeeping
jitter is arbitrarily large.

On Fri, Jan 15, 2016 at 2:54 PM, Timothy St. Clair <notifications@github.com

manager/container.go
timothysc: Done.

manager/container.go
timothysc: I've left the function signature, but defaulted on the call.

This should address the comments... I would like to leave the signature so I can rebuild and exercise during testing.  If at some point if it's deemed unnecessary we can easily collapse as it's a trivial change.

timothysc:
PTAL.

vishh:
LGTM.

vishh:
ok to test

Issue #1060, opened 2016-01-16
NPE in conversion.go


timstclair:
When querying manager.GetContainerInfoV2:
```
[restful] recover from panic situation: - runtime error: invalid memory address or nil pointer dereference
    /usr/src/go/src/runtime/asm_amd64.s:401
    /usr/src/go/src/runtime/panic.go:387
    /usr/src/go/src/runtime/panic.go:42
    /usr/src/go/src/runtime/sigpanic_unix.go:26
    /go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/info/v2/conversion.go:113
    /go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/manager/manager.go:427
    /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/pkg/kubelet/cadvisor/cadvisor_linux.go:141
```

Caused by uninitialized pointer fields introduced in https://github.com/google/cadvisor/commit/6dfdd8eca151469f92b2f0acf98afe3f46270511

/cc @vishh

jimmidyson:
Shouldn't this be picked up in tests? Do we have no coverage here?

timstclair:
Nope :( I'm adding a test.

jimmidyson:
Thanks!

PR #1061, opened 2016-01-17
Fix v2 stats conversion NPD & add test

timstclair:
Fix for https://github.com/google/cadvisor/issues/1060

### DIFF
--- a/info/v2/conversion.go
+++ b/info/v2/conversion.go
@@ -110,7 +110,10 @@ func ContainerStatsFromV1(spec *v1.ContainerSpec, stats []*v1.ContainerStats) []
 			stat.Memory = &val.Memory
 		}
 		if spec.HasNetwork {
-			stat.Network.Interfaces = val.Network.Interfaces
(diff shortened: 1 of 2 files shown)


timstclair:
@k8s-bot test this

jimmidyson:
LGTM thanks!

timstclair:
Thanks. Is it possible to cherrypick into a tagged release, or do we need to do another minor release?

jimmidyson:
We have a few different fixes now we need in Kubernetes that should be in next release, although I'm waiting on runc tag before submitting update to Kubernetes (sadly runc tag is waiting on a bug fix too...). I don't mind if we keep doing small cadvisor releases at all but we need to make sure we don't have overlapping cadvisor updates to Kubernetes.

timstclair:
Ok. I realized this is not actually blocking me at the moment, so I'll just wait for the next cAdvisor update to grab all the fixes.

vishh:
Thanks @timstclair. Missed this while rebasing.

PR #1071, opened 2016-01-21
Godeps mayhem

vishh:
cc @pwittrock


vishh:
cc @jimmidyson

vishh:
I need this merged to be able to update the deps in kubernetes repo.

vishh:
@k8s-bot: retest this please

yujuhong:
Looks good with one question: why do you need golang v1.5.3?

vishh:
[Security update ](https://groups.google.com/forum/#!topic/golang-announce/MEATuOi_ei4)

yujuhong:
LGTM

vishh:
@k8s-bot ok to test

Godeps/Godeps.json
```diff
@@ -38,29 +38,29 @@
...
 		{
-			"ImportPath": "github.com/docker/docker/pkg/longpath",
-			"Comment": "v1.9.1",
-			"Rev": "a34a1d598c6096ed8b5ce5219e77d68e5cd85462"
-		},
-		{
 			"ImportPath": "github.com/docker/docker/pkg/mount",
-			"Comment": "v1.9.1",
-			"Rev": "a34a1d598c6096ed8b5ce5219e77d68e5cd85462"
+			"Comment": "v1.9.1-rc1",
```
jimmidyson: Why downgrade to an RC of Docker?

jimmidyson:
What problem does this solve?

vishh:
@jimmidyson: It removes non-existent dependencies which is currently breaking `godep restore`.

I will fix revert the update to docker deps

jimmidyson:
This is a problem with godep not working when packages are removed from master branch but still exist in tagged versions obviously. In Kubernetes there is a convenience function [`preload-dep`](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-godeps.sh#L24) in `hack/verify-godeps.sh` that needs to be used in cases like this.

vishh:
Thanks for the pointer. I ended up writing a similar script. Vendoring in cadvisor into kube is failing because of this non-existent dependency. I'm attempting to move to glide from godeps to avoid similar issues in the future.

I have reverted the update to docker deps

vishh:
Self-merging to make progress on the kube side

jimmidyson:
But there is nothing wrong with these godeps afaik. Why didn't you just use kubes preload deps like I pointed out?

vishh:
I have updated cadvisor to not use a non-existent package. I don't see why
`godep` fails when one of the vendored dependencies doesn't exist. What is
the point of vendoring?
In any case, the pre-load option doesn't seem to work when updating godeps
in the kube side. That is the reason for fixing this.

PR #1073, opened 2016-01-21
Switch to using glide instead of godeps

vishh:
cc @jimmidyson @pwittrock 

Apologies for throwing a huge PR, but this has to be done in one shot. Kindly try it out manually and post your thoughts. 

From what I could glean, glide seems to be the best solution out there for deps management. If there are better alternatives, I'm all ears.


vishh:
Tip: Ignore all files under `vendor/` directory.

jimmidyson:
`go vet` is still trying to vet vendored packages & failing. Can you update scripts accordingly?

jimmidyson:
I'm happy to see this move but I do wonder how this affects godep vendoring for consumers like Kubernetes?

jimmidyson:
Pulled this PR branch locally & something isn't right:

```
$ glide list
[ERROR] Dependency /home/jdyson/go/src/github.com/google/cadvisor/vendor/github.com/stretchr/testify/mock failed to resolve: lstat /home/jdyson/go/src/github.com/google/cadvisor/vendor/github.com/stretchr/testify/mock: no such file or directory.
Oops! lstat /home/jdyson/go/src/github.com/google/cadvisor/vendor/github.com/stretchr/testify/mock: no such file or directory
```

jimmidyson:
Package list in Makefile might be easier to construct with `pkgs = $(shell $(GO) list ./... | grep -v /vendor/)`

timstclair:
Ping. What's the status of this?

vishh:
There are a couple of issues with glide that should ideally be resolved
before we can switch to it.

dashpole:
We probably want to move to using go modules instead.  Closing since this is outdated now.

PR #1077, opened 2016-01-26
Fix up networking stats for new docker network functionality

### LINKED ISSUE #1076: Problem with container in a network
Hi,
I have some containers running on an isolated network (not on bridge). The network is created using docker 1.9 new networking feature.
On cAdvisor UI I'm not able to see network section. I also tried to connect cAdvisor container to the container network, but nothing is changed.

Can you help me?

Thanks
Damiano

jimmidyson:
Fixes #1076 

@damianopezzotti This should fix your issue - works for me. If you can test that would be great but I can't see any problems with this approach.

/cc @vishh @pwittrock

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -226,16 +226,8 @@ func libcontainerConfigToContainerSpec(config *libcontainerconfigs.Config, mi *i
 	return spec
 }
 
-var (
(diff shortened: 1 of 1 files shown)


vishh:
LGTM

PR #1046, opened 2016-01-12
bump(github.com/opencontainers/runc/libcontainer)

jimmidyson:
Fixes issues with breaking changes to `GetPids` which is affecting
downstream consumers of cadvisor (e.g. Kubernetes).

Early PR to test functionality. Please don't merge until new libcontainer tag is
applied.

PR #1079, opened 2016-01-27
bump(github.com/opencontainers/runc/libcontainer)

jimmidyson:
runc was retagged at 0.0.7 - this uses the correct tag.

Also contains some fixes between the 2 0.0.7 tags - this was messy!


jimmidyson:
Self-merging as this is just a godep bump that we already agreed in #1046

vishh:
LGTM

PR #1083, opened 2016-01-30
Use storage dir from docker info for filesystem usage tracking.

vishh:
cc @pwittrock

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -15,6 +15,7 @@
 package docker
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"path"
@@ -40,6 +41,7 @@ var ArgDockerEndpoint = flag.String("docker", "unix:///var/run/docker.sock", "do
 var DockerNamespace = "docker"
 
 // Basepath to all container specific information that libcontainer stores.
+// TODO: Deprecate this flag
 var dockerRootDir = flag.String("docker_root", "/var/lib/docker", "Absolute path to the Docker state root directory (default: /var/lib/docker)")
 var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path to the Docker run directory (default: /var/run/docker)")
 
@@ -60,6 +62,10 @@ func DockerStateDir() string {
 // Whether the system is using Systemd.
 var useSystemd = false
 var check = sync.Once{}
+
+const (
+	dockerRootDirKey = "Root Dir"
+)
 
 func UseSystemd() bool {
 	check.Do(func() {
@@ -100,7 +106,8 @@ const (
 type dockerFactory struct {
 	machineInfoFactory info.MachineInfoFactory
 
-	storageDriver storageDriver
+	storageDriver    storageDriver
+	storageDriverDir string
 
 	client *docker.Client
 
@@ -129,6 +136,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		self.machineInfoFactory,
 		self.fsInfo,
 		self.storageDriver,
+		self.storageDriverDir,
 		&self.cgroupSubsystems,
 		inHostNamespace,
 		metadataEnvs,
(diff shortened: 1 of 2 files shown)


vishh:
@k8s-bot: ok to test

vishh:
Surprised that the tests passed without any changes to the PR. The e2e suite is now pretty flaky.

pwittrock:
Yes the e2e seems to get really flaky for a bit, and just when it is painful enough to drop what I am doing and fix it, it seems to clear up.

container/docker/factory.go
```diff
@@ -100,7 +102,8 @@ const (
 type dockerFactory struct {
 	machineInfoFactory info.MachineInfoFactory
 
-	storageDriver storageDriver
+	storageDriver    storageDriver
```
pwittrock: nit: is this gofmt'ed?  Usually it lines up the types

container/docker/factory.go
```diff
@@ -204,6 +208,35 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
...
+	for _, data := range storageDriverInfo {
+		if data == "" {
+			continue
+		}
+		var parts [][]string
+		if err := json.Unmarshal([]byte(data), &parts); err != nil {
+			return "", fmt.Errorf("failed to parse docker storage driver options - %+v", data)
+		}
+		for _, part := range parts {
+			if len(part) != 2 {
```
pwittrock: Since every element in parts is expected to be a string slice of 2 elements, can we unmarshal directly into a map?

container/docker/factory.go
```diff
@@ -204,6 +208,35 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
...
+				return "", fmt.Errorf("failed to parse docker storage driver options - %+v", part)
+			}
+			entries[part[0]] = part[1]
+		}
+	}
+	storageDir, exists := entries["Root Dir"]
```
pwittrock: Can we rely on these being static?  This is a good candidate for a const.

container/docker/factory.go
```diff
@@ -204,6 +208,35 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
...
+	if len(storageDriverInfo) == 0 {
+		return "", fmt.Errorf("failed to find docker storage driver options")
+	}
+
+	entries := map[string]string{}
+	for _, data := range storageDriverInfo {
+		if data == "" {
+			continue
+		}
+		var parts [][]string
```
pwittrock: Since we are just looking for one element ("Root Dir") we don't need to populate the map, and could just return when we find the entry we are looking for.  It is a pretty minor optimization, and only worth doing if it would improve the clarity of the code

container/docker/factory.go
```diff
@@ -232,18 +265,26 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
...
+	information, err := client.Info()
 	if err != nil {
 		return fmt.Errorf("failed to detect Docker info: %v", err)
 	}
-	execDriver, ok := information["ExecutionDriver"]
-	if !ok || !strings.HasPrefix(execDriver, "native") {
+
+	// Check that the libcontainer execdriver is used.
+	execDriver := information.Get("ExecutionDriver")
+	if execDriver == "" || !strings.HasPrefix(execDriver, "native") {
```
pwittrock: nit: is it necessary to check if execDriver == "", because that doesn't have the prefix "native"?

container/docker/factory.go
```diff
@@ -40,6 +41,7 @@ var ArgDockerEndpoint = flag.String("docker", "unix:///var/run/docker.sock", "do
 var DockerNamespace = "docker"
 
 // Basepath to all container specific information that libcontainer stores.
+// TODO: Deprecate this flag
```
pwittrock: Should we change the help documentation to state that this is deprecated and setting it has no effect?  Maybe "deprecated (no-op): docker root directory is now read from docker info"

container/docker/factory.go
vishh: Tests are supposed to fail if gofmt fails

vishh: Jenkins runs `./build/presubmit.sh`

container/docker/factory.go
vishh: But the string format is that of a slice right? How can we unmarshal a slice into a map?
Isn't the serialized format for map `"key": "value"`?

container/docker/factory.go
vishh: Yup done.

container/docker/factory.go
vishh: Makes sense. Done.

container/docker/factory.go
vishh: Yeah.

container/docker/factory.go
vishh: This is not yet deprecated. I was thinking of getting rid of this flag altogether. Attempting to be backwards compatible with that will only make the code messier.

container/docker/factory.go
pwittrock: You are probably right, I am not that familiar with the string format that docker provides or what the go deserialization libraries are capable of.  I was hoping the deserialization libraries would coerce a slice of slices len=2 into a map.

container/docker/factory.go
pwittrock: yeah.  Is it still used elsewhere then?

vishh: Yeah it is :(

pwittrock:
LGTM

vishh:
This PR should have failed on CoreOS nodes. Not sure how/why it passed. Reverting this PR.

vishh:
On retrospect, defaulting to flag defined value might be better. Making that change [here](https://github.com/vishh/cadvisor/commit/4eeafb2109d8682f30ce4cb3394abcf9af00bb90)

vishh:
FYI: Opened https://github.com/docker/docker/pull/19986 to fix this for real.

jimmidyson:
@vishh This needs reverting too. Checking the test logs it only passed because they timed out... This breaks stuff for devicemapper storage as it doesn't necessarily have a root storage dir.

vishh:
I noticed this and tried to fix in the next PR of mine. I defaulted to
using the flat value of root device wasn't found.

jimmidyson:
It was still broken - think this is fixed now in #1094 I think.

PR #1084, opened 2016-01-30
Track fs usage for docker versions >= 1.10

vishh:
Updated the v2.1 `/stats` endpoint to return `map[ContainerName]v2.ContainerInfo` instead of just `map[ContainerName]v2.ContainerStats`. 

Still Pending: Documenting the v2 API.
cc @pwittrock

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"path"
+	"strconv"
 	"strings"
 
 	v1 "github.com/google/cadvisor/info/v1"
@@ -85,6 +87,23 @@ func (self *Client) Attributes() (attr *v2.Attributes, err error) {
 	return
 }
 
+// Stats returns stats for the requested container.
+func (self *Client) Stats(name string, request *v2.RequestOptions) (map[string]v2.ContainerInfo, error) {
+	u := self.statsUrl(name)
+	ret := make(map[string]v2.ContainerInfo)
+	data := url.Values{
+		"type":      []string{request.IdType},
+		"count":     []string{strconv.Itoa(request.Count)},
+		"recursive": []string{strconv.FormatBool(request.Recursive)},
+	}
+
+	u = fmt.Sprintf("%s?%s", u, data.Encode())
+	if err := self.httpGetJsonData(&ret, nil, u, "stats"); err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
 func (self *Client) machineInfoUrl() string {
 	return self.baseUrl + path.Join("machine")
 }
@@ -101,7 +120,11 @@ func (self *Client) attributesUrl() string {
 	return self.baseUrl + path.Join("attributes")
 }
 
-func (self *Client) httpGetResponse(postData interface{}, url, infoName string) ([]byte, error) {
+func (self *Client) statsUrl(name string) string {
+	return path.Join(self.baseUrl, "stats", name)
+}
+
+func (self *Client) httpGetResponse(postData interface{}, urlPath, infoName string) ([]byte, error) {
 	var resp *http.Response
 	var err error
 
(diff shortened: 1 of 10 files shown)


vishh:
This PR might be failing on old versions of docker. I'm planning to update
our test infra to output docker version of the tests fail.

vishh:
Tested this PR locally with docker v1.10-rc2 and it works. e2e passed. AFter this PR goes in, we can add a node with docker v1.10 to the e2e suite.

container/docker/handler.go
```diff
@@ -82,6 +83,24 @@ type dockerContainerHandler struct {
...
 
+func getRwLayerID(containerID, storageDriverDir string, dockerVersion []int) (string, error) {
+	const (
+		// Docker version >=1.10.0 have a randomized ID for the root fs of a container.
+		randomizedRWLayerMinorVersion = 10
+		// Directory where the file containinig the randomized ID of the root fs of a container is stored in versions >= 1.10.0
+		rwLayerIDDir  = "../image/aufs/layerdb/mounts/"
+		rwLayerIDFile = "mount-id"
+	)
+	if dockerVersion[1] < randomizedRWLayerMinorVersion {
```
pwittrock: as discussed offline, check the major version as well

integration/framework/framework.go
```diff
@@ -255,6 +257,15 @@ func (self dockerActions) Run(args DockerRunArgs, cmd ...string) string {
 	return containerId
 }
 
+func (self dockerActions) Version() []string {
+	dockerCommand := []string{"docker", "version", "-f", "'{{.Server.Version}}'"}
+	output, _ := self.fm.Shell().Run("sudo", dockerCommand...)
+	if len(output) < 1 {
```
pwittrock: Should we check exactly 1 instead at least 1?

vishh:
@pwittrock: I have made several changes to PR including that of adding a much needed integration test. PTAL.
I have also added a check to include docker major version as well.

integration/tests/api/docker_test.go
```diff
@@ -270,3 +290,38 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+
+func TestDockerFilesystemStats(t *testing.T) {
+	fm := framework.New(t)
+	defer fm.Cleanup()
+
+	storageDriver := fm.Docker().StorageDriver()
+	switch storageDriver {
+	case framework.Aufs:
+	case framework.Overlay:
+		break
```
pwittrock: what does break do when it is not in a loop?

container/docker/handler.go
```diff
@@ -82,6 +83,24 @@ type dockerContainerHandler struct {
...
 
+func getRwLayerID(containerID, storageDriverDir string, dockerVersion []int) (string, error) {
+	const (
+		// Docker version >=1.10.0 have a randomized ID for the root fs of a container.
+		randomizedRWLayerMinorVersion = 10
+		// Directory where the file containinig the randomized ID of the root fs of a container is stored in versions >= 1.10.0
+		rwLayerIDDir  = "../image/aufs/layerdb/mounts/"
+		rwLayerIDFile = "mount-id"
+	)
+	if !(dockerVersion[0] >= 1 && dockerVersion[1] >= randomizedRWLayerMinorVersion) {
```
pwittrock: Should this be || instead of &&?
for docker 2.0 -> [2, 0] -> !(true && false) -> !(false) -> true

client/v2/client.go
```diff
@@ -101,7 +120,11 @@ func (self *Client) attributesUrl() string {
 	return self.baseUrl + path.Join("attributes")
 }
 
-func (self *Client) httpGetResponse(postData interface{}, url, infoName string) ([]byte, error) {
+func (self *Client) statsUrl(name string) string {
+	return self.baseUrl + path.Join("stats", name)
```
pwittrock: why not path.Join(self.baseUrl, "stats", name)

integration/tests/api/docker_test.go
```diff
@@ -270,3 +290,38 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+	time.Sleep(time.Minute)
+	require.NoError(t, err)
+	require.True(t, len(containerInfo) == 1)
+	var info v2.ContainerInfo
+	for _, cInfo := range containerInfo {
+		info = cInfo
+	}
+	sanityCheckV2(containerId, info, t)
+	require.NotNil(t, info.Stats[0].Filesystem.BaseUsageBytes)
+	assert.True(t, *info.Stats[0].Filesystem.BaseUsageBytes > (1<<6), "expected base fs usage to be greater than 0")
```
pwittrock: this string seems wrong, since we are checking if it is greater than 2^6

container/docker/handler.go
vishh: Good catch. Late night coding is clearly bad. Fixed it. Added a unit test as well.

client/v2/client.go
vishh: Yeah. Thats possible. Updated. I followed the existing templates, which are weird in retrospect.

integration/tests/api/docker_test.go
vishh: Fixed

vishh:
@pwittrock: Addressed your comments. PTAL

integration/framework/framework.go
vishh: Sure. Updated.

integration/tests/api/docker_test.go
vishh: Sweet. Updated.

pwittrock:
LGTM pending passing tests

jimmidyson:
Tests seemed to pass even though there were errors in the logs... `Test killed with quit: ran too long (10m0s).` So no confidence that this actually passed integration tests.

vishh:
Ohh no. That's bad. Time out should have been an error :(

PR #1087, opened 2016-02-01
Make e2e tests dump attributes and logs on failure.

vishh:


pwittrock:
@k8s-bot please test this

pwittrock:
@k8-bot test this please

pwittrock:
@k8s-bot test this please

integration/runner/runner.go
```diff
@@ -79,7 +104,7 @@ func PushAndRunTests(host, testDir string) error {
 	portStr := strconv.Itoa(*port)
 	errChan := make(chan error)
 	go func() {
-		args = common.GetGCComputeArgs("ssh", host, "--", fmt.Sprintf("sudo %s --port %s --logtostderr", path.Join(testDir, cadvisorBinary), portStr))
+		args = common.GetGCComputeArgs("ssh", host, "--", fmt.Sprintf("sudo %s --port %s --logtostderr &> %s/log.txt", path.Join(testDir, cadvisorBinary), portStr, testDir))
```
pwittrock: Is redirecting the output better than setting the log directory?

pwittrock: I believe this will only redirect stdout, and not stderr.  Is that your intent?  Could we use --alsologtostderr and then set the log file to <>/log.txt so that the error output still appears in Jenkins?

integration/runner/runner.go
```diff
@@ -141,9 +171,20 @@ func PushAndRunTests(host, testDir string) error {
 		}
 	}
 	if err != nil {
-		err = fmt.Errorf("error on host %s: %v", host, err)
+		err = fmt.Errorf("error on host %s: %v\n%+v", host, err, attributes)
+		// Copy logs from the host
+		args = common.GetGCComputeArgs("copy-files", fmt.Sprintf("%s:%s/log.txt", host, testDir), "./")
+		err = RunCommand("gcloud", args...)
+		if err != nil {
+			errChan <- fmt.Errorf("error fetching logs: %v", err)
```
pwittrock: Is there another routine listening to this channel?  If not this will block forever.

integration/runner/runner.go
vishh: setting the log directory will result in multiple log files. I tried it locally and for some reason the INFO logs did not have all the log data. For the sake of expediency, I went with this approach. Happy to reconsider once this PR works.

vishh:
@pwittrock: Can we merge this soon? I need this PR to figure out the test failures on my other PRs.

integration/runner/runner.go
vishh: Err. My bad. Fixing it.

vishh: Updated code. PTAL

integration/runner/runner.go
pwittrock: SGTM

pwittrock:
Ideally I'd prefer to grab the output directly from cadvisor, but it is a little hard to do since cadvisor isn't expected to stop.  I think this would be possible after killing it, but would require some updates to structure to capture the "error" caused by killing cadvisor.  If you agree, go ahead and add a TODO.

pwittrock:
LGTM

vishh:
@pwittrock: I like the changes you are proposing. It requires a bit more refactoring. Let's merge this for now and clean things up in a subsequent PR.

PR #1089, opened 2016-02-04
Releasing v0.21.0 tag.

vishh:


vishh:
Merging since this is just a version bump

PR #1095, opened 2016-02-05
Remove systemd docker container name check

jimmidyson:
@vishh @pwittrock Please review when you get a chance.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -21,17 +21,14 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/container/libcontainer"
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
-	"github.com/google/cadvisor/utils"
 
 	"github.com/fsouza/go-dockerclient"
 	"github.com/golang/glog"
-	"github.com/opencontainers/runc/libcontainer/cgroups"
 )
 
 var ArgDockerEndpoint = flag.String("docker", "unix:///var/run/docker.sock", "docker endpoint")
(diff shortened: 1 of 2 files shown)


container/docker/factory.go
```diff
@@ -273,6 +269,9 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 				return fmt.Errorf("cAdvisor requires docker version %v or above but we have found version %v reported as \"%v\"", expected_version, dockerVersion, version_string)
 			}
 		}
+		if dockerVersion[0] == 1 && dockerVersion[1] >= 10 {
```
vishh: This is not always true since we can enable systemd support optionally. How do we handle that?

vishh: AFAIK fedora and rhel will have systemd support enabled by default.

jimmidyson: I run fedora & this works for me. Docker has changed cgroup hierarchy in 1.10 it seems.

vishh: `sudo docker daemon --exec-opt="native.cgroupdriver=systemd" -s
devicemapper` will make the daemon use systemd. I wonder if we can use
docker info to figure this out.

jimmidyson: I didn't know that - thanks! Will take a look in the next few days.

jimmidyson:
@vishh This PR now simplifies stuff around systemd cgroups by ignoring the fact that systemd might be being used. I think the previous logic wasn't necessary at all - we can just limit the container name to containing the full Docker container ID format (`[a-z0-9]{64}`) & then validate the container ID via Docker inspect as we did before. This gets rid the of the need for the `nosystemd` flag too which is a good thing.

container/docker/factory.go
```diff
@@ -46,9 +43,7 @@ var dockerRunDir = flag.String("docker_run", "/var/run/docker", "Absolute path t
 
 // Regexp that identifies docker cgroups, containers started with
 // --cgroup-parent have another prefix than 'docker'
-var dockerCgroupRegexp = regexp.MustCompile(`.+-([a-z0-9]{64})\.scope$`)
-
-var noSystemd = flag.Bool("nosystemd", false, "Explicitly disable systemd support for Docker containers")
+var dockerCgroupRegexp = regexp.MustCompile(`([a-z0-9]{64})`)
```
vishh: Great thought :clap: :+1:

vishh:
LGTM.

PR #1094, opened 2016-02-05
Ensure filesystem usage isn't collected for devicemapper storage

jimmidyson:
@vishh Please review when you get a chance.

PR #1097, opened 2016-02-05
Fix fs stats handling for non-aufs storage drivers.

vishh:
@jimmidyson: Here's an alternate approach to fix #1094.
PTAL. It works with docker v1.10 device mapper backend on my workstation.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -83,18 +83,19 @@ type dockerContainerHandler struct {
 	fsHandler fsHandler
 }
 
-func getRwLayerID(containerID, storageDriverDir string, dockerVersion []int) (string, error) {
+func getRwLayerID(containerID, storageDir string, sd storageDriver, dockerVersion []int) (string, error) {
 	const (
 		// Docker version >=1.10.0 have a randomized ID for the root fs of a container.
 		randomizedRWLayerMinorVersion = 10
 		// Directory where the file containinig the randomized ID of the root fs of a container is stored in versions >= 1.10.0
-		rwLayerIDDir  = "../image/aufs/layerdb/mounts/"
-		rwLayerIDFile = "mount-id"
+		rwLayerIDDirTemplate = "image/%s/layerdb/mounts/"
+		rwLayerIDFile        = "mount-id"
 	)
 	if (dockerVersion[0] <= 1) && (dockerVersion[1] < randomizedRWLayerMinorVersion) {
 		return containerID, nil
 	}
-	bytes, err := ioutil.ReadFile(path.Join(storageDriverDir, rwLayerIDDir, containerID, rwLayerIDFile))
+
+	bytes, err := ioutil.ReadFile(path.Join(storageDir, "image", string(sd), "layerdb", "mounts", containerID, rwLayerIDFile))
 	if err != nil {
 		return "", fmt.Errorf("failed to identify the read-write layer ID for container %q. - %v", containerID, err)
 	}
@@ -107,7 +108,7 @@ func newDockerContainerHandler(
 	machineInfoFactory info.MachineInfoFactory,
 	fsInfo fs.FsInfo,
 	storageDriver storageDriver,
-	storageDriverDir string,
+	storageDir string,
 	cgroupSubsystems *containerlibcontainer.CgroupSubsystems,
 	inHostNamespace bool,
 	metadataEnvs []string,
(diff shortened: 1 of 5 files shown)


container/docker/handler.go
```diff
@@ -135,18 +135,18 @@ func newDockerContainerHandler(
 	id := ContainerNameToDockerId(name)
 
 	// Add the Containers dir where the log files are stored.
-	otherStorageDir := path.Join(path.Dir(storageDriverDir), pathToContainersDir, id)
+	otherStorageDir := path.Join(storageDir, pathToContainersDir, id)
 
-	rwLayerID, err := getRwLayerID(id, storageDriverDir, dockerVersion)
+	rwLayerID, err := getRwLayerID(id, storageDir, storageDriver, dockerVersion)
```
jimmidyson: Can we not run this fit devicemapper storage?

container/docker/handler.go
```diff
@@ -135,18 +135,18 @@ func newDockerContainerHandler(
...
 	if err != nil {
 		return nil, err
 	}
 	var rootfsStorageDir string
 	switch storageDriver {
```
jimmidyson: What about zfs storage that I think we support now?

vishh: This logic is common to all storage drivers AFAIK.

container/docker/handler.go
vishh: Yeah. That is possible. Any specific reason other than it being a no-op as of now?

container/docker/handler.go
vishh: @jimmidyson: Do you have access to a zfs machine? I'm trying to add a zfs machine to the test fleet.

vishh:
@k8s-bot: test again

vishh:
The disk usage test that I added recently is causing timeouts. I have temporarily disabled it. I will re-enable it in a subsequent PR. For now, this PR seems to work on aufs and devicemapper with docker v1.10. I'm planning to add a node for overlay with docker v1.10 and one with zfs. I think this is good to go for now, unless you have any objections @jimmidyson.

container/docker/handler.go
```diff
@@ -83,18 +83,18 @@ type dockerContainerHandler struct {
...
 
-func getRwLayerID(containerID, storageDriverDir string, dockerVersion []int) (string, error) {
+func getRwLayerID(containerID, storageDir string, sd storageDriver, dockerVersion []int) (string, error) {
 	const (
 		// Docker version >=1.10.0 have a randomized ID for the root fs of a container.
 		randomizedRWLayerMinorVersion = 10
 		// Directory where the file containinig the randomized ID of the root fs of a container is stored in versions >= 1.10.0
-		rwLayerIDDir  = "../image/aufs/layerdb/mounts/"
-		rwLayerIDFile = "mount-id"
+		rwLayerIDDirTemplate = "image/%s/layerdb/mounts/"
```
pwittrock: nit / optional:
there seems to be a widely copied / pasted way of doing this in github.  May want to use the standard approach:

https://github.com/search?q=%2Flayerdb%2Fmounts%2F&type=Code&utf8=%E2%9C%93

container/docker/handler.go
```diff
@@ -135,18 +135,18 @@ func newDockerContainerHandler(
 	id := ContainerNameToDockerId(name)
 
 	// Add the Containers dir where the log files are stored.
-	otherStorageDir := path.Join(path.Dir(storageDriverDir), pathToContainersDir, id)
+	otherStorageDir := path.Join(storageDir, pathToContainersDir, id)
```
pwittrock: not in this PR, but add a TODO to give more descriptive names than storageDir and otherStorageDir

container/docker/handler.go
vishh: Makes sense. I went with hard coding the path for now and avoided the constant.

container/docker/handler.go
vishh: Done.

pwittrock:
@jimmidyson I am going to lgtm this change so that it can be merged into k8s before the 1.2 feature freeze.  I'd prefer to wait for your approval since you have already commented, but doing so would cause this to miss the freeze deadline.  If you still have concerns we can discuss and update/revert if needed (would require a k8s 1.2 cherrypick).

pwittrock:
LGTM

jimmidyson:
Thanks for picking this up @pwittrock.

pwittrock:
@jimmidyson my pleasure

PR #1100, opened 2016-02-06
Adding a script that will run e2e tests on our jenkins builder.

vishh:
The script should make updates to the test infra explicit.

### DIFF
--- a/build/integration.sh
+++ b/build/integration.sh
@@ -14,6 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if [[ -n "${JENKINS_HOME}" ]]
+then 
+    exec ./jenkins_e2e.sh
+fi
 echo ">> starting cAdvisor locally"
 sudo ./cadvisor &
 echo ">> running integration tests against local cAdvisor"
--- a/build/jenkins_e2e.sh
+++ b/build/jenkins_e2e.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+godep go build github.com/google/cadvisor/integration/runner
+
+# Host Notes
+# e2e-cadvisor-ubuntu-trusty-docker110
+# - ubuntu 14.04
+# - docker 1.10
+# e2e-cadvisor-container-vm-v20160127-docker18
+# - docker 1.8.3
+# e2e-cadvisor-container-vm-v20151215-docker18
+# - docker 1.8.3
+# e2e-cadvisor-ubuntu-trusty-docker19
+# - ubunty 14.04
+# - docker 1.9.1
+# e2e-cadvisor-coreos-beta-docker19
+# - docker 1.9.1
+# e2e-cadvisor-rhel-7-docker19
+# - red hat 7
+# - docker 1.9.1
+# e2e-cadvisor-centos-v7
+# - docker 1.9.1
+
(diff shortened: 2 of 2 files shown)


build/jenkins_e2e.sh
```diff
@@ -0,0 +1,55 @@
...
+# e2e-cadvisor-coreos-beta
+# - docker 1.9.1
+# e2e-cadvisor-rhel-7
+# - red hat 7
+# - docker 1.9.1
+
+./runner --logtostderr --test-retry-count=8 --test-retry-whitelist=integration/runner/retrywhitelist.txt \
+e2e-cadvisor-ubuntu-trusty \
+e2e-cadvisor-coreos-beta \
+e2e-cadvisor-container-vm-v20151215 \
```
timstclair: No hostnotes for this one?

build/jenkins_e2e.sh
```diff
@@ -0,0 +1,55 @@
...
+e2e-cadvisor-container-vm-v20160127 \
+e2e-cadvisor-rhel-7
+
+# TODO: enable when docker 1.10  is working
+# e2e-cadvisor-ubuntu-trusty-docker110
```
timstclair: is this the same as `e2e-cadvisor-ubuntu-trusty-docker1-10`? Which is correct?

build/jenkins_e2e.sh
```diff
@@ -0,0 +1,55 @@
...
+# e2e-cadvisor-rhel-7
+# - red hat 7
+# - docker 1.9.1
+
+./runner --logtostderr --test-retry-count=8 --test-retry-whitelist=integration/runner/retrywhitelist.txt \
+e2e-cadvisor-ubuntu-trusty \
+e2e-cadvisor-coreos-beta \
+e2e-cadvisor-container-vm-v20151215 \
+e2e-cadvisor-container-vm-v20160127 \
+e2e-cadvisor-rhel-7
```
timstclair: I think it would be more readable if you make this list a variable

build/integration.sh
```diff
@@ -14,6 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if [[ ! -z "${JENKINS_HOME-}" ]]
```
timstclair: I think `! -z` is equivalent to `-n`?

build/integration.sh
```diff
@@ -14,6 +14,10 @@
...
+then 
+    exec ./jenkins_e2e.sh
```
timstclair: I think you need to call exit after this

build/integration.sh
jimmidyson: Is that extra `-` required?

build/integration.sh
jimmidyson: Don't think so - exec replaces current shell process & doesn't return to caller iirc.

Not sure why you need to exec it though? Just call it normally? Or are you relying on the exit code of the exec script?

jimmidyson:
Personally if prefer to move to a matrix build for testing on multiple distros. That way is easier to identify problems & build output is more readable. What do you think?

vishh:
@jimmidyson: Matrix build is what we need. Any ideas on how to get there?

build/integration.sh
vishh: The exit code is the reason for exec. I can catch the exit code and return that too, if exec has issues.

jimmidyson: I have no problem with exec here.

jimmidyson:
For matrix build you would need to set up each node as a Jenkins slave & then set up the matrix to run on each slave.

vishh:
We will have to periodically upgrade docker version and the base distributions. @pwittrock and I were brainstorming today to crowdsource testing if possible, similar to Go's [build dashboard](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=go%20build%20dashboard). WDYT?

I will address the comments in this PR now. This PR will help with the process until we have a better test infra. Disagree?

build/jenkins_e2e.sh
vishh: Fixed.

build/jenkins_e2e.sh
vishh: `110` is the valid name.

build/jenkins_e2e.sh
vishh: Done

build/integration.sh
vishh: Valid comments. Updated. PTAL.

vishh:
@timstclair @timstclair: PTAL.

vishh:
I want to merge this PR before adding a docker v1.10 node.

jimmidyson:
LGTM

Issue #997, opened 2015-12-03
Documentation of Go client is a bit broken


MattMattV:
Links to ContainerInfo struct are broken
Description is now outdated since ContainerInfo is storing not only an int but an int and 2 time.Time

Maybe more examples or explanations or updating would be nice to understand how to use it.

MattMattV:
bump

vishh:
@timstclair: Can you handle this issue?

timstclair:
Sure. I can work on this after k8s 1.2 is cut.

timstclair:
Fixed the main issues. I agree that there is a large issue with the documentation and examples as a whole, but I'm going to close this for now.

PR #1101, opened 2016-02-08
Fix out of date client/README.md

timstclair:
Fixes the worst of https://github.com/google/cadvisor/issues/997, but the overall documentation needs more work.

The documentation could use a complete rewrite with some additional examples, but I think it would be better to do that through better method comments, and directing users to the [godoc](https://godoc.org/github.com/google/cadvisor/client) instead.


vishh:
LGTM

PR #1104, opened 2016-02-09
hack rkt fs support into cadvisor

sjpotter:
I'm unsure why all these other commits were brought in, I don't have a merge commit in my logs.   (Educaiton would be appreciateD)

This is just a proof of concept so can just see my one commit at the end.  I don't like it as have to pass the RktPath in explicitly but unsure how else to do it.  so feedback on that would be welcomed as well

@vishh @yifan-gu @jonboulle

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -38,6 +38,7 @@ import (
 const (
 	LabelSystemRoot   = "root"
 	LabelDockerImages = "docker-images"
+	LabelRktImages    = "rkt-images"
 )
 
 type partition struct {
@@ -62,6 +63,7 @@ type Context struct {
 	// docker root directory.
 	DockerRoot string
 	DockerInfo map[string]string
+	RktPath    string
 }
 
 func NewFsInfo(context Context) (FsInfo, error) {
@@ -103,6 +105,10 @@ func NewFsInfo(context Context) (FsInfo, error) {
 	// need to call this before the log line below printing out the partitions, as this function may
 	// add a "partition" for devicemapper to fsInfo.partitions
 	fsInfo.addDockerImagesLabel(context)
+
+	if context.RktPath != "" {
+		fsInfo.addRktImagesLabel(context)
+	}
 
 	glog.Infof("Filesystem partitions: %+v", fsInfo.partitions)
 	fsInfo.addSystemRootLabel()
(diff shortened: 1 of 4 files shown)


vishh:
Can you rebase against upstream master? `git fetch upstream && git rebase
-i upstream/master`

sjpotter:
ok, that was a disaster.  unsure what happened here.

sjpotter:
ok, all better now.

sjpotter:
In order to do this nicer I'm thinking 

1) a basically nil rkt implementation in containers/ just enough for FS stuff

1a) this will involve some flags stuff, so will be a bit of dance with kubelet as it embeds it.

2) a basically nil rkt implementation in manager/ again just enough for FS stuff

2a) kubelet and cadvisor itself will specify which manager it wants based on flags passed in. 

as some of thse flags are going to be the same as what kubelet has now, there might be a bit of a dance to get them unified without breaking anything?

thoughts, though thats beyond what we talked about above.  just putting it out there as I find the above ugly :)

manager/manager.go
```diff
@@ -118,7 +118,7 @@ type Manager interface {
 }
 
 // New takes a memory storage and returns a new manager.
-func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingInterval time.Duration, allowDynamicHousekeeping bool) (Manager, error) {
+func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingInterval time.Duration, allowDynamicHousekeeping bool, rktPath string) (Manager, error) {
```
vishh: nit: Can we pass in a new `Config` object instead of expanding this interface? I'd like to see all the existing `bool` knobs be a part of that.

fs/fs.go
```diff
@@ -165,11 +171,19 @@ func (self *RealFsInfo) addDockerImagesLabel(context Context) {
 		dockerPaths := getDockerImagePaths(context)
 
 		for src, p := range self.partitions {
-			self.updateDockerImagesPath(src, p.mountpoint, dockerPaths)
+			self.updateContainerLabelImagesPath(LabelDockerImages, src, p.mountpoint, dockerPaths)
```
vishh: It is kind of weird to make `addDockerImagesLabel` update labels for `rkt` as well. This logic might require a more thorough refactoring.

sjpotter: See my other pr i do that there.

sjpotter:
This is superseded by https://github.com/google/cadvisor/pull/1120

PR #1105, opened 2016-02-10
a mostly null rkt manager implementation to support fs tracking

sjpotter:
prototype poc for kubernetes usage

so this this is a bit more involved than what I previously did, but probably also much cleaner.  In general for now users would use manager.New() to instantiate a normal docker manager, but for kubernetes we can use the NewRktManager() instead 

@yifan-gu @jonboulle @vishh


sjpotter:
same issue as before, trying to fix, will probably be a bad commit then a reset again

sjpotter:
I should note that this isn't ready there's a lot of interplay between container/ and manager/ which I haven't figured out yet, posted this now so that there can be discussion.

sjpotter:
@k8s-bot test this

container/rkt/factory.go
```diff
@@ -0,0 +1,67 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/google/cadvisor/container"
+	"github.com/google/cadvisor/fs"
+	info "github.com/google/cadvisor/info/v1"
+)
+
+// Basepath to all container specific information that libcontainer stores.
+var rktPath = flag.String("rktPath", "/var/lib/rkt", "Absolute path to the Rkt root directory (default: /var/lib/rkt)")
```
timstclair: (nit) flag adds the default value into the usage string, so you don't need it in the description (it's only needed when the default is filled in later).

timstclair: (nit) flags should be lowercase. Unfortunately there doesn't seem to be a convention for separating words with `_` `-` or no separator.

fs/fs.go
```diff
@@ -62,6 +63,7 @@ type Context struct {
 	// docker root directory.
 	DockerRoot string
 	DockerInfo map[string]string
+	RktPath    string
```
timstclair: Should this be `RktRoot` for consistency, or does this represent something different from `DockerRoot`?

fs/fs.go
```diff
@@ -102,7 +104,13 @@ func NewFsInfo(context Context) (FsInfo, error) {
...
 	// need to call this before the log line below printing out the partitions, as this function may
 	// add a "partition" for devicemapper to fsInfo.partitions
-	fsInfo.addDockerImagesLabel(context)
+	if context.DockerRoot != "" {
+		fsInfo.addDockerImagesLabel(context)
+	}
+
+	if context.RktPath != "" {
+		fsInfo.addRktImagesLabel(context)
+	}
```
timstclair: What if instead of hardcoding this, we created an `FsHandler` abstraction, and create a docker & rkt implemention in the respective Container directory, and have each implementation register itself? Then here you just loop over all the registered `FsHandler`s and call `handler.AddImagesLabel`?

manager/rkt_manager.go
```diff
@@ -0,0 +1,665 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
```
timstclair: (nit) 2016

manager/rkt_manager.go
```diff
@@ -0,0 +1,665 @@
...
+	cadvisorContainer        string
+	inHostNamespace          bool
+	loadReader               cpuload.CpuLoadReader
+	eventHandler             events.EventManager
+	startupTime              time.Time
+	maxHousekeepingInterval  time.Duration
+	allowDynamicHousekeeping bool
+}
+
+func NewRktManager(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingInterval time.Duration, allowDynamicHousekeeping bool, rktPath string) (Manager, error) {
```
timstclair: I don't think Manager is the right level of abstraction for splitting container runtimes. I know it currently integrates some docker concepts, but I'd rather see those concepts abstracted out. I think we should try to stick with ContainerHandler as the runtime abstraction, and move the problematic points of the manager to the ContainerHandlers.

container/rkt/factory.go
sjpotter: first nit just copied from manager.go (i.e. default listed there)

second nit, I'll do that.

fs/fs.go
sjpotter: Rkt seems to refer to it as RktPath so was trying to be consistent with that.

manager/rkt_manager.go
sjpotter: oh, I agree that there is a lot of duplicated code.  plan today was to try to refactor out the duplicate code and make a ManagerInternal interface that it would use.

wishing go had abstract classes though.

fs/fs.go
timstclair: Ok. I started my review at too fine a grain (sorry). We can ignore this fore now.

manager/rkt_manager.go
timstclair: Ok. I think it would be better to try and tackle this from the ContainerHandler interface before creating a second manager though. BTW, if you're still developing this PR, can you add `[WIP]` to the title?

sjpotter: and I'm thinking basically what you're saying is that the current manager.go is good enough but can stick in some switch statements to control which container handler is loaded?

sjpotter:
@timstclair is what I just pushed along the lines of what you were thinking?

sjpotter:
is the above jenkins error a failure in the testing infrastructure or in my code?  looks like in the testig infrastructure?

"F0210 20:23:56.256849   11661 runner.go:286] Error 0: error reading local log file: open ./log.txt: no such file or directory"

timstclair:
> @timstclair is what I just pushed along the lines of what you were thinking?

Closer, but rather than a bunch of switch statements I think it would be better to encapsulate those branch points in the ContainerHandler, and simply delegate to the container handler at that point. The manager might need a list of handlers (e.g. for Raw containers), but I'm not sure exactly how that would work. Maybe it would be good to schedule a VC meeting to discuss this in more detail?

sjpotter:
so instead of the the container handler being instantiated in the manager,
have it passed to the manager with some more methods to call the specific
functionality needed?  i.e. CreateFSContext() for that stuff?

timstclair:
Yeah, that's the gist of it. Or, the ContainerHandler could be instantiated in the manager, but I'm trying to reduce the number of places `switch containerRuntime { ... }` is needed. So, having one switch to instantiate the runtime (or delegate to a factory) would be acceptable.

container/rkt/factory.go
```diff
@@ -0,0 +1,67 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/google/cadvisor/container"
+	"github.com/google/cadvisor/fs"
+	info "github.com/google/cadvisor/info/v1"
+)
+
+// Basepath to all container specific information that libcontainer stores.
+var rktPath = flag.String("rkt_path", "/var/lib/rkt", "Absolute path to the Rkt root directory")
```
vishh: nit: The current code base is complex because flags are configured alongside regular code. Since you are adding new (and better) code, can we move the flags to a top level file and inject it into this package? 
So instead of providing a static method in this package, the path will only be returned from a `rkt` factory object.
I intend to perform a similar surgery on the docker factory.

Also, is it possible to auto-detect this path?

sjpotter: I'm hoping to get this info from the rkt api-service at some point in time,
but right now that service is limited.

vishh: If there is a plan to switch to the service, then how will the existing
method be modified to get the path from the service? We have the exact same
issue with docker factory :(

sjpotter:
the above commit (and a small fix that will come) pulls out the container configuration into its own factory method

sjpotter:
though apparently it seems to insist on me assigning copyright to google?

vishh:
Closing this in favor of #1120

PR #1108, opened 2016-02-11
Use vanilla ssh instead of gcloud ssh for e2e tests

pwittrock:
This is a proof of concept to get away from gcloud compute, which seems to cause a lot of flakes in the PR builder, and instead use ssh/scp directly with an ssh_config file.

Does this seem like the correct way to more forward.

TIP: You can generate a config file for gce machines using `gcloud compute config-ssh --dry-run`


pwittrock:
I will add a default value for ssh-config just so the pr build passes without updating jenkins

integration/runner/runner.go
```diff
@@ -68,6 +69,14 @@ func getAttributes(ipAddress, portStr string) (*cadvisorApi.Attributes, error) {
 	return &attributes, nil
 }
 
+func getIp(host string) (string, error) {
+	if *sshConfig == "" {
+		return host, nil
+	}
+	out, err := exec.Command("awk", fmt.Sprintf("/Host %s/ {getline; print $2}", host), *sshConfig).CombinedOutput()
```
timstclair: I'm not too familiar with awk, but I'm sure there's a native go way to do this?

integration/runner/runner.go
```diff
@@ -77,23 +86,35 @@ func RunCommand(cmd string, args ...string) error {
...
 	// Push binary.
 	glog.Infof("Pushing cAdvisor binary to %q...", host)
-	args := common.GetGCComputeArgs("ssh", host, "--", "mkdir", "-p", testDir)
-	err := RunCommand("gcloud", args...)
+	ipAddress, err := getIp(host)
+	if err != nil {
+		return fmt.Errorf("failed to get ip address of host: %s, %v", ipAddress, err)
+	}
+
+	err = RunSshCommand("ssh", host, "--", "mkdir", "-p", testDir)
```
timstclair: Alternatively, go has an [ssh client](https://godoc.org/golang.org/x/crypto/ssh). The advantage to using the go client is you wouldn't need to open and close multiple connections, so it seems like it would be less flaky? Although if you weren't using multiple connections, the connection would be open for longer, so I'm not sure which would be better.

pwittrock: golang ssh client looks like a better approach.  adding a todo.

integration/runner/runner.go
pwittrock: one would hope, but I wasn't able to find a better solution on SO or the man pages.  Happy to change if anyone with more ssh knowledge could suggest a better way.

pwittrock:
Adding a few debug lines for figuring out why e2e tests are failing

pwittrock:
Working on getting the tests to pass, having some issues with the Jenkins environment

integration/runner/runner.go
pwittrock: Fixed this by just requiring the hostname to resolve.  This works for gce and I think is an ok assumption to make.

pwittrock:
I changed this to use generic options to be more flexible.  I think instead of using an ssh_config, we will want to specify the options manually so we don't need to maintain a config file for all of our instances.

pwittrock:
Failures will be expected because I haven't defaulted the ssh options to the needed credentials

pwittrock:
Found an issue and repushed, will send a PTAL once the errors are the expected errors

integration/framework/framework.go
```diff
@@ -328,8 +320,11 @@ func (self shellActions) Run(command string, args ...string) (string, string) {
 		cmd = exec.Command(command, args...)
 	} else {
 		// We must SSH to the remote machine and run the command.
-		args = append(common.GetGCComputeArgs("ssh", self.fm.Hostname().GceInstanceName, "--", command), args...)
-		cmd = exec.Command("gcloud", args...)
+		args = append([]string{self.fm.Hostname().GceInstanceName, "--", command}, args...)
```
vishh: nit: Why is it called GceInstanceName?

pwittrock: Because I didn't clean up all the of the Gce specific code in the test runner :p .  Do you think that is important enough to do for this PR?

pwittrock: Done

pwittrock:
PTAL.  Issues should be resolved.

pwittrock:
@jimmidyson Hopefully this reduces that number of flake failures, as well as makes it simpler to test in non-gce environments.

integration/runner/runner.go
```diff
@@ -78,23 +78,29 @@ func RunCommand(cmd string, args ...string) error {
...
-		err := RunCommand("gcloud", args...)
+		err = RunSshCommand("ssh", host, "--", "rm", "-rf", testDir)
 		if err != nil {
 			glog.Errorf("Failed to cleanup test directory: %v", err)
 		}
 	}()
-	args = common.GetGCComputeArgs("copy-files", cadvisorBinary, fmt.Sprintf("%s:%s", host, testDir))
-	err = RunCommand("gcloud", args...)
+
+	err = RunSshCommand("scp", "-r", cadvisorBinary, fmt.Sprintf("%s:%s", host, testDir))
```
timstclair: I think scp takes different options from ssh - maybe this should just be RunCommand?

timstclair: If you do that, you could drop the `cmd` argument in `RunSshCommand`, and just always use the 'ssh' command.

pwittrock: scp is going to need to take the ssh options in order to properly authenticate

timstclair: I see, I guess we can just rely on the error checking if invalid options are passed in :)

timstclair:
LGTM

integration/runner/runner.go
pwittrock: Yeah, if the options aren't supported I expect the tests will fail.  :)

vishh:
Yayy... Thanks @pwittrock for making this change :bow:

PR #1109, opened 2016-02-11
cloudinfo: Adding AWS, Azure support and InstanceID

enoodle:


enoodle:
what is the Authorized Contributors Group for the CLA ?

enoodle:
@simon3z

vishh:
ok to test

vishh:
Overall lgtm

manager/machine.go
```diff
@@ -96,6 +96,7 @@ func getMachineInfo(sysFs sysfs.SysFs, fsInfo fs.FsInfo, inHostNamespace bool) (
 	realCloudInfo := cloudinfo.NewRealCloudInfo()
 	cloudProvider := realCloudInfo.GetCloudProvider()
 	instanceType := realCloudInfo.GetInstanceType()
+	instanceID := realCloudInfo.GetInstanceID()
```
simon3z: Can this take a long time? (e.g. is it polling ip addresses that may not be always available?)
Would that be a problem? Should this be async?

utils/cloudinfo/cloudinfo.go
```diff
@@ -23,19 +23,23 @@ import (
...
 type realCloudInfo struct {
 	cloudProvider info.CloudProvider
 	instanceType  info.InstanceType
+	instanceID    info.InstanceID
 }
 
 func NewRealCloudInfo() CloudInfo {
 	cloudProvider := detectCloudProvider()
 	instanceType := detectInstanceType(cloudProvider)
+	instanceID := detectInstanceID(cloudProvider)
```
simon3z: Can this take a long time? (e.g. is it polling ip addresses that may not be always available?)
Would that be a problem? Should this be async?

enoodle: I think detectCloudProvider could take some time. The others will be almost immediate as we are accessing the correct API server. In short this function could block for some time and its users should be aware of this as it was already blocking before this patch.

simon3z: @enoodle ok :+1:

utils/cloudinfo/azure.go
```diff
@@ -0,0 +1,48 @@
...
+	BiosUUIDFileName     = "/sys/class/dmi/id/product_uuid"
+	MicrosoftCorporation = "Microsoft Corporation"
+)
+
+func onAzure() bool {
+	data, err := ioutil.ReadFile(SysVendorFileName)
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(data), MicrosoftCorporation)
```
simon3z: @enoodle is this a stable check? Is it documented?

enoodle: No. I couldn't find any api service comparable to aws/gce 's for Azure, nor any other official documentation on ways to detect if an instance is running on Azure. I did find [1] which explains how to use dmidecode to get the instance ID (hence /sys/class/dmi/id/produect_uuid) and I noticed that on all my instances I had that string for sys_vendor so I used this.

[1] https://azure.microsoft.com/en-us/blog/accessing-and-using-azure-vm-unique-id/

enoodle:
@deads2k can you review this? Thanks

deads2k:
@googlebot @enoodle is covered under the Red Hat CLA.

deads2k:
@enoodle When you add the godep packages, can you name the commit `bump(github.com/aws/aws-sdk-go) 72440a9c5a884936f1c679591fcf04e31a9b772a` (that looks like what you were trying to get).  If the rest are just transitive that's fine, but that name will help track back to the target.

deads2k:
Is this still a WIP?  If so, what is remaining?

enoodle:
@deads2k Nothing more to implement. It was wip for the dependencies.

enoodle:
@deads2k I will rebase to change the dependencies commit name in a moment.

deads2k:
@vishh Implementation is done.  This enables fallback behavior for `ProviderID` determination in the kubelet (see https://github.com/kubernetes/kubernetes/pull/21373).  Take another look?

deads2k:
@jimmidyson ?

jimmidyson:
@deads2k Sorry on vacation.

@vishh Could you review when you get a chance please?

enoodle:
@vishh ping? you actually lgtm'ed it before, I only added a tiny fix to remove \n char when reading the azure instance ID from file.

vishh:
Sorry. This somehow fell off my radar. LGTM again

PR #1110, opened 2016-02-11
Stop leaking testing dependencies (and flags) in non-testing builds

timstclair:
Specifically, the "github.com/stretchr/testify/mock" pulls in "net/http/httptest" which adds the "httptest.serve" flag.


pwittrock:
fyi test failure:

/var/lib/jenkins/workspace/project/src/github.com/google/cadvisor/container/factory_test.go:49: undefined: FactoryForMockContainerHandler

timstclair:
I'm sorry, I could have sworn this worked the first time around... For the build tags to work, we would always need to specify `-tags test` when running unit tests, which is annoying. It looks like the go standard library just appends `_test` to test-only files (even if they don't contain any tests), so I'll go with that method unless there's any objections...

timstclair:
Actually, I was wrong again - `*_test.go` files aren't exported, which is problematic for `container/mock.go` which is required by the `manager` package. So, I think the only option without a more substantial refactoring (to move test utilities to their own package) is to use the test tag. WDYT?

timstclair:
PTAL

/cc @jimmidyson @vishh

vishh:
Can we move the mock/stubs to a separate package?

vishh:
This change is fine by me, although less ideal.

timstclair:
> Can we move the mock/stubs to a separate package?

It's not that simple unfortunately... just moving the file to a separate package creates a circular dependency. The circular dependency could be fixed by moving tests to a _test package, but then they lose access to the unexported types. The types could be exported in another _test.go file, but in the end I think the build tags are cleaner... WDYT?

vishh:
Got it. I'm fine with this change. It does not affect regular workflow unless we use `go` tools directly.

pwittrock:
yay more credential failures

pwittrock:
@k8s-bot test this please

pwittrock:
I am good with this as long as the tests pass

vishh:
Build is still failing. I suspect this is because the jenkins has not been updated to use `make` yet. I will make that change now.

timstclair:
> Build is still failing. I suspect this is because the jenkins has not been updated to use make yet. I will make that change now.

Where do you see that? I just see the auth failure...

pwittrock:
#1108 Should make these auth failures go away

pwittrock:
@k8s-bot please test this

pwittrock:
@k8s-bot test this please

pwittrock:
LGTM.  Ok to self merge

PR #1113, opened 2016-02-12
Proposal: Sandbox cAdvisor flags

timstclair:
NOTE: **This is a breaking change!** Anyone importing cAdvisor packages and using cAdvisor flags will now need to call `"utils/flag".AddAllFlags(...)`. How should we roll this out?

I am opening this approach up for discussion. This method accomplishes the goal (not leaking flags) with the minimum code changes (just an import), but it means that there are some implicit dependencies on import order when working with the "sandboxed" flags. The `utils/flag` library provides some
convenience functions for working with the cAdvisor flags, including
setting new (default) values and selectively adding flags.

Other alternatives to consider are:
- Moving flags to an `AddFlags(*flag.FlagSet)` method in each package. This approach makes the flag additions more explicit, but requires more code changes on both ends (cAdvisor & users of cAdvisor libraries).
- Moving flags to a config-like object. This approach moves away from the global flag variable pattern (though flags could still be provided), to a (possibly) global config object. This could make programmatic configuration a bit cleaner, but would require a lot of code changes, and could potentially obfuscated which config settings go with which package (or plugin).

Note that we could decide to move to an alternative solution at a later date, but the current implementation is the fastest way of controlling flag usage.

This also adds a new presubmit check to prevent new flags from being added to the CommandLine flags (this is how I caught the leaking `httptest.serve` flag - fixed in https://github.com/google/cadvisor/pull/1110). Note that the flag_check must be run as a separate command (not a test) to avoid pulling in test and test package flags.

xRef: https://github.com/kubernetes/kubernetes/issues/19432

/cc @vishh @pwittrock @jdef @jimmidyson


timstclair:
Failure is actually a good sign that the new presubmit check is working :) This PR depends on https://github.com/google/cadvisor/pull/1110

jdef:
Can you be more explicit about the implicit dependencies on import order?

timstclair:
@jdef were you asking for more of an explanation here, or to document it somewhere in cAdvisor?

To elaborate here, the flags are added by each package in the init() function, which is called the first time the package is imported. This means that if a caller is going to use `flag.Add(All)Flags` in an Init() function, the package that calls it must import all the cAdvisor packages that add desired flags. This can all be avoided by just calling `Add(All)Flags` outside the Init function, so it should not be much of an issue.

jdef:
Thanks for the explanation. Should be documented for sanity.

timstclair:
> Should be documented for sanity.

Done.

timstclair:
Closing in favor of solution proposed in https://github.com/google/cadvisor/issues/1175#issuecomment-199563817

Issue #501, opened 2015-02-09
Investigate cpu usage for scraping diskio stats


rjnagal:
There are many files and a lot of parsing. Internally, we do diskio stats scraping every 10s.
That might be the way to go.

cboggs:
:+1: 
I noticed a pretty hefty load increase when I had Cadvisor running yesterday on a docker host with a couple TB of local storage, and a couple dozen du's running pretty much all the time. Stopped the Cadvisor container and the du's went away, and load returned to expected values.

Happy to gather more detailed stats if needed.

vmarmol:
What version of cAdvisor were you running? Any way that you could get how much CPU cAdvisor was using? :) I don't doubt we're doing some silly things around this...

jamshid:
Every time I try cadvisor, after a while I have to disable it because of high cpu usage (eg, see below). Likely the same `du` issue, saw du's in `top` briefly. I did a fresh docker pull today of google/cadvisor. 
```
$ docker ps -q | xargs docker stats
CONTAINER           CPU %               MEM USAGE/LIMIT     MEM %               NET I/O
68e20776129f        0.00%               1.082 MiB/500 MiB   0.22%               452 KiB/46.16 KiB
7b31946e2c1e        638.20%             42 MiB/500 MiB      8.40%               21.02 MiB/516 MiB
8ab17fc74c77        3.47%               3.832 MiB/500 MiB   0.77%               8.042 KiB/648 B
...
$ docker ps -a | grep 7b31946e2c1e
7b31946e2c1e        google/cadvisor:latest                                                    "/usr/bin/cadvisor t   3 hours ago         Up 3 hours                     0.0.0.0:911->8080/tcp                                                                     buildenv_cadvisor_1

$ docker-machine ssh dev sudo top
Mem: 3650684K used, 402700K free, 0K shrd, 290K buff, 292K cached
CPU: 15.0% usr 78.1% sys  0.0% nic  6.3% idle  0.0% io  0.0% irq  0.3% sirq
Load average: 13.33 13.02 9.79 14/1423 6992
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
 2356   718 root     R     885m 22.3   7 70.4 /usr/bin/cadvisor true
[...]

PR #1117, opened 2016-02-16
Timeout `du` after a minute.

vishh:
cc @timstclair

### DIFF
--- a/container/docker/fsHandler.go
+++ b/container/docker/fsHandler.go
@@ -43,7 +43,10 @@ type realFsHandler struct {
 	stopChan chan struct{}
 }
 
-const longDu = time.Second
+const (
+	longDu    = time.Second
+	duTimeout = time.Minute
+)
 
 var _ fsHandler = &realFsHandler{}
 
@@ -66,12 +69,12 @@ func (fh *realFsHandler) needsUpdate() bool {
 
 func (fh *realFsHandler) update() error {
 	// TODO(vishh): Add support for external mounts.
-	baseUsage, err := fh.fsInfo.GetDirUsage(fh.rootfs)
+	baseUsage, err := fh.fsInfo.GetDirUsage(fh.rootfs, duTimeout)
 	if err != nil {
 		return err
 	}
 
-	extraDirUsage, err := fh.fsInfo.GetDirUsage(fh.extraDir)
+	extraDirUsage, err := fh.fsInfo.GetDirUsage(fh.extraDir, duTimeout)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


vishh:
@k8s-bot test this

container/docker/fsHandler.go
```diff
@@ -66,12 +69,12 @@ func (fh *realFsHandler) needsUpdate() bool {
...
 func (fh *realFsHandler) update() error {
 	// TODO(vishh): Add support for external mounts.
-	baseUsage, err := fh.fsInfo.GetDirUsage(fh.rootfs)
+	baseUsage, err := fh.fsInfo.GetDirUsage(fh.rootfs, duTimeout)
 	if err != nil {
 		return err
 	}
 
-	extraDirUsage, err := fh.fsInfo.GetDirUsage(fh.extraDir)
+	extraDirUsage, err := fh.fsInfo.GetDirUsage(fh.extraDir, duTimeout)
```
timstclair: I think we should log these errors at `Error` level (here: https://github.com/vishh/cadvisor/blob/timeout-du/container/docker/fsHandler.go#L99 - I can't figure out how to comment there on GH)

vishh: done

timstclair:
LGTM, but I'm concerned that if `du` times out, it may indicate a recurring problem (i.e. rerunning du may also timeout). If this seems to be the case, we should consider marking it as a permanent failure (maybe give it a few tries).

vishh:
How do we differentiate between transient and permanent failures? A disk
could become slow temporarily.

timstclair:
Maybe an exponential backoff on the retries would be better.

vishh:
@timstclair: I added exponential backoff. Limited it to a factor of `20` for now.

vishh:
@timstclair: I fixed #501 as part of this PR and re-enabled the disk e2e. PTAL.

vishh:
@timstclair: Addressed comments.

timstclair:
Commented on https://github.com/vishh/cadvisor/commit/a5decfeaad26fcf72ce7254e4516ef18b21eacb5

vishh:
@timstclair: This PR is ready to be reviewed again.

integration/tests/api/docker_test.go
```diff
@@ -292,36 +292,60 @@ func TestDockerContainerNetworkStats(t *testing.T) {
...
+	switch storageDriver {
+	case framework.Aufs, framework.Overlay:
+		needsBaseUsageCheck = true
+	}
+	pass := false
+	// We need to wait for the `dd` operation to complete.
+	for i := 0; i < 10; i++ {
+		containerInfo, err := fm.Cadvisor().ClientV2().Stats(containerId, request)
+		require.NoError(t, err)
+		require.True(t, len(containerInfo) == 1)
```
timstclair: (nit) prefer `require.Equal(t, 1, len(containerInfo))`

timstclair:
LGTM

a86c6f7964:
Any idea when this will be released, and/or if there is a good work around for the mean time?

vishh:
We will try to make a release by the end of this week.

integration/tests/api/docker_test.go
vishh: Fixing

vishh:
@timstclair: Addressed the two pending comments. This PR is ready to be merged now.

timstclair:
LGTM, ok to self merge once e2e passes.

PR #1121, opened 2016-02-23
Support opt out for metrics.

vishh:
This patch should let users disable disk and network metrics aggregation if required.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -86,6 +86,8 @@ type dockerFactory struct {
 	fsInfo fs.FsInfo
 
 	dockerVersion []int
+
+	ignoreMetrics container.MetricSet
 }
 
 func (self *dockerFactory) String() string {
@@ -111,6 +113,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		inHostNamespace,
 		metadataEnvs,
 		self.dockerVersion,
+		self.ignoreMetrics,
 	)
 	return
 }
@@ -178,7 +181,7 @@ func parseDockerVersion(full_version_string string) ([]int, error) {
 }
 
 // Register root container before running this function!
-func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
+func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo, ignoreMetrics container.MetricSet) error {
 	client, err := Client()
 	if err != nil {
 		return fmt.Errorf("unable to communicate with docker daemon: %v", err)
@@ -236,6 +239,7 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 		machineInfoFactory: factory,
 		storageDriver:      storageDriver(sd),
 		storageDir:         storageDir,
+		ignoreMetrics:      ignoreMetrics,
 	}
 
 	container.RegisterContainerHandlerFactory(f)
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -81,6 +81,8 @@ type dockerContainerHandler struct {
 
 	// Filesystem handler.
 	fsHandler fsHandler
+
+	ignoreMetrics container.MetricSet
 }
 
 func getRwLayerID(containerID, storageDir string, sd storageDriver, dockerVersion []int) (string, error) {
@@ -111,6 +113,7 @@ func newDockerContainerHandler(
 	inHostNamespace bool,
 	metadataEnvs []string,
 	dockerVersion []int,
+	ignoreMetrics container.MetricSet,
 ) (container.ContainerHandler, error) {
 	// Create the cgroup paths.
 	cgroupPaths := make(map[string]string, len(cgroupSubsystems.MountPoints))
@@ -161,8 +164,12 @@ func newDockerContainerHandler(
 		fsInfo:             fsInfo,
 		rootFs:             rootFs,
 		rootfsStorageDir:   rootfsStorageDir,
-		fsHandler:          newFsHandler(time.Minute, rootfsStorageDir, otherStorageDir, fsInfo),
 		envs:               make(map[string]string),
+		ignoreMetrics:      ignoreMetrics,
+	}
+
+	if !ignoreMetrics.Has(container.DiskUsageMetrics) {
+		handler.fsHandler = newFsHandler(time.Minute, rootfsStorageDir, otherStorageDir, fsInfo)
 	}
 
 	// We assume that if Inspect fails then the container is not known to docker.
@@ -255,8 +262,11 @@ func libcontainerConfigToContainerSpec(config *libcontainerconfigs.Config, mi *i
 	return spec
 }
 
-func hasNet(networkMode string) bool {
-	return !strings.HasPrefix(networkMode, "container:")
+func (self *dockerContainerHandler) needNet() bool {
+	if !self.ignoreMetrics.Has(container.NetworkUsageMetrics) {
+		return !strings.HasPrefix(self.networkMode, "container:")
+	}
+	return false
 }
 
 func (self *dockerContainerHandler) GetSpec() (info.ContainerSpec, error) {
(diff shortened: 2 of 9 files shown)


manager/manager.go
```diff
@@ -50,6 +51,35 @@ var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h",
...
+		return errors.New("disable_metrics already set")
+	}
+	for _, ip := range strings.Split(value, ",") {
+		if ip == container.DiskUsageMetrics.String() {
+			(*ml)[container.DiskUsageMetrics] = struct{}{}
+		} else if ip == container.NetworkUsageMetrics.String() {
+			(*ml)[container.NetworkUsageMetrics] = struct{}{}
+		} else if ip == container.NetworkTcpUsageMetrics.String() {
+			(*ml)[container.NetworkTcpUsageMetrics] = struct{}{}
+		}
```
dqminh: can we return an error here if the user specified an invalid value ?

container/docker/handler.go
```diff
@@ -160,8 +163,12 @@ func newDockerContainerHandler(
 		fsInfo:             fsInfo,
 		rootFs:             rootFs,
 		rootfsStorageDir:   rootfsStorageDir,
-		fsHandler:          newFsHandler(time.Minute, rootfsStorageDir, otherStorageDir, fsInfo),
 		envs:               make(map[string]string),
+		ignoreMetrics:      ignoreMetrics,
+	}
+
+	if _, ignore := ignoreMetrics[container.DiskUsageMetrics]; !ignore {
```
timstclair: (nit) If you make ignoreMetrics a `map[MetricKind]bool`, then these checks become cleaner. E.g.
```go
if !ignoreMetrics[container.DiskUsageMetrics] {
```

container/factory.go
```diff
@@ -35,6 +35,24 @@ type ContainerHandlerFactory interface {
...
+	DiskUsageMetrics       MetricKind = "disk"
+	NetworkUsageMetrics    MetricKind = "network"
+	NetworkTcpUsageMetrics MetricKind = "tcp"
+	AppMetrics             MetricKind = "app"
+)
+
+func (mk MetricKind) String() string {
+	return string(mk)
+}
+
```
timstclair: Consider declaring a MetricSet type here to clean up method signatures.

container/libcontainer/helpers.go
```diff
@@ -91,27 +93,30 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int) (*info.Cont
...
+		if _, ignore := ignoreMetrics[container.NetworkUsageMetrics]; !ignore {
+			netStats, err := networkStatsFromProc(rootFs, pid)
+			if err != nil {
+				glog.V(2).Infof("Unable to get network stats from pid %d: %v", pid, err)
+			} else {
+				stats.Network.Interfaces = append(stats.Network.Interfaces, netStats...)
+			}
 		}
+		if _, ignore := ignoreMetrics[container.NetworkTcpUsageMetrics]; !ignore {
+			// Commenting out to disable: too CPU intensive
```
timstclair: Delete this comment

container/docker/factory.go
```diff
@@ -236,6 +239,7 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo) error {
 		machineInfoFactory: factory,
 		storageDriver:      storageDriver(sd),
 		storageDir:         storageDir,
+		ignoreMetrics:      ignoreMetrics,
```
timstclair: (optional nit) I think this would all read a little more naturally if you inverted the logic (e.g. `includedMetrics` or `enabledMetrics`).

timstclair: Nevermind, I realize that this isn't an inclusive list. In that case, I have a slight preference for `disabledMetrics` but leave it up to you.

manager/manager.go
```diff
@@ -50,6 +51,35 @@ var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h",
...
+func (ml *metricsList) String() string {
+	return fmt.Sprint(*ml)
+}
+
+func (ml *metricsList) Set(value string) error {
+	if len(*ml) > 0 {
+		return errors.New("disable_metrics already set")
+	}
+	for _, ip := range strings.Split(value, ",") {
+		if ip == container.DiskUsageMetrics.String() {
```
timstclair: Use a switch statement. Or better yet, check the value against a whitelist (disk, network, tcp), and just cast the value:
```go
if !whitelist[ip] { return error }
(*ml)[container.MetricKind(ip)] = true
```

manager/manager.go
```diff
@@ -50,6 +51,35 @@ var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h",
...
+
+type metricsList map[container.MetricKind]struct{}
+
+func (ml *metricsList) String() string {
+	return fmt.Sprint(*ml)
+}
+
+func (ml *metricsList) Set(value string) error {
+	if len(*ml) > 0 {
+		return errors.New("disable_metrics already set")
```
timstclair: I don't think this should be an error. For instance, what if someone wanted to override the default value?

manager/manager.go
```diff
@@ -50,6 +51,35 @@ var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h",
...
+	}
+	for _, ip := range strings.Split(value, ",") {
```
timstclair: what does `ip` mean in this context?

manager/manager.go
```diff
@@ -199,18 +231,23 @@ type manager struct {
 	startupTime              time.Time
 	maxHousekeepingInterval  time.Duration
 	allowDynamicHousekeeping bool
+	ignoreMetrics            map[container.MetricKind]struct{}
+}
+
+func parseMetricKind(input string) (res map[container.MetricKind]struct{}) {
```
timstclair: unused?

timstclair:
Completed review pass.

manager/manager.go
vishh: Yup. Fixing it

container/docker/handler.go
vishh: `bool` is misleading since the zero value for an unknown key is also bool. I'm adding a complex type to make this cleaner.

container/factory.go
vishh: Yeah. Good idea. Doing it

container/libcontainer/helpers.go
vishh: Deleting.

manager/manager.go
vishh: +1 updating

manager/manager.go
vishh: Yeah. Thats true.

manager/manager.go
vishh: `input`.

vishh: updating it to `metric`

manager/manager.go
vishh: Yeah. Deleting this logic

vishh:
@timstclair: Addressed all comments. PTAL.

vishh:
@timstclair: PTAL

container/docker/handler.go
timstclair: I'm fine with the complex type, but just wanted to clarify my suggestion. Using a bool value that's always true works *because* the zero value is a bool (false). I recognize it can be a little ambiguous, but I think it's ok for contained usages. Since the set is used across files, I think your approach is good.

manager/manager.go
```diff
@@ -50,6 +50,42 @@ var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h",
 var eventStorageEventLimit = flag.String("event_storage_event_limit", "default=100000", "Max number of events to store (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is an integer. Default is applied to all non-specified event types")
 var applicationMetricsCountLimit = flag.Int("application_metrics_count_limit", 100, "Max number of application metrics to store (per container)")
 
+var (
+	// Metrics to be ignored.
+	ignoreMetrics metricSetValue = metricSetValue{container.MetricSet{}}
+	// List of metrics that can be ignored.
+	ignoreWhitelist = map[string]struct{}{
```
timstclair: (nit) use a MetricSet?

timstclair:
LGTM with 1 small nit. Self merge when ready :)

manager/manager.go
vishh: UPdating.

vishh:
Fixed the nits. Will merge once tests are green

a86c6f7964:
Would be nice to have this added to https://github.com/google/cadvisor/blob/master/docs/runtime_options.md

PR #1125, opened 2016-02-25
Evaluate labels from all mount points (rather than just "supported" fs)

aaronlevy:
My specific use-case is running the kubelet in a rkt container, where / will be overlayfs.

/cc @sjpotter @vishh

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -76,6 +76,10 @@ func NewFsInfo(context Context) (FsInfo, error) {
 		labels:     make(map[string]string, 0),
 		dmsetup:    &defaultDmsetupClient{},
 	}
+
+	fsInfo.addSystemRootLabel(mounts)
+	fsInfo.addDockerImagesLabel(context, mounts)
+
 	supportedFsType := map[string]bool{
 		// all ext systems are checked through prefix.
 		"btrfs": true,
@@ -102,12 +106,7 @@ func NewFsInfo(context Context) (FsInfo, error) {
 		}
 	}
 
-	// need to call this before the log line below printing out the partitions, as this function may
-	// add a "partition" for devicemapper to fsInfo.partitions
-	fsInfo.addDockerImagesLabel(context)
-
 	glog.Infof("Filesystem partitions: %+v", fsInfo.partitions)
-	fsInfo.addSystemRootLabel()
 	return fsInfo, nil
 }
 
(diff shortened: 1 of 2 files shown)


aaronlevy:
I have a CLA signed by employer (CoreOS). cla bot on kubernetes repo picks it up (or at least it has in the past).

sjpotter:
This goes to a better Q, why is the labels stuff so complicated?  

why don't we just say label-x maps to directory y and jus statfs that directory.  I'm sure there's some history behind it, just wondering.

fs/fs.go
```diff
@@ -78,9 +78,10 @@ func NewFsInfo(context Context) (FsInfo, error) {
 	}
 	supportedFsType := map[string]bool{
 		// all ext systems are checked through prefix.
-		"btrfs": true,
-		"xfs":   true,
-		"zfs":   true,
+		"btrfs":   true,
+		"overlay": true,
```
vishh: This will result in all the overlay filesystems being created for containers from being included in the list of filesystems on the host. It will make the API pretty much unusable.

aaronlevy:
I'm pretty unfamiliar with this code, so apologies for the PR that supposedly would break everything :)

What I'm seeing is that cadvisor (in kubelet) was not adding a "docker-images" label, because the rootfs (from the kubelet's perspective) was overlayfs. The kubelet would then keep logging `non-existent label "docker-images"`.

A quick test of the change in this PR resolved that issue, and the cadvisor output of /api/2.0/storage seemed reasonable (not adding a bunch of overlay filesystems). But again, really unfamiliar with what to look for.

Any thoughts on a way to move forward?

fs/fs.go
sjpotter: I think the Q is, what api is this trying to provide, I've tried to make heads or tails out of it, and I don't know.  

i.e. why isn't doing a statfs on a given dir good enough at all times?  why do we care about the partition table?

vishh: Ideally docker root fs should be detected using docker's APIs (info) and a statfs on that should be good enough. 
This code just needs some love.
This logic here was originally meant for discovering filesystems of interest at a global level. 
Layered filesystems that are ephemeral are not of interest to most users in the global context.

sjpotter: so for now perhaps say if mount point is / and its overlay we dont ignore
it?

vishh: I'd like this PR to use docker info + statfs for getting docker filesystem
stats. That way these labels don't matter.

vishh:
I'd like this PR to use docker info + statfs for getting docker filesystem
stats. That way these labels don't matter.

aaronlevy:
I wanted to get some feedback on the possibility of this proposed change.

I'm hesitant to go the route of changing the label/fsInfo interfaces because I'm not familiar with the motivations behind the design. When I took another look at this issue, it felt like I was guessing at what would be an acceptable way forward (because I don't know why certain decisions were made). Or, if I just introduced new interfaces to the manager (e.g. something like GetDockerFSInfo()) then I felt like it was exposing a new public API for something that wouldn't be tracked/exposed anywhere else (/api/v2.0/storage, for example).

Instead, I'm wondering if this approach might be acceptable in the interim. Essentially allowing labels to apply to (and add) partitions that might have otherwise been filtered due to filesystem or mountinfo ordering.

The former case is what kept the "root" label from being applied.
The latter case is what was causing the docker-images label to not be applied: /var/lib/foo was listed as mount before /var/lib/docker, so /var/lib/docker was ignored (partition tracks single mountpoint per device) and no "/" mount to fall back on.

Added WIP, as I could probably add some more tests for my specific case, and I still need to run it through some more end-to-end tests on my side ... but wanted to get feedback on this approach.

aaronlevy:
ps: may not have been clear, but I also added a new commit to this PR that takes a new approach / fixes the issues I was seeing running kubelet in a container. /cc @sjpotter @vishh

philips:
lgtm

fs/fs.go
```diff
@@ -190,22 +190,30 @@ func getDockerImagePaths(context Context) []string {
...
-				mnt := self.partitions[i].mountpoint
-				if len(mnt) < len(mountpoint) {
-					self.labels[LabelDockerImages] = source
-				}
-			} else {
-				self.labels[LabelDockerImages] = source
+func (self *RealFsInfo) updateDockerImagesPath(mounts []*mount.Info, dockerImagePaths []string) {
+	var useMount *mount.Info
+	for _, m := range mounts {
+		for _, p := range dockerImagePaths {
```
vishh: nit: To avoid this loop, can we make `getDockerImagePaths()` return a `map[string]struct{}` ?

vishh:
ok to test

vishh:
This change LGTM except for one nit.

PR #1137, opened 2016-02-27
use http to avoid not ssl support for wget in busybox

### LINKED ISSUE #1131: "wget: can't execute 'ssl_helper': No such file or directory" in integration test container
Junk in the TestDockerContainerNetworkStats test container

```
Every 1s: wget https://www.google.com/                      2016-02-27 03:35:07

Connecting to www.google.com (173.194.115.49:443)
wget: can't execute 'ssl_helper': No such file or directory
wget: error getting response: Connection reset by peer
```

Apparently it tries without ssl support an fails, but that generates network traffic which makes the test ultimately pass.   Worth cleaning up though.

sjenning:
Fixes #1131

### DIFF
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -269,7 +269,7 @@ func TestDockerContainerNetworkStats(t *testing.T) {
 	defer fm.Cleanup()
 
 	// Wait for the container to show up.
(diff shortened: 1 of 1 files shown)


jimmidyson:
OK to test

jimmidyson:
Thanks @sjenning

PR #1139, opened 2016-02-27
avoid stray timeout message

sjenning:
Presently, if the cmd returns and error the function immediately returns and doesn't stop the timer, leading to a stray timeout message a minute later for a command the has long since failed.


vishh:
ok to test

vishh:
LGTM

PR #1143, opened 2016-02-29
skip TestDockerFilesystemStats for devicemapper

### LINKED ISSUE #1135: TestDockerFilesystemStats fails on RHEL7 with HasFilesystem:false
HasFilesystem is expected to be true in the test

```
--- FAIL: TestDockerFilesystemStats (10.97s)
        Location:       docker_test.go:336
	Error:		Expected not to be nil.
	Messages:	got info: {Spec:{CreationTime:2016-02-27 03:58:05.94095891 +0000 UTC Aliases:[sharp_aryabhata 75c534be25bd074d7fdf9ae1f88698e8b389cf3fab7392c65023c1ea37f72b33] Namespace:docker Labels:map[] Envs:map[] HasCpu:true Cpu:{Limit:1024 MaxLimit:0 Mask:0-3 Quota:0 Period:0} HasMemory:true Memory:{Limit:18446744073709551615 Reservation:0 SwapLimit:18446744073709551615} HasCustomMetrics:false CustomMetrics:[] HasNetwork:true HasFilesystem:false HasDiskIo:true Image:busybox} Stats:[0xc2080561e0]}
		
FAIL
```

sjenning:
fixes #1135

### DIFF
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -295,6 +295,12 @@ func TestDockerFilesystemStats(t *testing.T) {
 	fm := framework.New(t)
 	defer fm.Cleanup()
 
+	storageDriver := fm.Docker().StorageDriver()
+	if storageDriver == framework.DeviceMapper {
+		// Filesystem stats not supported with devicemapper, yet
+		return
+	}
+
 	const (
 		ddUsage       = uint64(1 << 3) // 1 KB
 		sleepDuration = 10 * time.Second
@@ -307,7 +313,6 @@ func TestDockerFilesystemStats(t *testing.T) {
 		Count:  1,
 	}
 	needsBaseUsageCheck := false
-	storageDriver := fm.Docker().StorageDriver()
 	switch storageDriver {
 	case framework.Aufs, framework.Overlay:
 		needsBaseUsageCheck = true


container/docker/handler.go
```diff
@@ -168,7 +168,8 @@ func newDockerContainerHandler(
 		ignoreMetrics:      ignoreMetrics,
 	}
 
-	if !ignoreMetrics.Has(container.DiskUsageMetrics) {
+	if !ignoreMetrics.Has(container.DiskUsageMetrics) && storageDriver != devicemapperStorageDriver {
```
vishh: Is this necessary? `otherStorageDir` directory is common to all storage drivers and we do want to surface usage from that.

sjenning: You are saying that we should only not gather stats for the rootfsStorageDir?

vishh: Yeah. `rootfsStorageDir` is storage driver dependent.

sjenning:
oops, didn't mean to close. update pushed to just skip the TestDockerFilesystemStats if devicemapper is the storage driver.

integration/tests/api/docker_test.go
```diff
@@ -295,6 +295,12 @@ func TestDockerFilesystemStats(t *testing.T) {
 	fm := framework.New(t)
 	defer fm.Cleanup()
 
+	storageDriver := fm.Docker().StorageDriver()
+	if storageDriver == framework.DeviceMapper {
```
vishh: Logs usage is still supported with devicemapper. If you look below, only the root filesystem usage part is skipped for devicemapper.

vishh:
If we are ok with RHEL not being tested for filesystem stats, I'm fine with this change. We should come back and fix devicemapper support though since it is necessary for proper kube integration.

sjenning:
I think it is fine.

The problem is that ```info.Stats[0].Filesystem``` is nil (```container.Spec.HasFilesystem``` is false) so the test doesn't even get to the ```needsBaseUsageCheck``` part.

vishh:
Should we target ensure that the spec is valid? It should contain logs usage...

vishh:
ok to test

vishh:
Let's merge this for now. But I'd really appreciate any help with turning on the test for RHEL sometime soon. Disk accounting in kube v1.2 will remain untested on RHEL otherwise.

vishh:
@k8s-bot test this

PR #1136, opened 2016-02-27
fix dd string being treated as a single command

sjenning:
After running integration tests (and hacking the code to not remove the containers)

```
02b69f1b7f1d        busybox             "/bin/sh -c ''dd if=/" ...
# docker logs 02b69f1b7f1d
/bin/sh: dd if=/dev/zero of=/file count=2 bs=8 & sleep 10000: not found
```

This PR keeps the dd string from being treated as a single huge command, which of course isn't found.

Fixes #1130 (#1135 remains)

PR #1144, opened 2016-03-01
Fix test commands that are currently failing over ssh.

vishh:
cc @sjenning 

For #1136


pwittrock:
As stated offline, the test code is getting a bit hard to follow / validate.  We will want to rethink how we remotely execute test code.

LGTM

PR #1146, opened 2016-03-01
Ran `godep restore && godep save ./...`

vishh:
@pwittrock: This PR is an attempt to make godeps bump on the kube side succeed.


vishh:
Closing this PR since it does not help with kube side updates

PR #1152, opened 2016-03-08
Cadvisor hangs on startup

derekwaynecarr:
This came to my attention when debugging startup time of the kubelet when not on a cloud.

I would like to get this into kubernetes if possible as it negatively impacts our ability to test kubelet in its current state.

/cc @vishh @dchen1107

### DIFF
--- a/utils/cloudinfo/aws.go
+++ b/utils/cloudinfo/aws.go
@@ -15,6 +15,8 @@
 package cloudinfo
 
 import (
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -23,7 +25,13 @@ import (
 )
 
 func onAWS() bool {
-	client := ec2metadata.New(session.New(&aws.Config{}))
+	// the default client behavior retried the operation multiple times with a 5s timeout per attempt.
+	// if you were not on aws, you would block for 20s when invoking this operation.
+	// we reduce retries to 0 and set the timeout to 2s to reduce the time this blocks when not on aws.
+	client := ec2metadata.New(session.New(&aws.Config{MaxRetries: aws.Int(0)}))
+	if client.Config.HTTPClient != nil {
+		client.Config.HTTPClient.Timeout = time.Duration(2 * time.Second)
+	}
 	return client.Available()
 }
 
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -121,11 +121,11 @@ type NetInfo struct {
 type CloudProvider string
 
 const (
-	GCE            CloudProvider = "GCE"
-	AWS                          = "AWS"
-	Azure                        = "Azure"
-	Baremetal                    = "Baremetal"
-	UnkownProvider               = "Unknown"
+	GCE             CloudProvider = "GCE"
+	AWS                           = "AWS"
+	Azure                         = "Azure"
+	Baremetal                     = "Baremetal"
+	UnknownProvider               = "Unknown"
 )
 
 type InstanceType string
--- a/utils/cloudinfo/cloudinfo.go
+++ b/utils/cloudinfo/cloudinfo.go
@@ -66,7 +66,7 @@ func detectCloudProvider() info.CloudProvider {
 	case onBaremetal():
 		return info.Baremetal
 	}
-	return info.UnkownProvider
+	return info.UnknownProvider
 }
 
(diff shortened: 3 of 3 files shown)


dchen1107:
cc/ @justinsb on aws support. 

LGTM

justinsb:
LGTM, except I don't think there's any DNS involved, as on EC2 the metadata service is on 169.254.169.254.  If I'm correct the comment is a bit misleading...

There is also this alternative approach, though I haven't tried it in production (it seems to be accurate though): http://serverfault.com/a/700771

I'm fine with PR as is.  I don't think I've ever seen the metadata service fail to respond... I'm pretty sure it is implemented on the host machine, so it seems the best option, particularly if we want to get it into 1.2

derekwaynecarr:
I can update the comment in morning.

In general, I would love a solution that does not block for even 5s like
this so we can visit the alternate approaches post 1.2.

Thanks!

utils/cloudinfo/aws.go
```diff
@@ -23,7 +23,10 @@ import (
 )
 
 func onAWS() bool {
-	client := ec2metadata.New(session.New(&aws.Config{}))
+	// by default, we set retries to 0 for this operation.
+	// when not running on AWS, not setting retries can cause the operation to block
```
vishh: nit: Why is there a comment about not running on aws here?

vishh:
ok to test

utils/cloudinfo/aws.go
derekwaynecarr: I am going to clarify the comment.

The default aws behavior is to retry 4 times, each with a 5 second timeout.  If you run cadvisor outside of gce or aws, you end up waiting 20s because the cloud detection checks gce first, aws second, and then azure third, and then finally it says I give up.

dchen1107:
@justinsb and @vishh It is not a blocker, but it would be nice to get this into 1.2 release. Thanks!

utils/cloudinfo/aws.go
vishh: I wonder why the e2e tests did not catch this.

derekwaynecarr: I caught it running e2e not GCE or AWS.  It was very obvious.

derekwaynecarr:
@vishh updated the comment to try to improve clarity.

utils/cloudinfo/aws.go
```diff
@@ -23,7 +23,10 @@ import (
 )
 
 func onAWS() bool {
-	client := ec2metadata.New(session.New(&aws.Config{}))
+	// the default client behavior retried the operation 4 times with a 5s timeout per attempt.
+	// if you were not on aws, you would block for 20s when invoking this operation.
+	// we reduce retries to 0 to reduce the time this blocks when not on aws.
+	client := ec2metadata.New(session.New(&aws.Config{MaxRetries: aws.Int(0)}))
```
vishh: Can we also reduce the default timeout?

derekwaynecarr: yes, but I was apprehensive to doing that since I have no idea if the 5s was based on actual observation in aws.  that said, i would be happy to go to 2s or 1s if someone could vouch for that risk.

`client.Config.HTTPClient.Timeout` can be lowered after this line to something smaller...

derekwaynecarr: @justinsb - do you think its safe to decrease the default timeout to 2s?

vishh: I'd hope that the local metadata server can be resolved locally and that 2s
should be good enough.

derekwaynecarr:
@vishh updated to 2s timeout.

```
sudo ./cadvisor -logtostderr
I0309 16:26:12.040451   23094 storagedriver.go:45] Caching stats in memory for 2m0s
I0309 16:26:12.040596   23094 manager.go:168] cAdvisor running in container: "/"
I0309 16:26:12.181361   23094 fs.go:109] Filesystem partitions: map[/dev/mapper/fedora-root:{mountpoint:/ major:253 minor:2 fsType: blockSize:0} /dev/sda1:{mountpoint:/boot major:8 minor:1 fsType: blockSize:0} /dev/mapper/fedora-home:{mountpoint:/home major:253 minor:3 fsType: blockSize:0}]
I0309 16:26:14.241805   23094 manager.go:205] Machine: { ...
```

vishh:
LGTM

justinsb:
I'd say 2 seconds is likely fine, but I'm going to open an issue to detect using the hypervisor uuid trick so we can raise this back to the default values.

PR #1154, opened 2016-03-10
Part 2 of rkt support

sjpotter:
The actual handler

Since I originally wrote it and splitting it in 2 an ignore metrics map was created.  I think I captured it, but not 100% sure.

The design of this is to capture 2 of the cgroups as a rkt pod and a rkt container pair.

i.e. under the cgroup fs we'd have

machine-slice/<rkt pod>/system.slice/<rkt container>

in terms of raw containers this will ignore any system.slice directory under the rkt pod and will not be tracked by cadvisor by either.  I think this is correct, but feedback would be appreciated.

I'm also not convinced by the efforts of splitting the fs path, I was reccomended to do this over a regex but again feedback would be appreciated.


sjpotter:
@vishh @timstclair @yifan-gu @jonboulle

vishh:
The raw driver has been simple so far - it handles any cgroups that cannot be handled by other handlers. Any specific reason to change that?

sjpotter:
It still basically does that.  The idea is that "system.slice" cgroup itself isn't a "real" cgroup.

i.e. given a cgroup path

/sys/fs/cgroup/cpuset//machine.slice/machine-rkt\\x2d3c642e7b\\x2d7e44\\x2d42a1\\x2dba5f\\x2d2da88df3909c.scope/system.slice/alpine-sh.service

I want to consider the cgroups under machine.slice to be the "pod" (i.e. it will be the only one to report network stats as they are global to the pod) and the cgroups under system.slice to be the "containers". And I want those containers to be the direct subcontainers of the pod.

So, yea, I could say that the rkt handler doesn't accept or handle "system.slice" but it wouldn't have any parent or child from a raw container so would just disappear, seems better to say explicitly it is ignored.

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+	return nil, nil
+}
+
+func (self *rktContainerHandler) ListProcesses(listType container.ListType) ([]int, error) {
+	return libcontainer.GetProcesses(self.cgroupManager)
+}
+
+// Watches the specified directory and all subdirectories. Returns whether the path was
+// already being watched and an error (if any).
+func (self *rktContainerHandler) watchDirectory(dir string, containerName string) (bool, error) {
```
vishh: Why is the rkt handler watching for new containers? I'd expect to have a global watcher which then hands off new containers to one of the handlers.

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+
+func (self *rktContainerHandler) HasFilesystem() bool {
+	if !self.ignoreMetrics.Has(container.DiskUsageMetrics) {
+		return true
+	}
+	return false
+}
+
+func newRktContainerHandler(name string, cgroupSubsystems *libcontainer.CgroupSubsystems, machineInfoFactory info.MachineInfoFactory, fsInfo fs.FsInfo, watcher *common.InotifyWatcher, rootFs string, ignoreMetrics container.MetricSet) (container.ContainerHandler, error) {
+	rktPath := "/var/lib/rkt"
```
vishh: Is this configurable? Can users set it to an alternative path?

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+
+	glog.Infof("rkt cgroup name = %q", name)
```
vishh: nit: Reduce this log to at-least V(4).

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+
+		pids, err := cgroups.GetPids(cgroupPaths["cpu"])
+		if err == nil {
+			pid = pids[0]
+		} else {
+			glog.Infof("couldn't get processes as %v", err)
+		}
+	}
+
+	//SJP: unsure the point of this code, if it event does anything today?
```
vishh: I don't think this logic is needed for rkt handler. This logic was added for lxc integration a while back to the raw handler.

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+func (self *rktContainerHandler) GetSpec() (info.ContainerSpec, error) {
+	return common.GetSpec(self)
+}
+
+func (self *rktContainerHandler) getFsStats(stats *info.ContainerStats) error {
+	if self.ignoreMetrics.Has(container.DiskUsageMetrics) {
+		return nil
+	}
+
+	if self.isPod {
```
vishh: I assume logs will be pod level usage?

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	mi, err := self.machineInfoFactory.GetMachineInfo()
+	if err != nil {
+		return err
+	}
+	var limit uint64 = 0
+
+	// SJP: Docker does not impose any filesystem limits for containers. So it uses capacity as limit.
+	// Doing the same for Rkt.  is this true?
```
vishh: Yup. Filesystem limits aren't available.

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+func (self *rktContainerHandler) GetCgroupPath(resource string) (string, error) {
+	path, ok := self.cgroupPaths[resource]
+	if !ok {
+		return "", fmt.Errorf("could not find path for resource %q for container %q\n", resource, self.name)
+	}
+	return path, nil
+}
+
+func (self *rktContainerHandler) GetContainerLabels() map[string]string {
+	return map[string]string{}
```
vishh: Is rkt planning to implement labels?

container/rkt/handler.go
sjpotter: So this is the same exact code as the raw container, I didn't know if it was recursive or not.  If it is and it just works because of the raw container support, that's great.

container/rkt/handler.go
sjpotter: It will be.  At this point I just wanted to get it in wit ha static path, we've dded support for getting this info out of the api service, just need to figure out how to use its api.

container/rkt/handler.go
sjpotter: will be removed

container/rkt/handler.go
sjpotter: I think it is, need a pid in the cgroup to get networking stats to work.

container/rkt/handler.go
sjpotter: Don't have that done yet, but yes, need to figure out how to role that up (assuming this is the journald stuff we talked about previously?)

container/rkt/handler.go
sjpotter: yes, just haven't quite figured out what goes in there yet.

timstclair: Add TODO comments for cases like this.

container/rkt/factory.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/google/cadvisor/container"
+	"github.com/google/cadvisor/container/common"
+	"github.com/google/cadvisor/container/libcontainer"
+	"github.com/google/cadvisor/fs"
+	info "github.com/google/cadvisor/info/v1"
+
+	"github.com/golang/glog"
+)
+
+var RktNamespace = "rkt"
```
timstclair: (nit) const

container/rkt/factory.go
```diff
@@ -0,0 +1,98 @@
...
+
+type rktFactory struct {
+	// Factory for machine information.
```
timstclair: (nit) most of these comments are unnecessary

container/rkt/factory.go
```diff
@@ -0,0 +1,98 @@
...
+func (self *rktFactory) NewContainerHandler(name string, inHostNamespace bool) (container.ContainerHandler, error) {
+	rootFs := "/"
+	if !inHostNamespace {
+		rootFs = "/rootfs"
+	}
+	return newRktContainerHandler(name, self.cgroupSubsystems, self.machineInfoFactory, self.fsInfo, self.watcher, rootFs, self.ignoreMetrics)
+}
+
+func (self *rktFactory) CanHandleAndAccept(name string) (bool, bool, error) {
+	// SJP: will ignore all cgroup names that don't either correspond to the machine.slice that is the pod
```
timstclair: What is `SJP`?

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+	pid := os.Getpid()
+	hasNetwork := false
+	if isPod {
+		hasNetwork = true
+
+		pids, err := cgroups.GetPids(cgroupPaths["cpu"])
+		if err == nil {
+			pid = pids[0]
+		} else {
+			glog.Infof("couldn't get processes as %v", err)
```
timstclair: This should probably be a warning or an error

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+
+	//SJP: unsure the point of this code, if it event does anything today?
+	var externalMounts []common.Mount
+	for _, container := range cHints.AllHosts {
+		if name == container.FullName {
+			/*libcontainerState.NetworkState = network.NetworkState{
+				VethHost:  container.NetworkInterface.VethHost,
+				VethChild: container.NetworkInterface.VethChild,
+			}
+			hasNetwork = true*/
```
timstclair: delete this?

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+		}
+	}()
+
+	return nil
+}
+
+func (self *rktContainerHandler) StopWatchingSubcontainers() error {
+	// Rendezvous with the watcher thread.
+	self.stopWatcher <- nil
+	return <-self.stopWatcher
```
timstclair: If there is an error closing the watcher, this will never return.

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	// Process the events received from the kernel.
+	go func() {
+		for {
+			select {
+			case event := <-self.watcher.Event():
+				err := self.processEvent(event, events)
+				if err != nil {
+					glog.Warningf("Error while processing event (%+v): %v", event, err)
```
timstclair: Should this be error level?

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type parsedName struct {
+	Pod       string
+	Container string
+}
+
+func verifyName(name string) (bool, error) {
+	_, err := parseName(name)
+	if err != nil {
```
timstclair: (nit) prefer `if _, err := parseName(name); err != nil {`

timstclair: or alternatively:
```
_, err := parseName(name)
return err == nil, err
```

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// Parse cgroup name into a pod/container name struct
+func parseName(name string) (*parsedName, error) {
+	splits := strings.Split(name, "/")
+	if len(splits) == 3 || len(splits) == 5 {
```
timstclair: Can you add a comment with a few example paths here?

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return nil, fmt.Errorf("%s not handled by rkt handler", name)
+}
+
+// Gets a Rkt container's overlay upper dir
+func GetRootFs(root string, parsed *parsedName) string {
+	/* Example of where it stores the upper dir key
+	/var/lib/rkt/pods/run/bc793ec6-c48f-4480-99b5-6bec16d52210/appsinfo/alpine-sh/treeStoreID
+	*/
```
timstclair: (nit) Go style prefers `//` style comments (https://golang.org/doc/effective_go.html#commentary) in most places.

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+				return parsed, nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("%s not handled by rkt handler", name)
+}
+
+// Gets a Rkt container's overlay upper dir
+func GetRootFs(root string, parsed *parsedName) string {
```
timstclair: (nit) Be consistent with your exports. It looks like all these helpers are not used outside the rkt package, so should not be exported.

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	/* Example of where it stores the upper dir key
+	/var/lib/rkt/pods/run/bc793ec6-c48f-4480-99b5-6bec16d52210/appsinfo/alpine-sh/treeStoreID
+	*/
+	if parsed.Container == "" {
+		return ""
+	}
+
+	tree := path.Join(root, "pods/run", parsed.Pod, "appsinfo", parsed.Container, "treeStoreID")
+	glog.Infof("tree = %q", tree)
```
timstclair: I think this should be debug level (`V(4)`)

container/rkt/handler.go
```diff
@@ -0,0 +1,488 @@
...
+
+	// Rkt containers do not have subcontainers, only the "Pod" does.
+	if self.isPod == false {
+		var ret []info.ContainerReference
+		return ret, nil
+	}
+
+	// Turn the system.slice cgroups  into the Pod's subcontainers
+	for _, cgroupPath := range self.cgroupPaths {
+		glog.Infof("ListContainers: self.name = %q, cgroupPath = %q, listType = %v", self.name, cgroupPath, listType)
```
timstclair: I think this should also be debug level (`V(4)`)

container/rkt/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	bytes, err := ioutil.ReadFile(tree)
+	if err != nil {
+		glog.Infof("ReadFile failed: %v", err)
```
timstclair: error level?

container/rkt/factory.go
vishh: `TODO(SJP):` I guess...

container/rkt/handler.go
vishh: This PR needs to include the metrics selection (opt-out) patch that I posted recently.

container/rkt/factory.go
sjpotter: SJP being me.  J is for Joseph

container/rkt/handler.go
vishh: It is not needed here. IIUC, this logic is needed in the raw container handler mainly to watch all the sub-containers of `/`. 
This PR exposes one more abstraction issue. The raw handler must be made specialized, or the watching of subcontainers must be moved out of the handler interface into a something else, which the raw handler can implement.

container/rkt/handler.go
sjpotter: not needed.  was learning how it worked.

container/rkt/handler.go
sjpotter: this is straght up copied from raw, goes to other Q if this code is even needed if raw container does all the watching?

container/rkt/handler.go
sjpotter: again straight up copied from raw, so should look there too.

container/rkt/handler.go
sjpotter: so, logs will be pod level storage, not with the containers.  Is that ok?

vishh:
Ping @sjpotter..

sjpotter:
making progress, will update later today hopefully

container/rkt/handler.go
sjpotter: ok, removed and tested and you're right, works just fine.

container/rkt/handler.go
sjpotter: so, made some progress on this but ran into the godeps issue listed in https://github.com/google/cadvisor/issues/1161

container/rkt/handler.go
sjpotter: so ran into an issue here, the logic here of retrieving the pid isn't good enough, i.e. the cgroup is created before there is any data in that file and can be read before its filled in.  Unsure what the apporpriate way to deal with it, can treat it as os.Getpid() until their is something in the file (my current hack) but unsure that is best.

container/rkt/handler.go
sjpotter: using api service for this and the network namespace pid.

container/rkt/handler.go
sjpotter: using api service

container/rkt/factory.go
sjpotter: removed

container/rkt/handler.go
sjpotter: code fixed to not depend on this fragile code.

container/rkt/handler.go
sjpotter: Its been added.

container/rkt/handler.go
sjpotter: no more watcher in rkt handler

container/rkt/handler.go
sjpotter: no more watcher

container/rkt/handler.go
sjpotter: done

container/rkt/helpers.go
sjpotter: done

container/rkt/helpers.go
sjpotter: done

container/rkt/helpers.go
sjpotter: fixed

container/rkt/helpers.go
sjpotter: removed

sjpotter:
@k8s-bot test this

sjpotter:
@k8s-bot test this

yifan-gu:
@sjpotter Any ideas why the tests are failing?

yifan-gu:
@sjpotter 
Could you split the Godep changes from this commit `big change - rktapi service, pid fs stats, lots of nits`?
So we can ignore the Godep changes and focus on your actual touch of code?

sjpotter:
@yifan-gu per @vishh its a problem with jenkins not me.

sjpotter:
@yifan-gu split Godeps into its own commit

vishh:
The godeps changes are making it a pain to review. Can we squash all the non godep changes into a single commit?

container/rkt/client.go
```diff
@@ -0,0 +1,65 @@
...
+
+func RktPath() (string, error) {
+	client, err := Client()
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := client.GetInfo(context.Background(), &rktapi.GetInfoRequest{})
+	if err != nil {
+		return "", fmt.Errorf("couldn't GetInfo from rkt api servie: %v", err)
```
jonboulle: api service

container/rkt/factory.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (self *rktFactory) DebugInfo() map[string][]string {
+	return map[string][]string{}
+}
+
+func Register(machineInfoFactory info.MachineInfoFactory, fsInfo fs.FsInfo, ignoreMetrics container.MetricSet) error {
+	client, err := Client()
+	if err != nil {
+		return fmt.Errorf("unable to communicate with Rkt api service: %v", err)
```
jonboulle: nit, elsewhere you return this error directly, here you're wrapping

container/rkt/factory.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	resp, err := client.GetInfo(context.Background(), &rktapi.GetInfoRequest{})
+	if err != nil {
+		return fmt.Errorf("couldn't GetInfo from rkt api servie: %v", err)
```
jonboulle: service

container/rkt/handler.go
```diff
@@ -0,0 +1,339 @@
...
+func (self *rktContainerHandler) ContainerReference() (info.ContainerReference, error) {
+	return info.ContainerReference{
+		Name:      self.name,
+		Aliases:   self.aliases,
+		Namespace: RktNamespace,
+	}, nil
+}
+
+//SJP: Should a Rkt containe have have htis?
+func (self *rktContainerHandler) GetRootNetworkDevices() ([]info.NetInfo, error) {
```
jonboulle: does the raw one not work here?

container/rkt/helpers.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+type parsedName struct {
+	Pod       string
+	Container string
+}
+
+func verifyName(name string) (bool, error) {
+	_, err := parseName(name)
+	return err == nil, err
```
jonboulle: simpler to just inline on https://github.com/google/cadvisor/pull/1154/commits/206670a655ba587b6c794b1d238d6f1819ef69e8#diff-06ccdffbfc974d7801151a26ffee4f8fR66 IMHO

fs/fs.go
```diff
@@ -194,10 +211,10 @@ func getDockerImagePaths(context Context) map[string]struct{} {
 
 // This method compares the mountpoints with possible docker image mount points. If a match is found,
 // docker images label is added to the partition.
```
jonboulle: Needs update

manager/manager.go
```diff
@@ -135,7 +136,12 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 	if err != nil {
 		glog.Warningf("Unable to connect to Docker: %v", err)
 	}
-	context := fs.Context{DockerRoot: docker.RootDir(), DockerInfo: dockerInfo}
+	rktPath, err := rkt.RktPath()
+	if err != nil {
+		glog.Warningf("unable to connect to Rkt api service: %v", err)
```
jonboulle: So if we initialise a manager and rkt is unavailable, we just warn and then we're stuck with that for the lifetime of the manager? is that the same behaviour as with docker? surprised it doesn't try to recover...

container/rkt/client.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+var (
+	rktClient    rktapi.PublicAPIClient
+	rktClientErr error
+	once         sync.Once
+)
+
+func Client() (rktapi.PublicAPIClient, error) {
+	once.Do(func() {
```
vishh: Would we want to retry? What if the api service were to be restarting while cAdvisor is being started?
This is currently an issue with docker client as well, and it needs to be fixed there too.

container/rkt/factory.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	rktPath := resp.Info.GlobalFlags.Dir
```
vishh: nit: Why not use `RktPath()` instead?

container/rkt/handler.go
```diff
@@ -0,0 +1,339 @@
...
+	if parsed.Container == "" {
+		resp, err := rktClient.InspectPod(context.Background(), &rktapi.InspectPodRequest{
+			Id: parsed.Pod,
+		})
+		if err != nil {
+			return nil, err
+		}
+		pid = int(resp.Pod.Pid)
+	} else {
+		glog.Infof("skipping as Container")
```
vishh: nit: Give more context here. Skipped _which_ container?

container/rkt/handler.go
```diff
@@ -0,0 +1,339 @@
...
+		},
+		Paths: cgroupPaths,
+	}
+
+	hasNetwork := false
+	if isPod {
+		hasNetwork = true
+	}
+
+	//SJP: unsure the point of this code, if it event does anything today?
```
vishh: You don't have to include this logic here. It is used by some users to account for externally mounted volumes.

container/rkt/handler.go
vishh: No. I think the abstractions are incorrect. Only the raw handler that handler the machine itself is supposed to implement this method.

container/rkt/handler.go
```diff
@@ -0,0 +1,339 @@
...
+		rootFs:             rootFs,
+		isPod:              isPod,
+		aliases:            aliases,
+		pid:                pid,
+		rootfsStorageDir:   rootfsStorageDir,
+		ignoreMetrics:      ignoreMetrics,
+	}
+
+	if !ignoreMetrics.Has(container.DiskUsageMetrics) {
+		handler.fsHandler = common.NewFsHandler(time.Minute, rootfsStorageDir, "", fsInfo)
```
vishh: Are logs stored in some well known path?

container/rkt/handler.go
vishh: nit: Fix this and other comments/TODOs in this PR.

container/rkt/helpers.go
```diff
@@ -0,0 +1,92 @@
...
+	var tree string
+	if parsed.Container == "" {
+		tree = path.Join(root, "pods/run", parsed.Pod, "stage1TreeStoreID")
+	} else {
+		tree = path.Join(root, "pods/run", parsed.Pod, "appsinfo", parsed.Container, "treeStoreID")
+	}
+
+	bytes, err := ioutil.ReadFile(tree)
+	if err != nil {
+		glog.Infof("ReadFile failed: %v", err)
```
vishh: nit: Give more context. This line in the log file should make sense just by looking at it.

manager/manager.go
vishh: Recovery is needed. It has bitten us several times in the past.

vishh:
Completed another review pass. Ping me once you address all the comments @sjpotter. 
The main issue with the PR is the lack of resiliency to rkt apiservice failures. We can address that in a subsequent PR though...
How are we planning to add test coverage for the rkt integration?

container/rkt/factory.go
sjpotter: It's what the docker handler does

container/rkt/factory.go
sjpotter: fixed

container/rkt/factory.go
sjpotter: fixed

container/rkt/client.go
sjpotter: fixed

container/rkt/handler.go
sjpotter: So this isn't needed as a comment, the point of the code is that we only need the pid of the "pod" container, not a pid of each individual "app" container as the pid used only for networking which is at the pod layer.

container/rkt/handler.go
sjpotter: ok, removing hints

sjpotter: and the external mounts stuff, still need to be an empty slice to work with common code

container/rkt/handler.go
sjpotter: logs are stored in the journald in the pod's container and are accounted at the pod level.  is that sufficient?

container/rkt/handler.go
sjpotter: a little confused by what is being asked here?  Answering my Q?  so no code change really?

container/rkt/handler.go
sjpotter: you mean my SJP comments?

container/rkt/handler.go
vishh: I assume you will be removing this log then...

container/rkt/handler.go
vishh: Yeah. That sounds good.

container/rkt/handler.go
vishh: Yeah.

container/rkt/handler.go
sjpotter: yes.

container/rkt/handler.go
sjpotter: added a first cut at label support, not 100% sure its correct (i.e. we have Pod level labels and app level labels, applying them to the respective containers, but perhaps all pod labels should apply to apps as well)

container/rkt/helpers.go
sjpotter: fixed

fs/fs.go
sjpotter: fixed comment

manager/manager.go
sjpotter: out of scope for this PR though?

vishh:
LGTM. I didn't get a chance to test this PR yet. @sjpotter: What is the plan for testing rkt integration?

sjpotter:
@vishh need to figure that out, I'm not 100% sure how to do that.  In looking at the docker test cases, there isn't much there to inspire me

i can probably test individual functions (such as what I use to parse the cgroups), but unsure how to test every part of the e2e process.

In practice maybe can do integration tests like you have alrady (but aren't running)

vishh:
We can add a few nodes to the e2e setup that run rkt and have the e2e framework run either docker or rkt tests based on the environment.
The other option is for you to setup an independent e2e and post the status on github.

vishh:
@k8s-bot test this

sjpotter:
I've changed this PR a bit to use new rkt api service support to lookup rkt pods by cgroup path but it has to wait till 1.3 release this week (to be vendored properly).  This is necessary for it to be reliable in the presence of random slice definitions and pod cgroup names.

sjpotter:
@k8s-bot test this

vishh:
Tests are passing. Merging this PR. We can continue iterating on it.

Issue #1156, opened 2016-03-14
raw container GetStats not returning proper network stats in all cases


sjenning:
The raw container handler GetStats() is using the pid of the cadvisor process (or kubelet/origin) for getting the stats of every raw container.

```
func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
	stats, err := libcontainer.GetStats(self.cgroupManager, self.rootFs, os.Getpid(), self.ignoreMetrics)
```

The pid is currently only used for gathering network stats, but if the container is in a different network namespace, GetStats() will still return the net stats of the network namespace in which cadvisor is running.

In order to do this properly, the network stats from a pid in the ```tasks``` file of the cgroup should be used, assuming all pids in a cgroup are going to be in the same network namespace, which I don't think is a certainty, but is likely.

cc @ncdc @derekwaynecarr

ncdc:
cc @timothysc

ncdc:
cc @vishh

vishh:
cc @dchen1107

ncdc:
#1158 fixed this - we can close, right?

derekwaynecarr:
@ncdc :+1:

PR #1158, opened 2016-03-14
Avoid collecting network stats for non root cgroups in raw handler.

vishh:
cc @ncdc @sjenning @derekwaynecarr

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -92,29 +92,30 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int, ignoreMetri
 	stats := toContainerStats(libcontainerStats)
 
 	// If we know the pid then get network stats from /proc/<pid>/net/dev
-	if pid > 0 {
-		if !ignoreMetrics.Has(container.NetworkUsageMetrics) {
-			netStats, err := networkStatsFromProc(rootFs, pid)
-			if err != nil {
-				glog.V(2).Infof("Unable to get network stats from pid %d: %v", pid, err)
-			} else {
-				stats.Network.Interfaces = append(stats.Network.Interfaces, netStats...)
-			}
+	if pid == 0 {
+		return stats, nil
+	}
+	if !ignoreMetrics.Has(container.NetworkUsageMetrics) {
+		netStats, err := networkStatsFromProc(rootFs, pid)
+		if err != nil {
+			glog.V(2).Infof("Unable to get network stats from pid %d: %v", pid, err)
+		} else {
+			stats.Network.Interfaces = append(stats.Network.Interfaces, netStats...)
 		}
-		if !ignoreMetrics.Has(container.NetworkTcpUsageMetrics) {
-			t, err := tcpStatsFromProc(rootFs, pid, "net/tcp")
-			if err != nil {
-				glog.V(2).Infof("Unable to get tcp stats from pid %d: %v", pid, err)
-			} else {
-				stats.Network.Tcp = t
-			}
+	}
+	if !ignoreMetrics.Has(container.NetworkTcpUsageMetrics) {
+		t, err := tcpStatsFromProc(rootFs, pid, "net/tcp")
+		if err != nil {
+			glog.V(2).Infof("Unable to get tcp stats from pid %d: %v", pid, err)
+		} else {
+			stats.Network.Tcp = t
+		}
 
-			t6, err := tcpStatsFromProc(rootFs, pid, "net/tcp6")
-			if err != nil {
-				glog.V(2).Infof("Unable to get tcp6 stats from pid %d: %v", pid, err)
-			} else {
-				stats.Network.Tcp6 = t6
-			}
+		t6, err := tcpStatsFromProc(rootFs, pid, "net/tcp6")
+		if err != nil {
+			glog.V(2).Infof("Unable to get tcp6 stats from pid %d: %v", pid, err)
+		} else {
+			stats.Network.Tcp6 = t6
 		}
 	}
 
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -18,7 +18,6 @@ package raw
 import (
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path"
 	"strings"
 
@@ -62,6 +61,8 @@ type rawContainerHandler struct {
 
 	// Metrics to be ignored.
 	ignoreMetrics container.MetricSet
+
+	pid int
 }
 
 func (self *rawContainerHandler) GetCgroupPaths() map[string]string {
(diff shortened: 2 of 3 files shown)


ncdc:
Related issue is #1156

ncdc:
Looks like this should work

container/raw/handler.go
```diff
@@ -221,7 +221,12 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 }
 
 func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
-	stats, err := libcontainer.GetStats(self.cgroupManager, self.rootFs, os.Getpid(), self.ignoreMetrics)
+	pid := 0
+	// Do not collect network stats for non root cgroups.
+	if self.name == "/" {
```
sjenning: Nit: this will be the third instance of ```if self.name == "/" {``` check.  Maybe create a isRootCgroup() to help make this uniform and clear?

derekwaynecarr:
LGTM

derekwaynecarr:
@vishh - I think this should be cherry-picked into kube 1.2.  You agree?

vishh:
@derekwaynecarr: Yes.

container/raw/handler.go
vishh: Good idea. Done..

vishh:
ping @timstclair

container/raw/handler.go
```diff
@@ -88,6 +87,13 @@ func (self *rawContainerHandler) HasFilesystem() bool {
 	return false
 }
 
+func (self *rawContainerHandler) isRootCgroup() bool {
+	if self.name == "/" {
+		return true
+	}
+	return false
```
timstclair: Prefer:
```
return self.name == "/"
```

vishh: Yeah. Done.

container/raw/handler.go
```diff
@@ -221,7 +227,12 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 }
 
 func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
-	stats, err := libcontainer.GetStats(self.cgroupManager, self.rootFs, os.Getpid(), self.ignoreMetrics)
+	pid := 0
+	// pid is used to collect network stats. Set it only for root container.
+	if self.isRootCgroup() {
+		pid = 1
```
vishh: @derekwaynecarr @timstclair: I had incorrectly set the pid to be that of cAdvisor. I updated it to the init pid. FYI!

container/raw/handler.go
```diff
@@ -221,7 +227,12 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
...
+	}
```
timstclair: This seems like a bit of a hack, what if instead we add the network metrics to the set of ignoreMetrics we pass for non-root containers?

container/raw/handler.go
timstclair: I see. I guess my comment about adding network metrics to the ignoreMetrics is optional in that case.

container/raw/handler.go
vishh: Yeah. Makes sense. I updated the patch to do that.

vishh:
PTAL @timstclair

container/raw/handler.go
```diff
@@ -112,6 +117,16 @@ func newRawContainerHandler(name string, cgroupSubsystems *libcontainer.CgroupSu
 		}
 	}
 
+	if !isRootCgroup(name) {
+		// collect network metrics only for root cgroup
+		ignoreMetrics.Add(container.NetworkUsageMetrics)
```
timstclair: I think you need to make a deep copy of ignoreMetrics, or it will affect all containers.

vishh: :no_mouth: Fixed it.

timstclair: Your `Copy` is really a merge, so this isn't actually creating a new copy. I'd just update the copy function to initialize a new map, and return it.

vishh:
@k8s-bot test this

ncdc:
I'm seeing this in the logs:

```
--- FAIL: TestDockerContainerSpec (1.35s)
	assertions.go:150: 
	Location:	docker_test.go:220
	Error:		Should be true
	Messages:	Network should be isolated

--- FAIL: TestDockerContainerNetworkStats (10.89s)
	assertions.go:150: 
	Location:	docker_test.go:290
	Error:		Should not be equal
	Messages:	Network tx and rx bytes should not be equal

	assertions.go:150: 
	Location:	docker_test.go:291
	Error:		Should not be equal
	Messages:	Network tx and rx packets should not be equal
```

container/raw/handler.go
```diff
@@ -123,7 +133,8 @@ func newRawContainerHandler(name string, cgroupSubsystems *libcontainer.CgroupSu
 		externalMounts:     externalMounts,
 		watcher:            watcher,
 		rootFs:             rootFs,
-		ignoreMetrics:      ignoreMetrics,
+		ignoreMetrics:      newIgnoreMetrics,
```
ncdc: This needs fixing, right?

vishh:
@k8s-bot test this

container/raw/handler.go
vishh: Yeah. Sorry. Too much of context switching isn't helping with PRs..

vishh:
CI is happy now. PTAL @timstclair @ncdc

timstclair:
Did you remove the code to add network to the ignore metrics?

ncdc:
Wasn't that code that he removed essentially duplicating the pid > 0 check?

container/libcontainer/helpers.go
```diff
@@ -92,29 +92,30 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int, ignoreMetri
...
-			} else {
-				stats.Network.Interfaces = append(stats.Network.Interfaces, netStats...)
-			}
+	if pid == 0 {
+		return stats, nil
+	}
+	if !ignoreMetrics.Has(container.NetworkUsageMetrics) {
+		netStats, err := networkStatsFromProc(rootFs, pid)
+		if err != nil {
+			glog.V(2).Infof("Unable to get network stats from pid %d: %v", pid, err)
```
pmorie: I realize this is following the pattern that's in the code already, but why log the error instead of returning it, here?

vishh:
It should not be necessary since we set the pid.

pmorie:
K, this LGTM

container/libcontainer/helpers.go
vishh: Resiliency I guess. I can try changing that behavior in a new PR and see if any one complains..

vishh:
@derekwaynecarr: Can you verify this fix on a systemd machine? It should be straight forward. Run with `--profiling` flag to hit the pprof endpoints.

derekwaynecarr:
@vishh - I will pull and test now.

ncdc:
@vishh I didn't test this commit per se, but I did test a similar change in OpenShift's copy of cadvisor that only sets the pid nonzero (i.e. 1) for the root cgroup. And it did fix the CPU issue and I confirmed in pprof too.

derekwaynecarr:
@ncdc - i have it built locally, should take a few more secs for me to verify in the interest of safety.

derekwaynecarr:
LGTM, confirmed we are getting stats 1 time at the expected default 1s interval.

vishh:
Merging this. @timstclair I can address comments in a separate PR.

timstclair:
Yeah, this is fine. I just wanted to make sure that was a deliberate decision.

PR #1159, opened 2016-03-14
allow ignoreMetrics in new manager

sjenning:
kubelet would like to be able to set the ignore metrics but currently this is only supported when invoking cadvisor from the command line.

This allows the caller of manger.New() to pass in a set of metrics to ignore, falling back to the defaults if none are provided.

cc @ncdc @derekwaynecarr

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -22,9 +22,11 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"strings"
 	"syscall"
 	"time"
 
+	"github.com/google/cadvisor/container"
 	cadvisorhttp "github.com/google/cadvisor/http"
 	"github.com/google/cadvisor/manager"
 	"github.com/google/cadvisor/utils/sysfs"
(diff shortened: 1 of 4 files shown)


ncdc:
cc @vishh

derekwaynecarr:
LGTM

manager/manager.go
```diff
@@ -184,6 +184,10 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 		inHostNamespace = true
 	}
 
+	if ignoreMetricsSet == nil {
+		ignoreMetricsSet = ignoreMetrics.MetricSet
```
timstclair: I'd prefer that this was explicitly passed in from the main module rather than implicitly set here.

ncdc: Good idea - so move the flag defined in this file to cadvisor.go in the project root and pass it down from there?

timstclair: Yeah, sounds good.

sjenning:
@timothysc updated

manager/manager_test.go
```diff
@@ -206,12 +205,8 @@ func TestDockerContainersInfo(t *testing.T) {
...
 
 func TestNewNilManager(t *testing.T) {
-	_, err := New(nil, nil, 60*time.Second, true)
+	_, err := New(nil, nil, 60*time.Second, true, container.MetricSet{})
 	if err == nil {
 		t.Fatalf("Expected nil manager to return error")
 	}
 }
-
-func TestTcpMetricsAreDisabledByDefault(t *testing.T) {
```
vishh: nit: Can we keep this test around?

vishh:
Just one nit. Otherwise LGTM

cadvisor.go
```diff
@@ -52,8 +54,41 @@ var allowDynamicHousekeeping = flag.Bool("allow_dynamic_housekeeping", true, "Wh
...
+		}
+	}
+	return nil
+}
+
 func main() {
 	defer glog.Flush()
+	// Tcp metrics are ignored by default.
+	flag.Var(&ignoreMetrics, "disable_metrics", "comma-separated list of metrics to be disabled. Options are `disk`, `network`, `tcp`. Note: tcp is disabled by default due to high CPU usage.")
+	flag.Set("disable_metrics", "tcp")
```
timstclair: Rather than calling flag.Set, just add TCP to the default metrics on [line 58](https://github.com/google/cadvisor/pull/1159/files#diff-199ed097109d5fd1884fee60eefccfe8R58)

sjenning:
@vishh @timothysc updated

cadvisor.go
```diff
@@ -52,8 +54,40 @@ var allowDynamicHousekeeping = flag.Bool("allow_dynamic_housekeeping", true, "Wh
...
+		} else {
+			return fmt.Errorf("unsupported metric %q specified in disable_metrics", metric)
+		}
+	}
+	return nil
+}
+
 func main() {
 	defer glog.Flush()
+	// Tcp metrics are ignored by default.
```
timstclair: (nit) move the comment too.

cadvisor_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/cadvisor/container"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTcpMetricsAreDisabledByDefault(t *testing.T) {
+	assert.True(t, ignoreMetrics.Has(container.NetworkTcpUsageMetrics))
```
timstclair: This test isn't really doing anything anymore, since the ignoreMetrics flag is never actually added. I think the best way to solve this is by moving the flag declaration to an `init()` function, or an `addFlags()` function that you can call here.

Also, consider adding a `flag.Parse()` and verifying that the assertion still holds.

timstclair:
Thanks. BTW - I'm @timstclair not @timothysc - there are 2 of us :)

vishh:
LGTM.

sjenning:
@timstclair undetected name collision, my bad :) updated

timstclair:
LGTM

PR #1171, opened 2016-03-20
cloudinfo: bugfix openstack identified as aws

enoodle:
Because openstack and aws uses the exact same api,
the previous method to check for if running on aws
would have returned true on openstack as well.

### DIFF
--- a/utils/cloudinfo/aws.go
+++ b/utils/cloudinfo/aws.go
@@ -15,24 +15,26 @@
 package cloudinfo
 
 import (
-	"time"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"io/ioutil"
+	"strings"
 
 	info "github.com/google/cadvisor/info/v1"
 )
 
+const (
+	ProductVerFileName = "/sys/class/dmi/id/product_version"
+	Amazon             = "amazon"
+)
+
 func onAWS() bool {
-	// the default client behavior retried the operation multiple times with a 5s timeout per attempt.
-	// if you were not on aws, you would block for 20s when invoking this operation.
-	// we reduce retries to 0 and set the timeout to 2s to reduce the time this blocks when not on aws.
-	client := ec2metadata.New(session.New(&aws.Config{MaxRetries: aws.Int(0)}))
-	if client.Config.HTTPClient != nil {
-		client.Config.HTTPClient.Timeout = time.Duration(2 * time.Second)
+	data, err := ioutil.ReadFile(ProductVerFileName)
+	if err != nil {
+		return false
 	}
-	return client.Available()
+	return strings.Contains(string(data), Amazon)
 }
 
 func getAwsMetadata(name string) string {


simon3z:
Actually I looked into this as well and I couldn't spot anything better in the amazon documentation and api. (E.g. I heard someone is checking the mac address of the network interface, which seems even more fragile).

So for me at the moment is :+1: 

cc @vishh

ncdc:
cc @justinsb

derekwaynecarr:
:+1:  - this also removes a timeout delay at start-up...

vishh:
LGTM

vishh:
The same trick can be used for GCE as well. 

```shell
$ cat /sys/class/dmi/id/product_name 
Google
```

vishh:
ok to test

vishh:
@k8s-bot test again

jimmidyson:
When running in a container, do we want to prefix the path with path of mounted host root dir?

ncdc:
@jimmidyson at least when we run the OpenShift node containerized, we bind mount /sys as /sys into the container (see e.g. https://docs.openshift.org/latest/getting_started/administrators.html#running-in-a-docker-container). And the readme in this repo also suggests bind mounting /sys as /sys when running cadvisor as a container. Not sure about the Kubelet, though.

jimmidyson:
@ncdc Ah OK that's fine then.

jimmidyson:
retest this please

jimmidyson:
@k8s-bot test again

jimmidyson:
@vishh @pwittrock How do we rekick tests on PRs now? Not working for me...

jimmidyson:
ok to test

justinsb:
Is there any official documentation about this from AWS?  It seems the sort of thing that could break and be hard to diagnose & fix "in the field".  I'd be happier if there were others in the same boat as us, but I couldn't find anyone.

Also, is there some way to actively detect OpenStack?  Or to differentiate OpenStack's metadata service from AWS's (e.g. could we look at the availability-zone)?

One alternative: have a fast-path, and then slower but more complete fallback logic.  So dmi.product_version=amazon => AWS, dmi.product_name=Google => GCE etc, but if we don't identify any of them then we go through the official procedure.  I know it's ugly, but how about something like this:

```
func detectCloudProvider() info.CloudProvider {
       switch {
        case fastOnGCE():
                return info.GCE
        case fastOnAWS():
                return info.AWS
        case fastOnBaremetal():
                return info.Baremetal
        }
        glog.Infof("Fast cloud-provider detection could not determine the provider; falling back to slower approaches")
        switch {
        case onGCE():
                return info.GCE
        case onAWS():
                return info.AWS
        case onBaremetal():
                return info.Baremetal
        }
        return info.UnkownProvider
}

```

simon3z:
> One alternative: have a fast-path, and then slower but more complete fallback logic. So dmi.product_version=amazon => AWS, dmi.product_name=Google => GCE etc, but if we don't identify any of them then we go through the official procedure.

@justinsb let's assume the fast-path fails and you fallback to the official procedure... you would still hit this bug (identifying an OpenStack deployment as AWS).

The fail-safe addition of the slow-path that you're suggesting would just produce the same wrong results we have today.

vishh:
@justinsb @simon3z Are there any additional checks we can perform using the AWS client in the _slow_ path to avoid ambiguity against openstack?

simon3z:
@vishh I have no news since my last https://github.com/google/cadvisor/pull/1171#issuecomment-199247868

enoodle:
@vishh I didn't find any clear way to differentiate openstack from AWS with the REST API. I think a naive user could easily set an openstack availability-zone to the same name as the ones from AWS or give a public hostname that includes the string "amazon" or "aws".

vishh:
@k8s-bot test again

enoodle:
@k8s-bot test again

timstclair:
@k8s-bot test this

PR #1186, opened 2016-03-31
Remove CanHandleAndAccept from interface and include in NewContainerHandler

sjpotter:
This is about simplifying the inteface

change the ContainerHandlerFactory

removes
   CanHandleAndAccept(name string) (handle bool, accept bool, err error)

changes
   NewContainerHandler(name string, inHostNamespace bool) (c ContainerHandler, err error)
to
   NewContainerHandler(name string, inHostNamespace bool) (c ContainerHandler, ignore bool, err error)

NewContainerHandler will make the decision if its handling/accepting the container or if its explicitly ignoring it.  ignoring is if it could handle but decides not to accept.

if a container is ignored, its not handled by any other container.

so before we had a few cases

|old CanHandleAndAccept ||new NewContainerHandler||
| --- | --- | --- | --- |
|**handle**|**accept**|**handler**|**ignore**|
| true | false | nil | true |
| false | true | shouldn't happen | |
| false | false | nil | false |
| true | true | !nil | false |

this patch changes the interface and moves every container handler CanHandleAndAccept() into their own NewContainerHandler() and returns appropriately

also changed the tests to reflect these changes

The reason being is that deciding on acceptance and actually creating the handler often require the same information.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -94,15 +94,22 @@ func (self *dockerFactory) String() string {
 	return DockerNamespace
 }
 
-func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool) (handler container.ContainerHandler, ignoring bool, err error) {
+func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool) (handler container.ContainerHandler, err error) {
 	handle, accept, err := self.canHandleAndAccept(name)
-	if handle != true || accept != true || err != nil {
-		return nil, !accept, err
+	if handle != true {
+		return nil, nil
+	}
+	if accept != true {
+		ignore := container.NewIgnoreHandler(name)
+		return ignore, nil
+	}
+	if err != nil {
+		return nil, err
 	}
 
 	client, err := Client()
 	if err != nil {
-		return nil, !accept, err
+		return nil, err
 	}
 
 	metadataEnvs := strings.Split(*dockerEnvWhitelist, ",")
@@ -120,7 +127,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		self.dockerVersion,
 		self.ignoreMetrics,
 	)
-	return handler, !accept, err
+	return handler, err
 }
 
 // Returns the Docker ID from the full container name.
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -94,10 +94,15 @@ func (self *dockerFactory) String() string {
 	return DockerNamespace
 }
 
-func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool) (handler container.ContainerHandler, err error) {
+func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool) (handler container.ContainerHandler, ignoring bool, err error) {
+	handle, accept, err := self.canHandleAndAccept(name)
+	if handle != true || accept != true || err != nil {
+		return nil, !accept, err
+	}
+
 	client, err := Client()
 	if err != nil {
-		return
+		return nil, !accept, err
 	}
 
 	metadataEnvs := strings.Split(*dockerEnvWhitelist, ",")
@@ -115,7 +120,7 @@ func (self *dockerFactory) NewContainerHandler(name string, inHostNamespace bool
 		self.dockerVersion,
 		self.ignoreMetrics,
 	)
-	return
+	return handler, !accept, err
 }
 
 // Returns the Docker ID from the full container name.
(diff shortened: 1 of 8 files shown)


sjpotter:
@vishh @timstclair

container/factory.go
```diff
@@ -89,30 +86,27 @@ func HasFactories() bool {
...
-			if !canAccept {
-				glog.V(3).Infof("Factory %q can handle container %q, but ignoring.", factory, name)
-				return nil, false, nil
-			}
-			glog.V(3).Infof("Using factory %q for container %q", factory, name)
-			handle, err := factory.NewContainerHandler(name, inHostNamespace)
-			return handle, canAccept, err
+		if handler != nil {
+			return handler, err
+		} else if ignore {
```
vishh: Should we check for `ignore` before checking `if handler != nil` ?

vishh:
Just one comment. Otherwise LGTM. Thanks for the cleanup @sjpotter

container/factory.go
sjpotter: no, if we want to ignore this container then should return a nil handler.  ignore is only if we return a nil handler to tell the factory if it should ignore this container or continue checking the other factories

i.e. ignore is the equivalent of handle = true, accept = false in the previous code.

vishh: Ok. As an API consumer, the following seems easier to understand to me:

```go
if ignore {
 return nil
}
if handler == nil {
  return err
}
```

container/docker/factory.go
```diff
@@ -134,7 +139,7 @@ func isContainerName(name string) bool {
 }
 
 // Docker handles all containers under /docker
-func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
+func (self *dockerFactory) canHandleAndAccept(name string) (bool, bool, error) {
 	// docker factory accepts all containers it can handle.
 	canAccept := true
```
timstclair: (nit) const

container/factory.go
```diff
@@ -89,30 +86,27 @@ func HasFactories() bool {
...
+			glog.V(3).Infof("Factory %q can handle container %q, but ignoring.", factory, name)
+			return nil, nil
 		} else {
 			glog.V(4).Infof("Factory %q was unable to handle container %q", factory, name)
```
timstclair: I think we should log the error here.

timstclair: nvm, I see we log it above.

container/factory.go
```diff
@@ -89,30 +86,27 @@ func HasFactories() bool {
...
 // Create a new ContainerHandler for the specified container.
-func NewContainerHandler(name string, inHostNamespace bool) (ContainerHandler, bool, error) {
+func NewContainerHandler(name string, inHostNamespace bool) (ContainerHandler, error) {
 	factoriesLock.RLock()
 	defer factoriesLock.RUnlock()
 
 	// Create the ContainerHandler with the first factory that supports it.
 	for _, factory := range factories {
-		canHandle, canAccept, err := factory.CanHandleAndAccept(name)
+		handler, ignore, err := factory.NewContainerHandler(name, inHostNamespace)
```
timstclair: Consider: I like to avoid methods with more than 2 return parameters. What if we change the semantics to return if the factory returns an error? Then we can create an "IgnoreError", and simplify the logic here to just three cases: Return if `err != nil || handler != nil`, and continue if both are nil. You'd also need to update the factory methods to not return errors in the "expected" cases (e.g. in docker don't return an error when inspect fails) - which I think makes more sense anyway...

vishh: +1

sjpotter: but doesn't this prevent NewContainerHandler from returning other errors that might occur?

sjpotter: perhaps a better way is to return an IgnoreHandler instead of nil?

sjpotter:
this updates me above table

|old CanHandleAndAccept ||new NewContainerHandler|
| --- | --- | --- | 
|**handle**|**accept**|**handler**|
| true | false | ignoreHandler |
| false | true | shouldn't happen |
| false | false | nil |
| true | true | !ignoreHandler |

container/factory.go
sjpotter: see my current update

timstclair: I'm worried that the ignore handler  will unnecessary complicate things, especially if the results of some of those functions are assumed to be non-nill if the error is nil.

> but doesn't this prevent NewContainerHandler from returning other errors that might occur?

It means we'd change the behaviour of an error, and say that errors are returned rather than logged & ignored (in NewContainerHandler). If the error is bubbled up to manager the behaviour is still consistent ([manager.go#L774](https://github.com/google/cadvisor/blob/5d7c71a49536d56729c7c778daa70653c138f4a5/manager/manager.go#L774)). Tracing up through the code, I think this is actually preferable. For example, consider the case of creating the root container ([manager.go#L261](https://github.com/google/cadvisor/blob/5d7c71a49536d56729c7c778daa70653c138f4a5/manager/manager.go#L261)): ignoring the container in that case _should_ be an actual error, and not silently ignored. Everywhere else it will simply be logged (e.g. "E: Failed to create existing container: foo: ignoring container foo").

Issue #1185, opened 2016-03-31
[docker v1.11.0-rc2] container info missing metadata


timstclair:
Comparing the output of the stats query:
```
curl -X "POST" -d '{"containerName":"/","subcontainers":true,"num_stats":1}' $(k8s-node-proxy)/stats/container
```

From a cluster running on [docker-v1.11.0-rc2](https://gist.github.com/timstclair/78eddf7c36698fcefcd5ab0d99bb2e7a) vs [docker-v1.9.1](https://gist.github.com/timstclair/7cfbb357bc151803c3dd1c991bc54ca4), the docker-v1.11.0-rc2 response is missing
- id
- aliases
- namespace
- labels

xref https://github.com/kubernetes/kubernetes/issues/23397

timstclair:
Looks like the docker inspect API changed in https://github.com/docker/docker/commit/68fb7f4b744bf71206898d32fe203556a6261e5d, we might just need to update Godeps to the latest docker client (to include https://github.com/fsouza/go-dockerclient/pull/107)

vishh:
Docker does not support exec driver anymore. So cAdvisor needs to stop
checking for the `native` exec driver.

timstclair:
Fixed via https://github.com/google/cadvisor/issues/1190

PR #1190, opened 2016-04-05
Bump go-dockerclient godep version

timstclair:
Required for compatibility with docker v1.11 (see https://github.com/google/cadvisor/issues/1185)


container/docker/factory.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
 
-	"github.com/fsouza/go-dockerclient"
+	docker "github.com/fsouza/go-dockerclient"
```
vishh: this should not be necessary. I suspect that your goimports binary is causing trouble.

validate/validate.go
```diff
@@ -185,28 +185,24 @@ func validateCgroups() (string, string) {
...
+			desc += fmt.Sprintf("\tDocker container state directory %q is not accessible.\n", stateFile)
+			return Unsupported, desc
 		}
+		desc += fmt.Sprintf("\tDocker container state directory is at %q and is accessible.\n", stateFile)
+		return Recommended, desc
+	} else if strings.Contains(info.ExecutionDriver, "lxc") {
+		return Supported, desc
 	}
-	return Unknown, "Docker remote API not reachable\n\t"
+	return Unknown, desc
```
vishh: FYI: With docker v1.11, we should ignore the exec driver option all together.

vishh:
LGTM

container/docker/factory.go
timstclair: It's not necessary, but I think it makes it easier to parse the code (and figure out what `docker` refers to). We do it in some other places, but I can revert it if you prefer.

validate/validate.go
timstclair: ACK. I'm planning on sending a follow up for this.

container/docker/factory.go
vishh: I like it too for the record. Its just that we don't enforce it everywhere.

PR #1192, opened 2016-04-05
Fix cAdivsor docker validation

timstclair:
- Synchronize `Validate` policy with docker `factory.go` checks
- Don't enforce `ExecutionDriver == "native"` for docker >= v1.11

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -233,7 +233,6 @@ func (self *manager) Start() error {
 		glog.Errorf("Registration of the raw container factory failed: %v", err)
 	}
 
-	// FIXME - delete?
 	self.DockerInfo()
 	self.DockerImages()
 
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -163,6 +163,7 @@ var (
 	version_re            = regexp.MustCompile(version_regexp_string)
 )
 
+// TODO: switch to a semantic versioning library.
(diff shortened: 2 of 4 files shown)


timstclair:
@k8s-bot test this

manager/manager.go
```diff
@@ -233,7 +233,6 @@ func (self *manager) Start() error {
 		glog.Errorf("Registration of the raw container factory failed: %v", err)
 	}
 
-	// FIXME - delete?
```
vishh: why remove this?

vishh:
A couple of comments, otherwise LGTM

timstclair:
@k8s-bot test this

manager/manager.go
timstclair: I didn't mean to include it in https://github.com/google/cadvisor/pull/1190 so I removed it here.

timstclair:
@k8s-bot test this

timstclair:
Finally got tests passing - the issue is that `ServerVersion` doesn't get set in the info for docker < 1.9. I fixed it by falling back to the version API, and filling it in from that.

@vishh PTAL

timstclair:
Dropped https://github.com/google/cadvisor/pull/1192/commits/f9377267a833048649e62d5933aa24ac8627697a which had snuck in here.

PR #1196, opened 2016-04-06
Fix --disable_metrics parsing to accept empty string as input.

### LINKED ISSUE #1193: Can't clear the -disable_metrics flag
I can't find a way to clear the value of the `-disable_metrics` flag in order to reenable TCP connection stats.

```
$ ./cadvisor -disable_metrics=
invalid value "" for flag -disable_metrics: unsupported metric "" specified in disable_metrics
...
```

This appears to be because [the code that parses the string](https://github.com/google/cadvisor/blob/de1bd289cbeaff5a0d88df66086adafc0a54288f/cadvisor.go#L78-L87) is under the mistaken impression that `strings.Split(value, ",")` returns an empty slice when the input string is empty, but it actually returns `[]string{""}`.

I can certainly provide a PR for this but it'll take some time as I don't have a CLA set up.

vishh:
Fixes #1193

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -76,6 +76,10 @@ func (ml *metricSetValue) String() string {
 }
 
 func (ml *metricSetValue) Set(value string) error {
+	ignoreMetrics = metricSetValue{}
+	if value == "" {
+		return nil
+	}
 	for _, metric := range strings.Split(value, ",") {
 		if ignoreWhitelist.Has(container.MetricKind(metric)) {
 			(*ml).Add(container.MetricKind(metric))
--- a/cadvisor_test.go
+++ b/cadvisor_test.go
@@ -27,3 +27,9 @@ func TestTcpMetricsAreDisabledByDefault(t *testing.T) {
 	flag.Parse()
 	assert.True(t, ignoreMetrics.Has(container.NetworkTcpUsageMetrics))
 }
+
+func TestTcpMetricsAreEnabledOnDemand(t *testing.T) {
+	assert.True(t, ignoreMetrics.Has(container.NetworkTcpUsageMetrics))
+	ignoreMetrics.Set("")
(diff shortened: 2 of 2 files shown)


timstclair:
LGTM

cadvisor.go
```diff
@@ -76,6 +76,10 @@ func (ml *metricSetValue) String() string {
 }
 
 func (ml *metricSetValue) Set(value string) error {
+	ignoreMetrics = metricSetValue{}
```
sprohaska: I think this might cause a panic if `--disable_metrics` is not empy:

```
    goroutine 1 [running]:
    panic(0xb0c8c0, 0xc82020cdd0)
    	/usr/local/go/src/runtime/panic.go:481 +0x3e6
    main.(*metricSetValue).Set(0x15ac528, 0x7fffe9354f34, 0x4, 0x0, 0x0)
    	/go/src/github.com/google/cadvisor/cadvisor.go:85 +0x1da
    flag.(*FlagSet).parseOne(0xc820066060, 0xc820118801, 0x0, 0x0)
    	/usr/local/go/src/flag/flag.go:881 +0xdd9
    flag.(*FlagSet).Parse(0xc820066060, 0xc820064110, 0x3, 0x3, 0x0, 0x0)
    	/usr/local/go/src/flag/flag.go:900 +0x6e
    flag.Parse()
    	/usr/local/go/src/flag/flag.go:928 +0x6f
    main.main()
    	/go/src/github.com/google/cadvisor/cadvisor.go:99 +0x68
    panic: assignment to entry in nil map
```

I'm unsure about the details. I'm not very familiar with Go.

timstclair: Indeed! I'll send a fix, thanks for catching this.

timstclair: Fixed in https://github.com/google/cadvisor/pull/1259

PR #1200, opened 2016-04-08
Improve local cAdvisor integration test script

timstclair:
- Poll localhost:8080/healthz to wait for cAdvisor to start
- Set the script exit code to the test exit status
- Fix bash formatting
- Check for sudo access at the start

/cc @vishh


vishh:
LGTM

PR #1188, opened 2016-04-04
Use a dedicated CpuLoadReader per container

krallin:
This ensures each goroutine is given its own Netlink connection, and presumably avoids having a message destined for one goroutine read by another goroutine. 

This does means cAdvisor will use a larger number of connections (one per container being tracked), but I don't see any locking or serialization taking place in `cpuload`, so that seems like the lesser of two evils (the other evil being racy behavior).

I'm going to be running this patch on our end for a little while, but I'd be curious to know if you have feedback to provide so far! 

I presume this might address the issue I've reported in https://github.com/google/cadvisor/issues/1182 and https://github.com/google/cadvisor/issues/452, as well as another issue we've been experiencing here at Aptible (but haven't reported yet), where a container housekeeping goroutine gets stuck waiting on input from Netlink that never comes.

PR #1205, opened 2016-04-13
Netlink: aggressively close cgroup fds

krallin:
This is a very simple patch, but I've run into the issue described in the commit message when running the load tests described in #1188. This is, however, not a testing-only problem, as I've needed this patch to roll out a shorter collection interval for load probes in production.

The refactoring (of `getLoadStats`) is here to make it clearer that `getLoadStats` is effectively borrowing the open file and not just a number (i.e. it's important  that the file stays open while `getLoadStats` runs, otherwise we'll use an invalid FD, or even FD that refers to another cgroup..!). There isn't a great way to check that the FD is actually valid (that I know of), though. 

I'm happy to provide a much simpler version of that patch (https://github.com/aptible/cadvisor-factory/blob/master/patches/0009-Netlink-aggressively-close-cgroup-fds.patch is what I'm actually using at this time) if you don't think the refactoring is a real gain in clarity. 

---

Commit message follows:

Not closing the FDs manually means we have to rely on garbage collection
to run before cgroup FDs are closed. If the system is running a lot of
load probes at a high-frequency (i.e. dynamic housekeeping isn't backing
off because of load variations), we can end up hitting our FD limit due
to keeping around lots of (useless) FDs.

### DIFF
--- a/utils/cpuload/netlink/netlink.go
+++ b/utils/cpuload/netlink/netlink.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"os"
 	"syscall"
 
 	info "github.com/google/cadvisor/info/v1"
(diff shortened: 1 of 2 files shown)


timstclair:
ok to test

timstclair:
LGTM. Thanks for the patch! I'll merge it once the tests pass.

krallin:
Awesome; thanks @timstclair !

vishh:
ok to test

Issue #1142, opened 2016-02-29
No common performance graph for all containers


AndreLouisCaron:
I'm using cadvisor to monitor Docker containers for multiple application databases (MySQL, Redis, Mongo, Memcached) for a complex application.

I can see the memory & CPU usage for each database container individually, and I can also see the global memory & CPU usage per CPU core.

However, one thing I'm often interested in is seeing which container is stressed more heavily than the others.  There is currently no graph to plot this information, but it seems to me like all the information is available.

Any chance you can add this graph?

Thanks!

vishh:
If I parse correctly, you want a global view that is sorted based on CPU or
Memory usage. This does sound really useful. I don't think the core
maintainers have any spare cycles to work on this feature though. I'm
adding a `help-wanted` label and we will gladly accept PRs.

AndreLouisCaron:
I don't necessarily want the containers to be sorted.

I would mostly like a timeline graph showing CPU usage by container.  The home page shows this type of graph, but the usage is by CPU core (which only helps determine whether the docker daemon is distributing containers equally across all cores).

There is a detailed view by container where you can see the CPU usage for a single container, so I guess the information is already available and that we only need to combine the different "feeds" into a single graph on the home page.

I might consider putting some time into this, but I don't know anything about Go...

vishh:
Got it. Most of the UI elements are in javascript. The templates are in Go
though. Feel free to poke us if you run into any issues.

AndreLouisCaron:
Did a quick look into this today.  If I understand this correctly, `pages/containers_html.go` contains the page markup and `pages/static/containers_js.go` contains the JS code that powers it.  I don't see any tests for these files, are they stored elsewhere?

Also, sorry if this is a newbie question (I'm more fluent in Python), but why is the all the HTML and JS code in strings in Go files and how do you get syntax highlighting in your editor for this?

vishh:
There are no tests :(

Syntax highlighting is not possible. I usually copy the JS contents into a
separate file and then move it to its original location.

timstclair:
> I usually copy the JS contents into a
separate file and then move it to its original location.

We should consider doing this programmatically in the build process. I wonder if we could use `-ldflags`, or if the length of the data would be a problem...

PR #1207, opened 2016-04-13
Separated asset files into their own folder

ainsleyc:
As suggested in this issue:
https://github.com/google/cadvisor/issues/1142

I've separated out the js and css files into their own folder, and added a script to build them into the needed .go file using go-bindata. Should make them easier to work on.

Feel free to let me know if you don't think this is the right approach.


ainsleyc:
Just signed it.

Issue #1206, opened 2016-04-13
[docker v1.11.0] e2e tests fail


timstclair:
Lots of E2E tests are failing on docker v1.11(.0-rc5) with the same error:

```
E0413 01:52:15.500838   29558 framework.go:335] About to run - [sudo docker run -d kubernetes/pause]
E0413 01:52:15.819817   29558 framework.go:335] About to run - [sudo docker run -d kubernetes/pause]
E0413 01:52:20.971787   29558 framework.go:335] About to run - [sudo docker rm -f 9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830]
E0413 01:52:21.053403   29558 framework.go:335] About to run - [sudo docker rm -f c94b3273e8396604973da59e9561b56bc27d9f072639626f86a0c9b98035c63d]
--- FAIL: TestGetAllDockerContainers (5.64s)
	Location:	docker_test.go:63
	Error:		No error is expected but got request "http://localhost:8080/api/v1.3/docker/9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830" failed with error: "failed to get Docker container \"9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830\" with error: unable to find data for container /docker/9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830"
	Messages:	Timed out waiting for container "9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830" to be available in cAdvisor: request "http://localhost:8080/api/v1.3/docker/9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830" failed with error: "failed to get Docker container \"9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830\" with error: unable to find data for container /docker/9d23af2d53f21a55ecb9165622ac312f2ffd315d1784bf178d88af81a61b5830"
```

yujuhong:
If it helps, the error messages:
```
W0413 18:03:30.824396    3398 container.go:341] Failed to create summary reader for "/docker/f99bd2d7d876cd07743cbec1ef0f3204fc398f112d57442f4cbd7bcebcacd02d": none of the resources are being tracked.
[...]

PR #1208, opened 2016-04-13
Look for container state in containerd path

timstclair:
Fixes https://github.com/google/cadvisor/issues/1206

/cc @yujuhong

### DIFF
--- a/container/libcontainer/compatibility.go
+++ b/container/libcontainer/compatibility.go
@@ -178,8 +178,12 @@ type v1State struct {
 	Config v1Config `json:"config"`
 }
 
-// Relative path to the libcontainer execdriver directory.
-const libcontainerExecDriverPath = "execdriver/native"
+const (
+	// Relative path to the libcontainer execdriver directory.
+	libcontainerExecDriverPath = "execdriver/native"
+	// Absolute path of the containerd directory.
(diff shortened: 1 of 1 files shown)


container/libcontainer/compatibility.go
```diff
@@ -178,8 +178,12 @@ type v1State struct {
 	Config v1Config `json:"config"`
 }
 
-// Relative path to the libcontainer execdriver directory.
-const libcontainerExecDriverPath = "execdriver/native"
+const (
+	// Relative path to the libcontainer execdriver directory.
+	libcontainerExecDriverPath = "execdriver/native"
+	// Absolute path of the containerd directory.
+	containerdPath = "/run/containerd"
```
vishh: For the future, we should see if we can get all the information we need via containerd APIs.

vishh:
LGTM

container/libcontainer/compatibility.go
yujuhong: What information do we need from the config? Just curious. I can check the code of course :)

BTW, I think we can check the version of docker to decide where to look for the config, cant' we?

timstclair: Yeah, I'll try to clean this up.

PR #1213, opened 2016-04-15
Read docker container spec from cgroupfs, rather than libcontainer spec

timstclair:
With this change the cAdvisor integration tests pass on docker v1.11 (on ubuntu on GCE).

Also, refactor common container GetSpec (cc @sjpotter)
- Pull out the root container cases, since they're only relevant in the
  raw container handler
- Pass parameters rather than depending on AbstractContainerInterface


container/libcontainer/compatibility.go
```diff
@@ -15,354 +15,25 @@
...
-	err = json.Unmarshal(out, &state)
-	if err != nil {
-		return preAPIState{}, err
-	}
-
-	return state, nil
-}
-
 // Gets the path to the libcontainer configuration.
 func configPath(dockerRun, containerID string) string {
```
vishh: Why do we care about the state anymore?

timstclair: It's used by the [`Exists()`](https://github.com/google/cadvisor/blob/a8edca0c4672434537e0c9f8b3fbe126fe7ed872/container/docker/handler.go#L401) function to determine whether the container still exists. We might be able to switch to the same method that raw & rkt use:
```go
func (self *rawContainerHandler) Exists() bool {
	// If any cgroup exists, the container is still alive.
	for _, cgroupPath := range self.cgroupPaths {
		if utils.FileExists(cgroupPath) {
			return true
		}
	}
	return false
}
```
But I'm not sure what the full implications of that are...

timstclair: Thought about this some more, decided to go with it.

timstclair:
@yujuhong - could you take a look at this since Vish is out today?

yujuhong:
> @yujuhong - could you take a look at this since Vish is out today?

Sure.

yujuhong:
LGTM with less runtime-specific logic and more reusable code :-)

PR #1217, opened 2016-04-15
Separated asset files into their own folder

ainsleyc:
As suggested in this issue:
#1142

I've separated out the js and css files into their own folder, and added a script to build them into the needed .go file using go-bindata. Should make them easier to work on.

Feel free to let me know if you don't think this is the right approach.


ainsleyc:
Sorry for multiple requests, was trying to get the CLA check to work correctly.

Makefile
```diff
@@ -33,6 +33,7 @@ vet:
 
 build:
 	@echo ">> building binaries"
+	@./build/assets.sh
```
timstclair: How long does this take? If it's slow, we might want to make it a separate rule, since UI changes are relatively infrequent.

timstclair: Alternatively, you could update the script to only run if the modified time on source files is more recent than the output files.

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+    'oddTableRow': 'table-row'
+  };
+  var opts = {
+    alternatingRowStyle: true,
+    page: 'enable',
+    pageSize: numPages,
+    allowHtml: true,
+    sortColumn: sortIndex,
+    sortAscending: false,
+    cssClassNames: cssClassNames,
```
timstclair: nit: delete trailing comma

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  // animation.
+  var opts = {
+    curveType: 'function',
+    height: 300,
+    legend: {position: 'none'},
+    focusTarget: 'category',
+    vAxis: {
+      title: unit,
+      viewWindow: {
+        min: minWindow,
```
timstclair: delete trailing comma, and below

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+      },
+    },
+    legend: {
+      position: 'bottom',
+    },
+  };
+  // If the whole data series has the same value, try to center it in the chart.
+  if (min == max) {
+    opts.vAxis.viewWindow.max = 1.1 * max
```
timstclair: nit: missing semicolon, and below

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+    elements.push({v: processInfo[i].rss, f: humanizeIEC(processInfo[i].rss)});
+    elements.push({
+      v: processInfo[i].virtual_size,
+      f: humanizeIEC(processInfo[i].virtual_size)
+    });
+    elements.push(processInfo[i].status);
+    elements.push(processInfo[i].running_time);
+    elements.push(processInfo[i].cmd);
+    if (isRoot) {
+      var cgroup = processInfo[i].cgroup_path
```
timstclair: nit: missing semicolon

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  setTimeout(function() { stepExecute(steps); }, 0);
+}
+
+// Draw all the charts on the page.
+function drawCharts(machineInfo, containerInfo) {
+  var steps = [];
+
+  if (containerInfo.spec.has_cpu || containerInfo.spec.has_memory) {
+    steps.push(function() {
+      drawOverallUsage('usage-gauge', machineInfo, containerInfo)
```
timstclair: semicolon

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  }
+
+  // Custom Metrics
+  if (containerInfo.spec.has_custom_metrics) {
+    steps.push(function() {
+      getCustomMetrics(
+          window.cadvisor.rootDir, window.cadvisor.containerName,
+          function(metricsInfo) {
+            drawCustomMetrics(
+                'custom-metrics-chart', containerInfo, metricsInfo)
```
timstclair: semicolon

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+    if (window.cadvisor.metricLabelPair[i][0] == metricName) return i;
+  }
+  return -1;
+}
+
+function setLabel(metric, label) {
+  $('#' + metric + '-selection-text')
+      .empty()
+      .append($('<span>').text('Label: '))
+      .append($('<b>').text(label))
```
timstclair: `;`

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  return window.cadvisor.metricLabelPair[index][1];
+}
+
+function startCustomMetrics(elementId, containerInfo) {
+  var metricSpec = containerInfo.spec.custom_metrics;
+  var metricStats = containerInfo.stats.custom_metrics;
+  var el = $('<div>');
+
+  if (metricSpec.length < window.cadvisor.maxCustomMetrics)
+    window.cadvisor.maxCustomMetrics = metricSpec.length
```
timstclair: `;`

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  window.cadvisor.rootDir = rootDir;
+  window.cadvisor.containerName = containerName;
+
+  window.cadvisor.firstCustomCollection = true;
+  window.cadvisor.metricLabelPair = [];
+  window.cadvisor.maxCustomMetrics = 10;
+
+  // Draw process information at start and refresh every 60s.
+  getProcessInfo(rootDir, containerName, function(processInfo) {
+    drawProcesses(isRoot, rootDir, processInfo)
```
timstclair: `;`

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
...
+  });
+  setInterval(function() {
+    getProcessInfo(rootDir, containerName, function(processInfo) {
+      drawProcesses(isRoot, rootDir, processInfo)
```
timstclair: `;`

assets/js/containers.js
```diff
@@ -0,0 +1,914 @@
+
```
timstclair: Could you make all the changes to this file in a separate commit so that git sees this as a move and retains the file history. I added some comments for js problems, feel free to ignore them and I can do a follow up fix if you'd like.

timstclair:
Thanks for the nice cleanup! Can we nest the asset directory under `/pages/static` or just `/pages`? I'd like to keep all the UI code together.

ainsleyc:
Sure, thanks for the feedback! I'll make the suggested changes and resubmit.

ainsleyc:
I think that covers everything. I tried to squash some of the commits together, but it causes git to forget the containers.js history.

Feel free to let me know if there are any other changes you'd like. Thanks!

build/assets.sh
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+GIT_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+for f in $GIT_ROOT/pages/assets/**/*
+do
```
timstclair: nit: our bash style is to put do / then on the previous line with a semicolon (e.g. `for f in $FOO; do`).
Same for `if` below.

build/assets.sh
```diff
@@ -0,0 +1,27 @@
...
+  if [ "$f" -nt $GIT_ROOT/pages/static/assets.go ]
+  then
+    go get -u github.com/jteeuwen/go-bindata/...
+    go-bindata -o $GIT_ROOT/pages/static/assets.go -pkg static $GIT_ROOT/pages/assets/...
+    exit 0
```
timstclair: (nit) `exit $?`

timstclair:
Thanks for the fixes. I think you should be able to squash to 2 commits: The first separates the js from the go, but doesn't modify the code, the second commit cleans up the js & lint errors. If this turns out to be too complicated, don't waste too much time on it.

ainsleyc:
Cool, I didn't know you could manipulate git history to this degree. Thanks for the tip.

timstclair:
LGTM. I assume you've manually verified that the pages still work?

ainsleyc:
Yup, I've clicked through and believe that everything is loading correctly without errors in the browser console.

timstclair:
@vishh please add ok to test

timstclair:
ok to test

build/assets.sh
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+GIT_ROOT=$(dirname "${BASH_SOURCE}")/..
+
+for f in $GIT_ROOT/pages/assets/**/*; do
+  if [ "$f" -nt $GIT_ROOT/pages/static/assets.go ]; then
+    go get -u github.com/jteeuwen/go-bindata/...
```
timstclair: I think you need to add `&& go install github.com/jteeuwen/go-bindata` for this to be used.

timstclair:
```
+ ./build/check_gofmt.sh .
pages/static/assets.go pages/static/assets.go
```

I'm surprised this is necessary, but it looks like you need to add `gofmt` to your assets.sh script.

build/assets.sh
ainsleyc: Maybe I am misunderstanding how "go get" works, or the build environment is different somehow?

When I delete both the go-bindata src and executable in the bin from my $GOPATH and run the "go get" command, the script is working for me.

ainsleyc:
Ah. So how would you like to handle the boilerplate header for the generated file? I guess the options are to prepend it in the asset script, or omit that file from the boilerplate check script.

timstclair:
Oh yeah, the boilerplate should be prefixed to the generated file. As an example, see https://github.com/kubernetes/kubernetes/blob/master/hack/build-ui.sh. Feel free to copy a lot of that script into yours.

PR #1218, opened 2016-04-15
Dont block on grpc if nothing is listeing on rkt api service port

sjpotter:
Attempts an initial connection to the rkt api service via the golang net api.  Only try grpc if that succeeds.

### DIFF
--- a/container/rkt/client.go
+++ b/container/rkt/client.go
@@ -38,17 +38,19 @@ var (
 
 func Client() (rktapi.PublicAPIClient, error) {
 	once.Do(func() {
-		_, err := net.Dial("tcp", "localhost:15441")
+		conn, err := net.DialTimeout("tcp", defaultRktAPIServiceAddr, 2*time.Second)
 		if err != nil {
 			rktClient = nil
-			rktClientErr = fmt.Errorf("rkt: cannot connect to rkt api service: %v", err)
+			rktClientErr = fmt.Errorf("rkt: cannot tcp Dial rkt api service: %v", err)
 			return
 		}
+
+		conn.Close()
 
 		apisvcConn, err := grpc.Dial(defaultRktAPIServiceAddr, grpc.WithInsecure(), grpc.WithTimeout(5*time.Second))
 		if err != nil {
 			rktClient = nil
-			rktClientErr = fmt.Errorf("rkt: cannot connect to rkt api service: %v", err)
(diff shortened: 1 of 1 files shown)


container/rkt/client.go
```diff
@@ -37,6 +38,13 @@ var (
 
 func Client() (rktapi.PublicAPIClient, error) {
 	once.Do(func() {
+		_, err := net.Dial("tcp", "localhost:15441")
```
timstclair: Use `defaultRktAPIServiceAddr` here

timstclair: I think you should use `DialTimeout` with a reasonable value

timstclair: The connection should be closed

container/rkt/client.go
```diff
@@ -37,6 +38,13 @@ var (
...
+		if err != nil {
+			rktClient = nil
+			rktClientErr = fmt.Errorf("rkt: cannot connect to rkt api service: %v", err)
+			return
```
timstclair: Just do `return nil, fmt.Errorf...` instead

container/rkt/client.go
```diff
@@ -37,6 +38,13 @@ var (
 
 func Client() (rktapi.PublicAPIClient, error) {
 	once.Do(func() {
+		_, err := net.Dial("tcp", "localhost:15441")
+		if err != nil {
+			rktClient = nil
+			rktClientErr = fmt.Errorf("rkt: cannot connect to rkt api service: %v", err)
```
timstclair: Use a different error message here so the failures can be distinguished. Maybe "Unable to dial rkt api at localhost:15441"

timstclair:
Thanks for fixing this!

container/rkt/client.go
```diff
@@ -38,17 +38,19 @@ var (
 
 func Client() (rktapi.PublicAPIClient, error) {
 	once.Do(func() {
-		_, err := net.Dial("tcp", "localhost:15441")
+		conn, err := net.DialTimeout("tcp", defaultRktAPIServiceAddr, 2*time.Second)
```
timstclair: Any reason you chose a different timeout from the one below? If not, can you create a timeout const to use in both places instead.

timstclair:
Looks like you need to run `gofmt`, then LGTM.

PR #1220, opened 2016-04-15
Fix docker GetSpec to include image, labels, and env vars

timstclair:
Regressed in https://github.com/google/cadvisor/pull/1213

Added an integration test to catch this in the future.

Thanks @sjpotter for spotting this!


vishh:
LGTM

PR #1221, opened 2016-04-15
Cleanup manager package

timstclair:
- Refactor manager/machine.go -> utils/machine/info.go
- Refactor docker-specific functions from manager to docker
- Delete unused ManagerMock


timstclair:
Rebased.

timstclair:
@k8s-bot test this

manager/manager.go
```diff
@@ -110,10 +112,10 @@ type Manager interface {
 	CloseEventChannel(watch_id int)
 
 	// Get status information about docker.
-	DockerInfo() (DockerStatus, error)
+	DockerInfo() (docker.DockerStatus, error)
```
vishh: Given that kubelet consumes these methods, should the API types be kept in a separate package?

utils/machine/info.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
```
vishh: nit: Why push this to `utils` package? Why not a top level `machine` package?

vishh:
Just two comments. Otherwise LGTM

manager/manager.go
timstclair: Done. I moved it to `info/v1` for now, since that is our "internal" version at the moment. Once I get time to cleanup the versions properly, I'll add it to v2 as well.

utils/machine/info.go
timstclair: Because it already existed :)  I moved the whole `utils/machine` package to a top-level `machine` package (we should reconsider where this package should live when https://github.com/google/cadvisor/issues/1250 is implemented)

timstclair:
Addressed comments. Thanks!

vishh:
lgtm

PR #1222, opened 2016-04-15
RFC: Use rkt api service to verify container

sjpotter:
This doesn't parse the file system but asks the rkt api service to determine if its responsible for a specific cgroup.

However, there is a slight race when systemd is used to create the initial cgroup so the "pod" cgroup can be missed.  With that said, this seems good enough for k8s hpa as those are focused on the app containers which work fine

### DIFF
--- a/container/rkt/client.go
+++ b/container/rkt/client.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	rktapi "github.com/coreos/rkt/api/v1alpha"
 
 	"golang.org/x/net/context"
@@ -27,6 +28,7 @@ import (
 )
 
 const (
+	minimumRktBinVersion     = "1.4.0"
 	defaultRktAPIServiceAddr = "localhost:15441"
 	timeout                  = 2 * time.Second
 )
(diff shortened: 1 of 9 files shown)


sjpotter:
also includes my other PR of the rkt bump

sjpotter:
@timstclair

sjpotter:
refreshed PR as one commit was accepted

container/rkt/client.go
```diff
@@ -37,25 +39,83 @@ var (
 	once         sync.Once
 )
 
+type rktVersion struct {
+	*semver.Version
+}
+
+func (r rktVersion) Compare(other string) (int, error) {
```
timstclair: (optional) This is already implemented by "github.com/blang/semver".Version.Compare

container/rkt/client.go
```diff
@@ -37,25 +39,83 @@ var (
...
 		conn, err := net.DialTimeout("tcp", defaultRktAPIServiceAddr, timeout)
 		if err != nil {
 			rktClient = nil
 			rktClientErr = fmt.Errorf("rkt: cannot tcp Dial rkt api service: %v", err)
 			return
 		}
-
 		conn.Close()
 
+		rktClient = nil
```
timstclair: unnecessary?

timstclair:
I like the idea, but I'm nervous about the potential race condition. Can you elaborate more on how that can happen? Is it possible to fix it within the rkt API?

container/rkt/client.go
sjpotter: was just mirroring what is done for rkt in k8s

container/rkt/client.go
sjpotter: is it bad unnecessary?  I was just trying to be clear with the code.

sjpotter:
So as of now it can't be easily fixed.  the api service doesn't launch rkt pods, it just monitors the state, so its possible for cgroups to be created (such as when they are created by systemd) before rkt has done anything and hence nothing for hte api-service to return.  

People seem to believe that once coreos ship systemd 229 (and hence will be part of the stage1) that we will be able to handle this better, but healthy skepticism till I see it.

I'd note that the old code works for cases where a user did a "rkt run" on the command line, but doesn't work when rkt is called from a systemd unit at all (ala k8s usage)

sjpotter:
As of now, think this is going to be deprecated in favor of my other PR, just not closing it yet

sjpotter:
other approach was used

PR #1223, opened 2016-04-15
Move mocks to testing package to remove +build tags

timstclair:
Some go tools (e.g. godef, gorename) don't handle +build tags well, so
I refactored some packages to remove the test tags from cAdvisor.


vishh:
LGTM

Issue #1175, opened 2016-03-22
Eliminate flag globals


timstclair:
cAdvisor flags are added to the go flag set at package init. This means that binaries (such as Kubelet) which include cAdvisor as a library cannot control which flags are added. It is also difficult to override for those packages to override the default values.

xRef: kubernetes/kubernetes#19432

timstclair:
Proposed solution:
- Create a configuration package which holds a cAdvisor configuration struct.
- Library integration points (e.g. manager) should accept a configuration object instead of relying on global variables.
- Provide a DefaultConfiguration global with the same defaults used by the standalone cAdvisor binary.
- Provide a mechanism for adding flags for (parts) of the configuration object.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #1224, opened 2016-04-16
RFC: Refactor cAdvisor flags to config package

timstclair:
For https://github.com/google/cadvisor/issues/1175

In an effort to clean up cAdvisor configuration, specifically for cAdvisor as a library, I'm proposing moving all flags into a method in the config package, and moving the config to a global config object. Eventually, I would like to make the config more granular, and pass it through as a parameter rather than global state, but this is a more modest intermediate step.

This PR still needs to be cleaned up before it is thoroughly reviewed, but I'm looking for comments on the design before I invest that time.

/cc @vishh


cadvisor.go
```diff
@@ -35,25 +35,8 @@ import (
 	"github.com/golang/glog"
 )
 
-var argIp = flag.String("listen_ip", "", "IP to listen on, defaults to all IPs")
-var argPort = flag.Int("port", 8080, "port to listen")
-var maxProcs = flag.Int("max_procs", 0, "max number of CPUs that can be used simultaneously. Less than 1 for default (number of cores).")
-
-var argDbDriver = flag.String("storage_driver", "", "storage driver to use. Data is always cached shortly in memory, this controls where data is pushed besides the local cache. Empty means none. Options are: <empty> (default), bigquery, influxdb, and kafka")
 var versionFlag = flag.Bool("version", false, "print cAdvisor version and exit")
```
grobie: This should be removed as well, right?

grobie:
LGTM. Still a bit too much global state for my taste, but not writing into the global flag state anymore is a huge win already.

cadvisor.go
timstclair: I left this one since it's not configuration. If this flag is set, cAdvisor just prints the version to stdout and exits immediately.

timstclair:
Thanks @grobie . I agree about the global state, and don't consider this work finished, but rather it's a step in the right direction.

timstclair:
+cc @mtaufen since this is similar to the kubelet config work you just did

cadvisor.go
grobie: But it will collide with version flags set by other programs, no? I'd move that into main()

timstclair: This is in the `main` package, AFAIK it can't be imported into another program.

grobie: You're right of course, my bad.

mtaufen:
A few initial thoughts :)
- I am a big fan of this kind of cleanup work. Thank you.
- I second previous comments wrt global state. But this is definitely a step in the right direction.
- It looks like there's some initialization of components going on in `main`, are these things that will have to move to a separate e.g. `Init` function to make this a library? If you have an `Init` code path, you'll be able to pass config into that and from there to sub-components and can probably get rid of the global state. 
- It might be worth adding an explicit validation stage for the config before any sub-components start running.

timstclair:
Thanks for the comments!

> It looks like there's some initialization of components going on in main, are these things that will have to move to a separate e.g. Init function to make this a library?

We already use cAdvisor in "library form" in the kubelet, mostly accessed through the `manager` package. Generally, the `main` package is doing similar things to what the kubelet sets up with the manager, plus a few extras that aren't relevant to us. I think this is working ok.

> e.g. Init function to make this a library? If you have an Init code path, you'll be able to pass config into that and from there to sub-components and can probably get rid of the global state.

I think you mean component constructors, as opposed to an actual golang init function? The golang init function adds _more_ global state, and should be avoided... To reduce the global state, I would pass the config through to the various component constructors. 

> It might be worth adding an explicit validation stage for the config before any sub-components start running.

I like this idea, but from the library perspective, it's not clear where this validation would go. It could be a separate function that is the responsibility of the library-user to call before starting any cAdvisor components. I think a better option is to just make each sub-component validate the relevant part of the config (I think it already works this way to some extent).

---

Looking at this PR again, one change I'd like to add a way to control which flags are added, rather than the current all-or-nothing approach.

I'll clean this up and target it for the Kubernetes 1.5 (cAdvisor v0.25) timeframe.

mtaufen:
> We already use cAdvisor in "library form" in the kubelet, mostly accessed through the manager package. Generally, the main package is doing similar things to what the kubelet sets up with the manager, plus a few extras that aren't relevant to us. I think this is working ok.

Ack.

> I think you mean component constructors, as opposed to an actual golang init function? The golang init function adds more global state, and should be avoided... To reduce the global state, I would pass the config through to the various component constructors.

Yup. Constructors. Definitely not golang's init. Sorry for the confusion.

> I like this idea, but from the library perspective, it's not clear where this validation would go. It could be a separate function that is the responsibility of the library-user to call before starting any cAdvisor components. I think a better option is to just make each sub-component validate the relevant part of the config (I think it already works this way to some extent).

Yeah, I suppose that's the easiest path to having validation and people being able to pick-and-choose what components they use. It might be nice to expose a validation function for each component that a library user could independently call to validate config without actually constructing the component. That way people can still front-load their validation if they want to.

timstclair:
I dropped the changes to the storage backends, since I have a separate proposal addressing them (https://github.com/google/cadvisor/issues/1458).

dashpole:
Closing this as outdated.  Feel free to reopen if you are still working on it

PR #1225, opened 2016-04-18
Add support for UNIX sockets

krallin:
This lets cAdvisor listen on a UNIX socket instead of a TCP socket. This can be practical in environments where untrusted users have access to the TCP ports, and where one would rather rely on user permissions rather than HTTP auth to control access to cAdvisor (in our case, this is simply because we'd rather not have to manage those user / passwords if we don't have to).

There is definitely a bit of a race condition in setting up the permissions on the socket, since we only set the permissions on it after creating it. However, there doesn't appear to be better way of doing so on Linux (FWIW, Docker does exactly the same thing), and the recommended practice for users that have a problem with that is to change the listening socket to listen in a directory with restricted permissions.

This does not change the default behavior, which remains to listen on a TCP port.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -35,6 +36,7 @@ import (
 	"github.com/golang/glog"
 )
 
+var argPath = flag.String("listen_path", "", "Path to listen on (UNIX socket), defaults to empty (use TCP instead)")
 var argIp = flag.String("listen_ip", "", "IP to listen on, defaults to all IPs")
 var argPort = flag.Int("port", 8080, "port to listen")
 var maxProcs = flag.Int("max_procs", 0, "max number of CPUs that can be used simultaneously. Less than 1 for default (number of cores).")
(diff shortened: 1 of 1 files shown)


cadvisor.go
```diff
@@ -35,6 +36,7 @@ import (
 	"github.com/golang/glog"
 )
 
+var argPath = flag.String("listen_path", "", "Path to listen on (UNIX socket), defaults to empty (use TCP instead)")
```
vishh: nit: Rename this to `argUnixSocketPath`

vishh:
ok to test

vishh:
Just one nit, otherwise LGTM

cadvisor.go
timstclair: flag name too, maybe `unix_socket`

timstclair: It's not clear from the description that this overrides listening on TCP. Maybe we should deprecate `listen_ip` and `port`, and switch to a similar approach to docker, a list of paths / URIs to listen to (e.g. `--listen=tcp://localhost:8080,unix:///var/run/cadvisor.sock`)

krallin: @timstclair Just to be sure I'm understanding you correctly, is this (the `--listen` flag) something you'd expect me to post as part of this patch, or something you'd like to look at later?

Cheers,

timstclair: Removing flags is a lot harder than adding them, so if we want to use a `--listen` flag, I'd prefer not to add a `--listen_path` first.

In other words, yes, would you mind adding it in this patch?

krallin: Sure. I'm out next week, but I can take fare of this when I come back.

Do we want backwards compatibility with the existing port flag?

timstclair: Yeah, we should maintain backwards compatibility. Thanks!

PR #1226, opened 2016-04-18
FsHandler: Add support for mounts, usage with df, and skipped devices

krallin:
This PR includes several commits, but they all boil down to one main feature, which is to let `FsHandler` report on arbitrary directories, which is used to report on the disk usage of Docker mounts. 

I've updated the Docker handler to include a container's mounts in reporting, but I have not touched the `rkt` handler beyond updating it to use the new API for FsHandler (I'm not a `rkt` user myself).

This also adds a set of flags that can be used to control the behavior of `FsHandler`:
- `disk_skip_du` can be used to use disk metrics instead of `du` to evaluate the usage of a directory. This can be useful in cases where "is this disk running out of space?" is a more important question than "how much disk usage is this specific container responsible for?" It's also a little faster for that purpose, since it sidesteps the need for running `du`, which can be expensive in time and CPU for large directory trees.
- `disk_skip_devices` can be used to skip specific devices when calculating disk usage and reporting. This can be useful in cases where the usage of mounts matters, but the usage of e.g. `/var/lib/docker` does not (this is particualrly useful with `disk_skip_du`, since you'd otherwise end up reporting the usage of `/var/lib/docker` for every single container, which isn't particularly relevant). The flag is a regex, and defaults to not excluding any devices.

Now, there's also a set of supporting changes to actually make this work:
- First, the list of partitions on the system used to be cached. If we start monitoring mounts, this breaks down, because new partitions can of course be added after cAdvisor starts. This PR adds a partition cache that will refresh when asked for a device or a label it can't find. It also refreshes periodically to catch new volumes that might have been added but whose device is unknown (there are a few queries made to the cache that walk the partition list, and in those cases the partition cache can't know whether the searcher found what it was looking for).
- Second, this PR adds a separate cache for disk statistics. This helps ensure this PR doesn't impact performance negatively (we used to use the machine info partition list to find disk sizes, but that's no longer the case in this PR).

Hope this makes sense. This has been working well for us, but there are a few things I'd like to call out:
- First, cAdvisor sometimes returns unpredictable names for devices because `/proc/self/mountinfo` (which Docker mounts uses) sometimes returns symlinks for devices (e.g. `/dev/by-id/...` for `/dev/sda1`, or `/dev/mapper/...` for `/dev/dm-0`). I do have a fix for this (which is to de-reference symlinks: https://github.com/google/cadvisor/commit/65c6382f3a54ba09a523a4c478048c29eeb2a56a), but that doesn't work when cAdvisor is running within a Docker container (we'd need to prefix the link we de-reference with `/rootfs/` in that case). This somewhat diminishes the usefulness of `disk_skip_devices`, so I'm happy to provide a patch that does this separately; just let me know.
- Second, as of this PR, the FsHandler complains when it's asked for metrics it doesn't have. In practice this will often happen when a container was just added: we'll try to get the disk metrics, but they aren't ready yet. This will generate one message in the logs per container indicating the disk metrics aren't ready yet. I'm happy to work on suppressing that first message when it shows if needed.
- Third, I left a few comments with `REVIEW` inline, for things that might need more 

I used [this test benchmark](https://github.com/krallin/cadvisor-bench) to compare the performance of the existing cAdvisor to this PR. The memory usage on both versions appears to be equivalent, and it looks like this PR uses a little less CPU. I'm still investigating why (it's not immediately clear to me why it would do so).

Let me know what you think!


vishh:
cc @pmorie

vishh:
ok to test

vishh:
@krallin Thanks for the PR. I won't have time to look at this until Wednesday though.

krallin:
No worries; thanks for the heads up @vishh!

krallin:
Looks like there was a test that failed here. I'll look into what happened shortly 😄

krallin:
@vishh 

I think the failing test is due to a change in behavior I made that causes the `fsHandler` to stop reporting stats if none are available. Indeed, the `fsHandler` runs on its own schedule in a separate goroutine, so if the container is asked for metrics before the `fsHandler` is done running, then the `fsHandler` has two options:

- Return incomplete / invalid data. This is what cAdvisor used to do. It would report usage as 0, although it did correctly report disk size. Unfortunately, that's not really a viable option once you stop pulling disk size from a static cache (which we used to). In practice, 0 is not nil, so that would explain why the test used to pass.
- Return nothing. The `fsHandler` will eventually update, and it'll eventually return data, but it'll won't return any until it has updated. This is what it does with this patch. For clarity, that's what I was referring to in my "Second, as of this PR, the FsHandler complains when it's asked for metrics it doesn't have" comment.

---

I'm not sure how I can reproduce the test failure to confirm. Perhaps we should update [the test code to provide more relevant information](https://github.com/google/cadvisor/blob/bbd05e4baa6fd3675a1f5bffb382f0e60842bf1c/integration/tests/api/docker_test.go#L360), e.g. use:

```
require.NotNil(t, info.Stats[0].Filesystem, "got info: %+v", info.Stats[0])
```

Let me know what you think the next steps are here?

Thanks!

vishh:
I agree that returning `nil` is the right approach. The `v1` API is not
compatible with nil values and that is the reason why the existing logic
doesn't use `nil` instead of `0`. I'm ok with updating the test to wait
until the output is non-zero.

Issue #1211, opened 2016-04-14
cAdvisor should ignore .mount cgroups when on systemd


derekwaynecarr:
A previous fix was made in Kubernetes to ignore .mount units

https://github.com/kubernetes/kubernetes/pull/22656

I think .mount cgroups should be ignored on systemd entirely by cAdvisor.

derekwaynecarr:
I was getting errors like the following from here:

https://github.com/kubernetes/kubernetes/blob/master/Godeps/_workspace/src/github.com/google/cadvisor/cache/memory/memory.go#L104

```
unable to find data for container /system.slice/var-lib-docker-devicemapper-mnt-6b4fa4a130bbfa32390d93562453da45350787058d449b15ee1f75c1a56a1c1f\x2dinit.mount 
```

vishh:
Can we instead have a `systemd` plugin in cAdvisor?

derekwaynecarr:
@vishh - that works, is there a particular interface you have in mind that we could implement?

vishh:
Existing rkt integration will be a good place to start - https://github.com/google/cadvisor/tree/master/container/rkt

derekwaynecarr:
I need to look into this some more, but isn't the issue that the raw container handler is what would need to be updated?

derekwaynecarr:
So the factories are registered in-order, if I have a systemd factory registered before the raw factory, are you saying I could say that the systemd can handle .mount cgroups, but it doesn't accept them and the raw factory would just ignore them as a consequence?

derekwaynecarr:
I guess that is what the code currently is doing... ok, I can buy that this may work.

vishh:
Yeah. The raw drive is the fallback option. I'd like to see more systemd
specific code go into the systemd handler.

derekwaynecarr:
Yeah, I want to add more to it as time permits

PR #1235, opened 2016-04-21
on systemd, we should ignore .mount cgroups

derekwaynecarr:
Fixes https://github.com/google/cadvisor/issues/1211

/cc @vishh @pmorie - per our conversation today, this is really basic for now.

FYI @ncdc @smarterclayton  - This was the source of random errors when invoking the kubelet summary stats api when tearing down or bringing up new containers on the node.  Putting this on your radar if you see any other reports, but I suspect I was the one primarily exercising this API right now.


vishh:
ok to test

vishh:
LGTM

PR #1242, opened 2016-04-25
Optimize TimedStore to avoid sort on every data add

timstclair:


utils/timed_store.go
```diff
@@ -58,19 +58,23 @@ func NewTimedStore(age time.Duration, maxItems int) *TimedStore {
...
 		timestamp: timestamp,
-		data:      copied,
-	})
+		data:      item,
+	}
+	// Common case: data is added in order.
+	if len(self.buffer) == 0 || !timestamp.Before(self.buffer[len(self.buffer)-1].timestamp) {
+		self.buffer = append(self.buffer, data)
+	} else {
+		// Data is out of order; insert it in the correct position.
```
jimmidyson: What causes this to happen?

utils/timed_store.go
```diff
@@ -81,6 +85,11 @@ func (self *TimedStore) Add(timestamp time.Time, item interface{}) {
 		self.buffer = self.buffer[index:]
 	}
 
+	// Remove any elements if over our max size.
+	if self.maxItems >= 0 && len(self.buffer) > self.maxItems {
```
jimmidyson: Can we use a ring buffer here as a fixed size list?

utils/timed_store.go
timstclair: Creation events for containers that were running before cAdvisor started.

utils/timed_store.go
timstclair: Yeah, I was playing around with that. It adds a bunch of complexity for relatively little gain, but I'll think about cleaning it up and sending a patch.

utils/timed_store.go
jimmidyson: Can we only do this once when starting & remove the need for this check when running normally?

timstclair: I'd rather not. I'm not positive there aren't any other times that events could arrive out of order, and I think the "common case" check makes the cost of this very low.

timstclair:
Reassigning to Vish - I'd like to get this in to k8s v1.3.

timstclair:
@k8s-bot test this

jimmidyson:
Sorry this slipped off my radar. LGTM.

timstclair:
Thanks! I had heard a rumor you weren't working on cAdvisor anymore.

jimmidyson:
@timstclair Rumours... I'm still around, just not contributing as much as I'd like to be able to right now.

PR #1243, opened 2016-04-25
Don't create a EventStore if the event limit is set to 0

timstclair:
For https://github.com/kubernetes/kubernetes/issues/24698


timstclair:
Friendly ping, this is blocking https://github.com/kubernetes/kubernetes/issues/24698

vishh:
LGTM. Sorry for the delay!

PR #1246, opened 2016-04-27
Ring timed store

timstclair:
I'm not sure the additional complexity is worth the performance gains, but here is an implementation of TimedStore using a ring buffer. I couldn't find a decent library for this, but let me know if there is one I missed.

Benchmarks:

```
Iters = 10000
Limit = 10

BenchmarkSliceAddSizeLimit-12            500       2529468 ns/op
BenchmarkRingAddSizeLimit-12            1000       2053455 ns/op
BenchmarkSliceAddAgeLimit-12            1000       2440000 ns/op
BenchmarkRingAddAgeLimit-12             1000       2024899 ns/op
BenchmarkSliceAddRandom-12               300       4127581 ns/op
BenchmarkRingAddRandom-12                500       2988560 ns/op


Iters = 10000
Limit = 100

BenchmarkSliceAddSizeLimit-12            500       2545626 ns/op
BenchmarkRingAddSizeLimit-12            1000       2332148 ns/op
BenchmarkSliceAddAgeLimit-12             500       2438340 ns/op
BenchmarkRingAddAgeLimit-12             1000       2320046 ns/op
BenchmarkSliceAddRandom-12               300       5750130 ns/op
BenchmarkRingAddRandom-12                500       3555849 ns/op

Iters = 10000
Limit = 1000

BenchmarkSliceAddSizeLimit-12            500       2707495 ns/op
BenchmarkRingAddSizeLimit-12             500       2563681 ns/op
BenchmarkSliceAddAgeLimit-12             500       2704203 ns/op
BenchmarkRingAddAgeLimit-12              500       2529219 ns/op
BenchmarkSliceAddRandom-12               100      21935956 ns/op
BenchmarkRingAddRandom-12                300       4272770 ns/op

Iters = 10000
Limit = none

BenchmarkSliceAddRandomUnlimited-12       10    1592485215 ns/op
BenchmarkRingAddRandomUnlimited-12       200      11563374 ns/op
```

The ring buffer performs significantly better in the Unlimited case, but the gains in the limited case are diminishing as the buffer size increases.

/cc @vishh 

Note: This includes https://github.com/google/cadvisor/pull/1242

### DIFF
--- a/utils/timed_store.go
+++ b/utils/timed_store.go
@@ -19,24 +19,191 @@ import (
 	"time"
 )
 
-type timedStoreDataSlice []timedStoreData
-
-func (t timedStoreDataSlice) Less(i, j int) bool {
-	return t[i].timestamp.Before(t[j].timestamp)
+type timedStoreRingBuffer struct {
+	start   int
+	end     int // Last element index + 1
+	buffer  []timedStoreData
+	limited bool // Whether the buffer should grow or overwrite
+	empty   bool // Necessary to distinguish full & empty
 }
 
-func (t timedStoreDataSlice) Len() int {
-	return len(t)
+func newTimedStoreRingBuffer(cap int, limited bool) timedStoreRingBuffer {
+	return timedStoreRingBuffer{
+		start:   0,
+		end:     0,
+		buffer:  make([]timedStoreData, cap),
+		limited: limited,
+		empty:   true,
+	}
 }
 
-func (t timedStoreDataSlice) Swap(i, j int) {
-	t[i], t[j] = t[j], t[i]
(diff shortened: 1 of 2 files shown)


vishh:
The reality is that we will seldom add elements out of order. Do you think its worth optimizing the sort?

timstclair:
The sort is expensive enough that it shows up in cAdvisor profiles, but that can be fixed by the much simpler changes in https://github.com/google/cadvisor/pull/1242. This PR deals mostly with optimizing the adding & removing elements. Or were you suggesting we just eat the cost of a full sort instead of insert when an append won't do?

PR #1251, opened 2016-05-02
Use docker to build cAdvisor. 

vishh:
`make` now works on OS X.

cc @timstclair @jimmidyson

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,15 +13,33 @@
 
 GO := godep go
 pkgs  = $(shell $(GO) list ./...)
+SOURCEDIR = $(shell pwd)
+SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
+DOCKER = $(shell command -v docker 2> /dev/null)
+build-image = cadvisor-build-image
+DOCKERMAKE = $(DOCKER) run --rm -i -v $(SOURCEDIR):/go/src/github.com/google/cadvisor $(build-image) make
 
 all: format build test
 
-test:
+all-docker: format build-docker test-docker
+
+build-image: build/Dockerfile
+	@$(DOCKER) build -t $(build-image) build/
+	@touch $@
+
+test-docker: $(SOURCES) build-image
+	@echo ">> running tests using docker"
+	@$(DOCKERMAKE) test
+
+test: $(SOURCES)
 	@echo ">> running tests"
 	@$(GO) test -tags test -short -race $(pkgs)
 
 test-integration: build test
 	@./build/integration.sh
+
+test-integration-docker: 
+	@$(DOCKERMAKE) test-integration
 
 format:
 	@echo ">> formatting code"
@@ -31,10 +49,14 @@ vet:
 	@echo ">> vetting code"
 	@$(GO) vet $(pkgs)
 
-build:
+build-docker: $(SOURCES) build-image 
+	@echo ">> building binaries using docker"
+	@$(DOCKERMAKE) build
+
+build: $(SOURCES)
 	@echo ">> building binaries"
-	@./build/assets.sh
-	@./build/build.sh
+	@./build/assets.sh && ./build/build.sh
+	@touch $@
 
 release: build
 	@./build/release.sh
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,0 +1,5 @@
+FROM golang:1.6.2
+MAINTAINER vishnuk@google.com
+
+RUN go get github.com/tools/godep
+WORKDIR /go/src/github.com/google/cadvisor


Makefile
```diff
@@ -12,13 +12,14 @@
...
+goversion = 1.6.2
 pkgs  = $(shell $(GO) list ./...)
 
 all: format build test
 
 test:
-	@echo ">> running tests"
-	@$(GO) test -tags test -short -race $(pkgs)
+	@echo ">> running tests using docker"
+	@docker run -i -v $(PWD):/go/src/github.com/google/cadvisor golang:$(goversion) /bin/sh -c "cd /go/src/github.com/google/cadvisor && go get github.com/tools/godep && $(GO) test -tags test -short -race $(pkgs)"
```
thaJeztah: Perhaps use `-w` (workdir) instead of `"cd /go/src/github.com/google/cadvisor"`, and add `--rm` so that the container is removed after it's finished;

```Makefile
@docker run -i --rm -v $(PWD):/go/src/github.com/google/cadvisor -w /go/src/github.com/google/cadvisor golang:$(goversion) /bin/sh -c "go get github.com/tools/godep && $(GO) test -tags test -short -race $(pkgs)"
```

Makefile
```diff
@@ -32,9 +33,8 @@ vet:
 	@$(GO) vet $(pkgs)
 
 build:
-	@echo ">> building binaries"
-	@./build/assets.sh
-	@./build/build.sh
+	@echo ">> building binaries using docker"
+	@docker run -i -v $(PWD):/go/src/github.com/google/cadvisor golang:$(goversion) /bin/sh -c "cd /go/src/github.com/google/cadvisor && go get github.com/tools/godep && ./build/assets.sh && ./build/build.sh"
```
thaJeztah: same here;

```Makefile
@docker run -i --rm -v $(PWD):/go/src/github.com/google/cadvisor -w /go/src/github.com/google/cadvisor golang:$(goversion) /bin/sh -c "go get github.com/tools/godep && ./build/assets.sh && ./build/build.sh"
```

Makefile
thaJeztah: or (slightly less DRY), use a variable for workdir;

```Makefile
@docker run -i --rm -v $(PWD):$(workdir) -w $(workdir) golang:$(goversion) /bin/sh -c "go get github.com/tools/godep && $(GO) test -tags test -short -race $(pkgs)"
```

vishh: Ahh. That makes sense. Updating PR now.

Makefile
vishh: Updating!

timstclair: How does this play with https://github.com/google/cadvisor/pull/1253? Is it still able to reuse the intermediate build results?

timstclair: Would it make sense to move some of this (e.g. godep install) into a Dockerfile? Maybe add another make rule to build the docker image first?

thaJeztah: For speed, that would probably help yes

vishh: @timstclair 

#1253 will still be in effect. Adding a Dockerfile makes sense. I'm updating the PR now.

vishh:
@timstclair @thaJeztah PTAL

Makefile
timstclair: I don't think it is. I think you need to mount the $GOPATH/pkg directory for it to work.

With your changes (irrelevant output omitted):
```
[(c262eb7...)] cadvisor: time make build
>> building binaries using docker
 >   cadvisor

real	0m14.201s
user	0m0.066s
sys	0m0.032s

[(c262eb7...)] cadvisor: time make build
>> building binaries using docker
 >   cadvisor

real	0m14.279s
user	0m0.073s
sys	0m0.027s
```

Without your changes:
```
[install] cadvisor: time make build
>> building binaries
 >   cadvisor

real	0m6.658s
user	0m13.909s
sys	0m1.622s

[install] cadvisor: time make build
>> building binaries
 >   cadvisor

real	0m0.781s
user	0m0.343s
sys	0m0.188s
```

Notice how the second build is super fast once everything is cached.

vishh: Given that we build entirely using vendor'ed code, I'm weary of mapping in
the entire `$GOPATH/pkg` directory. On top of that since we install godep,
it will overwrite the existing `godep` in the host's $GOBIN path, which is
undesirable.

timstclair:
Can we do something like the solution in http://stackoverflow.com/questions/26118303/docker-with-make-build-image-on-dockerfile-change so the docker image doesn't need to be rebuilt on every execution?

Makefile
timstclair: Hmm, what if we create an `_output/pkg` directory in cAdvisor which we map in instead?

Makefile
```diff
@@ -12,13 +12,18 @@
 # See the License for the specific language governing permissions and
 
 GO := godep go
+goversion = 1.6.2
```
thaJeztah: you can remove this now

thaJeztah:
@timstclair I don't think that would make much difference; the docker build cache should make building near instant. The only thing that can be done to improve is to add a `.dockerignore` file, so that `docker build` doesn't upload the source files to the daemon

timstclair:
Partly I was concerned about the output spam, but maybe if we clean that up the build time is a non-issue.

Makefile
vishh: Done

vishh:
@timstclair made a few changes
* docker based builds is now optional - I wish we make it the default because it will *just* work for everyone (assuming docker is available)
* no rebuilds will be triggered for the image unless the build image Dockerfile changes
* re-use makefile inside docker container

PTAL

vishh:
@timstclair 
The thing I'm having trouble with is having make not run `docker run` unless the sources files have changes. I got it working for the build image, but not for source and test files. Inputs welcome :)
In any case we can now safely improve this.

chenchun:
Happened to do the same thing. Maybe adding another make target `docker-build` instead of depending on `DOCKER` env makes makefile cleaner?

vishh:
@chenchun good suggestion. I went too deep in one path and did not think of having an alternate build rule 👍

thaJeztah:
Tried this in a "clean" environment; not sure to what extend you want it to "_just_ _work_", but `make all-docker` still triggers `format`, which calls `godep` on the host.

```bash
$ make all-docker
make: godep: Command not found
>> formatting code
make: godep: No such file or directory
make: *** [format] Error 1
```

Oh, not related to this pull request, but now that there's quite some options in the Makefile, this is a really cool trick to implement a `make help`: https://github.com/docker/docker/blob/9e7651db4d464649bc32cf346fdee8c6900ebe0f/Makefile#L114-L116

timstclair:
I'm ok with what you have, but here's an alternative solution: create a script which: 1) builds the docker image if necessary (easier to just do it in a script than make...) 2) execute the make command in docker (like you have). Then, you could just control the use of docker with a `USE_DOCKER` variable, something like:
```
USE_DOCKER =
ifdef USE_DOCKER
  DOCKERMAKE = ...
endif

build:
  $(DOCKERMAKE) ./build/build.sh

etc.
```

Does that make sense? I'll leave it up to you.

vishh:
@thaJeztah 

My intention was to make it possible to build and test cAdvisor outside of linux. Since `godep` is available across OSes and platforms, I did not' bother running format inside a docker container. But I can do that too :)

@timstclair 

I'm starting to like make rules over conditionals. If we can get make to not invoke `docker run` when no files have changes, I'd be interested in switching the default to use docker.

thaJeztah:
@vishh alright, feel free to ignore then, I was just wondering to what extend you wanted to make it "work" anywhere. Making that work would be a "nice to have" then, not strictly necessary

Issue #1248, opened 2016-04-28
OOM monitoring fails when logs are rotated


sjenning:
The OOM monitoring code monitors a single log file once the monitoring starts.  If the logs are rotated, cadvisor doesn't switch to the new log file.

PR #1252, opened 2016-05-02
prefer using systemd for oom monitoring

sjenning:


vishh:
ok to test

vishh:
This change LGTM! Can you clarify the rationale for this change in this PR?

sjenning:
While I'm looking to add log rotation support (#1248) to the path that watches log files, preferring systemd bypasses the issue entirely by reading from journalctl if it is available.  Some systems will have the log files AND systemd and in that case, we want to read from the journal since it hides the log rotation issue.

utils/oomparser/oomparser.go
```diff
@@ -205,13 +205,17 @@ func getSystemFile() (string, error) {
 // initializes an OomParser object and calls getSystemFile to set the systemFile
 // attribute.  Returns and OomParser object and an error
 func New() (*OomParser, error) {
+	parser, err := trySystemd()
+	if err == nil {
+		return parser, err
```
timstclair: (nit) return `nil` instead `err` for clarity

vishh:
Given that there are still other distros that are not running systemd yet, it will still be helpful to fix the log rotation issue.

sjenning:
@timstclair updated

timstclair:
Thanks, LGTM

PR #1253, opened 2016-05-02
Switch to go install for faster incremental builds

timstclair:


vishh:
LGTM. I was thinking about this yesterday 👍

PR #1254, opened 2016-05-02
fix chan recieve block on oom event

sjenning:
SteamOoms() blocks on a chan receive on the last line of the OOM kernel message block here:
https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser.go#L124

This results in the oom event not being reported until another kernel log line comes in after the oom block.

### DIFF
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -144,17 +144,18 @@ func (self *OomParser) StreamOoms(outStream chan *OomInstance) {
 			oomCurrentInstance := &OomInstance{
 				ContainerName: "/",
 			}
-			finished := false
-			for !finished {
+			for line := range lineChannel {
 				err := getContainerName(line, oomCurrentInstance)
 				if err != nil {
 					glog.Errorf("%v", err)
 				}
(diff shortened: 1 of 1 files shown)


utils/oomparser/oomparser.go
```diff
@@ -144,17 +144,18 @@ func (self *OomParser) StreamOoms(outStream chan *OomInstance) {
 			oomCurrentInstance := &OomInstance{
 				ContainerName: "/",
 			}
-			finished := false
-			for !finished {
+			for line := range lineChannel {
```
timstclair: This is a bit different, since it doesn't check the start line. Is the desired information ever on that first line? If not this is fine, otherwise you can keep the original loop and just check `finished` before receiving the line.

sjenning: @timstclair  The desired information is never in the start line so no problem.   Good you brought it up though.

timstclair:
ok to test

sjenning:
@timstclair jenkins failure looks like a jenkins issue

timstclair:
LGTM

PR #1259, opened 2016-05-03
Fixes for --disable_metrics flag

timstclair:
1. Fix for writing to nil map when value is non-empty.
2. Fix the usage string: go flag package uses backticks for the flag
name.
3. Render the flag default value in the same format as the flag accepts.


vishh:
LGTM. Thanks for the cleanup @timstclair!

PR #1261, opened 2016-05-03
move ListContainers code to common and make rkt play nice

sjpotter:
playing nice means not elliding the system.slice cgroup but allowing the raw handler to handle it


sjpotter:
@timstclair @vishh

vishh:
Good catch. LGTM

PR #1263, opened 2016-05-04
Refactor container watching out of raw handler into its own inteface / package

sjpotter:
So this PR is about trying to enable other methods of identifying container creation that can run in parallel with cgroup based detection as well as enabling one to override a cgroup based detection handler (either to operate outside the cgroup hierarchy, as might be case with VMs) or simply override the cgroup based handler.

idea is to use get it so the we an determine when rkt pods are created not via cgroup creation but via the rkt api service itself.

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -27,23 +27,6 @@ const (
 	ListRecursive
 )
 
-// SubcontainerEventType indicates an addition or deletion event.
-type SubcontainerEventType int
-
-const (
-	SubcontainerAdd SubcontainerEventType = iota
-	SubcontainerDelete
-)
-
-// SubcontainerEvent represents a
-type SubcontainerEvent struct {
-	// The type of event that occurred.
-	EventType SubcontainerEventType
-
-	// The full container name of the container where the event occurred.
-	Name string
-}
-
 // Interface for container operation handlers.
 type ContainerHandler interface {
 	// Returns the ContainerReference
@@ -60,12 +43,6 @@ type ContainerHandler interface {
 
 	// Returns the processes inside this container.
 	ListProcesses(listType ListType) ([]int, error)
-
-	// Registers a channel to listen for events affecting subcontainers (recursively).
-	WatchSubcontainers(events chan SubcontainerEvent) error
-
-	// Stops watching for subcontainer changes.
-	StopWatchingSubcontainers() error
 
 	// Returns absolute cgroup path for the requested resource.
 	GetCgroupPath(resource string) (string, error)
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -37,6 +37,8 @@ import (
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/info/v2"
 	"github.com/google/cadvisor/machine"
+	"github.com/google/cadvisor/manager/watcher"
+	rawwatcher "github.com/google/cadvisor/manager/watcher/raw"
 	"github.com/google/cadvisor/utils/cpuload"
 	"github.com/google/cadvisor/utils/oomparser"
 	"github.com/google/cadvisor/utils/sysfs"
@@ -163,6 +165,9 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 		inHostNamespace = true
 	}
 
+	// Register for new subcontainers.
+	eventsChannel := make(chan watcher.ContainerEvent, 16)
+
 	newManager := &manager{
 		containers:               make(map[namespacedContainerName]*containerData),
 		quitChannels:             make([]chan error, 0, 2),
@@ -174,6 +179,8 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 		maxHousekeepingInterval:  maxHousekeepingInterval,
 		allowDynamicHousekeeping: allowDynamicHousekeeping,
 		ignoreMetrics:            ignoreMetricsSet,
+		containerWatchers:        []watcher.ContainerWatcher{},
+		eventsChannel:            eventsChannel,
 	}
 
 	machineInfo, err := machine.Info(sysfs, fsInfo, inHostNamespace)
@@ -217,6 +224,8 @@ type manager struct {
 	maxHousekeepingInterval  time.Duration
 	allowDynamicHousekeeping bool
 	ignoreMetrics            container.MetricSet
+	containerWatchers        []watcher.ContainerWatcher
+	eventsChannel            chan watcher.ContainerEvent
 }
 
 // Start the container manager.
@@ -240,6 +249,12 @@ func (self *manager) Start() error {
 	if err != nil {
 		glog.Errorf("Registration of the raw container factory failed: %v", err)
 	}
+
+	rawWatcher, err := rawwatcher.NewRawContainerWatcher()
+	if err != nil {
+		return err
+	}
+	self.containerWatchers = append(self.containerWatchers, rawWatcher)
 
 	if *enableLoadReader {
 		// Create cpu load reader.
@@ -269,7 +284,7 @@ func (self *manager) Start() error {
 	}
 
 	// Create root and then recover all containers.
-	err = self.createContainer("/")
+	err = self.createContainer("/", watcher.Raw)
 	if err != nil {
 		return err
 	}
@@ -769,10 +784,14 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 }
 
 // Create a container.
-func (m *manager) createContainer(containerName string) error {
+func (m *manager) createContainer(containerName string, watchSource watcher.ContainerWatchSource) error {
 	m.containersLock.Lock()
 	defer m.containersLock.Unlock()
 
+	return m.createContainerLocked(containerName, watchSource)
+}
+
+func (m *manager) createContainerLocked(containerName string, watchSource watcher.ContainerWatchSource) error {
 	namespacedName := namespacedContainerName{
 		Name: containerName,
 	}
@@ -782,7 +801,7 @@ func (m *manager) createContainer(containerName string) error {
 		return nil
 	}
 
-	handler, accept, err := container.NewContainerHandler(containerName, m.inHostNamespace)
+	handler, accept, err := container.NewContainerHandler(containerName, watchSource, m.inHostNamespace)
 	if err != nil {
 		return err
 	}
@@ -849,6 +868,10 @@ func (m *manager) destroyContainer(containerName string) error {
 	m.containersLock.Lock()
 	defer m.containersLock.Unlock()
 
+	return m.destroyContainerLocked(containerName)
+}
+
+func (m *manager) destroyContainerLocked(containerName string) error {
 	namespacedName := namespacedContainerName{
 		Name: containerName,
 	}
@@ -946,7 +969,7 @@ func (m *manager) detectSubcontainers(containerName string) error {
 
 	// Add the new containers.
 	for _, cont := range added {
-		err = m.createContainer(cont.Name)
+		err = m.createContainer(cont.Name, watcher.Raw)
 		if err != nil {
 			glog.Errorf("Failed to create existing container: %s: %s", cont.Name, err)
 		}
@@ -965,28 +988,15 @@ func (m *manager) detectSubcontainers(containerName string) error {
 
 // Watches for new containers started in the system. Runs forever unless there is a setup error.
 func (self *manager) watchForNewContainers(quit chan error) error {
-	var root *containerData
-	var ok bool
-	func() {
-		self.containersLock.RLock()
-		defer self.containersLock.RUnlock()
-		root, ok = self.containers[namespacedContainerName{
-			Name: "/",
-		}]
-	}()
-	if !ok {
-		return fmt.Errorf("root container does not exist when watching for new containers")
-	}
-
-	// Register for new subcontainers.
-	eventsChannel := make(chan container.SubcontainerEvent, 16)
-	err := root.handler.WatchSubcontainers(eventsChannel)
-	if err != nil {
-		return err
+	for _, watcher := range self.containerWatchers {
+		err := watcher.Start(self.eventsChannel)
+		if err != nil {
+			return err
+		}
 	}
 
 	// There is a race between starting the watch and new container creation so we do a detection before we read new containers.
-	err = self.detectSubcontainers("/")
+	err := self.detectSubcontainers("/")
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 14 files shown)


sjpotter:
@timstclair

sjpotter:
it contains 2 parts

1) pass a "detection" type into the CanHandleAndAccept function, i.e. "raw" "rkt" but can be more in future

2) break up the createContainer/destroyContainer into unlocked version (along with wrappers that lock them) as well as an updateContainer method that can destroy an old handler while creating a new one.

manager/manager.go
```diff
@@ -769,11 +769,35 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 	return nil
 }
 
+// As we are proposing having multiple ways to detect a container (in addition to cgroups)
+// need the ability to delete
```
timstclair: I don't understand this comment.

timstclair:
I don't like passing around the detection type. I think it would be better if factories are registered per detection mechanism instead. Rather than registering the factories against a global (https://github.com/google/cadvisor/blob/master/container/factory.go#L76), we could create a "ContainerWatcher" interface, something like:
```go
type ContainerWatcher interface {
  Start()
  Stop()
  RegisterFactory(factory ContainerHandlerFactory)
}
```

manager/manager.go
sjpotter: so currently we detect containers via raw cgroups (what I'll call the "raw" method).  I'm proposing the ability to detect containers via other methods, such as the rkt api service.

if a raw handler was already created for a cgroup path for a rkt container (via raw detection), we want the rkt path to overwrite/update it to the rkt handler.  Or at least that's what I thought you wanted.  i.e. the raw handler will be destroyed and a rkt handler will take its place.  all in one function to ensure atomic nature.

if the rkt handler is created first, the raw handler wont be created as createContainer() skips if the cgroup is already a key in the map.

sjpotter:
ok, I'll try to work someting up tomorrow, it will basically mirror the existing factory system.

sjpotter:
so this commit I'm not sure is exactly where you want it, but instead of passing the detectionType all the way through it's just passed to container.NewContainerHandler (i.e. the factory method).

When we register factories we say what type of detection they support (I used a slice so they can technically support more than one, but unsure that's really an issue)

And I expanded the SubcontainerEvent to include the detection type, that way I hope we can just have multiple "detectors" writing to the same channel.   Next commit will add a new event called Update which will use the update path from the original commit

sjpotter:
@timstclair @vishh so lots of surgery, but this moves the raw cgroup container watcher to its own package and enables registering other watchers with the manager to use the same channel to write on.

I have a polling rkt implementation here https://github.com/sjpotter/cadvisor/tree/new-handler-detection-with-rkt/

manager/manager.go
timstclair: Ok, update the comment with that explanation? :)
I think the comment is not a proper sentence.

container/container.go
```diff
@@ -42,6 +43,9 @@ type SubcontainerEvent struct {
 
 	// The full container name of the container where the event occurred.
 	Name string
+
+	//who detected it
+	DetectionType string
```
timstclair: I still don't like having "detection type". It just feels messy and not the right abstraction. Why is it necessary? I don't think container factories should care how the container was detected, their job is to just create the handler for it. It should be up to the manager (or a child module) to ensure that the correct factories are hooked up to the right detection sources.

watcher/raw/raw.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package container defines types for sub-container events and also
+// defines an interface for container operation handlers.
+package raw
```
timstclair: I think the raw sub-package is probably unnecessary (can just be in watcher?). Also, can we make watcher a sub-package of manager?

manager/manager.go
sjpotter: done

// Enables overwriting an existing containerData/Handler object for a given containerName
// createContainer just returns if a given containerName has a handler already
// Ex: rkt handler will want to take priority over the raw handler, but the raw handler might be created first

container/container.go
sjpotter: I'm trying to minimize the code duplication, so only have them all going through the same event loop which goes to the same factory loop.

However, since we are dealing with the same container names if we don't distinguish where the container name came from, rkt wouldn't be able to determine if it was for it or not and would have to wait to make sure.  it's just labeling which watcher found it.

I'm not convinced this is right, but I'm trying to keep the code simple and not overly complicated

watcher/raw/raw.go
sjpotter: yea, it can be sub manager, but not 100% sure why you want the raw bumped up, looses the symetry with any other watcher that is added

sjpotter:
@k8s-bot test this

manager/watcher/raw/raw.go
```diff
@@ -0,0 +1,215 @@
...
+			break
+		}
+	}
+	if containerName == "" {
+		return fmt.Errorf("unable to detect container from watch event on directory %q", event.Name)
+	}
+
+	// Maintain the watch for the new or deleted container.
+	switch {
+	case eventType == container.SubcontainerAdd:
```
timstclair: switch on eventType

container/container.go
```diff
@@ -42,6 +42,9 @@ type SubcontainerEvent struct {
 
 	// The full container name of the container where the event occurred.
 	Name string
+
+	//who detected it
```
yifan-gu: More description on the comment?

manager/manager.go
```diff
@@ -163,6 +165,9 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 		inHostNamespace = true
 	}
 
+	// Register for new subcontainers.
+	eventsChannel := make(chan watcher.SubcontainerEvent, 16)
```
yifan-gu: How is the channel size (16)  been decided? @sjpotter

yifan-gu: Ok, I saw that you copied from the old code.

manager/manager.go
```diff
@@ -768,11 +783,44 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 	return nil
 }
 
+// Enables overwriting an existing containerData/Handler object for a given containerName
+// createContainer just returns if a given containerName has a handler already
```
yifan-gu: The documentation here is incomplete.

yifan-gu:
LGTM overall, with some nits. So this PR splits the watcher from the container handler, right? @sjpotter

sjpotter:
@yifan-gu corrcet, this is just the refactor, the rkt support will be a separate drop.  It will be a small rework of the handler along with a new "watcher" mechanism that use the rkt api service.

manager/manager.go
sjpotter: I'm not sure what you mean, what is missing?

sjpotter:
@k8s-bot test this

manager/manager.go
```diff
@@ -995,21 +1034,23 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
 				switch {
-				case event.EventType == container.SubcontainerAdd:
-					err = self.createContainer(event.Name)
-				case event.EventType == container.SubcontainerDelete:
+				case event.EventType == watcher.SubcontainerAdd:
+					err = self.createContainer(event.Name, event.WatchSource)
+				case event.EventType == watcher.SubcontainerDelete:
 					err = self.destroyContainer(event.Name)
+				case event.EventType == watcher.SubcontainerOverride:
+					err = self.overrideContainer(event.Name, event.WatchSource)
```
timstclair: I have some concerns around how override is handled, but it might make more sense once the rkt watcher is added. Can we just remove the override event & method from this PR, and I'll review it when it's actually used?

manager/manager.go
```diff
@@ -995,21 +1034,23 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
 					glog.Warningf("Failed to process watch event %+v: %v", event, err)
 				}
 			case <-quit:
 				// Stop processing events if asked to quit.
-				err := root.handler.StopWatchingSubcontainers()
-				quit <- err
-				if err == nil {
+				for _, watcher := range self.containerWatchers {
+					err := watcher.StopWatchingSubcontainers()
+					quit <- err
```
timstclair: I'm pretty sure this will deadlock. Please add a test for this case.

manager/watcher/watcher.go
```diff
@@ -0,0 +1,52 @@
...
+	// The full container name of the container where the event occurred.
+	Name string
+
+	// The watcher that detected this change event
+	WatchSource SubcontainerWatchSource
+}
+
+type ContainerWatcher interface {
+	// Registers a channel to listen for events affecting subcontainers (recursively).
+	WatchSubcontainers(events chan SubcontainerEvent) error
```
timstclair: Can we s/Subcontainer/Container/ everywhere? I don't think subcontainer is really accurate everywhere, and I think it just adds confusion.

manager/watcher/watcher.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Stops watching for subcontainer changes.
+	StopWatchingSubcontainers() error
```
timstclair: nit: I think you can just call these methods `Start` and `Stop`

manager/manager.go
yifan-gu: @sjpotter I don't understand why we talks about createContainer here?

manager/manager.go
yifan-gu: Hmm, not a fan of using a chan for both sending errors and signaling.

manager/watcher/watcher.go
sjpotter: done

manager/watcher/watcher.go
sjpotter: done

manager/manager.go
sjpotter: this is the exact same code as before.  We spoke before about this being sort of pointless, perhaps its ok to not return an error and just signal by closing the quit channel?

manager/manager.go
sjpotter: removed the event for now

manager/manager.go
sjpotter: trying to justify this method as an addition to createContainer()

manager/manager.go
timstclair: It's not the same, because you're doing it in a loop. The calling code does this:
```go
quit <- nil
err := <- quit
```
Which will catch the first result, but if there are multiple watchers will get stuck.

@yifan-gu I totally agree, in particular I think the [`Stop()`](https://github.com/google/cadvisor/blob/master/manager/manager.go#L300) needs to be fixed. @sjpotter it's up to you whether you want to tackle that, but either way this code should not deadlock.

manager/manager.go
yifan-gu: @sjpotter Got you, can we add some comma or period, e.g. before `createContainer just...` ? cc @joshix for more suggestions :)

manager/manager.go
```diff
@@ -768,11 +783,44 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
...
+	namespacedName := namespacedContainerName{
+		Name: containerName,
+	}
+
+	if _, ok := m.containers[namespacedName]; ok {
+		containerData := m.containers[namespacedName]
+
+		if containerData.handler.String() != "raw" {
+			return nil
+		}
```
timstclair: Why is this check needed?

manager/manager.go
```diff
@@ -995,21 +1034,35 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+					if err != nil {
+						errors = append(errors, err)
+					}
+				}
+
+				if len(errors) > 0 {
+					err_str := fmt.Sprintf("%v", errors[0])
+					for i := 1; i < len(errors); i++ {
+						err_str = fmt.Sprintf("%v, %v", err_str, errors[i])
+					}
```
timstclair: nit: make `errors` a `[]string`, then just do `strings.Join` here.

timstclair:
Ok, I still don't like passing around the `WatchSource` but I think fixing it would be a more substantial change, so I'm willing to unblock this with the WatchSource. If you clean up the remaining issues, we can get this merged. Also, please update the PR title.

manager/manager.go
```diff
@@ -768,11 +783,44 @@ func (m *manager) registerCollectors(collectorConfigs map[string]string, cont *c
 	return nil
 }
 
+// Enables overwriting an existing containerData/Handler object for a given containerName.
+// can't use createContainer as it just returns if a given containerName has a handler already
+// Ex: rkt handler will want to take priority over the raw handler, but the raw handler might be created first
+
+// Only allow raw handler to be overriden
+func (m *manager) overrideContainer(containerName string, watchSource watcher.ContainerWatchSource) error {
```
timstclair: I thought I said this elsewhere, but can't find the comment.... Can we remove overrrideContainer for now (since it's unused), so that I can review it when I can see how it's actually used?

manager/manager.go
sjpotter: Its now removed in this branch and is part of the rkt addons branch, but the reason is to only allow overriding a raw handler.

i.e. cgroup gets a raw handler, then we say it should get a rkt handler.  it just to enforce.  If you dont think its neccessary can remove it

manager/manager.go
sjpotter: you did, and I did, and just pushed it.

manager/manager.go
sjpotter: done

manager/manager.go
```diff
@@ -995,21 +1005,32 @@ func (self *manager) watchForNewContainers(quit chan error) error {
...
+				errors := []string{}
+
 				// Stop processing events if asked to quit.
-				err := root.handler.StopWatchingSubcontainers()
-				quit <- err
-				if err == nil {
+				for _, watcher := range self.containerWatchers {
+					err := watcher.Stop()
+					if err != nil {
+						errors = append(errors, fmt.Sprintf("%v", err))
```
timstclair: just do err.Error()

container/container.go
```diff
@@ -82,4 +59,7 @@ type ContainerHandler interface {
 	// Start starts any necessary background goroutines - must be cleaned up in Cleanup().
 	// It is expected that most implementations will be a no-op.
 	Start()
+
+	//Name of handler
+	String() string
```
timstclair: Can we remove this?

sjpotter: we could for now, but its needed for identifying handlers later on

timstclair: Let's add it when it's needed.

sjpotter: ok, in progress

yifan-gu:
@timstclair Are we able to get those rkt support PRs in and bump cadvisor in 7 days?

timstclair:
Yes. Please make sure that any remaining issues / PRs are marked with the [kubernetes v1.3 milestone](https://github.com/google/cadvisor/milestones/Kubernetes%20v1.3)

sjpotter:
Once this pr is merged i have another pr with the rkt implementation ready
to post.

Testing it with k8s now.

PR #1264, opened 2016-05-04
add log file tailing and logrotate support

### LINKED ISSUE #1248: OOM monitoring fails when logs are rotated

sjenning:
Fixes #1248

Add some "tail -F" support (via polling and reopening) so that cadvisor can continue to track kernel log messages across log rotations.

Looked at doing this with fsnotify but 1) fsnotify depends on golang.org/x/sys/unix and that is about 10k LOC and 2) I couldn't find a tail implementation that actually worked with fsnotify.

### DIFF
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -18,7 +18,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"os"
 	"os/exec"
 	"path"
 	"regexp"
@@ -26,6 +25,7 @@ import (
 	"time"
 
 	"github.com/google/cadvisor/utils"
+	"github.com/google/cadvisor/utils/tail"
 
 	"github.com/golang/glog"
 )
@@ -105,27 +105,29 @@ func checkIfStartOfOomMessages(line string) bool {
 // Should prevent EOF errors that occur when lines are read before being fully
 // written to the log. It reads line by line splitting on
 // the "\n" character.
-func readLinesFromFile(lineChannel chan string, ioreader *bufio.Reader) {
+func readLinesFromFile(lineChannel chan string, ioreader *bufio.Reader) error {
 	linefragment := ""
 	var line string
 	var err error
 	for true {
 		line, err = ioreader.ReadString('\n')
-		if err == io.EOF {
-			if line != "" {
-				linefragment += line
-			}
-			time.Sleep(100 * time.Millisecond)
-		} else if err == nil {
-			if linefragment != "" {
-				line = linefragment + line
-				linefragment = ""
-			}
-			lineChannel <- line
-		} else if err != nil && err != io.EOF {
+		if err != nil && err != io.EOF {
 			glog.Errorf("exiting analyzeLinesHelper with error %v", err)
+			close(lineChannel)
+			break
+		}
+		if line == "" {
+			time.Sleep(100 * time.Millisecond)
+			continue
+		}
+		if err == nil {
+			lineChannel <- linefragment + line
+			linefragment = ""
+		} else { // err == io.EOF
+			linefragment += line
 		}
 	}
+	return err
 }
 
 // Calls goroutine for readLinesFromFile, which feeds it complete lines.
@@ -160,7 +162,7 @@ func (self *OomParser) StreamOoms(outStream chan *OomInstance) {
 			outStream <- oomCurrentInstance
 		}
 	}
-	glog.Infof("exiting analyzeLines")
+	glog.Infof("exiting analyzeLines. OOM events will not be reported.")
 }
 
 func callJournalctl() (io.ReadCloser, error) {
@@ -184,7 +186,6 @@ func trySystemd() (*OomParser, error) {
 	return &OomParser{
 		ioreader: bufio.NewReader(readcloser),
 	}, nil
-
 }
 
 // List of possible kernel log files. These are prioritized in order so that
@@ -193,7 +194,7 @@ var kernelLogFiles = []string{"/var/log/kern.log", "/var/log/messages", "/var/lo
 
 // looks for system files that contain kernel messages and if one is found, sets
 // the systemFile attribute of the OomParser object
-func getSystemFile() (string, error) {
+func getLogFile() (string, error) {
 	for _, logFile := range kernelLogFiles {
 		if utils.FileExists(logFile) {
 			glog.Infof("OOM parser using kernel log file: %q", logFile)
@@ -203,22 +204,29 @@ func getSystemFile() (string, error) {
 	return "", fmt.Errorf("unable to find any kernel log file available from our set: %v", kernelLogFiles)
 }
 
-// initializes an OomParser object and calls getSystemFile to set the systemFile
-// attribute.  Returns and OomParser object and an error
+func tryLogFile() (*OomParser, error) {
+	logFile, err := getLogFile()
+	if err != nil {
+		return nil, err
+	}
+	tail, err := tail.NewTail(logFile)
+	if err != nil {
+		return nil, err
+	}
+	return &OomParser{
+		ioreader: bufio.NewReader(tail),
+	}, nil
+}
+
+// initializes an OomParser object. Returns an OomParser object and an error.
 func New() (*OomParser, error) {
 	parser, err := trySystemd()
 	if err == nil {
 		return parser, nil
 	}
-	systemFile, err := getSystemFile()
-	if err != nil {
-		return nil, err
+	parser, err = tryLogFile()
+	if err == nil {
+		return parser, nil
 	}
-	file, err := os.Open(systemFile)
-	if err != nil {
-		return nil, err
-	}
-	return &OomParser{
-		ioreader: bufio.NewReader(file),
-	}, nil
+	return nil, err
 }
--- a/utils/tail/tail.go
+++ b/utils/tail/tail.go
@@ -1,0 +1,157 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package tail implements "tail -F" functionality following rotated logs
+package tail
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"golang.org/x/exp/inotify"
+)
+
+const (
+	readerStateOpening = 1 << iota
+	readerStateOpened
+	readerStateError
+)
+
+type Tail struct {
+	reader      *bufio.Reader
+	readerState int
+	readerLock  sync.Mutex
+	filename    string
+	file        *os.File
+	stop        chan bool
+	watcher     *inotify.Watcher
+}
+
+const retryOpenInterval = time.Second
+const maxOpenAttempts = 3
+
+// NewTail starts opens the given file and watches it for deletion/rotation
+func NewTail(filename string) (*Tail, error) {
+	t := &Tail{
+		filename: filename,
+	}
+	var err error
+	t.stop = make(chan bool)
+	t.watcher, err = inotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("inotify init failed on %s: %v", t.filename, err)
+	}
+	t.readerState = readerStateOpening
+	go t.watchLoop()
+	return t, nil
+}
+
+// Read implements the io.Reader interface for Tail
+func (t *Tail) Read(p []byte) (int, error) {
+	t.readerLock.Lock()
+	defer t.readerLock.Unlock()
+	if t.reader == nil {
+		if t.readerState == readerStateOpening {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("can't open log file %s", t.filename)
+	}
+	return t.reader.Read(p)
+}
+
+var _ io.Reader = &Tail{}
+
+// Close stops watching and closes the file
+func (t *Tail) Close() {
+	close(t.stop)
+}
+
+func isEvent(event *inotify.Event, flag uint32) bool {
+	return event.Mask&flag == flag
+}
+
+func (t *Tail) fileChanged() error {
+	for {
+		select {
+		case event := <-t.watcher.Event:
+			// We don't get IN_DELETE because we are holding the file open
+			if isEvent(event, inotify.IN_ATTRIB) || isEvent(event, inotify.IN_MOVE_SELF) {
+				return nil
+			}
+		case <-t.stop:
+			return fmt.Errorf("watch was cancelled")
+		}
+	}
+}
+
+func (t *Tail) attemptOpen() (err error) {
+	t.readerLock.Lock()
+	defer t.readerLock.Unlock()
+	for attempt := 1; attempt <= maxOpenAttempts; attempt++ {
+		glog.V(4).Infof("Opening %s (attempt %d of %d)", t.filename, attempt, maxOpenAttempts)
+		t.file, err = os.Open(t.filename)
+		if err == nil {
+			// TODO: not interested in old events?
+			//t.file.Seek(0, os.SEEK_END)
+			t.reader = bufio.NewReader(t.file)
+			t.readerState = readerStateOpened
+			return nil
+		}
+		select {
+		case <-time.After(retryOpenInterval):
+		case <-t.stop:
+			t.readerState = readerStateError
+			return fmt.Errorf("watch was cancelled")
+		}
+	}
+	t.readerState = readerStateError
+	return err
+}
+
+func (t *Tail) watchLoop() {
+	for {
+		err := t.watchFile()
+		if err != nil {
+			glog.Errorf("Tail failed on %s: %v", t.filename, err)
+			break
+		}
+	}
+}
+
+func (t *Tail) watchFile() error {
+	err := t.attemptOpen()
+	if err != nil {
+		return err
+	}
+	defer t.file.Close()
+	err = t.watcher.Watch(t.filename)
+	if err != nil {
+		return err
+	}
+	defer t.watcher.RemoveWatch(t.filename)
+	err = t.fileChanged()
+	if err != nil {
+		return err
+	}
+	glog.V(4).Infof("Log file %s moved/deleted", t.filename)
+	t.readerLock.Lock()
+	defer t.readerLock.Unlock()
+	t.readerState = readerStateOpening
+	return nil
+}


utils/oomparser/oomparser.go
```diff
@@ -202,22 +201,29 @@ func getSystemFile() (string, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	return &OomParser{
-		ioreader: bufio.NewReader(file),
+		ioreader: bufio.NewReader(tail),
 	}, nil
 }
+
+// initializes an OomParser object. Returns and OomParser object and an error.
```
ncdc: s/and/an/

utils/tail/tail.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+			fs, err := os.Stat(t.filename)
+			if err != nil || !os.SameFile(origfs, fs) {
+				glog.Infof("File %s deleted/moved", t.filename)
+				break
+			}
+		}
+		t.file.Close()
+		attempt := 1
+		for attempt <= maxReopenAttempts {
```
timstclair: Use a `for atempt := 0; attempt < maxReopenAttempts; attempt++` style loop (I don't think you're incrementing attempt)

utils/tail/tail.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (t *Tail) watchFile() {
+	for {
+		origfs, err := os.Stat(t.filename)
+		if err != nil {
+			glog.Errorf("Stat failed on %s", t.filename)
+			return
+		}
+		for {
+			time.Sleep(pollInterval)
```
timstclair: delete this?

utils/tail/tail.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		t.file.Close()
+		attempt := 1
+		for attempt <= maxReopenAttempts {
+			glog.Infof("Reopening %s (attempt %d of %d)", t.filename, attempt, maxReopenAttempts)
+			t.file, err = os.Open(t.filename)
+			if err == nil {
+				break
+			}
+			time.Sleep(pollInterval)
```
timstclair: Stop should work here.

timstclair:
Can we just use inotify?

utils/tail/tail.go
```diff
@@ -0,0 +1,106 @@
...
+func NewTail(filename string) (t *Tail, err error) {
+	t = &Tail{
+		filename: filename,
+	}
+	t.file, err = os.Open(t.filename)
+	if err != nil {
+		glog.Errorf("Open failed on %s", t.filename)
+		return nil, err
+	}
+	t.file.Seek(0, os.SEEK_END)
```
timstclair: This is a change in behavior, as we won't get old events.

utils/tail/tail.go
```diff
@@ -0,0 +1,106 @@
...
+
+const maxReopenAttempts = 3
+
+func (t *Tail) watchFile() {
+	for {
+		origfs, err := os.Stat(t.filename)
+		if err != nil {
+			glog.Errorf("Stat failed on %s", t.filename)
+			return
+		}
```
timstclair: I think this check is redundant.

utils/tail/tail.go
sjenning: that is true. i figured it was bug before, sending oom events for things that may have happened days ago in the logs.

sjenning:
@timstclair i'm looking into using inotify.  might be less of a hassle than fsnotify was.

utils/tail/tail.go
timstclair: Things from days ago are probably not relevant, but events from seconds ago might be. Either way, I think it's out of scope for this PR (feel free to leave a TODO)

sjenning:
@timstclair @ncdc updated with inotify instead of polling

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+				// we don't get IN_DELETE because we are holding the file open
+				if isEvent(event, inotify.IN_ATTRIB) || isEvent(event, inotify.IN_MOVE_SELF) {
+					reopen = true
+				}
+			case <-t.stop:
+				watcher.Close()
+				return
+			}
+			if reopen {
+				break
```
ncdc: I'm pretty sure this will break you out of the outer loop. I'd recommend adding some unit tests :)

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+		}
+		glog.Infof("Log file %s moved/deleted", t.filename)
+		watcher.RemoveWatch(t.filename)
+		// delay the close of the old file until we have the new reader up
+		oldfile := t.file
+		// wait a little before trying reopen or the first attempt will always fail
+		time.Sleep(100 * time.Millisecond)
```
ncdc: Artificial initial delays are bad. Can we do something else? Run os.Stat() in a loop or something?

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+const maxReopenAttempts = 3
+
+// NewTail starts opens the given file and watches it for deletion/rotation
+func NewTail(filename string) (t *Tail, err error) {
+	t = &Tail{
+		filename: filename,
+	}
+	t.file, err = os.Open(t.filename)
+	if err != nil {
+		glog.Errorf("Open failed on %s", t.filename)
```
timstclair: nit: I prefer to only log where there error is actually captured, other wise you end up with the same error being logged multiple times. Instead, consider adding some context to the error (e.g. `fmt.Errorf("open failed on %s: %v", t.filename, err)`)

ncdc: +1

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+				return
+			}
+			if reopen {
+				break
+			}
+		}
+		glog.Infof("Log file %s moved/deleted", t.filename)
+		watcher.RemoveWatch(t.filename)
+		// delay the close of the old file until we have the new reader up
+		oldfile := t.file
```
timstclair: Why don't you move this whole block (here to the end) into a function, that way you can just use `defer oldfile.close()` here.

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+		// wait a little before trying reopen or the first attempt will always fail
+		time.Sleep(100 * time.Millisecond)
+		for attempt := 1; attempt <= maxReopenAttempts; attempt++ {
+			glog.Infof("Reopening %s (attempt %d of %d)", t.filename, attempt, maxReopenAttempts)
```
timstclair: nit: Let's make this `V(4)`

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+func (t *Tail) Read(p []byte) (n int, err error) {
+	return t.reader.Read(p)
+}
+
+// type assertion
+var _ io.Reader = &Tail{}
+
+// Close stops watching and closes the file
+func (t *Tail) Close() {
+	t.file.Close()
```
timstclair: `t.file` could be nil here. Probably best to just let the watchFile routine handle closing.

utils/tail/tail.go
sjenning: it breaks me out of the for loop starting line 90 which is what i want.  i've done tests so i know it works.  i'll have to think about how to unit test it.

ncdc: My bad, had it backwards

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+			select {
+			case <-time.After(retryReopenInterval):
+			case <-t.stop:
+				oldfile.Close()
+				return
+			}
+		}
+		if err != nil {
+			glog.Errorf("Reopen failed on %s", t.filename)
+			// don't close old file or the oomparser will just hot loop
```
timstclair: I'd prefer to close the file here, and fix the hot-looping behavior. I think you should modify `Read` to return EOF in this case.

utils/tail/tail.go
```diff
@@ -0,0 +1,133 @@
...
+	watcher, err := inotify.NewWatcher()
+	if err != nil {
+		glog.Errorf("Inotify init failed: %s", err.Error())
+		return
+	}
+	for {
+		err := watcher.Watch(t.filename)
+		if err != nil {
+			glog.Errorf("Inotify watch failed: %s", err.Error())
+			return
```
timstclair: I think there's a lot that needs to be closed here, and similarly for returns below. I think this can all be refactored into several functions and rely on defers to close everything.

utils/tail/tail.go
sjenning: in a hot loop?  while the normal case is that the file will be recreated in a short amount of time, it also might not be.  how long would we do this loop before timing out?  seems like it's just an idle wait vs a busy one.

ncdc: not a hot loop. but i'd rather try immediately and then sleep instead of sleeping up front

sjenning:
@timstclair new update that i think captures all the previous comments. PTAL when you can.

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+	// TODO: not interested in old events?
+	//t.file.Seek(0, os.SEEK_END)
+	t.reader = bufio.NewReader(t.file)
+	t.stop = make(chan bool)
+	t.watcher, err = inotify.NewWatcher()
+	if err != nil {
+		glog.Errorf("Inotify init failed on %s: %v", t.filename, err)
+		return
```
timstclair: nit: [avoid naked returns](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	go t.watchFile()
+	return t, nil
+}
+
+// Read implements the io.Reader interface for Tail
+func (t *Tail) Read(p []byte) (n int, err error) {
+	n, err = t.reader.Read(p)
+	if err != nil {
+		return 0, io.EOF
```
timstclair: EOF should be used to signal there is no more data, but in this case it is used to signal that `readLinesFromFile` should retry. I think it would be better to just return 0 bytes read and nil error in this case (unless `err` is an actual error and not EOF). The logic in `readLinesFromFile`  will need to be updated too.

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		glog.Errorf("Inotify init failed on %s: %v", t.filename, err)
+		return
+	}
+	go t.watchFile()
+	return t, nil
+}
+
+// Read implements the io.Reader interface for Tail
+func (t *Tail) Read(p []byte) (n int, err error) {
```
timstclair: nit: these named returns don't give any additional information

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+
+var _ io.Reader = &Tail{}
+
+// Close stops watching and closes the file
+func (t *Tail) Close() {
+	close(t.stop)
+}
+
+func isEvent(event *inotify.Event, flag uint32) bool {
+	if event.Mask&flag == flag {
```
timstclair: nit: `return event.Mask&flag == flag`

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+			if isEvent(event, inotify.IN_ATTRIB) || isEvent(event, inotify.IN_MOVE_SELF) {
+				return nil
+			}
+		case <-t.stop:
+			return fmt.Errorf("Watch was cancelled")
+		}
+	}
+}
+
+func (t *Tail) attemptReopen() (err error) {
```
timstclair: nit: named result again ([don't stutter](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters))

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+		t.file, err = os.Open(t.filename)
+		if err == nil {
+			return nil
+		}
+	}
+	return err
+}
+
+func (t *Tail) watchFile() {
+	defer t.file.Close()
```
timstclair: This will close whatever `t.file` was when `WatchFile` _started_, not when it _finished_. See [defer](https://golang.org/doc/effective_go.html#defer).

I suggest flipping the inner loop segments (open the file then watch then close -- rather than watch, close, open), and encapsulating the loop body in a function, and doing `defer t.file.Close()` in that function.

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+				return nil
+			}
+		case <-t.stop:
+			return fmt.Errorf("Watch was cancelled")
+		}
+	}
+}
+
+func (t *Tail) attemptReopen() (err error) {
+	for attempt := 1; attempt <= maxReopenAttempts; attempt++ {
```
timstclair: nit: count from 0

utils/tail/tail.go
```diff
@@ -0,0 +1,141 @@
...
+		glog.Errorf("Inotify init failed on %s: %v", t.filename, err)
+		return
+	}
+	go t.watchFile()
+	return t, nil
+}
+
+// Read implements the io.Reader interface for Tail
+func (t *Tail) Read(p []byte) (n int, err error) {
+	n, err = t.reader.Read(p)
```
timstclair: You have a race here on `t.reader` with the `watchFile` goroutine. Please add a unit test to catch this.

timstclair:
What's the status on this? It would be good to get this fix into k8s 1.3.

timstclair:
BTW - why did you decide not to use https://github.com/hpcloud/tail?

timstclair:
ok to test

sjenning:
@timstclair i would need to figure out how to incorperate your feedback.  some of the things you point out, i don't know how to deal with cleanly.

re: hpcloud/tail it brought in 10k lines of vendor code and the fsnotify implementation is broken:
https://github.com/hpcloud/tail/issues/21

ncdc:
@sjenning fyi there's a contrib PR (https://github.com/kubernetes/contrib/pull/937) that uses hpcloud/tail so I don't think it would be unreasonable if the fsnotify bug were fixed.

timstclair:
Which parts are you having trouble with? The race condition can be avoided by protecting the reader with a mutex, and you can punt on the unit test for now.

timstclair:
FYI, we're hoping to cut a v0.23.2 release tomorrow (2016-05-18) for kubernetes 1.3, and I'd like to get this change in. Please ping me if there's anything I can help with to move this along.

sjenning:
I'll fix it up asap. Update within the hour.

timstclair:
Great, thanks!

utils/tail/tail.go
sjenning: i counted from one here since the message "attempt 0 of 3" isn't human friendly

utils/tail/tail.go
sjenning: how do i avoid these?

timstclair: Just change this line to `return t, err` or `return nil, err`. Also change the return type to `(*Tail, error)` since the variable names don't add anything in that case.

IMHO there are only 2 reasons to give return types names:
1. If the return type is something like `(int, int, int, int)`, names distinguish the different ints (though I would usually prefer a struct for this)
2. If you need to modify a return value in a `defer` statement (again, this should be avoided if possible)

utils/tail/tail.go
sjenning: i'm not following here

sjenning:
@timstclair I'm close to having this done.  Another half hour.

utils/tail/tail.go
timstclair: Nevermind. You may want to consider propagating the same error though (i.e. `return 0, err`), or was there a reason you always return EOF here?

sjenning:
@timstclair i've pushed an update.  i reworked watchFile, but now the mutex handling is nasty in there.  recommendations welcome.

utils/tail/tail.go
```diff
@@ -0,0 +1,142 @@
...
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"golang.org/x/exp/inotify"
+)
+
+type Tail struct {
+	reader     *bufio.Reader
+	sync.Mutex // protects reader
```
timstclair: Call this `readerLock` (if it's embedded it's less clear what it's protecting)

utils/tail/tail.go
```diff
@@ -0,0 +1,142 @@
...
+	t := &Tail{
+		filename: filename,
+	}
+	var err error
+	t.stop = make(chan bool)
+	t.watcher, err = inotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("inotify init failed on %s: %v", t.filename, err)
+	}
+	t.Lock()
```
timstclair: Why are you locking here?

sjenning: To block readers before the file is open

utils/tail/tail.go
```diff
@@ -0,0 +1,142 @@
...
+func (t *Tail) watchFile() error {
+	err := t.attemptOpen()
+	if err != nil {
+		t.Unlock()
+		return err
+	}
+	defer t.file.Close()
+	// TODO: not interested in old events?
+	//t.file.Seek(0, os.SEEK_END)
+	t.reader = bufio.NewReader(t.file)
```
timstclair: Move this to attemptOpen. Just lock the reader for the whole method (I think that's safer anyway since we don't want to read from the old reader), and unlock with defer.

utils/tail/tail.go
```diff
@@ -0,0 +1,142 @@
...
+	t.Unlock()
+	err = t.watcher.Watch(t.filename)
+	if err != nil {
```
timstclair: nit: shorten this to
```
if t.watcher.Watch(t.filename) != nil { ...
```

Same below.

sjenning: if i don't capture the error, i can't return it in the block

sjenning:
@timstclair ok, i incorporated your comments and introduced a readerState so that the locking can be cleaner.  the real purpose is to avoid the race between a reader and the log file opening.

utils/tail/tail.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	defer t.watcher.RemoveWatch(t.filename)
+	err = t.fileChanged()
+	if err != nil {
+		return err
+	}
+	glog.V(4).Infof("Log file %s moved/deleted", t.filename)
+	t.readerLock.Lock()
+	defer t.readerLock.Unlock()
+	t.readerState = readerStateOpening
```
timstclair: Did you mean to set reader to nil here as well? I don't think setting the state does anything otherwise.

utils/tail/tail.go
```diff
@@ -0,0 +1,157 @@
...
+	t := &Tail{
+		filename: filename,
+	}
+	var err error
+	t.stop = make(chan bool)
+	t.watcher, err = inotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("inotify init failed on %s: %v", t.filename, err)
+	}
+	t.readerState = readerStateOpening
```
timstclair: I'd prefer to call this `readerErr`, and have `nil` represent the working (opening) state. That way if there is an error, you can provide more context around it.

timstclair:
Thanks, this looks much better. Just a couple small things, then LGTM.

timstclair:
Taking this as-is, will follow up with a couple fixes. Thanks!

PR #1265, opened 2016-05-05
Clean up storage driver docs

timstclair:
Also fix `-storage_driver` flag usage to show all the registered drivers as options:
```
-storage_driver driver
  storage driver to use. Data is always cached shortly in memory, this controls where data is pushed besides the local cache. Empty means none. Options are: <empty>, bigquery, elasticsearch, influxdb, kafka, redis, statsd, stdout
```


docs/storage/README.md
```diff
@@ -0,0 +1,14 @@
+# cAdvisor Storage Plugins
+
+cAdvisor supports exporting stats to various storage driver plugins. To enable a storage driver, set the `-storage_driver` flag.
```
pwittrock: 1 or 2 dashes (-)?

storagedriver.go
```diff
@@ -31,18 +34,28 @@ import (
...
+	storageDriver   string
+	storageDuration = flag.Duration("storage_duration", 2*time.Minute, "How long to keep data stored (Default: 2min).")
+)
+
+func init() {
+	// Add storage driver flag.
+	options := storage.ListDrivers()
+	sort.Strings(options)
+	storageDriverUsage := fmt.Sprintf("Storage `driver` to use. Data is always cached shortly in memory, this controls where data is pushed besides the local cache. Empty means none. Options are: <empty>, %s", strings.Join(options, ", "))
+	flag.StringVar(&storageDriver, "storage_driver", "", storageDriverUsage)
```
pwittrock: why is this flag initialized differently than storage_duration?

timstclair: Because I wanted to dynamically create the usage message. Options for making it more consistent:
1. Keep storageDriver a `*string`, and use the `flag.String` method
2. Move storageDuration initiallization into init()
3. Make a function for getting the sorted options, inline the usage creation

WDYT?

docs/storage/README.md
timstclair: Doesn't matter (I don't like go's flag library), but 1-dash seems more consistent with the other storage docs.

storagedriver.go
pwittrock: I'd vote for 2) if you wouldn't mind doing it.

timstclair: Went with 3.

timstclair:
Addressed comments.

pwittrock:
lgtm

PR #1269, opened 2016-05-05
version checking for rkt + godep changes

sjpotter:
so it seems some of our godeps no longer need to be there.  add blang/semver and added version checking for rkt

pulled it mostly from https://github.com/google/cadvisor/pull/1222 as changed approach a bit


container/rkt/client.go
```diff
@@ -55,7 +57,31 @@ func Client() (rktapi.PublicAPIClient, error) {
...
+			rktClientErr = fmt.Errorf("rkt: couldn't parse RtVersion: %v", err)
+			return
+		}
+		minVersion, err := semver.Make(minimumRktBinVersion)
+		if err != nil {
+			rktClientErr = fmt.Errorf("rkt: couldn't parse minimumRktBinVersion: %v", err)
+			return
+		}
+		result := binVersion.Compare(minVersion)
+		if result < 0 {
```
timstclair: nit: just use `if binVersion.LT(minVersion) {`

container/rkt/client.go
```diff
@@ -55,7 +57,31 @@ func Client() (rktapi.PublicAPIClient, error) {
...
+		binVersion, err := semver.Make(resp.Info.RktVersion)
+		if err != nil {
+			rktClientErr = fmt.Errorf("rkt: couldn't parse RtVersion: %v", err)
+			return
+		}
+		minVersion, err := semver.Make(minimumRktBinVersion)
+		if err != nil {
+			rktClientErr = fmt.Errorf("rkt: couldn't parse minimumRktBinVersion: %v", err)
+			return
+		}
```
timstclair: just use MustParse, since it's hard coded (add a unit test)

timstclair:
LGTM

PR #1272, opened 2016-05-05
Added per-subcontainer charts to the UI

### LINKED ISSUE #1142: No common performance graph for all containers

ainsleyc:
I believe this resolves #1142.


vishh:
ok to test

info/v1/container.go
```diff
@@ -110,6 +110,10 @@ type ContainerInfoRequest struct {
 	// End time for which to query information.
 	// If ommitted, current time is assumed.
 	End time.Time `json:"end,omitempty"`
+
+	// Flag to determine if expanded subcontainer stats are requested.
+	// If ommitted, assume false.
+	ExpandSubcontainers bool `json:"expand_subcontainers,omitempty"`
```
timstclair: There's a separate endpoint for this: `http://cadvisor:8080/api/v1.1/subcontainers` (note that it's not available in 1.0).

pages/assets/js/containers.js
```diff
@@ -282,6 +296,58 @@ function drawCpuUsageBreakdown(elementId, machineInfo, containerInfo) {
...
+  }
+
+  var titles = ['Time'];
+  var data = [];
+  for (var i = 0; i < subcontainerInfos.length; i++) {
+    titles.push(subcontainerInfos[i].name);
+    for (var j = 1; j < subcontainerInfos[i].stats.length; j++) {
+      var cur = subcontainerInfos[i].stats[j];
+      var prev = subcontainerInfos[i].stats[j - 1];
+      var intervalInNs = getInterval(cur.timestamp, prev.timestamp);
```
timstclair: What is "Ns" here?

pages/assets/js/containers.js
```diff
@@ -282,6 +296,58 @@ function drawCpuUsageBreakdown(elementId, machineInfo, containerInfo) {
...
+
+      var elements = [cur.timestamp];
+      subcontainerInfos.forEach(function() {
+        elements.push(null);
```
timstclair: I don't understand what this is doing.

pages/assets/js/containers.js
timstclair: Ah, nanoseconds. Drop the "In"? Just `intervalNs`

pages/assets/js/containers.js
```diff
@@ -282,6 +296,58 @@ function drawCpuUsageBreakdown(elementId, machineInfo, containerInfo) {
...
 
+// Draw the graph for per-subcontainer CPU usage.
+function drawCpuPerSubcontainerUsage(elementId, subcontainerInfos) {
+  if (!hasResourceForAll(subcontainerInfos, 'cpu')) {
+    return;
+  }
+
+  var titles = ['Time'];
+  var data = [];
+  for (var i = 0; i < subcontainerInfos.length; i++) {
```
timstclair: How does this handle 200+ containers? Perhaps it should only include the top N.

pages/assets/js/containers.js
```diff
@@ -282,6 +296,58 @@ function drawCpuUsageBreakdown(elementId, machineInfo, containerInfo) {
...
+      var cur = subcontainerInfos[i].stats[j];
+
+      var elements = [cur.timestamp];
+      subcontainerInfos.forEach(function() {
+        elements.push(null);
+      });
+      elements[i+1] = cur.memory.usage / oneMegabyte;
+      data.push(elements);
+    }
+  }
```
timstclair: nit: Dedup this with the above method, i.e. `function getSubcontainerChartData(subcontainerInfos, dataFn)`, where `dataFn` is a function that generates the data points.

timstclair:
Review pass complete.

pages/assets/js/containers.js
ainsleyc: For the subcontainers, the timestamps of the metrics are not aligned. If I understand the DataTables API correctly, this means I need to generate a series of sparse arrays of equal length, one for each unique timestamp. Ex:

[timestamp1, null, null, 23],
[timestamp2, null, 322, null],
[timestamp3, 664, null, null],
...

ainsleyc:
Addressed comments from review.

ainsleyc:
Haha... seems the disk on the machine Jenkins is running on is full?

timstclair:
> Haha... seems the disk on the machine Jenkins is running on is full?
￼
@pwittrock - how do we fix this? Is there something we can do to prevent this from happening again?

pages/assets/js/containers.js
timstclair: I see. Please add a comment explaining this. Also, does it matter if the data points are out of order?

pages/assets/js/containers.js
```diff
@@ -270,18 +287,71 @@ function drawCpuUsageBreakdown(elementId, machineInfo, containerInfo) {
...
+}
+
+// Draw the graph for per-subcontainer CPU usage.
+function drawCpuPerSubcontainerUsage(elementId, subcontainerInfos) {
+  if (!hasResourceForAll(subcontainerInfos, 'cpu')) {
+    return;
+  }
+
+  var chartData = getSubcontainerChartData(subcontainerInfos,
+      function(cur, prev, intervalNs) {
```
timstclair: (nit) drop intervalNs from the function API, and just compute it here.

timstclair:
LGTM, just 2 nits

pages/assets/js/containers.js
ainsleyc: I don't believe the order matters, the library must handle it for you.

ainsleyc:
Addressed comments. Thanks!

timstclair:
Awesome! I tested this locally, and ran into a few issues. Let me know if you'd like to address them in this PR or another.

1. The graph would occasionally sporatically jump, which I suspect might be caused by a different container becoming a "top N" user (and bumping another from the graph).
2. The graph tail disappearing, like this:
3yspe1jytpm
I think this is due to an offset in container collection frequency (the blue line lags), but I'm not sure what to do about it.
3. The legend is paginated with more containers (see screenshot below), but the pagination is reset each time the graph is updated.
swozwzeghmf

ainsleyc:
1. Yeah, this is probably caused by the sorting based on the simplistic average cpu/memory. Let me try re-sorting the sliced top N containers by name before displaying them, that will definitely reduce the churn at least in the likely common case where N > number of containers.

2. As you say, I think this one is caused by time offset between the subcontainers. I could try to use some heuristic to detect how much to trim, but might be discarding information relevant to the user. I'm inclined to leave it as-is, unless you have a suggestion how it might be done?

3. I think this is true of all the charts. Happy to take it on, though let's do it in a separate commit.

timstclair:
SGTM, let me know once you've made the change to sort by name, and I'll merge it.

ainsleyc:
Added the sorting of the slices by name, let me know if that's not working for your use case.

timstclair:
LGTM

PR #1274, opened 2016-05-07
Delete unused ListThreads container method

timstclair:


vishh:
LGTM. Although a better change would be to implement it :smile:

timstclair:
> a better change would be to implement it 
It's not just a matter of implementing it though. To be useful, it also needs to be wired through to all the endpoints: REST API, clients, and UI. If you think it's valuable we should open an issue for it, but as is I think it was just a confusing bit for new runtimes (rkt) to deal with.

vishh:
Yeah. We haven't had an explicit request for it *yet*. So this PR makes
sense for now...

PR #1275, opened 2016-05-07
Docker state cleanup

timstclair:
- Don't validate docker state file, since it's no longer used
- Clean up docker root dir

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/container/libcontainer"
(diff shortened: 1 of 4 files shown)


timstclair:
Failure was from a merge conflict.

container/docker/factory.go
```diff
@@ -34,31 +35,33 @@ import (
...
-func DockerStateDir() string {
-	return libcontainer.DockerStateDir(*dockerRootDir)
-}
+var (
+	// Basepath to all container specific information that libcontainer stores.
+	dockerRootDir string
 
-const (
-	dockerRootDirKey = "Root Dir"
+	dockerRootDirFlag = flag.String("docker_root", "/var/lib/docker", "DEPRECATED: docker root is read from docker info (this is a fallback, default: /var/lib/docker)")
```
vishh: Ideally, we should get rid of this flag and use docker CLI directly. That requires a retry from other consumers until docker is available though.

vishh:
LGTM

container/docker/factory.go
timstclair: I'm confused by this comment - I think that's what this PR does? I left the flag in (and marked it deprecated) in case anyone was using it (to not break the CLI), but I think we should just remove it in a couple releases.

vishh: I meant that we should not fallback at all ideally. If docker is not even around, why do we even need docker's root directory. The current code structure where pkg `fs` depends on `container/docker` makes this difficult. 
We cannot deprecate it until the clients of this package can deal with the fact that docker might not be available when cAdvisor is initialized.

timstclair: I see. That makes sense, but do we need to support this case? It's already the case that if cAdvisor can't connect to docker at start-up the docker factory isn't registered (https://github.com/google/cadvisor/blob/master/container/docker/factory.go#L176)

PR #1204, opened 2016-04-13
devicemapper thin_ls support

pmorie:
WIP

@vishh

PR #1277, opened 2016-05-09
Add support to report container's filesystem stats

shilpamayanna:
This fix includes support to report container’s filesystem stats. 
Added support to identify container’s mounts and report disk statistics for the same.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -202,6 +202,24 @@ func newDockerContainerHandler(
 		}
 	}
 
+	dockerStatus, err := Status()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dockerStatus : %v", id, err)
+	}
+
+	fsContext := fs.Context{
+		Docker: fs.DockerContext{
+			Root:         RootDir(),
+			Driver:       dockerStatus.Driver,
+			DriverStatus: dockerStatus.DriverStatus,
+		},
+	}
+
+	handler.fsInfo, err = fs.ContainerFsInfo(fsContext, handler.pid)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get mounted filesystems for container %q: %v", id, err)
+	}
+
 	return handler, nil
 }
 
@@ -249,8 +267,15 @@ func (self *dockerContainerHandler) getFsStats(stats *info.ContainerStats) error
 	if self.ignoreMetrics.Has(container.DiskUsageMetrics) {
 		return nil
 	}
+
+	if self.fsInfo == nil {
+		return nil
+	}
+
 	switch self.storageDriver {
 	case aufsStorageDriver, overlayStorageDriver, zfsStorageDriver:
+	case devicemapperStorageDriver:
+		return self.getContainerFsStats(stats)
 	default:
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


vishh:
@shilpamayanna Thanks for the PR. I don't think the logic is valid. Take a look at #1204 which is an attempt to add filesystem metrics support for device mapper storage backend.

shilpamayanna:
Thanks @vishh. With this PR, we are trying to add support to obtain filesystem stats for all device files used by container. 
We can leverage the support from PR #1204 for thin-provisioned devices.
However, the implementation doesn’t seem to handle the block devices shown in the example below.
nvme1n1                     259:0    0    10G  0 disk /var/lib/kubelet/pods/d3162804-1222-11e6-9aea-2c600c82eccc/volumes/datawise.io~volume/data
nvme2n1                     259:1    0    10G  0 disk /var/lib/kubelet/pods/d30a9040-1222-11e6-9aea-2c600c82eccc/volumes/datawise.io~volume/data

Could you please let me know if there is any active work around this area to obtain stats for these devices?

vishh:
@shilpamayanna IIUC, you intend to add support for getting stats for volumes mounted into containers. Given that volumes are essentially mounts, there is no guarantee that these mounts will be backed by devices. They can be arbitrary directories on disk, that are potentially shared between containers.

If you are using kubernetes, take a look at the output of summary API - `http://localhost:10255/stats/summary` on your nodes. Kubelet collects volumes stats. We can extend that to collect stats for your volume type easily. Its basically a TODO.

shilpamayanna:
@vishh Thanks for the reply. There is a check to ensure that the statistics are retrieved for block devices only and other mounts that are not backed by devices are skipped. If the mounts are shared between containers, the stats are collected for all the containers sharing the device.

Summary API provides metrics for the filesystem usage and capacity only. 
Is there a plan to add support for disk metrics?

vishh:
@shilpamayanna This PR is incompatible with the v2 version of the API. In v2, [`Filesystem`](https://github.com/google/cadvisor/blob/master/info/v2/container.go#L293) stats do not include Disk metrics. We decided to expose disk metrics as part of the entire machine only. 

Why do you need disk stats for volumes?

shilpamayanna:
@vishh We use disk stats to calculate IOPs used by containers. Is there any other way to compute IOPs?

vishh:
@shilpamayanna [DiskIO](https://github.com/google/cadvisor/blob/master/info/v1/container.go#L301) stats are already exposed per container.

shilpamayanna:
@vishh Thanks for the providing the pointers. I observed that DiskIO stats reported contains device major and minor number. In a cluster with multiple nodes, these numbers could overlap across nodes. We need a way to map these stats to the volume. May be container’s mountpoint would help. 

Also it would be helpful to export the DiskIO stats to heapster.

vishh:
@shilpamayanna As part of overall machine stats you get a dump of all devices with major and minor. You can use that information to identify devices.

I'd suggest opening an issue against heapster to support DiskIO stats.

shilpamayanna:
@vishh Thanks. Will open issue against heapster. 
How can I get the machine stats?

vishh:
`/api/v2.1/machinestats`

vishh:
I'm closing this PR under the assumption that there is no change required from cAdvisor. Feel free to re-open if you disagree @shilpamayanna!

PR #1282, opened 2016-05-12
Add 'release mode' to build script

timstclair:
Build releases with:
```
RELEASE=true build/build.sh
```

Differences in release mode are:
- Don't use cached build artifacts (`go build` instead of `go install`)
- Don't include hostname in the binary, use git email instead
- Day (instead of second) granularity for the BuildDate

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -16,14 +16,27 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


vishh:
LGTM

Issue #1201, opened 2016-04-08
cAdvisor trying to connect to non-existent rkt API on startup


timstclair:
On my local machine (not running rkt), starting cadvisor prints several error messages:
```
$ sudo ./cadvisor
2016/04/08 16:37:51 grpc: Conn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:15441: getsockopt: connection refused"; Reconnecting to "localhost:15441"
2016/04/08 16:37:53 grpc: Conn.resetTransport failed to create client transport: connection error: desc = "transport: dial tcp 127.0.0.1:15441: getsockopt: connection refused"; Reconnecting to "localhost:15441"
2016/04/08 16:37:53 Failed to dial localhost:15441: grpc: timed out trying to connect; please retry.
E0408 16:37:55.525431   18016 manager.go:228] Registration of the rkt container factory failed: unable to get the RktPath variable couldn't GetInfo from rkt api service: rpc error: code = 2 desc = "grpc: the client connection is closing"
```
Also note that it takes ~4 seconds to fail to register the rkt runtime, 

I suggest adding a `-container_handlers` flag which takes a comma-separated list of container handlers to register. The existing flag `-docker_only` (which rkt doesn't currently respect) would be deprecated in favor of the new flag. The raw container handler would still be registered to handle the root container, but would ignore all others (same behavior as `-docker_only`).

/cc @sjpotter @vishh

vishh:
Would it ok if we just suppress those errors? From a usability stand point, it is great if cadvisor just works on a node that is running rkt.

sjpotter:
I guess this is just a function of grpc.  i.e. when you try to connect to the docker daemon and cant, its doing the same thing internally.

sjpotter:
https://github.com/grpc/grpc-go/blob/master/grpclog/logger.go

does it make sense for gprc to use its own logger instead of glog?  worth filing an issue against it?

timstclair:
> Would it ok if we just suppress those errors? From a usability stand point, it is great if cadvisor just works on a node that is running rkt.

I'm less concerned about the error messages, and more concerned with the startup latency. It blocks cAdvisor startup for about 4 seconds while grpc retries (hence the need for https://github.com/google/cadvisor/pull/1200). If we could register the handlers in separate go routines it might not be a problem, but I haven't worked through the implications of doing it concurrently.

sjpotter:
let me try it with a 0s timeout

[...]

PR #1283, opened 2016-05-12
rename dockerOnly flag as its not accurate description anymore

sjpotter:
Should be fairly self explanitory

### DIFF
--- a/container/raw/factory.go
+++ b/container/raw/factory.go
@@ -28,6 +28,7 @@ import (
 	"github.com/golang/glog"
 )
 
+var dockerOnly = flag.Bool("docker_only", false, "Only report docker containers in addition to root stats (deprecated)")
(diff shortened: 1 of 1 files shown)


vishh:
Should we mark the flag as deprecated before removing it?

sjpotter:
can do that too. is their a normal way to do that?

vishh:
We use golang flags package. So deprecation will have to handled manually. Just post a message saying that the flag will be removed in a subsequent release and switch to using the new flag should do.

timstclair:
See my comment on https://github.com/google/cadvisor/issues/1201#issue-147065503

I think `-runtime_only` is a bit weird, and we should just give the option to control which container handlers are enabled (default should be all). As for deprecating, I'd just update the usage message to be `DEPRECATED: Use -container_handlers instead` (or whatever the new flag is).

PR #1286, opened 2016-05-14
Make manager multi-container functions robust to partial failures

timstclair:
This is a slight change in behavior, as cases that would previously return `nil, err` will now include a non-nil result. It is common go practice to ignore the result if there's an error so this shouldn't existing users. However, to take advantage of the partial results will require a code change in most cases.

The exception is `GetRequestedContainersInfo`, which was discarding the errors and returning the partial result, and now returns a `partialFailure`. The change in behavior is reflected in `api/versions.go`, which now needs to check for a result and handle the error in that case.

Addresses https://github.com/kubernetes/kubernetes/issues/25131

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -69,6 +69,8 @@ type Manager interface {
 	GetContainerInfo(containerName string, query *info.ContainerInfoRequest) (*info.ContainerInfo, error)
 
 	// Get V2 information about a container.
+	// Recursive (subcontainer) requests are best-effort, and may return a partial result alongside an
+	// error in the partial failure case.
 	GetContainerInfoV2(containerName string, options v2.RequestOptions) (map[string]v2.ContainerInfo, error)
 
 	// Get information about all subcontainers of the specified container (includes self).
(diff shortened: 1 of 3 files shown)


manager/manager.go
```diff
@@ -1202,3 +1211,19 @@ func getVersionInfo() (*info.VersionInfo, error) {
 		CadvisorRevision:   version.Info["revision"],
 	}, nil
 }
+
+// Helper for accumulating partial failures.
+type partialFailure []string
+
+func (f *partialFailure) append(id, operation string, err error) {
+	*f = append(*f, id, operation, err.Error())
```
vishh: Why not append a string form of the input instead?

vishh:
Just one nit. Otherwise LGTM!

manager/manager.go
timstclair: Yeah, that's way better. No idea what I was thinking...

timstclair:
Thanks, done.

timstclair:
@k8s-bot test this

PR #1288, opened 2016-05-18
Run the integration tests locally with the race detector enabled

timstclair:
The integration test script was changed so that it aborts the testing
when the cAdvisor binary exits with a non-zero error code.

Also in this PR:
- clean up makefile to print more informative output
- change make release to build the release binary

I will follow up with another PR to enable the race detector on the jenkins jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ test:
 	@echo ">> running tests"
 	@$(GO) test -tags test -short -race $(pkgs)
 
-test-integration: build test
+test-integration:
 	@./build/integration.sh
 
 format:
@@ -43,4 +43,4 @@ release:
 docker:
 	@docker build -t cadvisor:$(shell git rev-parse --short HEAD) -f deploy/Dockerfile .
 
-.PHONY: all build docker format release test vet
+.PHONY: all build docker format release test test-integration vet
--- a/build/integration.sh
+++ b/build/integration.sh
@@ -18,17 +18,38 @@ if [[ -n "${JENKINS_HOME}" ]]; then
     exec ./build/jenkins_e2e.sh
 fi
 
-sudo -v || exit 1
+set -e
 
-echo ">> starting cAdvisor locally"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -18,28 +18,30 @@ all: format build test
...
 vet:
 	@echo ">> vetting code"
-	@$(GO) vet $(pkgs)
+	$(GO) vet $(pkgs)
 
 build:
 	@echo ">> building binaries"
-	@./build/assets.sh
-	@./build/build.sh
+	./build/assets.sh
```
vishh: Without `@` doesn't `make` echo all these lines?

build/integration.sh
```diff
@@ -18,17 +18,40 @@ if [[ -n "${JENKINS_HOME}" ]]; then
     exec ./build/jenkins_e2e.sh
 fi
 
-sudo -v || exit 1
+set -e
 
-echo ">> starting cAdvisor locally"
-sudo ./cadvisor --docker_env_metadata_whitelist=TEST_VAR &
+sudo -v  # Refresh sudo credentials
```
vishh: This fails on GCE VMs. Can we just remove it?

Makefile
timstclair: Yeah, do you think that's undesireable? I thought printing the commands that were executing was better for understanding what was happening, but I don't feel too strongly about it.

build/integration.sh
timstclair: Done.

Makefile
vishh: The name of the scripts are not really that descriptive. instead we can have these scripts log what they are doing..

vishh:
LGTM. Feel free to self-merge based on what you decide for https://github.com/google/cadvisor/pull/1288#discussion_r63938199

Makefile
timstclair: Done (reverted Makefile changes).

timstclair:
Rebased & reverted Makefile changes. Will merge once tests pass.

PR #1291, opened 2016-05-18
Release v0.23.2

timstclair:


vishh:
Device mapper needs a kernel module I think. It is worth mentioning that module in the release notes... cc @pmorie @sjenning 

Otherwise lgtm

timstclair:
Updated to mention Device Mapper module. PTAL

vishh:
LGTM

PR #1292, opened 2016-05-19
Fixes for log tailing

timstclair:
IN_ATTRIB inotify events are generated when atime / mtime is changed,
which would cause the tail to be reset, and reread the same log
again (generating duplicate events). Instead, watch the directory for
file delete / move.

Also, use an exponential backoff when retrying opening the file.

/cc @sjenning

### DIFF
--- a/utils/tail/tail.go
+++ b/utils/tail/tail.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -37,8 +38,10 @@ type Tail struct {
 	watcher    *inotify.Watcher
 }
 
-const retryOpenInterval = time.Second
-const maxOpenAttempts = 3
+const (
+	defaultRetryInterval = 100 * time.Millisecond
+	maxRetryInterval     = 30 * time.Second
+)
 
 // NewTail starts opens the given file and watches it for deletion/rotation
 func NewTail(filename string) (*Tail, error) {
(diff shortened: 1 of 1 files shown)


sjenning:
wfm and lgtm, thanks

utils/tail/tail.go
```diff
@@ -105,8 +92,11 @@ func (t *Tail) attemptOpen() error {
 			t.reader = bufio.NewReader(t.file)
```
vishh: Why is the line above commented?

utils/tail/tail.go
```diff
@@ -133,15 +123,24 @@ func (t *Tail) watchFile() error {
 		return err
 	}
 	defer t.file.Close()
-	err = t.watcher.Watch(t.filename)
+
+	watchDir := filepath.Dir(t.filename)
+	err = t.watcher.AddWatch(watchDir, inotify.IN_MOVED_FROM|inotify.IN_DELETE)
```
vishh: Are we sure that a rename is what logrotation does? Sounds most likely, but just checking..

vishh:
LGTM

utils/tail/tail.go
timstclair: It's tied to the TODO. I actually don't think we want to do that...

utils/tail/tail.go
timstclair: Manually verified this works by forcing a logrotate.

PR #1294, opened 2016-05-19
Fix nil-interface partialFailure bug

timstclair:
This slipped through the cracks because `require.NoError` handles nil interfaces differently from a standard `err != nil` check (filed https://github.com/stretchr/testify/issues/311).


vishh:
LGTM

PR #1237, opened 2016-04-22
Compile cAdvisor statically

luxas:
Very simple fix for #1227
Also partially required for #1236

@vishh @djtm

PR #1299, opened 2016-05-20
Link cAdvisor statically.

vishh:
cc @mikedanese

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -45,6 +45,7 @@ if [ "${go_version:0:3}" = "1.4" ]; then
 fi
 
 ldflags="
+  -extldflags '-static'
   -X ${repo_path}/version.Version${ldseparator}${version}
   -X ${repo_path}/version.Revision${ldseparator}${revision}
   -X ${repo_path}/version.Branch${ldseparator}${branch}


timstclair:
What are the concerns around building this way? Is it just that it increases our shipping binary size?

vishh:
It makes it possible to distribute cAdvisor & kubelet without requiring a
controlled base image.

timstclair:
LGTM

timstclair:
This won't affect kubelet though.

vishh:
This PR won't. I'm working on a separate patch for k8s. AFAIK cAdvisor was
the main reason why we did not link statically.

luxas:
Seems like you forgot #1237, which was opened ages ago :( and failed due to internal flakes
However, I'm glad this is in, and I've been experimenting with linking kubelet statically...

vishh:
Oops. Apologies. I sent out the PR while discussing kubelet distribution
issues with my colleagues.
I'm happy to revert this PR in favor of #1237 :)

luxas:
No no. This is good as-is :)
Instead, please review kubernetes/kubernetes#26028 :)

djtm:
Ha, great. :+1:

PR #1301, opened 2016-05-20
Downgrade failure to rgeister runtime factory to warning

timstclair:
It is not an error to fail to register the Docker factory on a system
running only rkt, and vice-versa, so these failures are downgraded from
an Error to a Warning. The raw handler should always be registered.

/cc @sjpotter


vishh:
LGTM

Issue #1300, opened 2016-05-20
[scalability] failed to collect filesystem stats - failed to exec du


timstclair:
Full error:
```
E0520 20:38:01.903710   16716 fsHandler.go:106] failed to collect filesystem stats - failed to exec du - fork/exec /usr/bin/nice: cannot allocate memory
```

To reproduce:
```
$ sudo ./cadvisor --logtostderr &
$ for i in `seq 1 500`; do docker run -d kubernetes/pause; done
```

I suggest we limit the number of parallel execs to a reasonable number.

vishh:
+1

PR #1303, opened 2016-05-20
Cap the maximum consecutive du commands

timstclair:
I picked the number `20` pretty arbitrarily. LMK if you prefer a different number, or have a suggestion for choosing in a scientific manner (e.g. roll 4 d20s).

Fixes https://github.com/google/cadvisor/issues/1300

Also noteworthy, this causes cAdvisor to startup much faster, and not crash docker if there are already 500 containers running.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -44,6 +44,26 @@ const (
 	LabelRktImages    = "rkt-images"
 )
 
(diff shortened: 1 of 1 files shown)


vishh:
LGTM

Issue #1279, opened 2016-05-11
Prometheus label "name" conflicts with docker label "name" of running Image


ZimboBoyd:
We use prometheus to pull the metrics from cadvisor. 

The Host ist running Docker containers based on centos7 

in The Dockerfile of Centos7 there is 
LABEL name="CentOS Base Image" 
here is the Dockerfile  https://github.com/CentOS/sig-cloud-instance-images/blob/f0b1738ab545dccaae07a57eab0fc7356a96d404/docker/Dockerfile 

Now all images based on this Dockerimage will report an additional name="Centos Base Image" as label to prometheus. 

See the double name in the Output of /metrics :

container_tasks_state{build_date="2016-03-04",id="/system.slice/docker-ba3cc714541ccf7d990e9d2553d5f47bb3ffcfb81c47715f7d6fe314d1be4327.scope",image="xxx:5000/nginx",license="GPLv2",name="nginx-2",name="CentOS Base Image",state="uninterruptible",vendor="CentOS"} 0

This confusing Prometheus!

I talked to the prometheus people, and they said it would be best to prefix the label comnig from the docker info with some value like "label_"

Or we could use a flag to ignore the Docker labels

PR #1305, opened 2016-05-23
Preifix the Labels with "label_"

ZimboBoyd:
See https://github.com/google/cadvisor/issues/1279


timstclair:
This should be fixed in the prometheus handler ([here](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L531-L534)), not the docker handler. If you change it in the docker handler, it will break other clients expecting to get the labels un-altered. Also, fixing it in prometheus will fix it for all container runtimes.

PR #1307, opened 2016-05-24
Work to add rkt tests

sjpotter:
So this is not ready as is, I want to add rkt support to the integration tests but they can't be added as is

the PR right now 

1) adds rkt to the integration testing framework (but no tests)

2) create generic namespace version of DockerContainer() AllDockerContainer() manager functions.

3) exposes those namespace version in a v1.4 api calls

4) modifies the rkt handler to work slightly better with them (prefix alias with / like it does to docker)


sjpotter:
@timstclair this can probably be reviewed now, no tests yet, but infrastructure is there

sjpotter:
@k8s-bot test this

api/versions.go
```diff
@@ -62,10 +63,11 @@ func getApiVersions() []ApiVersion {
 	v1_1 := newVersion1_1(v1_0)
 	v1_2 := newVersion1_2(v1_1)
 	v1_3 := newVersion1_3(v1_2)
+	v1_4 := newVersion1_4(v1_3)
 	v2_0 := newVersion2_0()
 	v2_1 := newVersion2_1(v2_0)
 
-	return []ApiVersion{v1_0, v1_1, v1_2, v1_3, v2_0, v2_1}
+	return []ApiVersion{v1_0, v1_1, v1_2, v1_3, v1_4, v2_0, v2_1}
```
timstclair: You don't need to add a new version for this, since it's not changing exsiting behavior. Also, can we expose this as part of th 2.x API instead of 1.x?

api/versions.go
```diff
@@ -28,20 +28,21 @@ import (
...
+	summaryApi             = "summary"
+	statsApi               = "stats"
+	specApi                = "spec"
+	eventsApi              = "events"
+	storageApi             = "storage"
+	attributesApi          = "attributes"
+	versionApi             = "version"
+	psApi                  = "ps"
+	customMetricsApi       = "appmetrics"
+	namespacedContainerApi = "namespacedcontainers"
```
timstclair: I know I said to add a new endpoint, but what do you think about modifying the `containersApi` instead? It looks like it's missing a v2 version, so you could add the v2 version and use the [RequestOptions](https://github.com/google/cadvisor/blob/master/info/v2/container.go#L221) to specify the namespace (the `IdType` field).

integration/framework/framework.go
```diff
@@ -95,21 +101,31 @@ const (
...
 )
 
+type RunArgs struct {
+	// Image to use.
+	Image string
+
+	// Arguments to the CLI.
+	Args []string
+
+	InnerArgs []string
```
timstclair: What are InnerArgs?

integration/framework/framework.go
```diff
@@ -209,37 +244,89 @@ func (self *realFramework) ClientV2() *v2.Client {
 	return self.cadvisorClientV2
 }
 
-func (self dockerActions) RunPause() string {
-	return self.Run(DockerRunArgs{
+func (self rktActions) RuntPause() string {
```
timstclair: I think it would be good to split this file up, and put the rkt-specific code in rkt.go

integration/framework/framework.go
sjpotter: that's just a rename of the DockerRunArgs.  I assumed that args is to the runtime while innerargs are passed to the container?

sjpotter:
So I've made progress on rkt tests, except tht rkt has to run as root.

sjpotter:
@timstclair the above error is expected as jenkins now fails test as rkt isn't running on the machines, what do we want to do now?

timstclair:
From offline discussions, here's how we'd like to handle rkt testing:
- Eventually we'd like to move to a distributed testing model where CoreOS maintains a CI project which we add to our build monitoring
- For the near-term, if you create a GCE VM image (with rkt setup) and share it with us, we'll add it to our e2e test nodes (this way it will hopefully be easier to debug failures in a consistent manner)

With regards to this PR, I think whether rkt tests are run should be explicitly controlled with a test flag (either a build tag or a per-host parameter to the runner).

/cc @vishh

PR #1309, opened 2016-05-24
add cgropu swap usage and export as prometheus metric

f0:
Hi ,

this adds the swap information from memory.stat and exports this via prometheus metrics.

Question: Atm this is implemented with the memory.stat file, maybe ists better to use memory_usage_in_bytes and memsw_usage_in_bytes ?

regards f0

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -338,6 +338,10 @@ type MemoryStats struct {
 	// Units: Bytes.
 	RSS uint64 `json:"rss"`
 
+	// The amount of swap currently used by the processes in this cgroup
+	// Units: Bytes.
+	Swap uint64 `json:"swap"`
+
 	// The amount of working set memory, this includes recently accessed memory,
 	// dirty memory, and kernel memory. Working set is <= "usage".
 	// Units: Bytes.
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -167,6 +167,13 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					return metricValues{{value: float64(s.Memory.RSS)}}
 				},
 			}, {
+				name:      "container_memory_swap",
+				help:      "Container swap usage in bytes.",
+				valueType: prometheus.GaugeValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Memory.Swap)}}
+				},
+			}, {
 				name:      "container_memory_failcnt",
 				help:      "Number of memory usage hits limits",
 				valueType: prometheus.CounterValue,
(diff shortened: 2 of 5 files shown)


f0:
if i take a look at the documentation https://lwn.net/Articles/432224/  i think the toalt_{rss,cache,swap} values should be used, but cadvisor uses the not total prefixed once, is there a reason?

f0:
@vishh  any opinion about the total vs non total statistics?

vishh:
Isn't total_ meant for hierarchical stats instead of just this cgroups' stats?

f0:
@vishh 
 thats true, but the cgroup casn have children cgroups or not?

vishh:
They can. Current cAdvisor API doesn't provide hierarchical stats AFAIK.

f0:
ok, switched back to the non total value

f0:
any chance that this can be merged ?

metrics/prometheus.go
```diff
@@ -437,7 +444,26 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					}
 					return values
 				},
-			}, {
+			},
+
+			{
+				name:      "container_tcp_connections_established",
+				help:      "XXX.",
```
timstclair: TODO?

metrics/prometheus.go
```diff
@@ -437,7 +444,26 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					}
 					return values
 				},
-			}, {
+			},
+
+			{
+				name:      "container_tcp_connections_established",
```
timstclair: This looks unrelated to this PR? Same below.

info/v1/container.go
```diff
@@ -324,6 +324,10 @@ type MemoryStats struct {
 	// Units: Bytes.
 	RSS uint64 `json:"rss"`
 
+	// The amount of swap currently used by the processes in this cgroup
+	// Units: Bytes.
+	Swap uint64 `json:"swap"`
```
timstclair: This should be added to v2 as well.

metrics/prometheus.go
```diff
@@ -142,6 +142,13 @@ func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFu
 					return metricValues{{value: float64(s.Memory.RSS)}}
 				},
 			}, {
+				name:      "container_memory_swap",
+				help:      "container swap usage in bytes.",
```
timstclair: (nit) capitalize container

f0:
comments fixed, unrelated stuff removed

info/v1/container.go
f0: ok i take a look

f0: @timstclair  hm v2 uses the v1 values

f0:
@timstclair  @vishh any outstanding issues?

timstclair:
LGTM. Needs a rebase though.

f0:
@timstclair  rebased

PR #1315, opened 2016-05-31
Prefix Labels with "labels_" for Prometheus

ZimboBoyd:
See https://github.com/google/cadvisor/issues/1279

This time the fix is in the Prometheus Exporter

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -529,7 +529,7 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 
 		for k, v := range container.Spec.Labels {
-			baseLabels = append(baseLabels, sanitizeLabelName(k))
+			baseLabels = append(baseLabels, "label_"+sanitizeLabelName(k))
 			baseLabelValues = append(baseLabelValues, v)
 		}
 		for k, v := range container.Spec.Envs {


ZimboBoyd:
I signed it!

timstclair:
@vishh - This is a breaking change for the cAdvisor-prometheus API but it sounds like a necessary fix. Do you have any additional concerns about this change, or should we just clearly document it in the next release?

vishh:
I'm fine with this change. We should document it though in the next release.

timstclair:
LGTM

timstclair:
Sorry, the tests were never run against this. Please rebase, then I will kick off the tests.

metrics/prometheus.go
```diff
@@ -529,7 +529,7 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 
 		for k, v := range container.Spec.Labels {
-			baseLabels = append(baseLabels, sanitizeLabelName(k))
+			baseLabels = append(baseLabels, "label_"+sanitizeLabelName(k))
```
jimmidyson: Probably need to check for existing matching labels so we don't get any collisions?

jimmidyson:
This is a necessary fix that needs to go in IMO. Just need to protect against collisions & then LGTM.

jimmidyson:
@ZimboBoyd Are you still working on this PR? It not, then do you mind if I pick this up & send in a separate updated PR with the minor change detailed above?

grobie:
👍 Also happy to look into this otherwise, just found a similar problem.

jimmidyson:
@grobie What do you think about the constant prefix, even with no collisions? Happy with this approach?

grobie:
I'm actually in favor of a consistent prefix, makes relabeling a lot easier. And with relabeling you can also still promote labels to non-prefixed ones. I'd be interested to learn why @fabxc finds this a "shaky" behavior though.

A more descriptive prefix might work better than the generic `label_`. Something like `com_docker_label_`?

grobie:
@timstclair Can be closed as https://github.com/google/cadvisor/pull/1402 is merged.

PR #1318, opened 2016-06-03
Fixes a panic when a prometheus endpoint ends with a newline

thomaso-mirodin:
We were experimenting with custom metrics earlier today. When we enabled it on one of our kubelets the next time our prometheus server polled the cadvisor endpoint for metrics the kubelet paniced.

The panic that we got in the logs started with this:

```
panic: runtime error: slice bounds out of range
goroutine 602 [running]:
	github.com/google/cadvisor/collector.(*PrometheusCollector).GetSpec(0xc2082f2d20, 0x0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/collector/prometheus_collector.go:129 +0x7bc
	github.com/google/cadvisor/collector.(*GenericCollectorManager).GetSpec(0xc20888a660, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/collector/collector_manager.go:67 +0x118
	github.com/google/cadvisor/manager.(*containerData).updateSpec(0xc208c925a0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/manager/container.go:465 +0x17e
	github.com/google/cadvisor/manager.(*containerData).GetInfo(0xc208c925a0, 0x0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/manager/container.go:118 +0x83
	github.com/google/cadvisor/manager.(*manager).containerDataToContainerInfo(0xc208578000, 0xc208c925a0, 0xc208b92840, 0xc208f037c0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/manager/manager.go:440 +0x4e
	github.com/google/cadvisor/manager.(*manager).containerDataSliceToContainerInfoSlice(0xc208578000, 0xc208bb4500, 0x92, 0x92, 0xc208b92840, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/k8s.io/kubernetes/Godeps/_workspace/src/github.com/google/cadvisor/manager/manager.go:560 +0x189
[...]

PR #1324, opened 2016-06-08
Cherry picks for Kubernetes 1.3

timstclair:
https://github.com/google/cadvisor/pull/1303
https://github.com/google/cadvisor/pull/1318


timstclair:
@vishh How do I handle the CLA with cherrypicks?

vishh:
We can ignore CLA here. I think we can add a cla human approved label like in the case of kubernetes.

vishh:
LGTM

PR #1328, opened 2016-06-10
Adds in support for collecting custom metrics from a Jolokia endpoint.

mwringe:
This adds support for collecting metrics from a Jolokia endpoint (used for monitoring Java application).

As part of this support it adds in a few new methods to the ContainerHandler:

- GetContainerIPAddress allows us to get the containers ip address (either from the docker image directly if it has it, or from the pod if using Kubernetes). This allows us to not rely on using hostports when running in a Kubernetes environment

- ReadFile allows us to read files from within the docker image. This is useful for accessing secured endpoint as we can fetch the ca cert when dealing with self signed certificates

I also added a tls.config to the manager. We need some way to pass certificates to the http client used to fetch metrics so that we can do things such as certificate based authentication. This allows Kubernetes/OpenShift to pass their own certs which can be used by the system.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/cadvisor/utils/sysfs"
 	"github.com/google/cadvisor/version"
 
+	"crypto/tls"
 	"github.com/golang/glog"
 )
 
@@ -118,7 +119,7 @@ func main() {
 		glog.Fatalf("Failed to create a system interface: %s", err)
 	}
 
-	containerManager, err := manager.New(memoryStorage, sysFs, *maxHousekeepingInterval, *allowDynamicHousekeeping, ignoreMetrics.MetricSet)
+	containerManager, err := manager.New(memoryStorage, sysFs, *maxHousekeepingInterval, *allowDynamicHousekeeping, ignoreMetrics.MetricSet, tls.Config{})
 	if err != nil {
 		glog.Fatalf("Failed to create a Container Manager: %s", err)
 	}
--- a/collector/config.go
+++ b/collector/config.go
@@ -17,6 +17,7 @@ package collector
 import (
 	"time"
 
+	"encoding/json"
 	"github.com/google/cadvisor/info/v1"
 )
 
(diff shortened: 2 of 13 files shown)


mwringe:
Anything I can do to get this looked at?

jimmidyson:
Have you looked at using [Prometheus jmx exporter](https://github.com/prometheus/jmx_exporter) & exposing Prometheus endpoint, reusing existing cadvisor custom metric collection?

This is what we do in fabric8, via [Agent Bond](https://github.com/fabric8io/agent-bond), which combines both jolokia & jmx exporter into a single Java agent.

mwringe:
@jimmidyson Interesting, that can be a useful option for some people. Not eveyone is going to want to add in an extra Prometheus endpoints, especially if they are already using Jolokia endpoints.

The system already support HTTP based endpoints and Prometheus ones. Is there some policy against supporting another endpoint type?

jimmidyson:
There is no policy per se, but adding in an extra custom endpoint type of course adds maintenance going forward. Jolokia only targets a small percentage of target users (JVM containers that use Jolokia). Both http & Prometheus cover a wider target. I'm not dead against adding this in, just want to make sure it's agreed & that there is agreement to maintain it going forward.

Don't get me wrong: I LOVE Jolokia personally (my team mate & friend is the creator of Jolokia), but I'm not sure if it is suitable for cadvisor.

Thoughts @vishh @timstclair?

mwringe:
This PR also fixes a few issues with the current implementation which makes using custom metrics cumbersome, regardless of which endpoint type you are using. 

- we can now gather the ip address directly from the container instead of having to specify the exact url. This means we don't need to know before hand which ip address or hostname the container will be running under. It also means we don't need to use use hostports when dealing with kubernetes pods.

- ability to pass certificates to be used by cadvisor when fetching metrics from endpoints. This allows the endpoint to be secured with certificate based authentication.  We probably still would want to add in more security here (eg basic auth) and we would probably also want to add this to the non-Jolokia endpoints as well

jimmidyson:
Could you split those fixes into separate PRs please, one for the IP address & one for the auth fixes? We can discuss the jolokia endpoint separately, but those fixes sound like ones we would like to get in please.

mwringe:
OK, I have created a new PR here https://github.com/google/cadvisor/pull/1349 which adds in support for gathering the ip address of docker containers and being able to read files within docker containers.

Once we get consensus and approval for that PR, I will update this one with only the Jolokia parts

PR #1340, opened 2016-06-21
Move deps to vendor, use go-1.6+

imikushin:
I've been building cadvisor for ARM and found that godep was missing some dependency files. 
So I used my own tool (https://github.com/rancher/trash) to populate ./vendor dir. 

I'm probably missing something, so please advise where I need to improve 
(if this PR is of any value, of course).


imikushin:
I've signed it!

vishh:
ok to test

timstclair:
It looks like you're doing 2 different things here:

1. Move Godeps/_workspace to vendor
2. Replace godep with trash for dependency management

The first sounds reasonable to me, as godep now supports the vendor directory and Kubernetes has already made the move. As for the second, I would like to keep the cAdvisor tooling as close to the Kubernetes tooling as possible. Some people have been investigating replacing `godep` with `glide`, but I'm not sure what the status is. If you think `trash` is a supperior option, I recommend opening a discussion on the [kubernetes](https://github.com/kubernetes/kubernetes) repo.

vishh:
+1 for not moving away from `godep` for now.

timstclair:
Superseded by https://github.com/google/cadvisor/pull/1369

PR #1351, opened 2016-06-23
Cherrypick #1348: Fix detection of devmapper metadata reservation

sjenning:
Cherrypick https://github.com/google/cadvisor/pull/1348 to v0.23

Because new versions of devicemapper may add additional fields to the output of 'dmsetup status',
change the check to use SplitN to get the minimum fields required, and ignore any additional fields
after the one we care about.

@ncdc @timstclair @yujuhong


timstclair:
ok to test

PR #1353, opened 2016-06-24
Including the number of free inodes in FsInfo

ronnielai:
This info helps Kubelet decide on declaring OutOfDisk

### DIFF
--- a/info/v2/container.go
+++ b/info/v2/container.go
@@ -216,6 +216,9 @@ type FsInfo struct {
 
 	// Labels associated with this filesystem.
 	Labels []string `json:"labels"`
+
+	// Number of available Inodes.
+	InodesFree uint64 `json:"inodes_free"`
 }
 
 type RequestOptions struct {
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -687,6 +687,7 @@ func (self *manager) GetFsInfo(label string) ([]v2.FsInfo, error) {
 			Usage:      fs.Usage,
 			Available:  fs.Available,
 			Labels:     labels,
+			InodesFree: fs.InodesFree,
 		}
 		fsInfo = append(fsInfo, fi)
 	}


vishh:
ok to test

info/v2/container.go
```diff
@@ -216,6 +216,9 @@ type FsInfo struct {
 
 	// Labels associated with this filesystem.
 	Labels []string `json:"labels"`
+
+	// Number of available Inodes
```
timstclair: (nit) end in a period for consistency

timstclair:
LGTM, but needs a rebase

info/v2/container.go
ronnielai: Fixed.

timstclair:
LGTM.

info/v2/container.go
```diff
@@ -216,6 +216,9 @@ type FsInfo struct {
 
 	// Labels associated with this filesystem.
 	Labels []string `json:"labels"`
+
+	// Number of available Inodes.
+	InodesFree uint64 `json:"inodes_free"`
```
derekwaynecarr: this is problematic with devicemapper where inodes_free is not knowable since it operates at a block level.  I think we should make this a pointer.

manager/manager.go
```diff
@@ -687,6 +687,7 @@ func (self *manager) GetFsInfo(label string) ([]v2.FsInfo, error) {
 			Usage:      fs.Usage,
 			Available:  fs.Available,
 			Labels:     labels,
+			InodesFree: fs.InodesFree,
```
derekwaynecarr: I think we need to have a way on `FsStats` to know if inodes information is knowable in order to support block level things like devicemapper properly.  We would then only set `InodesFree` if `fs.InodesKnowable` was true for example.  I am not sure how much backwards compatibility concerns there are around `FsStats` struct.

info/v2/container.go
timstclair: SGTM

manager/manager.go
timstclair: Do you mean the FsInfo struct? I think we should try to maintain backwards compatibility between releases, but I think we don't need to worry about backwards compatibility with HEAD between releases.

derekwaynecarr: look at https://github.com/google/cadvisor/pull/1368

PR #1356, opened 2016-06-28
Simplify cAdvisor release versioning

timstclair:
- Eliminate version/VERSION in favor of parsing git versions
- Enforce that releases are built against tagged versions
- Delete the broken release script, change `make release` to build the release

/cc @ronnielai


timstclair:
Rebased.

vishh:
This is great! LGTM @timstclair !

PR #1365, opened 2016-07-05
Expose total inodes

derekwaynecarr:
Our operators want to support inode-based disk eviction policies in Kubernetes as a percentage of available inodes rather than a fixed value.  In order to support building a percentage, we need to expose total inodes and inodes free.

Related: https://github.com/kubernetes/kubernetes/pull/28501

/cc @vishh @ronnielai @ncdc @pmorie

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -418,6 +418,9 @@ type FsStats struct {
 	// HasInodes when true, indicates that Inodes info will be available.
 	HasInodes bool `json:"has_inodes"`
 
+	// Number of Inodes
+	Inodes uint64 `json:"inodes"`
+
 	// Number of available Inodes
 	InodesFree uint64 `json:"inodes_free"`
 
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -166,9 +166,11 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 }
 
 func fsToFsStats(fs *fs.Fs) info.FsStats {
+	inodes := uint64(0)
 	inodesFree := uint64(0)
 	hasInodes := fs.InodesFree != nil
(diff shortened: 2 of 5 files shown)


info/v1/container.go
```diff
@@ -415,6 +415,9 @@ type FsStats struct {
 	// Number of bytes available for non-root user.
 	Available uint64 `json:"available"`
 
+	// Number of Inodes
+	Inodes uint64 `json:"inodes"`
```
derekwaynecarr: we need to potentially treat this and inodes_free to be optional fields so we can distinguish the field as being set versus unset in cases where its not knowable.

derekwaynecarr:
Will rebase this to https://github.com/google/cadvisor/pull/1368

derekwaynecarr:
This is now rebased to https://github.com/google/cadvisor/pull/1368 , PTAL

derekwaynecarr:
Nag :-).  I want to consume this in Kubelet this week if possible to
support inode based eviction.

timstclair:
@ronnielai Can you take this?

ronnielai:
@derekwaynecarr could you add some tests?

derekwaynecarr:
@ronnielai - what kind of test do you recommend?  I see no obvious pattern to integrate with existing tests...

pmorie:
I am also wondering what can be done to test this -- we could write some very simple unit tests for stuff like `fsToFsStats` but I'm really uncertain about integration or e2es

ronnielai:
I was referring to simple unit tests :)

derekwaynecarr:
ok, will add unit test tomorrow in `fsToFsStats`

derekwaynecarr:
@ronnielai - added a unit test.

Issue #1366, opened 2016-07-06
Determine how to handle inodes with block level fs


derekwaynecarr:
cAdvisor returns inodes_free in `ContainerInfo` and `FsStats`

* inodes_free needs to be a pointer in `ContainerInfo`
* `FsStats` could have a `bool` that says if inode information is knowable, or we need to ensure we only set `inodes_free` or really any `inodes` related info based on `FsType`.

See: https://github.com/kubernetes/kubernetes/issues/28481#issuecomment-230799217

Related (report total inodes):
https://github.com/google/cadvisor/pull/1365

I am happy to send a PR to fix this today if we can get agreement:

@vishh @ronnielai @ncdc @pmorie

ncdc:
I'm guessing that changing to pointers probably isn't backwards compatible, so I'm fine with `inodesReported bool` or something to that effect

ronnielai:
@timstclair

derekwaynecarr:
To summarize:

* inodes and inodes_free in `v2.FsInfo` will be pointers
* inodes and inodes_free will switch to use pointers in `fs.Fs`
* we will not change `v1.FsStats.InodesFree` (which means it will be wrong for devicemapper)
* we will not change `v1.FsInfo.Inodes` (which means it will be wrong for devicemapper)
* kube will use the v2 API to handle inodes related info

I will send a PR today or tomorrow.

derekwaynecarr:
In trying to fix this:

https://github.com/google/cadvisor/blob/master/manager/manager.go#L690

Internally, cadvisor `InMemoryCache` is working against the v1 API.

An option I thought about is the following:

* inodes and inodes_free in `v2.FsInfo` will be pointers
* inodes and inodes_free will switch to use pointers in `fs.Fs`
* add `HasInodes bool` to `v1.FsStats`
* add `HasInodes bool` to `v1.FsInfo`

The above would allow us to change the code here:
https://github.com/google/cadvisor/blob/master/manager/manager.go#L690

It has advantages:
* we can convert v1 to v2 format object correctly and vice-versa while limit code churn
* we can return the proper value for devicemapper in v1 api

@vishh - please ack

vishh:
https://github.com/google/cadvisor/issues/1366#issuecomment-230896690 SGTM

PR #1368, opened 2016-07-06
Allow clients to know if inodes data is available

derekwaynecarr:
Fixes https://github.com/google/cadvisor/issues/1366

/cc @ncdc @vishh @ronnielai @timstclair

### DIFF
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -165,6 +165,34 @@ func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	return spec, nil
 }
 
+func fsToFsStats(fs *fs.Fs) info.FsStats {
+	inodesFree := uint64(0)
+	hasInodes := fs.InodesFree != nil
+	if hasInodes {
+		inodesFree = *fs.InodesFree
+	}
+	return info.FsStats{
+		Device:          fs.Device,
+		Type:            fs.Type.String(),
+		Limit:           fs.Capacity,
+		Usage:           fs.Capacity - fs.Free,
+		HasInodes:       hasInodes,
+		InodesFree:      inodesFree,
+		Available:       fs.Available,
+		ReadsCompleted:  fs.DiskStats.ReadsCompleted,
+		ReadsMerged:     fs.DiskStats.ReadsMerged,
+		SectorsRead:     fs.DiskStats.SectorsRead,
+		ReadTime:        fs.DiskStats.ReadTime,
+		WritesCompleted: fs.DiskStats.WritesCompleted,
+		WritesMerged:    fs.DiskStats.WritesMerged,
+		SectorsWritten:  fs.DiskStats.SectorsWritten,
+		WriteTime:       fs.DiskStats.WriteTime,
+		IoInProgress:    fs.DiskStats.IoInProgress,
+		IoTime:          fs.DiskStats.IoTime,
+		WeightedIoTime:  fs.DiskStats.WeightedIoTime,
+	}
+}
+
 func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 	// Get Filesystem information only for the root cgroup.
 	if isRootCgroup(self.name) {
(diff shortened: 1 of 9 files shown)


container/raw/handler.go
```diff
@@ -173,14 +173,19 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 			return err
 		}
 		for _, fs := range filesystems {
+			inodesFree := uint64(0)
```
ronnielai: Since this logic is repeated in 2 places, maybe put them in a utility method?

derekwaynecarr: ack, will update

info/v1/machine.go
```diff
@@ -26,6 +26,9 @@ type FsInfo struct {
 
 	// Total number of inodes available on the filesystem.
 	Inodes uint64 `json:"inodes"`
+
+	// HasInodes when true, indicates that Inodes info will be available.
+	HasInodes bool `json:"has_inodes"`
```
timstclair: I think this should go in a `FilesystemSpec` struct nested in `ContainerSpec`. We should prioritize removing the v1 API...

timstclair: Oops, that was meant for the container FsStats, but the same applies here too.

derekwaynecarr: hmm, can you model out what you would like to see to save some time?

The main intention in adding this was to keep this code from getting messy here:
https://github.com/google/cadvisor/blob/master/manager/manager.go#L671

I am not sure how to pull back the ContainerSpec cleanly from that context, but am happy to make the update with a pointer.

info/v1/container.go
```diff
@@ -415,6 +415,9 @@ type FsStats struct {
 	// Number of bytes available for non-root user.
 	Available uint64 `json:"available"`
 
+	// HasInodes when true, indicates that Inodes info will be available.
+	HasInodes bool `json:"has_inodes"`
```
timstclair: I was suggesting to move this to [ContainerSpec](https://github.com/google/cadvisor/blob/3977cb8bd1fb734369513336aa0c20a47330d2e4/info/v1/container.go#L44), or better yet, create a `FsSpec` struct in that file with this field, and link to that struct from the `ContainerSpec`.

info/v1/machine.go
timstclair: Sorry, you're right: there's no precedent for doing this with Machine. I added a comment above re: Container FsInfo.

info/v1/container.go
derekwaynecarr: So I have tried this change in my local workspace, but I am having trouble figuring out the best way to set it on `ContainerSpec` since best I can tell wherever we call `GetSpec` https://github.com/google/cadvisor/blob/master/container/common/helpers.go#L48, we don't yet appear to actually know if the associated filesystem supports inodes or not yet.

info/v1/machine.go
derekwaynecarr: ok, i will leave machine as is.  see my other question on how to properly initialize for Container `FsSpec`

info/v1/container.go
derekwaynecarr: trying to think how to handle this for container spec:

* zfs, devicemapper do not support inodes at all
* btrfs, xfs have inode limits that can dynamically grow

not sure where in code we want to populate the `FsSpec`

info/v1/machine.go
derekwaynecarr: per our discussion, moving to Container `FsSpec` in v1 API proved problematic as you have to basically fetch all fs stats, and treat it as a `[]FsSpec` since a container has multiple filesystems each of which may or may not support inodes.  the outcome was to stick with keeping a `has_inodes` throughout the v1 API.

info/v1/container.go
derekwaynecarr: per earlier comment, moved away from the `FsSpec` proposal in v1 API.

derekwaynecarr:
@timstclair - this should be good to go now.  verified all API endpoints on a host with and without devicemapper.  note I also cleaned up some other iterators to ensure we don't get a repeat of the earlier bug we saw on referencing values.  PTAL

timstclair:
LGTM

Issue #1181, opened 2016-03-28
Switch to native go vendoring instead of godeps


bluecmd:
Hi,

With Go 1.6 there is no need to use godeps. Please consider switching to the vendor structure native to 1.6.

Details: https://golang.org/cmd/go/#hdr-Vendor_Directories

Thanks

timstclair:
xref: https://github.com/google/cadvisor/pull/1073

vishh:
I will make some time this week to wrap up #1073.

PR #1369, opened 2016-07-11
Switch to using vendor directory for deps.

vishh:
Fixes https://github.com/google/cadvisor/issues/1181


vishh:
cc @timstclair

vendor/github.com/Shopify/sarama/.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 go:
 - 1.4.3
-- 1.5.2
```
timstclair: Why did `Shopify/sarama` files change? I don't see any corresponding changes in `Godeps.json`

vendor/github.com/abbot/go-http-auth/digest.go
```diff
@@ -22,7 +22,7 @@ type DigestAuth struct {
 	Secrets          SecretProvider
 	PlainTextSecrets bool
 
-	/*
```
timstclair: ditto RE: misc changes

docs/development/build.md
```diff
@@ -14,24 +14,18 @@ After setting up Go, you should be able to `go get` cAdvisor as expected (we use
 $ go get -d github.com/google/cadvisor
```
timstclair: There's conflicting information above: "Ensure that your version of Go is at least 1.3." and "**Note**: cAdvisor requires Go 1.5 to build.". Could you fix it? Does the vendor directory work by default on go 1.5 or do you need to set a flag?

build/presubmit.sh
```diff
@@ -19,6 +19,4 @@ set -x
 
 ./build/check_gofmt.sh .
 ./build/check_boilerplate.sh
-go vet github.com/google/cadvisor/...
-godep go test -v -race -test.short github.com/google/cadvisor/...
-godep go build github.com/google/cadvisor
+make
```
timstclair: how about instead of putting this here, we add a presubmit make rule that depends on the necessary other rules? (then change the presubmit call in the jenkins script to call make instead)

vishh: Great idea. Can we do that in a separate PR?

deploy/build.sh
```diff
@@ -17,6 +17,6 @@
 set -e
 set -x
 
-godep go build -a github.com/google/cadvisor
+make
```
timstclair: I'd prefer `make build` here. I don't think `fmt` `vet` and `test` need to be part of this script.

docs/development/build.md
vishh: Good catch. Fixing it. I'm going to make go 1.6 minimum requirement.

deploy/canary/Dockerfile
```diff
@@ -3,8 +3,7 @@ MAINTAINER vmarmol@google.com
 
 RUN apt-get install -y git dmsetup
 RUN git clone https://github.com/google/cadvisor.git /go/src/github.com/google/cadvisor
-RUN go get github.com/tools/godep
-RUN cd /go/src/github.com/google/cadvisor && godep go build .
+RUN cd /go/src/github.com/google/cadvisor && make
```
timstclair: ditto: use `make build` here. Are you sure the Makefile should be the source of truth for all our build commands? I'm inclined to prefer calling `build/build.sh` directly, but don't have any concrete reasons for it.

deploy/canary/Dockerfile
```diff
@@ -3,8 +3,7 @@ MAINTAINER vmarmol@google.com
 
 RUN apt-get install -y git dmsetup
```
timstclair: Can we just delete this file?  `make docker` seems sufficient for local development and uses `deploy/Dockerfile` instead.

docs/development/build.md
```diff
@@ -50,6 +44,6 @@ If you modify files in the /assets folder, you will need to rebuild the assets:
 
 ```
 $GOPATH/src/github.com/google/cadvisor $ ./build/assets.sh
-$GOPATH/src/github.com/google/cadvisor $ godep go build .
+$GOPATH/src/github.com/google/cadvisor $ make
```
timstclair: `make build` includes ./build/assets.sh, so the above line is redundant. I've been meaning to split out a separate `make assets` rule that build depends on...

docs/development/build.md
```diff
@@ -14,24 +14,18 @@ After setting up Go, you should be able to `go get` cAdvisor as expected (we use
...
-$ go get github.com/tools/godep
-```
-
 ## Building from Source
 
 At this point you can build cAdvisor from the source folder:
 
 ```
-$GOPATH/src/github.com/google/cadvisor $ godep go build .
+$GOPATH/src/github.com/google/cadvisor $ make
```
timstclair: Again, I'd prefer to be explicit about which rule to execute (do we really need `all` here?)

timstclair:
Thanks for fixing this!

build/presubmit.sh
timstclair: As long as you promise to do it :) I'm worried that leaving all these naked `make` calls around is fragile.

vendor/github.com/Shopify/sarama/.travis.yml
vishh: Honestly, I have no idea how I managed to pull in a newer version of this dep. Do you think its worth fixing?

docs/development/build.md
vishh: Fixing it.

deploy/canary/Dockerfile
vishh: This is for our jenkins automation that builds a canary docker image from HEAD every night. We can update it I guess to use existing make rules. Again can we do that in a separate PR?

docs/development/build.md
vishh: Ack. Removing this section.

docs/development/build.md
vishh: Depends on the context I think. We definitely want contributors to run `make all` before posting patches. But it is not necessary for building though.

deploy/canary/Dockerfile
vishh: I will clean this up in a separate PR since it needs some work on the jenkins side as well

build/presubmit.sh
vishh: Ok. I'm doing it in this patch itself :)

vishh:
@timstclair posted a new patch. PTAL

deploy/canary/Dockerfile
vishh: https://github.com/google/cadvisor/pull/1372

deploy/canary/Dockerfile
timstclair: Ah, that makes sense, sorry I missed that before. Do you think it's better to build cAdvisor in the docker container, like it was before? Or does it not make a difference? I'm OK with whichever you think is best :)

vendor/github.com/Shopify/sarama/.travis.yml
timstclair: I was able to reproduce this in the old tree with:
```
godep restore
rm -rf Godeps
export  GO15VENDOREXPERIMENT=0
godep save ./...
```
I filed https://github.com/google/cadvisor/issues/1373 to prevent this in the future, but I think you can leave this as is.

vendor/github.com/abbot/go-http-auth/digest.go
timstclair: <ignore>

timstclair:
LGTM. Go ahead and self-merge after squashing the comment commits.

PR #1372, opened 2016-07-11
add a canary build rule. Delete canary dockerfile

vishh:
Signed-off-by: Vishnu kannan vishnuk@google.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@
 
 GO := go
 pkgs  = $(shell $(GO) list ./... | grep -v vendor)
+canary = google/cadvisor:canary
 
 all: presubmit build test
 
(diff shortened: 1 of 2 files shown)


vishh:
https://github.com/kubernetes/test-infra/pull/273 should fix jenkins CI.

timstclair:
Sorry, I'm losing track of which comments I left where:

> Ah, that makes sense, sorry I missed that before. Do you think it's better to build cAdvisor in the docker container, like it was before? Or does it not make a difference? I'm OK with whichever you think is best :)

This LGTM, if you decide this approach is the way to go (if you do, go ahead with a self-merge)

vishh:
@timstclair Portable option is to use a docker container for builds. We need to separate builds from real containers though. So I think this PR is still safe to be merged. I have a pending PR to run builds inside a docker container. If I can get that to work, then we have hit the holy grail.

timstclair:
Actually, I think to do this without breaking CI we need to split it into 2 PRs that are merged in this order:

1. Add the make rule
2. Update the CI to use the make rule
3. Delete the dockerfile

vishh:
@timstclair Given that a single CI failure is not a critical failure, do you still think it is necessary to break this PR?

timstclair:
Fair. Do you have merge permissions on test-infra? If we merge both at roughly the same time we shouldn't have any failures anyway...

vishh:
I do. Let's give the test infra team a heads up before merging. WDYT?

timstclair:
SGTM.

PR #1376, opened 2016-07-13
Update collectors to be able to directly use the ip address of containers

mwringe:
Allows collectors to directly use the ip address of containers.

This means that custom metrics no longer have to know the ip address of the container before they are deployed.

In Kubernetes environments, it greatly makes handling custom metrics much easier. We no longer have to deal with special permissions for hostports and we can now deploy multiple pods of the same type without having to manually manage the ports in use.

### DIFF
--- a/collector/prometheus_collector.go
+++ b/collector/prometheus_collector.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/info/v1"
 )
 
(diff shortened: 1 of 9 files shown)


collector/prometheus_collector.go
```diff
@@ -55,6 +56,13 @@ func NewPrometheusCollector(collectorName string, configFile []byte, metricCount
 
 	minPollingFrequency := configInJSON.PollingFrequency
 
+	//If the exact URL was not specified, generate it based on the ip address of the container.
+	endpoint := configInJSON.Endpoint
+	if endpoint.URL == "" {
+		ipAddress := containerHandler.GetContainerIPAddress()
+		configInJSON.Endpoint.URL = endpoint.URLConfig.Protocol + "://" + ipAddress + ":" + endpoint.URLConfig.Port.String() + endpoint.URLConfig.Path
+	}
+
```
timstclair: Consider moving the configFile parsing to a shared method so you don't need this duplicate code.

timstclair:
LGTM with one nit

mwringe:
Ok, the requested change has been applied.

timstclair:
Thanks, LGTM.

PR #1379, opened 2016-07-15
Update collectors to use a customized httpClient.

mwringe:
Updates collectors to use a customized httpClient.

This client has been configured to be able to read from https endpoints which in which cAdvisor cannot entirely trust. For instance, in the case of endpoints where its secured with self signed certificates or do not have the right ip address as part of the certificate. We do not need to verify that endpoints have properly certificates, this is just to make reading endpoints regardless if they are http or https based.

Expose as a configuration option the ability to specify a certificate to be used by the collectors when accessing endpoints. This means that endpoints can be configured to be use client certificate based authentication, allowing endpoints to only allow access to custom metrics if they trust the cAdvisor certificates.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/cadvisor/utils/sysfs"
 	"github.com/google/cadvisor/version"
 
+	"crypto/tls"
 	"github.com/golang/glog"
 )
 
@@ -52,6 +53,9 @@ var maxHousekeepingInterval = flag.Duration("max_housekeeping_interval", 60*time
 var allowDynamicHousekeeping = flag.Bool("allow_dynamic_housekeeping", true, "Whether to allow the housekeeping interval to be dynamic")
 
 var enableProfiling = flag.Bool("profiling", false, "Enable profiling via web interface host:port/debug/pprof/")
+
+var collectorCert = flag.String("collector_cert", "", "Collector's certificate, exposed to endpoints for certificate based authentication.")
+var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
 
 var (
 	// Metrics to be ignored.
@@ -118,7 +122,9 @@ func main() {
 		glog.Fatalf("Failed to create a system interface: %s", err)
 	}
 
-	containerManager, err := manager.New(memoryStorage, sysFs, *maxHousekeepingInterval, *allowDynamicHousekeeping, ignoreMetrics.MetricSet)
+	collectorHttpClient := createCollectorHttpClient(*collectorCert, *collectorKey)
+
+	containerManager, err := manager.New(memoryStorage, sysFs, *maxHousekeepingInterval, *allowDynamicHousekeeping, ignoreMetrics.MetricSet, &collectorHttpClient)
 	if err != nil {
 		glog.Fatalf("Failed to create a Container Manager: %s", err)
 	}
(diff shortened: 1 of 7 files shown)


mwringe:
@timstclair any opinion on this?

cadvisor.go
```diff
@@ -186,3 +192,29 @@ func installSignalHandler(containerManager manager.Manager) {
...
 	}()
 }
+
+func createCollectorHttpClient(collectorCert, collectorKey string) http.Client {
+	//Enable accessing insecure endpoints. We should be able to access metrics from any endpoint
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: true,
+	}
+
+	if collectorCert != "" {
```
timstclair: I'm not a TLS expert, but if a certificate is provided why do we want to skip verification?

manager/manager.go
```diff
@@ -226,6 +228,7 @@ type manager struct {
 	ignoreMetrics            container.MetricSet
 	containerWatchers        []watcher.ContainerWatcher
 	eventsChannel            chan watcher.ContainerEvent
+	httpClient               *http.Client
```
timstclair: I think you should call this `collectorHTTPClient`, since the flags are specifically specifying collector certs.

timstclair:
Approach looks good, but I'm concerned about the `InsecureSkipVerify`. Can you explain the need for that in more detail?

mwringe:
Essentially the need for `InsecureSkipVerify` is because not having it set it doesn't actually improve security and would otherwise require a bunch of hoops for developers to jump through.

- cAdvisor can read metrics from a http endpoints, it doesn't verify the authenticity of the container exposing the metrics. If a developer were to switch on or off https, it doesn't change the security of how metrics are stored within cAdvisor.

- it is common for containers to self-sign their own certificates. In this case, the container would have to pass to cAdvisor the CA certificate to trust. But since both are coming from the container, this doesn't make anything more secure (this was the reason why originally I had exposed the option to read a file from a container, to pass the CA cert).

- if we are accessing the endpoint via its ip address, in order for the endpoint to be acceptable, it has to have the ip address listed under its certificate. Since the ip address of the container is most likely not known before the container is started, this can make it difficult to configure.

In the end I couldn't figure out a reason why we shouldn't enabled `InsecureSkipVerify`. We are not actually verifying the authenticity of the container which expose metric endpoints.

In terms of why a container would want to expose an https endpoint instead of an http one. There are a few reasons:

- they may already have an https endpoint already configured for some other process and they may not want to expose another http one just for cAdvisor.

- security. If the endpoint wanted to expose metrics in a secure fashion, they would need to do so over an https connection. This could mean either by only trusting connections from clients with the cAdvisor certificate, or requiring the client to verify themselves (in which they would have to provide the username/password in the endpoint configuration file). With either of these, having `InsecureSkipVerify` shouldn't have an affect on the security of the system.

mwringe:
I made the change to rename the httpClients to collectorHttpClient to make it more apparent its only meant for the collectors. And also rebased to master.

In terms of the `InsecureSkipVerify` concern, would making this a configurable option for each individual endpoint be a more palatable option? Eg we add a 'InsecureSkipVerify' option to https://github.com/google/cadvisor/blob/master/collector/config.go in which the configuration can either set it to true/false (with a default of false)

timstclair:
Thanks for the explanation. I think this looks good as is, we can always add in the option later if there is demand for it. Needs another rebase though.

mwringe:
Rebased.

PR #1380, opened 2016-07-15
Modify working set memory stats calculation

mtaufen:
Change working set calculation to usage - total_inactive_file, rather than usage - total_inactive_anon - total_inactive_file. Since writes to tmpfs get tracked as total_inactive_anon when swap is disabled, the old calculation would under-report memory pressure.

See these for context:
https://github.com/kubernetes/kubernetes/issues/28619
https://github.com/kubernetes/kubernetes/pull/28693

cc @vishh @timstclair


vishh:
LGTM

PR #1381, opened 2016-07-17
Continue watching other sub directories when there is watch error.

Random-Liu:
Address the `TODO` in `manager/watcher/raw/raw.go`.
```
TODO(vmarmol): We don't have to fail here, maybe we can recover and try to get as many registrations as we can.
```

Many reasons could cause `watch` to fail, such as file permission issue, file removed before watching etc.
In the original code, once a `watch` operation failed, the error will be returned through the recursive call stack and finally cause cadvsior start failure, which causes issues like https://github.com/kubernetes/kubernetes/issues/28997.

As the `TODO` said, we should keep on watching other subdirectories when error happens.

@vishh @timstclair

### DIFF
--- a/manager/watcher/raw/raw.go
+++ b/manager/watcher/raw/raw.go
@@ -19,6 +19,7 @@ package raw
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 	"strings"
 
@@ -133,9 +134,15 @@ func (self *rawContainerWatcher) watchDirectory(dir string, containerName string
 	}
 	for _, entry := range entries {
 		if entry.IsDir() {
-			// TODO(vmarmol): We don't have to fail here, maybe we can recover and try to get as many registrations as we can.
-			_, err = self.watchDirectory(path.Join(dir, entry.Name()), path.Join(containerName, entry.Name()))
+			entryPath := path.Join(dir, entry.Name())
+			_, err = self.watchDirectory(entryPath, path.Join(containerName, entry.Name()))
 			if err != nil {
+				glog.Errorf("Failed to watch directory %q: %v", entryPath, err)
+				if os.IsNotExist(err) {
+					// The directory may have been removed before watching. Try to watch the other
+					// subdirectories. (https://github.com/kubernetes/kubernetes/issues/28997)
+					continue
+				}
 				return alreadyWatching, err
 			}
 		}


gmarek:
@vishh @timstclair - this is causing startup errors in our test jobs, can one of you review this?

manager/watcher/raw/raw.go
```diff
@@ -133,10 +133,12 @@ func (self *rawContainerWatcher) watchDirectory(dir string, containerName string
 	}
 	for _, entry := range entries {
 		if entry.IsDir() {
-			// TODO(vmarmol): We don't have to fail here, maybe we can recover and try to get as many registrations as we can.
-			_, err = self.watchDirectory(path.Join(dir, entry.Name()), path.Join(containerName, entry.Name()))
+			entryPath := path.Join(dir, entry.Name())
+			_, err = self.watchDirectory(entryPath, path.Join(containerName, entry.Name()))
```
yujuhong: can we check what type of error this is? It probably only makes sense to continue if the path doesn't exist...

Random-Liu: Here, it is checking all sub directories, even though one directory returns error, we can still check other directories.

timstclair: Yes, I agree with @Random-Liu. It doesn't ignore the error on the base directory, but if there is an error on a sub-directory, I think continuing at a best-effort is reasonable.

timstclair:
LGTM, if @yujuhong is OK with ignoring the error type.

manager/watcher/raw/raw.go
yujuhong: One example of potential issues is that kubelet hits the limits of the inotify watches, but it continues to start up without watching any pods. We can argue that this may happen anytime when kubelet is running, so there is no need to fail the startup. I am ok with that argument, but just want to point out that this is possible and we should try ensuring there is a recovery path. For example, maybe cadvisor should be not ready if errors like this keep happening, and kubelet should report node not ready?

I don't want to block this PR, but maybe we should open up an issue to discuss propagating the errors better. On the other hand, if cadvisor does **not** solely rely on inotify watches for notifications (e.g., it lists everything periodically), just ignore what I said :-)

Random-Liu: > One example of potential issues is that kubelet hits the limits of the inotify watches, but it continues to start up without watching any pods. We can argue that this may happen anytime when kubelet is running, so there is no need to fail the startup. I am ok with that argument, but just want to point out that this is possible and we should try ensuring there is a recovery path. For example, maybe cadvisor should be not ready if errors like this keep happening, and kubelet should report node not ready?

We don't have a good way to expose errors now. Failing cadvisor makes dubug easier, because the system will fail early when there is unexpected error.
However, failing cadvisor will make the system less robust.

For now, maybe we can log the error only if the path doesn't exist to fix current issue, and fail cadvisor otherwise. In the future, if we find that there are a lot more temporary errors, we can consider how to handle the errors better.

@yujuhong @timstclair WDYT?

Random-Liu:
Update the PR to address https://github.com/google/cadvisor/pull/1381#discussion_r71241593 only continue when the error is "file not exist".
@timstclair @yujuhong PTAL~

yujuhong:
LGTM. Thanks!

PR #1386, opened 2016-07-19
Expose cpu cgroup CFS prometheus metrics

### LINKED ISSUE #1337: Export cgroup throttled data
As far as I can tell cadvisor doesn't currently expose when processes in a cgroup are throttled:

https://github.com/google/cadvisor/blob/master/container/libcontainer/helpers.go

Is there a reason for that? Otherwise I am happy to raise PR to add it as we'd really like this data aggregated in prometheus.

grobie:
If CPU quota is configured (cpu.cfs_quota != -1) the CFS will provide
stats about elapsed periods and throtting in cpu.stats. This change
makes these information available as container_cpu_cfs_* metrics.

Fixes #1337

@chbatey @matthiasr @rtreffer

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -266,7 +266,7 @@ type LoadStats struct {
 // CPU usage time statistics.
 type CpuUsage struct {
 	// Total CPU usage.
-	// Units: nanoseconds
+	// Unit: nanoseconds.
 	Total uint64 `json:"total"`
 
 	// Per CPU/core usage of the container.
@@ -274,17 +274,31 @@ type CpuUsage struct {
 	PerCpu []uint64 `json:"per_cpu_usage,omitempty"`
 
 	// Time spent in user space.
-	// Unit: nanoseconds
+	// Unit: nanoseconds.
 	User uint64 `json:"user"`
 
 	// Time spent in kernel space.
-	// Unit: nanoseconds
+	// Unit: nanoseconds.
 	System uint64 `json:"system"`
+}
+
+// Cpu Completely Fair Scheduler statistics.
+type CpuCFS struct {
+	// Total number of elapsed enforcement intervals.
+	Periods uint64 `json:"periods"`
+
+	// Total number of times tasks in the cgroup have been throttled.
+	ThrottledPeriods uint64 `json:"throttled_periods"`
+
+	// Total time duration for which tasks in the cgroup have been throttled.
+	// Unit: nanoseconds.
+	ThrottledTime uint64 `json:"throttled_time"`
 }
 
 // All CPU usage metrics are cumulative from the creation of the container
 type CpuStats struct {
 	Usage CpuUsage `json:"usage"`
+	CFS   CpuCFS   `json:"cfs"`
 	// Smoothed average of number of runnable threads x 1000.
 	// We multiply by thousand to avoid using floats, but preserving precision.
 	// Load is smoothed over the last 10 seconds. Instantaneous value can be read
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -89,7 +89,7 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int, ignoreMetri
 	libcontainerStats := &libcontainer.Stats{
 		CgroupStats: cgroupStats,
 	}
-	stats := toContainerStats(libcontainerStats)
+	stats := newContainerStats(libcontainerStats)
 
 	// If we know the pid then get network stats from /proc/<pid>/net/dev
 	if pid == 0 {
@@ -350,7 +350,7 @@ func DiskStatsCopy(blkio_stats []cgroups.BlkioStatEntry) (stat []info.PerDiskSta
 }
 
 // Convert libcontainer stats to info.ContainerStats.
-func toContainerStats0(s *cgroups.Stats, ret *info.ContainerStats) {
+func setCpuStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	ret.Cpu.Usage.User = s.CpuStats.CpuUsage.UsageInUsermode
 	ret.Cpu.Usage.System = s.CpuStats.CpuUsage.UsageInKernelmode
 	n := len(s.CpuStats.CpuUsage.PercpuUsage)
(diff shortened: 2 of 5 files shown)


grobie:
I updated my commit to use my company email (@soundcloud.com), we have signed the CLA already.

grobie:
@vishh Any feedback? It's a really big issue for us that Kubernetes/cAdvisor doesn't have any metrics on CPU throttling. Happy to fix whatever problem you see (I tried to follow the cAdvisor code style wherever possible).

jimmidyson:
LGTM... but I no longer seem to be a maintainer so can't kick off test builds or merge PRs...

@timstclair Would you be able to review if @vishh is unavailable?

info/v1/container.go
```diff
@@ -266,25 +266,39 @@ type LoadStats struct {
 // CPU usage time statistics.
```
vishh: Can you update `v2` APi as well? We plan on deprecating `v1` sometime soon.

vishh:
ok to test

vishh:
@jimmidyson I can fix the maintainer issue.

vishh:
@jimmidyson The jenkins admin list was updated recently. Now you are in that list.

info/v1/container.go
grobie: Sure thing.

jimmidyson:
@vishh Thanks!

info/v1/container.go
grobie: @vishh Ah, I had looked for that actually already. v2 uses the v1 CPU struct and automatically includes my changes: https://github.com/google/cadvisor/blob/master/info/v2/container.go#L136

info/v1/container.go
```diff
@@ -266,25 +266,39 @@ type LoadStats struct {
...
+
+	// Total time duration for which tasks in the cgroup have been throttled.
+	// Unit: nanoseconds.
+	ThrottledTime uint64 `json:"throttled_time"`
+}
+
 // All CPU usage metrics are cumulative from the creation of the container
 type CpuStats struct {
 	Usage CpuUsage `json:"usage"`
+	CFS   CpuCFS   `json:"cfs"`
```
grobie: I'm still a bit unsure how to namespace these stats. They get read from a file called `cpu.stats`. The implementation is done by using a completely fair scheduler. After reading the kernel documentation https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt, maybe `Bandwidth` would be better?

What do you think @jimmidyson @vishh?

grobie: I'd apply any naming change also to the Prometheus metric names. Currently it's `container_cpu_cfs_*`.

grobie:
@vishh Anything missing?

grobie:
I signed it!

grobie:
@vishh @timstclair What is missing here?

timstclair:
The CLA bot isn't picking up your signature... Did you sign it with the same email address you used with your commit?

matthiasr:
We supposedly signed a corporate CLA, however, I have no idea how to figure out whether that is actually in place. The CLA pages show nothing.

grobie:
@timstclair I did this as part of my job at SoundCloud, so I changed my commit to use my soundcloud email address. As @matthiasr said, we signed it as a company. Please advise what to do here.

timstclair:
Do you have your soundcloud email address connected to your github account? (i.e. is it listed here? https://github.com/settings/emails)

grobie:
Yes, it's even my primary email address.

matthiasr:
It seems the signing stalled on our end. Hold on while I follow up …

info/v1/container.go
```diff
@@ -266,25 +266,39 @@ type LoadStats struct {
...
 	// Time spent in kernel space.
-	// Unit: nanoseconds
+	// Unit: nanoseconds.
 	System uint64 `json:"system"`
 }
 
+// Cpu Completely Fair Scheduler statistics.
+type CpuCFS struct {
+	// Total number of elapsed enforcement intervals.
+	Periods uint64 `json:"periods"`
```
vishh: Are any of these fields optional? If yes, add an `omitempty` tag

vishh: Ignore my comment. What is the enforcement interval?

info/v1/container.go
vishh: From user perspective, grouping it under cpu sgtm. After all it is related to CPU.

info/v1/container.go
vishh: Ah. My bad. I thought we had cleaned that up :(

info/v1/container.go
vishh: Also, would it make sense to instead expose CPU hard limits based on quota settings and then expose the number and duration of hardcapping events?
Essentially hide the cgroups API a bit for make the cAdvisor API user friendly

matthiasr:
The CLA is really in place now, sorry about that.

f0:
@grobie the exposed Cpu Period is the CPU reference time (found e.g in cgroups)
https://github.com/google/cadvisor/commit/669bc4abfab349e18bc4efbec0f8d13b13602e29 

This PR breakt this

grobie:
@f0 What behavior gets broken here? I assume you are referring to the period change, it's true that I changed it to the value used by Kubernetes to make the example more clear. Given this is just a test setup, what problem do you see here?

f0:
@grobie  hm you are right, you only break the tests  so yes there is no real problem sorry

info/v1/container.go
grobie: The configured limits are already available under the spec namespace, see quota and period there. The resulting prometheus metrics can be seen here: https://github.com/google/cadvisor/blob/master/metrics/testdata/prometheus_metrics#L119-L124

The number of throttled periods and the total resulting throttled time already get exposed verbatim with the change. What exactly would you like to see?

I always prefer exporting metrics more or less directly. Most of the documentation users will find will discuss the standard cgroups metrics, I'd find it confusing to first read the code of cadvisor to understand how they were changed and exported instead.

info/v1/container.go
grobie: If `cpu.cfs_quota` is set to a value greater -1, the CPU usage of that cgroup will be limited to that amount of nanoseconds per period. The period length is defined by `cpu.cfs_period_us`. Both of these values get exported already under the `spec` namespace. The periods here describe how many periods any process in the cgroup requested some cpu cycles. As far as I understand, a completely idle cgroup will not cause this counter to go up.

grobie:
@vishh I rebased my change against current master. Do you have any other questions or change requests?

info/v1/container.go
vishh: Ok. If we expect that looking up kernel documentation is the best option as of now, I'm ok with this change.

vishh:
lgtm

grobie:
Thanks so much @vishh!

PR #1370, opened 2016-07-11
Updating libcontainer deps

vishh:
fixes https://github.com/kubernetes/kubernetes/issues/28577#issuecomment-231800140

PR #1387, opened 2016-07-19
Update libcontainer deps

vishh:


timstclair:
Is this different from https://github.com/google/cadvisor/pull/1370?

vishh:
Closing #1370

timstclair:
My same comment applies here: should we update to an official tagged release rather than an untagged commit?

vishh:
I'm not sure when runc will post a tag. Meanwhile, head has some build fixes and will also help with unifying dependencies across cadvisor and kube.

timstclair:
Ok, LGTM.

PR #1393, opened 2016-07-21
Added CPU Usage details to statsd

bah2830:
Added CPU user, system stats as well as the load average stat.

### DIFF
--- a/storage/statsd/statsd.go
+++ b/storage/statsd/statsd.go
@@ -36,7 +36,7 @@ const (
 	// CPU user
 	colCpuUsageUser string = "cpu_usage_user"
 	// CPU average load
-	colLoadAverage string = "load_average"
+	colCpuLoadAverage string = "cpu_load_average"
 	// Memory Usage
 	colMemoryUsage string = "memory_usage"
 	// Working set size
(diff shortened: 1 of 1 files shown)


bah2830:
Is there anything else I need to do to get k8s-bot testing approved?

timstclair:
ok to test

storage/statsd/statsd.go
```diff
@@ -31,6 +31,12 @@ type statsdStorage struct {
 
 const (
 	colCpuCumulativeUsage string = "cpu_cumulative_usage"
+	// CPU system
+	colCpuUsageSystem string = "cpu_usage_system"
+	// CPU user
+	colCpuUsageUser string = "cpu_usage_user"
+	// CPU average load
+	colLoadAverage string = "load_average"
```
timstclair: Would it be better to call this `"cpu_load_average"` to group it?

timstclair:
LGTM, with one question.

storage/statsd/statsd.go
bah2830: I don't have a problem with that. I can get it updated when I get home from work.

Thanks

PR #1398, opened 2016-07-27
enable FilesystemStat test for devicemapper

### LINKED ISSUE #1147: (Re)Enable TestDockerFilesystemStats test for RHEL integration test
This was disabled in #1143 to fix #1130 and allow the integration tests to pass on RHEL with devicemapper storage driver.  Need to get this working properly, though.

sjenning:
Fixes #1147 

This test works for devicemapper now

@ncdc @pmorie

### DIFF
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -307,12 +307,6 @@ func TestDockerFilesystemStats(t *testing.T) {
 	fm := framework.New(t)
 	defer fm.Cleanup()
 
-	storageDriver := fm.Docker().StorageDriver()
-	if storageDriver == framework.DeviceMapper {
-		// Filesystem stats not supported with devicemapper, yet
-		return
-	}
-
 	const (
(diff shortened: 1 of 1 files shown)


sjenning:
Well nevermind! I tried this on Fedora but RHEL doesn't have thin_ls yet.  device-mapper-persistent-data update is scheduled to be pushed out to the repos for 7.2 on Aug 2.  I'll reopen then.

dashpole:
@sjenning, do you plan to pick this up again?

sjenning:
@dashpole no.  DM per-container fs stats ended up being a YUGE mess, mostly due to the implementation details of the tool we were using to get them, `thin_ls`.  tl;dr `thin_ls` would max out IOPs on many cloud instances if the DM metadata got too big. It was disabled in cadvisor PR  https://github.com/google/cadvisor/pull/1588


This is dead for the foreseeable future. :disappointed:

PR #1402, opened 2016-07-30
Prefix Docker labels & env vars in Prometheus metrics to prevent conflicts

### LINKED ISSUE #1279: Prometheus label "name" conflicts with docker label "name" of running Image

jimmidyson:
Fixes #1279. Supersedes #1315 (it's become inactive sadly).

@grobie What do you think? I decided to go with consistent, plain `docker_{label,env}_` prefix as the actual Docker label itself will (should) be namespaced. This also fixes a potential for conflicts in env vars used in Prometheus labels.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -500,6 +500,11 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
 	c.errors.Collect(ch)
 }
 
+const (
+	containerLabelPrefix = "container_label_"
+	containerEnvPrefix   = "container_env_"
+)
+
 func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric) {
 	containers, err := c.infoProvider.SubcontainersInfo("/", &info.ContainerInfoRequest{NumStats: 1})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


grobie:
:+1: Looks great, the namespace makes sense. Thanks a lot!

jimmidyson:
@timstclair Would you be able to take a look & merge if you're happy please?

metrics/prometheus.go
```diff
@@ -500,6 +500,11 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
 	c.errors.Collect(ch)
 }
 
+const (
+	dockerLabelPrefix = "docker_label_"
+	dockerEnvPrefix   = "docker_env_"
```
timstclair: What about other runtimes (rkt)? You could use the namespace from the ContainerReference.

jimmidyson: How about just `container_{label,env}_`? At least that will be consistent for dashboards, etc across docker/rkt envs.

jimmidyson: Changed to `container_{label,env}_`, which makes sense to me at least.

timstclair:
LGTM

jimmidyson:
Thanks @timstclair!

jimmidyson:
@timstclair This is affecting a number of users with Kubernetes-Prometheus integration when using base images with LABELs in them (e.g. CentOS). Shall I send in a PR to cherry-pick this into release-0.23 & get a new release tagged to be upgraded in Kubernetes?

jimmidyson:
Created cherry-picked PR #1405

timstclair:
I can handle the release.

PR #1423, opened 2016-08-16
GetSpec missing memory reservation

chenchun:
I was using cadvisor package in my project and found memory reservation is missing.


timstclair:
ok to test

timstclair:
LGTM. Thanks!

Issue #1312, opened 2016-05-26
Container restart count is a label


fabxc:
The number of restarts is attached as a label to all metrics for a container. This causes all time series for a container to break as it restarts as identity is defined by the full set of labels.

The restart count is a time series itself and should only be exported as such.

vishh:
+1

grobie:
This is a kubernetes issue and not a cAdvisor one. https://github.com/kubernetes/kubernetes/blob/b0ea89c2f6b954a27ba0f545da955db4f25009d0/pkg/kubelet/kuberuntime/labels.go#L38

grobie:
I'm starting to understand what's going on here. Kubernetes attaches all these docker labels as a fallback to retrieve them later again. I guess the easiest would be to filter all labels which start with `io.kubernetes.`. What do you think @timstclair @vishh?

grobie:
By the way, I believe cAdvisor shouldn't try to parse these labels and create a restart counter from it. This would be the job of Kubernetes, also as Docker containers don't have the concept of restarts.

timstclair:
Yes, the labels should all be treated as opaque by cAdvisor.

rikatz:
Actually, this isn't a Kubernetes Label, but a Docker Label also (when using --restart=always directive):

```
docker inspect nginx1 |grep -i restartcount
        "RestartCount": 2,
```

This is a pure Docker 17.03 running, no Kubernetes. I'm going to check what's the impact to change cAdvisor code to also set a metric containing the RestartCount of the containers.

Magikon:
kubectl describe po nginx1 | grep "Restart Count"
      Restart Count:  0

fgrzadkowski:
This is still not fixed for users who are using cadvisor directly. RestartCount is a label, which breaks everything and makes it pretty impossible to write alerts based on it (e.g. alert if restart happened more than X times in the last Y min)

fgrzadkowski:
@fabxc @dashpole @vishh @timstclair

caligo-mentis:
Any news here? It would be useful to have container_restart_count metric to setup alerting when using bare docker engine without kubernetes.

OloBo-MSK:
Any news here?

Go-Pomegranate:
Any news?

dashpole:
[...]

PR #1426, opened 2016-08-18
Filter internal kubernetes labels from Prometheus metrics

grobie:
Kubernetes sets a bunch of internal labels on each docker container in
order to recover information in case of loss of the pod information
after a crash. These labels are not meaningful for Prometheus monitoring
and will create new timeseries in case they change (like restartCount).

With @jimmidyson's recent change to prefix labels with `container_label_`, users would have seen `container_label_io_kubernetes_container_hash`. 

@fabxc @timstclair @vishh

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -17,6 +17,7 @@ package metrics
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	info "github.com/google/cadvisor/info/v1"
@@ -535,6 +536,16 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
 const (
 	containerLabelPrefix = "container_label_"
 	containerEnvPrefix   = "container_env_"
+	// kubernetesInternalLabelPrefix is the common prefix of all labels set by
+	// Kubernetes in order to store internal information on docker containers.
+	// See https://github.com/kubernetes/kubernetes/blob/7523669/pkg/kubelet/dockertools/labels.go.
+	kubernetesInternalLabelPrefix = "io.kubernetes."
+)
+
+var (
+	// kubernetesInternalLabelWhitelist defines a regular expression such
+	// Kubernetes labels need to match in order to be attached to metrics.
+	kubernetesInternalLabelWhitelist = regexp.MustCompile(`^io\.kubernetes\.(container\.name|pod\.(name|namespace))$`)
 )
 
 func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric) {
@@ -567,6 +578,10 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 
 		for k, v := range container.Spec.Labels {
+			if strings.HasPrefix(k, kubernetesInternalLabelPrefix) &&
+				!kubernetesInternalLabelWhitelist.MatchString(k) {
+				continue
+			}
 			baseLabels = append(baseLabels, sanitizeLabelName(containerLabelPrefix+k))
 			baseLabelValues = append(baseLabelValues, v)
 		}
--- a/metrics/prometheus_test.go
+++ b/metrics/prometheus_test.go
@@ -65,7 +65,15 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 				},
 				CreationTime: time.Unix(1257894000, 0),
 				Labels: map[string]string{
-					"foo.label": "bar",
+					"foo.label":                                      "bar",
+					"io.kubernetes.container.hash":                   "e2b4295e",
+					"io.kubernetes.container.name":                   "POD",
+					"io.kubernetes.container.restartCount":           "0",
+					"io.kubernetes.container.terminationMessagePath": "",
+					"io.kubernetes.pod.name":                         "foo-bar-1985164696-y8y0d",
+					"io.kubernetes.pod.namespace":                    "qux",
+					"io.kubernetes.pod.terminationGracePeriod":       "30",
+					"io.kubernetes.pod.uid":                          "6a291e48-47c4-11e6-84a4-c81f66bdf8bd",
 				},
 				Envs: map[string]string{
 					"foo+env": "prod",
--- a/metrics/testdata/prometheus_metrics
+++ b/metrics/testdata/prometheus_metrics
@@ -3,150 +3,150 @@
 cadvisor_version_info{cadvisorRevision="abcdef",cadvisorVersion="0.16.0",dockerVersion="1.8.1",kernelVersion="4.1.6-200.fc22.x86_64",osVersion="Fedora 22 (Twenty Two)"} 1
 # HELP container_cpu_cfs_periods_total Number of elapsed enforcement period intervals.
 # TYPE container_cpu_cfs_periods_total counter
-container_cpu_cfs_periods_total{container_env_foo_env="prod",container_label_foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 723
+container_cpu_cfs_periods_total{container_env_foo_env="prod",container_label_foo_label="bar",container_label_io_kubernetes_container_name="POD",container_label_io_kubernetes_pod_name="foo-bar-1985164696-y8y0d",container_label_io_kubernetes_pod_namespace="qux",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 723
 # HELP container_cpu_cfs_throttled_periods_total Number of throttled period intervals.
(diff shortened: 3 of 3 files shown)


grobie:
Fixes #1312

jimmidyson:
We need the pod name, container name, etc that come from these labels (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/labels.go#L20). Perhaps we can just whitelist those labels & ignore others?

grobie:
We don't rely on these internal labels, but use the standard Prometheus relabel configs to build these from the available meta labels https://prometheus.io/docs/operating/configuration/#<kubernetes_sd_config>. I assumed that is the preferred way, but I can also create more specific filter if you think that's better. 

The current resulting label names like `container_label_io_kubernetes_pod_namespace` are a bit too long to be useful anyways.

jimmidyson:
I'm not at a computer at all this week to check but would you be able to share a snippet from exported kubelet /metrics from a kubernetes build from master? Just want to check what the raw pod metrics labels look like.

jimmidyson:
I'm pretty sure this will affect pod discovery which uses the container labels from exported metrics for pod name, etc but as I'm out on PTO all well I can't check myself.

grobie:
I'd be surprised if the Prometheus Kubernetes discovery client would rely
on labels exported by cAdvisor. As far as I can see, it reads these labels
from the Kubernetes API
https://github.com/prometheus/prometheus/blob/master/retrieval/discovery/kubernetes/pod.go#L294-L304.
This PR won't touch that at all, it only reduces the additional label noise
generated by cAdvisor.

metrics/prometheus.go
```diff
@@ -535,6 +536,10 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
 const (
 	containerLabelPrefix = "container_label_"
 	containerEnvPrefix   = "container_env_"
+	// kubernetesInternalLabelPrefix is the common prefix of all labels set by
+	// Kubernetes in order to store internal information on docker containers.
+	// See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/labels.go.
```
fabxc: Fix this link to the commit hash?

jimmidyson:
Doh I'm not talking about pod discovery but pod cgroup metrics. Pretty sure in kubernetes master, pod/container names will only be exposed via these container labels linked above. Just need to see a dump of /metrics from a kubelet built from master to confirm. We cleaned up duplicate label names iirc, relying on these container labels instead of deriving them from the docker container name.

metrics/prometheus.go
grobie: @fabxc done

grobie:
@jimmidyson I understand now what you mean. Given Prometheus scrapes one cadvisor instance per node, it's not possible to use relabeling to attach pod name/namespace labels to the metrics. I'll rework the PR.

grobie:
Here is an example of one metric exported by cadvisor (slightly reformatted to increase visibility).

google/cadvisor@2ed7198 kubelet@v1.3.5
```
container_cpu_cfs_periods_total{
  container_label_io_kubernetes_container_hash="5af8c3b4",
  container_label_io_kubernetes_container_name="sync",
  container_label_io_kubernetes_container_restartCount="1",
  container_label_io_kubernetes_container_terminationMessagePath="/dev/termination-log",
  container_label_io_kubernetes_pod_name="popularsearches-web-3165456836-2bfey",
  container_label_io_kubernetes_pod_namespace="popularsearches",
  container_label_io_kubernetes_pod_terminationGracePeriod="30",
  container_label_io_kubernetes_pod_uid="6a291e48-47c4-11e6-84a4-c81f66bdf8bd",
  id="/docker/68e1f15353921f4d6d4d998fa7293306c4ac828d04d1284e410ddaa75cf8cf25",
  image="redacted.com/popularsearches:42-16-ba6bd88",
  name="k8s_sync.5af8c3b4_popularsearches-web-3165456836-2bfey_popularsearches_6a291e48-47c4-11e6-84a4-c81f66bdf8bd_c02d3775"
} 72819
```

timstclair:
ok to test

timstclair:
Thus far we've avoided leaking Kubernetes concepts into cAdvisor, so I'm concerned about the precedent this is setting. I think I still don't quite understand what the problem here is. It sounds to me like the bug is that Kubernetes is abusing container labels, and this shouldn't be cAdvisor's problem.

If we really do need the logic in cAdvisor, I think a better approach would be to accept a list of labels to ignore, and have the Kubelet pass the list in rather than hardcoding it in cAdvisor. I also think the pod name, namespace, and container name are important to the identity and _should_ remain on the container.

grobie:
Alright, I've reworked the implementation and added a whitelist, allowing `io.kubernetes.pod.name` and `io.kubernetes.pod.namespace` still to be applied to all container metrics. There is currently no other way to attach such labels to cadvisor metrics.

grobie/cadvisor@12e1d20 kubelet@v1.3.5
```
container_cpu_cfs_periods_total{
  container_label_io_kubernetes_pod_name="popularsearches-web-3165456836-2bfey",
  container_label_io_kubernetes_pod_namespace="popularsearches",
  id="/docker/68e1f15353921f4d6d4d998fa7293306c4ac828d04d1284e410ddaa75cf8cf25",
  image="redacted.com/popularsearches:42-16-ba6bd88",
  name="k8s_sync.5af8c3b4_popularsearches-web-3165456836-2bfey_popularsearches_6a291e48-47c4-11e6-84a4-c81f66bdf8bd_c02d3775"
} 73700
```

grobie:
@timstclair I completely agree with you on leaking Kubernetes knowledge into cAdvisor. See my very first comment on the issue which started this discussion: https://github.com/google/cadvisor/issues/1312#issuecomment-240599649

That being said, without fundamentally changing Kubernetes, I don't see another way to work around this for now. cAdvisor reads these labels directly from the Docker API, so there is no other way to filter them then adding the knowledge to cADvisor. 

The huge problems with these labels is that they generate new timeseries every time they change. Prometheus is designed under the assumption that a labelset uniquely identifies one entity. These changed labels (e.g. when the restartCount gets increased) result in incorrect aggregation results for example. Some of the now filtered labels are static, but just noise and make it a lot harder to work with the cadvisor metrics.

Briefly before your comment I changed the implementation similar to your suggestion. I'm using a whitelist instead of a blacklist though. Pod name, namespace as well as the container name remain on the generated metrics as you can see in the example above.

PTAL again @jimmidyson @fabxc

grobie:
Alright, I didn't read your comment carefully enough. I'll rework it once again so that cadvisor simply accepts a filter function which can be set by kubernetes.

timstclair:
Yes, that was my suggestion. Specifically, it should be passed to [this function](https://github.com/google/cadvisor/blob/2ed7198f77395ee9a172878a0a7ab92ab59a2cfd/http/handlers.go#L92).

I would suggest renaming `ContainerNameToLabelsFunc` to `ContainerLablesFunc`, and have it take a few more parameters:
```
type ContainerNameToLabelsFunc func(containerName string, containerLabels map[string]string, containerEnv map[string]string) map[string]string
```
(also, make the sanitizeLabelName func exported, so it can be reused by implementers of the above function).

Finally, I would move all the existing label logic into an else condition. I.e., if that function is set, just use the output of that. Otherwise, do what we do today.

WDYT?

grobie:
I like the approach, any objections to make it even more generic and pass in the container object? There are also labels set based on `container.Aliases` and `container.Spec.Image` for example.

I'd then do something like `if labelsFunc == nil { labelsFunc = defaultLabelsFunc }` in the constructor.

jimmidyson:
SGTM. ContainerNameToLabelsFunc isn't used in kubernetes any more iirc. It was a stop gap until kubernetes implemented proper docker labels for this stuff.

muradm:
Will this fix the same behavior when using Docker Composer? Or we need another PR to remove container_label_com_docker_composer_* ?

grobie:
This change did not alter the behavior around metric labels. By default, cAdvisor will continue to export all available Docker labels using a container_label_* prefix. Kubernetes embeds cAdvisor and we pass a custom containerLabelsFunc there to only set a handful of explicitly chosen labels. I'm not aware of how cAdvisor is usually used in combination with docker compose, but something similar could probably be done. I suspect we could alternatively also provide a flag to let users explicitly set the list of labels they want to attach.

muradm:
Last proposal, with configurable set of labels seems to be most reasonable. In addition to the list it could be a map, something like:

`[-container.labels <label_name_as_set_on_container[:label_name_as_should_be_exported]> ... ]`

where `label_name_as_should_be_exported` is optional, if it is needed to change the name of exported label.

That would be great. Shall we open an issue for that?

Thanks in advance

grobie:
I guess opening a new issue and discussing this separate topic there makes
sense in either way.

PR #1429, opened 2016-08-20
Provide option to control Prometheus labels

### LINKED ISSUE #1312: Container restart count is a label

grobie:
This change generalizes the existing ContainerNameToLabelsFunc to allow the user to fully control all labels attached to exported Prometheus metrics. The existing behavior is available as DefaultContainerLabels and is used if no custom function is provided.

This will allow Kubernetes to filter out its internal Docker labels.

Fixes #1312. Once Kubernetes provides a custom function to filter its internal labels implemented in kubernetes/kubernetes#31064.

PTAL @jimmidyson @fabxc @vishh @timstclair 

@vishh Tim and I discussed that the current label settings classify as bug and that this should be included in Kubernetes v1.4.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -25,13 +25,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-// This will usually be manager.Manager, but can be swapped out for testing.
+// infoProvider will usually be manager.Manager, but can be swapped out for testing.
 type infoProvider interface {
-	// Get information about all subcontainers of the specified container (includes self).
+	// SubcontainersInfo provides information about all subcontainers of the
+	// specified container including itself.
 	SubcontainersInfo(containerName string, query *info.ContainerInfoRequest) ([]*info.ContainerInfo, error)
-	// Get information about the version.
+	// GetVersionInfo provides information about the version.
 	GetVersionInfo() (*info.VersionInfo, error)
-	// Get information about the machine.
+	// GetMachineInfo provides information about the machine.
 	GetMachineInfo() (*info.MachineInfo, error)
 }
 
@@ -56,8 +57,8 @@ func fsValues(fsStats []info.FsStats, valueFn func(*info.FsStats) float64) metri
 	return values
 }
 
-// A containerMetric describes a multi-dimensional metric used for exposing
-// a certain type of container statistic.
+// containerMetric describes a multi-dimensional metric used for exposing a
+// certain type of container statistic.
 type containerMetric struct {
 	name        string
 	help        string
@@ -71,21 +72,29 @@ func (cm *containerMetric) desc(baseLabels []string) *prometheus.Desc {
 	return prometheus.NewDesc(cm.name, cm.help, append(baseLabels, cm.extraLabels...), nil)
 }
 
-type ContainerNameToLabelsFunc func(containerName string) map[string]string
+// ContainerLabelsFunc defines all base labels and their values attached to
+// each metric exported by cAdvisor.
+type ContainerLabelsFunc func(*info.ContainerInfo) map[string]string
 
 // PrometheusCollector implements prometheus.Collector.
 type PrometheusCollector struct {
-	infoProvider          infoProvider
-	errors                prometheus.Gauge
-	containerMetrics      []containerMetric
-	containerNameToLabels ContainerNameToLabelsFunc
+	infoProvider        infoProvider
+	errors              prometheus.Gauge
+	containerMetrics    []containerMetric
+	containerLabelsFunc ContainerLabelsFunc
 }
 
-// NewPrometheusCollector returns a new PrometheusCollector.
-func NewPrometheusCollector(infoProvider infoProvider, f ContainerNameToLabelsFunc) *PrometheusCollector {
+// NewPrometheusCollector returns a new PrometheusCollector. The passed
+// ContainerLabelsFunc specifies which base labels will be attached to all
+// exported metrics. If left to nil, the DefaultContainerLabels function
+// will be used instead.
+func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCollector {
+	if f == nil {
+		f = DefaultContainerLabels
+	}
 	c := &PrometheusCollector{
-		infoProvider:          infoProvider,
-		containerNameToLabels: f,
+		infoProvider:        i,
+		containerLabelsFunc: f,
 		errors: prometheus.NewGauge(prometheus.GaugeOpts{
 			Namespace: "container",
 			Name:      "scrape_error",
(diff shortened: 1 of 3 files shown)


metrics/prometheus.go
```diff
@@ -537,6 +543,26 @@ const (
...
+// values.
+func DefaultContainerLabels(container *info.ContainerInfo) map[string]string {
+	set := map[string]string{"id": container.Name}
+	if len(container.Aliases) > 0 {
+		set["name"] = container.Aliases[0]
+	}
+	if image := container.Spec.Image; len(image) > 0 {
+		set["image"] = image
+	}
+	for k, v := range container.Spec.Labels {
```
jimmidyson: Should containerLabelPrefix be exported so custom mills can use it?

metrics/prometheus.go
```diff
@@ -537,6 +543,26 @@ const (
 	containerEnvPrefix   = "container_env_"
 )
 
+// DefaultContainerLabels implements ContainerLabelsFunc. It exports the
+// container name, first alias, image name as well as all its env and label
+// values.
+func DefaultContainerLabels(container *info.ContainerInfo) map[string]string {
+	set := map[string]string{"id": container.Name}
```
jimmidyson: So this now handles all labels, including id name etc. I personally like that control for custom impls but it needs to be documented somewhere as well as release noted.

grobie: Sure! Any suggestions/pointers where to document that?

jimmidyson: Let's just makes it clear in the godoc comment for now & make sure it's clear when used in kubernetes too.

metrics/prometheus.go
grobie: Done.

metrics/prometheus.go
grobie: I extended the documentation a bit, but I'm not completely sure what you'd like to see here. I had hoped the comments on ContainerLabelsFunc, NewPrometheusCollector and DefaultContainerLabels were helpful. Please let me know what else you'd like to see added here.

metrics/prometheus.go
```diff
@@ -533,10 +541,34 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
...
+	// ContainerEnvPrefix is the prefix added to all env variables added to
+	// exported metrics.
+	ContainerEnvPrefix = "container_env_"
 )
 
+// DefaultContainerLabels implements ContainerLabelsFunc. It exports the
+// container name, first alias, image name as well as all its env and label
+// values.
+func DefaultContainerLabels(container *info.ContainerInfo) map[string]string {
+	set := map[string]string{"id": container.Name}
```
jimmidyson: That's good enough for me - thanks!

jimmidyson:
LGTM - @timstclair?

fabxc:
LGTM

jimmidyson:
@timstclair Can we get this in 0.24 please?

timstclair:
ok to test

timstclair:
@jimmidyson Yes. I just retagged `v0.24.0` to `v0.24.0-alpha1` (I know it's bad to change tags, but that was the original intent). We can cherrypick this into the `release-v0.24` branch so it will be included in the actual v0.24 release in a couple weeks.

timstclair:
Oh yeah, our cAdvisor e2e tests are hosed right now. I'll retest this once they're back up and running.

metrics/prometheus.go
```diff
@@ -533,10 +541,34 @@ func (c *PrometheusCollector) Collect(ch chan<- prometheus.Metric) {
...
+	if len(container.Aliases) > 0 {
+		set["name"] = container.Aliases[0]
+	}
+	if image := container.Spec.Image; len(image) > 0 {
+		set["image"] = image
+	}
```
timstclair: Should we export these intermediate steps for reuse? Or at least make the keys (`image`, `name`, `id`) exported?

grobie:
I can re-open this PR against the release-v0.24 branch if that helps to
avoid cherry-picking?

timstclair:
No, it should be merged into master too.

timstclair:
Just a couple small comments. Mostly looks good.

timstclair:
@k8s-bot test this

grobie:
@timstclair of course. I thought the general pattern for bug fixes is to apply them against the release branch and then merge that release branch into master. but whatever works for you.

timstclair:
Yeah, we could consider adopting that practice, but for now I'm just mirroring the practices used by Kubernetes (which is to cherry-pick from master into release).

metrics/prometheus.go
grobie: I'm a bit unsure about exposing so many parts. If someone overwrites the function, it applies to their whole setup anyway, why would they need to care about how the labels in other clusters look like?

I could split up the id+name+image part and the labels+env parts into two functions we expose each? I'm also fine with exposing constants for the labels though.

Lastly, actually we're doing it all wrong here by applying so many labels to all metrics. It should be only one identifier (probably id) and export another metric containing the mapping. See http://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas/ and http://www.robustperception.io/exposing-the-software-version-to-prometheus/.

grobie: Very last part, the env labels have never worked for me so far, they're always empty (I think that's a bug somewhere in cadvisor). When I use `docker inspect <container>` I see many envs listed for kubernetes managed containers, but I've never seen any env labels on our metrics.

Shall we just remove the env part as we're already working on that part? @jimmidyson

timstclair:
@k8s-bot test this

metrics/prometheus.go
timstclair: > I could split up the id+name+image part and the labels+env parts into two functions we expose each? I'm also fine with exposing constants for the labels though. 

Thinking about this some more, I think just exposing the constants is the way to go.

> Lastly, actually we're doing it all wrong here by applying so many labels to all metrics. It should be only one identifier (probably id) and export another metric containing the mapping. See http://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas/ and http://www.robustperception.io/exposing-the-software-version-to-prometheus/.

Ack. I think this is too big a change for this late in the release though. Could you open an issue with a proposal to fix it, and we can try to address it in the next (v0.25) release. Thanks!

jimmidyson: > the env labels have never worked for me so far, they're always empty

IIRC env vars are whitelisted by the `docker_env_metadata_whitelist` flag. By default that is empty as env vars could potentially contain sensitive data so must be consciously whitelisted.

grobie: Ah, ok. Thanks for the explanation. I still think it's a bad practice to expose such as labels, but I'll not introduce more breaking behavior here and mention that as well in a follow up issue.

grobie: @timstclair Done. Extracted the id/name/image to be exported constants.

timstclair:
LGTM. Waiting for tests to merge.

grobie:
Cool, thanks. Once this is merged, I'll update my kubernetes PR to use the constants and the real master commit ID in godeps.json.

timstclair:
ok to test

timstclair:
@k8s-bot test this

timstclair:
Please send a PR with this change against the `release-v0.24` branch, then you can update the k8s godeps from there.

PR #1457, opened 2016-09-13
Wavefront Storage Driver for cAdvisor

ezeev:


wmturner:
Can we please see some traction on this PR?  It is currently very frustrating using the non-official repo especially when some infrastructure is not in WF.

Issue #1458, opened 2016-09-13
Rethink storage driver model


timstclair:
Our current model for cAdvisor storage backends involves checking in code that translates the cAdvisor API and pushes out or exposes the stats for each storage backend. The problem with this approach is that we don't have a good story for maintenance of all the storage drivers, and expecting the current cAdvisor maintainers to maintain a driver for every storage solution on the market is unsalable.

For example, we currently have 6 PRs open to add new storage drivers: https://github.com/google/cadvisor/pulls?q=is%3Aopen+is%3Apr+label%3Aarea%2Fstorage

There are a couple ways we could handle storage drivers if we moved them out of cAdvisor

1. Implement the drivers as a shim process which gets stats pushes from cAdvisor (e.g. through IPC or a REST endpoint), and takes the appropriate action. This could be implemented as a separate process in the same container, or another container (a "sidecar").

2. Encourage wrapper implementations or forks of cAdvisor which add the necessary functionality for each driver. We could create an "official" list of forks organized by drivers to make these more discoverable. We could also reorganize the cAdvisor main package to make this easier.

Any other ideas for how to deal with this issue?

@google/cadvisor @vishh @dchen1107 @jimmidyson @derekwaynecarr @pmorie @ncdc

ncdc:
I'm leaning towards option 1.

Having had the experience we have being a downstream consumer of Kubernetes in OpenShift, I am really scared of what it would entail to keep forks up to date as cadvisor continues to evolve. If we do end up choosing option 2, we would need to have explicit guarantees about API compatibility, so fork maintainers have as few headaches as possible when it comes time to rebase cadvisor into their fork.

vishh:
+1 for `1` as well. It should be fairly straightforward to make the
existing storage plugins logic a standalone entity.
Polling might introduce time sync issues, but we can hopefully address that
with a general purpose library that all storage plugins can re-use.

timstclair:
Yeah, a general purpose library sounds good. We could implement it so that it uses the exact same StorageDriver interface we already have, and all that would need to be done is build a binary out of it.

timstclair:
I'll put together a PoC. Any objections to using go's standard `net/rpc` package over a unix socket?

vishh:
Why not use the existing REST API? We have enough trouble maintaining that
one. We are yet to deprecate the v1 API and move to a v2 after completing
it.

timstclair:
[...]

PR #1462, opened 2016-09-17
RFC: Client/server storage driver over RPC

timstclair:
Proof-of-concept storage driver over RPC. A server implementation is as simple as:

``` go
package main

import "github.com/google/cadvisor/storage/plugin/server"

func main() {
  driver := newStorageDriver() // Implements the StorageDriver interface
  server.Run(driver)
}
```

Then, start cAdvisor with the `--storage_driver=plugin` option to connect.

Eventually I'd like to move all storage drivers to this model. It could either be the users responsibility to run them, or the binary could be passed to cAdvisor (through a flag), and cAdvisor would handle starting (and restarting, if needed) the server.

For https://github.com/google/cadvisor/issues/1458

/cc @vishh @ncdc

### DIFF
--- a/storage/plugin/server/server.go
+++ b/storage/plugin/server/server.go
@@ -1,0 +1,139 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


storage/plugin/server/server.go
```diff
@@ -0,0 +1,139 @@
...
+
+	conn io.Closer
+}
+
+var (
+	// The default socket to listen on.
+	DefaultSocket = "/var/run/cadvisor-storage.sock"
+
+	// Signals to listen on to shut down the server.
+	TerminationSignals = []os.Signal{syscall.SIGINT, syscall.SIGKILL, syscall.SIGHUP}
```
ncdc: Did you intend to use this on line 93? Also, you can't trap `SIGKILL`, can you?

timstclair: Yup. Done.

PR #1464, opened 2016-09-19
Cherrypick 1.3 changelog

timstclair:
Cherrypick release notes from `release-v0.23` branch into `master`

/cc @vishh @dchen1107 @sjenning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 0.23.9 (2016-08-09)
+
+- Cherry-pick release:
+  - Ensure minimum kernel version for thin_ls
+
 ### 0.23.8 (2016-08-02)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,41 @@
 # Changelog
 
+### 0.23.9 (2016-08-09)
```
ncdc: Do you want whitespace after each version, or not (like 0.23.3 and 4)?

timstclair: GitHub renders lists without the whitespace, but it's not "proper markdown", and some engines do not, so I think we should use whitespace. For this PR, I'd prefer not to touch the original commits, but I'll send a follow-up to clean up the file.

ncdc: SGTM

PR #1473, opened 2016-09-20
Switch to Prometheus decoders for Prometheus custom metric endpoint parsing

### LINKED ISSUE #1472: Prometheus collector does not handle spaces in label values
The Prometheus metrics collector does not appear to handle spaces in the label values. For example, this metric:
```
my_metric{name="Network Agent"} 0
```

Will cause the following exception:
```
Failed to update stats for container "/docker/c039": Error 0: Error 0: strconv.ParseFloat: parsing "Agent\"} 0"
```

This is a valid label value according to the [Prometheus docs](https://prometheus.io/docs/concepts/data_model/): "Label values may contain any Unicode characters."

Problem seems to be that the parsing code treats the first space as the delimiter between the metric and its value, regardless of where it appears on the line: https://github.com/google/cadvisor/blob/552d2578ecc1d51118da5804bd19e832666b361d/collector/prometheus_collector.go#L188

jimmidyson:
Fixes #1472

@fabxc This is a hand-rolled parser to collect application metrics from containers. Forgetting the question or application metrics & cadvisor, I assume the better approach for this would be to use Prometheus' own parser? If so, can you point me to it please?

### DIFF
--- a/collector/prometheus_collector.go
+++ b/collector/prometheus_collector.go
@@ -17,12 +17,13 @@ package collector
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"math"
+	"io"
 	"net/http"
-	"strconv"
-	"strings"
 	"time"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
 
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/info/v1"
@@ -100,60 +101,61 @@ func (collector *PrometheusCollector) Name() string {
 	return collector.name
 }
 
-func getMetricData(line string) string {
-	fields := strings.Fields(line)
-	data := fields[3]
-	if len(fields) > 4 {
-		for i := range fields {
-			if i > 3 {
-				data = data + "_" + fields[i]
-			}
-		}
-	}
-	return strings.TrimSpace(data)
-}
-
 func (collector *PrometheusCollector) GetSpec() []v1.MetricSpec {
-	specs := []v1.MetricSpec{}
 
 	response, err := collector.httpClient.Get(collector.configFile.Endpoint.URL)
 	if err != nil {
-		return specs
+		return nil
 	}
 	defer response.Body.Close()
 
-	pageContent, err := ioutil.ReadAll(response.Body)
-	if err != nil {
-		return specs
+	if response.StatusCode != http.StatusOK {
+		return nil
 	}
 
-	lines := strings.Split(string(pageContent), "\n")
-	lineCount := len(lines)
-	for i, line := range lines {
-		if strings.HasPrefix(line, "# HELP") {
-			if i+2 >= lineCount {
-				break
-			}
+	dec := expfmt.NewDecoder(response.Body, expfmt.ResponseFormat(response.Header))
 
-			stopIndex := strings.IndexAny(lines[i+2], "{ ")
-			if stopIndex == -1 {
-				continue
-			}
+	var specs []v1.MetricSpec
 
-			name := strings.TrimSpace(lines[i+2][0:stopIndex])
-			if _, ok := collector.metricsSet[name]; collector.metricsSet != nil && !ok {
-				continue
-			}
-			spec := v1.MetricSpec{
-				Name:   name,
-				Type:   v1.MetricType(getMetricData(lines[i+1])),
-				Format: "float",
-				Units:  getMetricData(lines[i]),
-			}
-			specs = append(specs, spec)
+	for {
+		d := dto.MetricFamily{}
+		if err = dec.Decode(&d); err != nil {
+			break
 		}
+		name := d.GetName()
+		if len(name) == 0 {
+			continue
+		}
+		if _, ok := collector.metricsSet[name]; collector.metricsSet != nil && !ok {
+			continue
+		}
+		spec := v1.MetricSpec{
+			Name:   name,
+			Type:   metricType(d.GetType()),
+			Format: v1.FloatType,
+		}
+		specs = append(specs, spec)
 	}
+
+	if err != nil && err != io.EOF {
+		return nil
+	}
+
 	return specs
+}
+
+// metricType converts Prometheus metric type to cadvisor metric type.
+// If there is no mapping then just return the name of the Prometheus metric type.
+func metricType(t dto.MetricType) v1.MetricType {
+	switch t {
+	case dto.MetricType_COUNTER:
+		return v1.MetricCumulative
+	case dto.MetricType_GAUGE:
+		return v1.MetricGauge
+	default:
+		return v1.MetricType(t.String())
+	}
+
 }
 
 //Returns collected metrics and the next collection time of the collector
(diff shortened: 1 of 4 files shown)


jimmidyson:
`mkdir: cannot create directory ‘/tmp/cadvisor-768’: No space left on device` - @timstclair any chance you could sort that out?

fabxc:
For reference, the parser is located in https://github.com/prometheus/common/tree/master/expfmt

jimmidyson:
@fabxc Would be you be able to take a look if you have some time? Switched to expfmt decoders & seems to work nicely (few tweaks to tests but nothing fundamental changing in functionality).

timstclair:
@jimmidyson I'm aware of the issue. I'll try to have it sorted out by EOD

timstclair:
@k8s-bot test this

collector/prometheus_collector.go
```diff
@@ -168,59 +170,54 @@ func (collector *PrometheusCollector) Collect(metrics map[string][]v1.MetricVal)
...
+				continue
 			}
 
 			metric := v1.MetricVal{
-				Label:      metLabel,
-				FloatValue: metVal,
-				Timestamp:  currentTime,
+				FloatValue: float64(sample.Value),
+				Timestamp:  sample.Timestamp.Time(),
 			}
```
fabxc: Not sure I entirely understand this. We seem to ignore the labels of the samples entire and just append timestamp/value pairs for the metric name?

jimmidyson: Ah yeah I forgot to ask what the purpose of this label was. Considering there are likely to be multiple labels on each metric, what should this be set to? @timstclair?

collector/prometheus_collector.go
```diff
@@ -17,13 +17,14 @@ package collector
...
 	"fmt"
-	"io/ioutil"
-	"math"
+	"io"
 	"net/http"
-	"strconv"
-	"strings"
 	"time"
 
+	dto "github.com/prometheus/client_model/go"
```
timstclair: what does `dto` stand for?

collector/prometheus_collector.go
```diff
@@ -168,59 +170,54 @@ func (collector *PrometheusCollector) Collect(metrics map[string][]v1.MetricVal)
...
+		Dec: expfmt.NewDecoder(response.Body, expfmt.ResponseFormat(response.Header)),
+		Opts: &expfmt.DecodeOptions{
+			Timestamp: model.TimeFromUnixNano(currentTime.UnixNano()),
+		},
+	}
 
-	for _, line := range lines {
-		if line == "" {
+	var (
+		decSamples = make(model.Vector, 0, 50)
```
timstclair: How'd you pick 50? What happens if this overflows?

fabxc: It's just the initial capacity. The slice is appended to, so it just allocates more if we go beyond.
50 SGTM. It's a number of metrics we can expect from virtually any endpoint, grow allocations beyond that are fine. But we basically save 6 re-allocations for the general case.

collector/prometheus_collector.go
fabxc: Data transfer object. Kinda nasty, but better than `go` and kinda what we use everywhere else in Prometheus land.

collector/prometheus_collector.go
jimmidyson: @timstclair @DirectXMan12 What should the value of label be in this case? A concatenation of all Prometheus labels?

collector/prometheus_collector.go
```diff
@@ -100,62 +101,63 @@ func (collector *PrometheusCollector) Name() string {
...
+// metricType converts Prometheus metric type to cadvisor metric type.
+// If there is no mapping then just return the name of the Prometheus metric type.
+func metricType(t dto.MetricType) v1.MetricType {
+	switch t {
+	case dto.MetricType_COUNTER:
+		return v1.MetricCumulative
+	case dto.MetricType_GAUGE:
+		return v1.MetricGauge
+	default:
+		return v1.MetricType(t.String())
```
DirectXMan12: uhh..., if people are expecting this to be only one of these set values, this is going to throw them for a loop.
`v1.MetricType`  seems to be an implicit enumeration (since Go doesn't have explicit enumerations :-/), so I really don't think it's a good idea to arbitrarily add extra values to the enum at runtime.

collector/prometheus_collector.go
```diff
@@ -100,62 +101,63 @@ func (collector *PrometheusCollector) Name() string {
...
+	for {
+		d := dto.MetricFamily{}
+		if err = dec.Decode(&d); err != nil {
+			break
+		}
+		name := d.GetName()
+		if len(name) == 0 {
+			continue
 		}
+		if _, ok := collector.metricsSet[name]; collector.metricsSet != nil && !ok {
```
DirectXMan12: this could really use a comment to indicate why it's checking this (e.g. `// skip any metrics not on our list to collect`).  I realize that the old code doesn't have a comment, but we should fix that (I blame myself :-P).

collector/prometheus_collector.go
DirectXMan12: you should probably add a comment noting that it's just a guess at a useful initial capacity

collector/prometheus_collector.go
DirectXMan12: >  A concatenation of all Prometheus labels?

Yeah, seems like it.  That seems... poor.  It works *ok* for one label, but multiple labels could come in any order.  If we wanted to make sure this was comparable, we could sort them first.  Doesn't look like the label is actually carried through to the Kube summary API.

collector/prometheus_collector.go
DirectXMan12: none-the-less, something more expressive might be in order (perhaps `rawmodel` or `rawtypes`?)

collector/prometheus_collector.go
timstclair: Yeah, I agree this is not ideal. Unfortunately this requires an API change to fix properly. Can you file an issue? Sorting + concatenating sounds good for now. Check the allowed symbols on prometheus labels to decide how to concatenate.

fabxc: In label values all unicode runes are allowed. In Prometheus itself we hance us `\0xff` as an invalid rune as a separator for unambiguous concatenation.

collector/prometheus_collector.go
fabxc: Agreed

collector/prometheus_collector.go
jimmidyson: Types aren't arbitrary, it's just that cadvisor doesn't support the same metric types as Prometheus & there's no way to map e.g. a Prometheus histogram to a cadvisor metric type.

collector/prometheus_collector.go
jimmidyson: Done.

collector/prometheus_collector.go
jimmidyson: Still not sure how to handle this nicely... This PR doesn't fix the existing problem, but doesn't make it worse so will raise a separate issue to figure out how to handle it, but assume that to be in a separate PR.

collector/prometheus_collector.go
jimmidyson: Done.

collector/prometheus_collector.go
jimmidyson: Done.

collector/prometheus_collector.go
jimmidyson: Hope what I've done it right... always confused by unicode stuff like this. Added tests too.

jimmidyson:
Ready for a second round review.

jimmidyson:
@timstclair What's left to get this merged?

timstclair:
LGTM, feel free to merge.

PR #1488, opened 2016-10-02
Add udp and udp6 network statistics

tcolgate:
This is a first crack at adding UDP stats collection support.
One possible issue here is that the dropped packets is a coutner.
if many sockets come and go withint he container, the sum of the
dropped packet counters will not be monotonically increasing.

### DIFF
--- a/info/v2/container.go
+++ b/info/v2/container.go
@@ -269,6 +269,10 @@ type NetworkStats struct {
 	Tcp TcpStat `json:"tcp"`
 	// TCP6 connection stats (Established, Listen...)
 	Tcp6 TcpStat `json:"tcp6"`
+	// UDP connection stats
+	Udp v1.UdpStat `json:"udp"`
+	// UDP6 connection stats
+	Udp6 v1.UdpStat `json:"udp6"`
 }
 
 // Instantaneous CPU stats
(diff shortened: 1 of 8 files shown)


tcolgate:
/proc/net/stat may be a better source of a single counter

dashpole:
@k8s-bot ok to test

info/v2/container.go
```diff
@@ -262,13 +262,23 @@ type TcpStat struct {
 	Closing     uint64
 }
 
+type UdpStat struct {
```
dashpole: I realize we duplicate structs for TcpStat, but I would prefer if you removed this struct, and used v1.UdpStat in this file

cmluciano:
Is the test failure a known flake?

dashpole:
Yes, something is up with our test infrastructure.  I have been looking into it

tcolgate:
Please don't merge, I'll add the rx_queue and tx_queue, and think it could do with a bit ore error checking.

tcolgate:
Fixed dropped to actuall do the right thing.
Add RxQueued and TxQueued
Added a test.

dashpole:
ignore the test failure for now.  It is an infrastructure issue

tcolgate:
@dashpole anything else I need to do for this? Is it just a case of waiting for someone to merge it?

dashpole:
@k8s-bot test this

tcolgate:
@dashpole pushed a rebase

dashpole:
Looks good!

PR #1489, opened 2016-10-03
Cadvisor now publishes per-container inode stats

dashpole:
Per-container stats available in the v2 api in the FileSystemStats.InodeUsage field.  They are available in the v1 stats in the Filesystem.Inodes field.  Per-container Inode stats collected using the command "find . -xdev | wc -l".  The collection of inode stats uses the same back-off strategy as measuring per container bytes used, so that slow calls to find do not severely impact performance.

### DIFF
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -16,6 +16,7 @@
 package common
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
@@ -26,71 +27,85 @@ import (
 
 type FsHandler interface {
 	Start()
-	Usage() (baseUsageBytes uint64, totalUsageBytes uint64)
+	Usage() FsUsage
 	Stop()
+}
+
+type FsUsage struct {
+	BaseUsageBytes  uint64
+	TotalUsageBytes uint64
+	InodeUsage      uint64
 }
 
 type realFsHandler struct {
 	sync.RWMutex
-	lastUpdate     time.Time
-	usageBytes     uint64
-	baseUsageBytes uint64
-	period         time.Duration
-	minPeriod      time.Duration
-	rootfs         string
-	extraDir       string
-	fsInfo         fs.FsInfo
+	lastUpdate time.Time
+	usage      FsUsage
+	period     time.Duration
+	minPeriod  time.Duration
+	rootfs     string
+	extraDir   string
+	fsInfo     fs.FsInfo
 	// Tells the container to stop.
 	stopChan chan struct{}
 }
 
 const (
-	longDu             = time.Second
-	duTimeout          = time.Minute
-	maxDuBackoffFactor = 20
+	longOp           = time.Second
+	timeout          = 2 * time.Minute
+	maxBackoffFactor = 20
 )
+
+const DefaultPeriod = time.Minute
 
 var _ FsHandler = &realFsHandler{}
 
 func NewFsHandler(period time.Duration, rootfs, extraDir string, fsInfo fs.FsInfo) FsHandler {
 	return &realFsHandler{
-		lastUpdate:     time.Time{},
-		usageBytes:     0,
-		baseUsageBytes: 0,
-		period:         period,
-		minPeriod:      period,
-		rootfs:         rootfs,
-		extraDir:       extraDir,
-		fsInfo:         fsInfo,
-		stopChan:       make(chan struct{}, 1),
+		lastUpdate: time.Time{},
+		usage:      FsUsage{},
+		period:     period,
+		minPeriod:  period,
+		rootfs:     rootfs,
+		extraDir:   extraDir,
+		fsInfo:     fsInfo,
+		stopChan:   make(chan struct{}, 1),
 	}
 }
 
 func (fh *realFsHandler) update() error {
 	var (
-		baseUsage, extraDirUsage uint64
-		err                      error
+		baseUsage, extraDirUsage, inodeUsage    uint64
+		rootDiskErr, rootInodeErr, extraDiskErr error
 	)
 	// TODO(vishh): Add support for external mounts.
 	if fh.rootfs != "" {
-		baseUsage, err = fh.fsInfo.GetDirUsage(fh.rootfs, duTimeout)
-		if err != nil {
-			return err
-		}
+		baseUsage, rootDiskErr = fh.fsInfo.GetDirDiskUsage(fh.rootfs, timeout)
+		inodeUsage, rootInodeErr = fh.fsInfo.GetDirInodeUsage(fh.rootfs, timeout)
 	}
 
 	if fh.extraDir != "" {
-		extraDirUsage, err = fh.fsInfo.GetDirUsage(fh.extraDir, duTimeout)
-		if err != nil {
-			return err
-		}
+		extraDirUsage, extraDiskErr = fh.fsInfo.GetDirDiskUsage(fh.extraDir, timeout)
 	}
 
+	// Wait to handle errors until after all operartions are run.
+	// An error in one will not cause an early return, skipping others
 	fh.Lock()
 	defer fh.Unlock()
 	fh.lastUpdate = time.Now()
-	fh.usageBytes = baseUsage + extraDirUsage
-	fh.baseUsageBytes = baseUsage
+	if rootDiskErr == nil && fh.rootfs != "" {
+		fh.usage.InodeUsage = inodeUsage
+	}
+	if rootInodeErr == nil && fh.rootfs != "" {
+		fh.usage.TotalUsageBytes = baseUsage + extraDirUsage
+	}
+	if extraDiskErr == nil && fh.extraDir != "" {
+		fh.usage.BaseUsageBytes = baseUsage
+	}
+	// Combine errors into a single error to return
+	if rootDiskErr != nil || rootInodeErr != nil || extraDiskErr != nil {
+		return fmt.Errorf("rootDiskErr: %v, rootInodeErr: %v, extraDiskErr: %v", rootDiskErr, rootInodeErr, extraDiskErr)
+	}
 	return nil
 }
 
@@ -105,15 +120,15 @@ func (fh *realFsHandler) trackUsage() {
 			if err := fh.update(); err != nil {
 				glog.Errorf("failed to collect filesystem stats - %v", err)
 				fh.period = fh.period * 2
-				if fh.period > maxDuBackoffFactor*fh.minPeriod {
-					fh.period = maxDuBackoffFactor * fh.minPeriod
+				if fh.period > maxBackoffFactor*fh.minPeriod {
+					fh.period = maxBackoffFactor * fh.minPeriod
 				}
 			} else {
 				fh.period = fh.minPeriod
 			}
 			duration := time.Since(start)
-			if duration > longDu {
-				glog.V(2).Infof("`du` on following dirs took %v: %v", duration, []string{fh.rootfs, fh.extraDir})
+			if duration > longOp {
+				glog.V(2).Infof("du and find on following dirs took %v: %v", duration, []string{fh.rootfs, fh.extraDir})
 			}
 		}
 	}
@@ -127,8 +142,8 @@ func (fh *realFsHandler) Stop() {
 	close(fh.stopChan)
 }
 
-func (fh *realFsHandler) Usage() (baseUsageBytes, totalUsageBytes uint64) {
+func (fh *realFsHandler) Usage() FsUsage {
 	fh.RLock()
 	defer fh.RUnlock()
-	return fh.baseUsageBytes, fh.usageBytes
+	return fh.usage
 }
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -19,6 +19,7 @@ package fs
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -44,24 +45,24 @@ const (
 	LabelRktImages    = "rkt-images"
 )
 
-// The maximum number of `du` tasks that can be running at once.
-const maxConsecutiveDus = 20
+// The maximum number of `du` and `find` tasks that can be running at once.
+const maxConcurrentOps = 20
 
-// A pool for restricting the number of consecutive `du` tasks running.
-var duPool = make(chan struct{}, maxConsecutiveDus)
+// A pool for restricting the number of consecutive `du` and `find` tasks running.
+var pool = make(chan struct{}, maxConcurrentOps)
 
 func init() {
-	for i := 0; i < maxConsecutiveDus; i++ {
-		releaseDuToken()
+	for i := 0; i < maxConcurrentOps; i++ {
+		releaseToken()
 	}
 }
 
-func claimDuToken() {
-	<-duPool
+func claimToken() {
+	<-pool
 }
 
-func releaseDuToken() {
-	duPool <- struct{}{}
+func releaseToken() {
+	pool <- struct{}{}
 }
 
 type partition struct {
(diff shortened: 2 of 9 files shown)


dashpole:
Issue: https://github.com/kubernetes/kubernetes/issues/33382

container/common/fsHandler.go
```diff
@@ -119,16 +149,42 @@ func (fh *realFsHandler) trackUsage() {
...
+			} else {
+				fh.findPeriod = fh.minFindPeriod
+			}
+			duration := time.Since(start)
+			if duration > longFind {
+				glog.V(2).Infof("`find` on following dirs took %v: %v", duration, []string{fh.rootfs, fh.extraDir})
+			}
+		}
+	}
+}
```
timstclair: Actually, on further thought I'm wondering if this should be in the same routine as the `du` call. The advantages are that containers with very large filesystems wouldn't hog as many exec calls, and there might be less disk thrashing.

fs/fs.go
```diff
@@ -456,19 +473,78 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
 	return usageInKb * 1024, nil
 }
 
+func (self *RealFsInfo) GetDirInodeUsage(dir string, timeout time.Duration) (uint64, error) {
+	if dir == "" {
+		return 0, fmt.Errorf("invalid directory")
+	}
+	claimFindToken()
+	defer releaseFindToken()
+	findCmd := exec.Command("find", dir, "-xdev")
```
timstclair: When I tried this, I found `find $dir -xdev -printf '.' | wc -c` to be about twice as fast. Might be a premature optimization though...

container/common/fsHandler.go
```diff
@@ -119,16 +149,42 @@ func (fh *realFsHandler) trackUsage() {
 	}
 }
 
+func (fh *realFsHandler) trackInodeUsage() {
+	fh.updateInodeUsage()
```
timstclair: don't ignore the error

fs/fs.go
```diff
@@ -456,19 +473,78 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
+	if err := wcCmd.Start(); err != nil {
+		return 0, fmt.Errorf("failed to exec cmd %v - %v", wcCmd.Args, err)
+	}
+	stdoutb, souterr := ioutil.ReadAll(stdoutp)
+	findStderrb, _ := ioutil.ReadAll(findStderrp)
+	wcStderrb, _ := ioutil.ReadAll(wcStderrp)
+	timer := time.AfterFunc(timeout, func() {
+		glog.Infof("killing cmd %v, and cmd %v due to timeout(%s)", findCmd.Args, wcCmd.Args, timeout.String())
+		wcCmd.Process.Kill()
+		findCmd.Process.Kill()
```
timstclair: A timeout should be an error. Does killing the findCmd cause `Wait()` to return an error?

fs/fs.go
```diff
@@ -456,19 +473,78 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
+	if err != nil {
+		return 0, fmt.Errorf("cmd %v failed with output stdout: %s, stderr: %s - %v", findCmd.Args, string(stdoutb), string(findStderrb), err)
+	}
+	err = wcCmd.Wait()
+	if err != nil {
+		return 0, fmt.Errorf("cmd %v failed with output stdout: %s, stderr: %s - %v", wcCmd.Args, string(stdoutb), string(wcStderrb), err)
+	}
+	timer.Stop()
+	stdout := string(stdoutb)
+	if souterr != nil {
```
timstclair: nit: move the error handling to just below the function that returns the error.

fs/fs.go
```diff
@@ -456,19 +473,78 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
+		glog.Errorf("failed to read from stdout for cmds: %v, %v - %v", findCmd.Args, wcCmd.Args, souterr)
+	}
+	// return 0, fmt.Errorf("cmd %v output %s", cmd.Args, stdout)
```
timstclair: delete this

fs/fs.go
```diff
@@ -456,19 +473,78 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
+	inodeUsage, err := strconv.ParseUint(strings.Fields(stdout)[0], 10, 64)
```
timstclair: Is the fields call necessary?

container/common/fsHandler.go
```diff
@@ -119,16 +149,42 @@ func (fh *realFsHandler) trackUsage() {
...
+				glog.V(2).Infof("`find` on following dirs took %v: %v", duration, []string{fh.rootfs, fh.extraDir})
+			}
+		}
+	}
+}
+
 func (fh *realFsHandler) Start() {
-	go fh.trackUsage()
+	go fh.trackDiskUsage()
+	go fh.trackInodeUsage()
```
timstclair: Does it make sense to track inode usage for devicemapper? If not we should skip it.

fs/fs.go
timstclair: I think this will double count inodes on base layers shared across containers. That's probably what we want, but it means killing a countainer will not reclaim all it's reported inodes.

fs/fs_test.go
```diff
@@ -100,11 +100,28 @@ func TestDirUsage(t *testing.T) {
...
+	as.NoError(err)
+	defer os.RemoveAll(dir)
+	numFiles := 100
+	for i := 0; i < numFiles; i++ {
+		_, err := ioutil.TempFile(dir, "")
+		as.NoError(err)
+	}
+	inodes, err := fsInfo.GetDirInodeUsage(dir, time.Minute)
+	as.NoError(err)
+	as.True(uint64(numFiles) <= inodes, "expected inodes in dir to be at-least %d; got inodes: %d", numFiles, inodes)
```
timstclair: shouldn't `numFiles == inodes`? Under what circumstances would it be different?

info/v2/conversion.go
```diff
@@ -127,6 +127,7 @@ func ContainerStatsFromV1(spec *v1.ContainerSpec, stats []*v1.ContainerStats) []
 				stat.Filesystem = &FilesystemStats{
 					TotalUsageBytes: &val.Filesystem[0].Usage,
 					BaseUsageBytes:  &val.Filesystem[0].BaseUsage,
+					InodeUsage:      &val.Filesystem[0].Inodes,
```
timstclair: If we skip inode collection for some storage drivers, then we should probably leave `InodeUsage` nil here when it's not collected.

fs/fs_test.go
dashpole: I believe it is actually, numFiles + 1 == inodes, since the directory I created counts as well.  I could change it to that, but I wasn't sure if it would consistently be the case.

timstclair:
ok to test

timstclair:
@k8s-bot test this

ixdy:
@k8s-bot ok to test

spxtr:
@k8s-bot test this

It didn't have permissions to set the status, but the test ran :P

container/common/fsHandler.go
vishh: AFAIK, inodes cannot be tracked for devicemapper.

fs/fs.go
vishh: It should not be doing that. `dir` here is expected to be the writable layer of a container. We need to test this.

info/v2/conversion.go
vishh: +1

fs/fs.go
dashpole: Just manually tested this since I couldn't find it on the web.  Killing the command causes Wait to return an error.
Also, while testing, I found that the timeouts were not working properly for this command.  
ioutil.ReadAll blocks until the end of the stream, so timeouts were waiting until the command finished to start...
Ill add a test to ensure that timeouts function properly.

fs/fs.go
dashpole: stdout = <int64>\n, so the Fields call was just to remove the newline character.

info/v2/conversion.go
dashpole: I think that is what actually happens.  &val.Filesystem[0].Inodes may be nil if we skipped collection.  In that case, it should also set InodeUsage to nil as well?

fs/fs.go
dashpole: I am very likely wrong, but I think the "rootfs" directory that it is pointed at is the container's modifiable directory, and shouldn't contain the base layers.  At least in most of my tests, the containers start out reporting 0 inodes used, which would indicate to me that they are not counting any shared resources in base layers.

dashpole: Ill go ahead and try it out.  It shouldnt hurt to make it faster, and it should still work.

timstclair: If that's the case, shouldn't it be zero for a new container? That's not what I'm seeing.

info/v2/conversion.go
timstclair: I don't think so. If you're taking the address of a variable, it will always be non-nil. The value might be zero, but if possible we should distinguish between 0 and not set (nil).

dashpole:
pushed a commit addressing all comments except device-mapper related ones.

fs/fs_test.go
dashpole: Confirmed, I changed it to numFiles +1 == inodes

info/v2/conversion.go
dashpole: Wouldn't it actually be more accurate to set InodesUsed to 0 in that case?  We only skip collection for some storage drivers if they do not use inodes, which means they use 0 of the inodes shared by the rest of the pods.

container/common/fsHandler.go
dashpole: See the note I made in the Issue.  I was able to test this, and the find command works correctly (and takes almost no time, since it always finds 0 inodes) with devicemapper.

fs/fs.go
dashpole: I tested this.  It only includes the writeable layer.

info/v2/conversion.go
dashpole: Just to document, since these storage drivers do not use inodes, I am leaving this the way it is, since using 0 inodes is equivalent to using no inodes.  Also, making a change here would require chaning the v1 api, or a large refactor of the code.

container/common/fsHandler.go
```diff
@@ -26,71 +27,89 @@ import (
...
-		rootfs:         rootfs,
-		extraDir:       extraDir,
-		fsInfo:         fsInfo,
-		stopChan:       make(chan struct{}, 1),
+		lastUpdate: time.Time{},
+		usage: FsUsage{
+			TotalUsageBytes: 0,
+			BaseUsageBytes:  0,
+			InodeUsage:      0,
+		},
```
timstclair: nit: no need to specify this (values default to 0)

container/common/fsHandler.go
```diff
@@ -26,71 +27,89 @@ import (
...
+	if rootDiskErr == nil && fh.rootfs != "" {
+		fh.usage.InodeUsage = inodeUsage
+	}
+	if rootInodeErr == nil && fh.rootfs != "" {
+		fh.usage.TotalUsageBytes = baseUsage + extraDirUsage
+	}
+	if extraDiskErr == nil && fh.extraDir != "" {
+		fh.usage.BaseUsageBytes = baseUsage
+	}
+	//combine errors into a single error to return
```
timstclair: nit: space after `//`, start comments with a Capital.

fs/fs.go
```diff
@@ -45,15 +45,24 @@ const (
 )
 
 // The maximum number of `du` tasks that can be running at once.
-const maxConsecutiveDus = 20
+const maxConcurrentDus = 20
+
+// The maximum number of `find` tasks that can be running at once.
+const maxConcurrentFinds = 20
```
timstclair: Do we need a separate pool for each of these? On the one hand, if one operation is significantly slower, it prevents it from hogging the tokens. On the other hand, the operations happen in sequence, so the one operation will be blocking anyway. I'm leaning towards a single "exec" pool.

fs/fs.go
```diff
@@ -447,28 +464,86 @@ func (self *RealFsInfo) GetDirUsage(dir string, timeout time.Duration) (uint64,
...
+	if dir == "" {
+		return 0, fmt.Errorf("invalid directory")
+	}
+	claimFindToken()
+	defer releaseFindToken()
+	findCmd := exec.Command("find", dir, "-xdev", "-printf", ".")
+	wcCmd := exec.Command("wc", "-c")
+	wcCmd.Stdin, _ = findCmd.StdoutPipe()
+
+	stdoutp, err := wcCmd.StdoutPipe()
```
timstclair: I think it would be better to assign a `bytes.Buffer` to stdout and stderr. I'm not positive, but I tihnk the process could hang if the pipe fills up, waiting for it to be drained.

fs/fs_test.go
```diff
@@ -100,11 +100,63 @@ func TestDirUsage(t *testing.T) {
 	fi, err := f.Stat()
 	as.NoError(err)
 	expectedSize := uint64(fi.Size())
-	size, err := fsInfo.GetDirUsage(dir, time.Minute)
+	size, err := fsInfo.GetDirDiskUsage(dir, time.Minute)
 	as.NoError(err)
 	as.True(expectedSize <= size, "expected dir size to be at-least %d; got size: %d", expectedSize, size)
 }
 
+//make sure that the timeout is actually being triggered (found this bug in PR#1489)
```
timstclair: nit: spaces & capitals, same below.

fs/fs_test.go
```diff
@@ -100,11 +100,63 @@ func TestDirUsage(t *testing.T) {
...
+func TestDirDiskUsageTimeout(t *testing.T) {
+	as := assert.New(t)
+	fsInfo, err := NewFsInfo(Context{})
+	as.NoError(err)
+	dir, err := ioutil.TempDir(os.TempDir(), "")
+	as.NoError(err)
+	defer os.RemoveAll(dir)
+	dataSize := 1024 * 10000 //1000 KB bigger to make sure it triggers tihe timeout
```
timstclair: s/tihe/the/

Did you mean 1,000 KB or 10,000 KB?

fs/fs_test.go
```diff
@@ -100,11 +100,63 @@ func TestDirUsage(t *testing.T) {
...
+	as := assert.New(t)
+	fsInfo, err := NewFsInfo(Context{})
+	as.NoError(err)
+	dir, err := ioutil.TempDir(os.TempDir(), "")
+	as.NoError(err)
+	defer os.RemoveAll(dir)
+	numFiles := 1000
+	for i := 0; i < numFiles; i++ {
+		_, err := ioutil.TempFile(dir, "")
+		as.NoError(err)
```
timstclair: I think you should use a require here, otherwise this could be a lot of error spam.

fs/fs_test.go
```diff
@@ -100,11 +100,63 @@ func TestDirUsage(t *testing.T) {
...
+	as := assert.New(t)
+	fsInfo, err := NewFsInfo(Context{})
+	as.NoError(err)
+	dir, err := ioutil.TempDir(os.TempDir(), "")
+	as.NoError(err)
+	defer os.RemoveAll(dir)
+	numFiles := 100000 //make sure we actually trigger the timeout
+	for i := 0; i < numFiles; i++ {
+		_, err := ioutil.TempFile(dir, "")
+		as.NoError(err)
```
timstclair: ditto

fs/fs_test.go
timstclair: looks like you missed the typo

timstclair:
LGTM once you fix the typo and squash the commits.

container/common/fsHandler.go
dashpole: done

fs/fs.go
dashpole: done

fs/fs.go
dashpole: done

fs/fs_test.go
dashpole: done

PR #1491, opened 2016-10-05
Add support for Docker Labels on ES Storage

rikatz:
This PR adds support for Docker Labels on Elasticsearch Storage.

This support is only available for ES 5 and later, due to restrictions on ES to '.' characters on field names.


rikatz:
I signed it!!

PR #1504, opened 2016-10-18
fixed go vet bug

dashpole:
This was probably due to a change in go vet.  It does not allow Mutex's to be copied.  Testify.Mock contains a Mutex, and cannot be copied.  Because of this, we cannot change a testify.Mock object after it is added to a list for all viewers of the list.  Since we cannot change objects after they are added to the list, I had to pass a boolean to the function that initially creates the list.  When true, the boolean causes GetSpec to return an error, rather than the normal return value, and cases Exists to return true.

### DIFF
--- a/manager/manager_test.go
+++ b/manager/manager_test.go
@@ -115,7 +115,7 @@ func expectManagerWithContainers(containers []string, query *info.ContainerInfoR
 			h.On("GetSpec").Return(
 				spec,
 				nil,
-			)
+			).Once()
 			handlerMap[h.Name] = h
 		},
 		t,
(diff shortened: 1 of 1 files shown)


spxtr:
@k8s-bot test this

manager/manager_test.go
```diff
@@ -206,22 +221,14 @@ func TestGetContainerInfoV2Failure(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("RemoveContainer failed: %v", err)
 	}
 
-	// Make GetSpec fail on /c2
-	mockErr := fmt.Errorf("intentional GetSpec failure")
-	failingHandler := containertest.NewMockContainerHandler(failing)
-	failingHandler.On("GetSpec").Return(info.ContainerSpec{}, mockErr)
-	failingHandler.On("Exists").Return(true)
-	*handlerMap[failing] = *failingHandler
```
timstclair: `Mock.On(...)` just appends a call to `Mock.ExpectedCalls` (https://github.com/stretchr/testify/blob/master/mock/mock.go#L202). It might be simpler to just clear out ExpectedCalls. Alternatively, you could just modify the call directly, something like:
```go
failingHandler.ExpectedCalls[0].Return(info.ContainerSpec{}, mockErr)
...
```

timstclair:
For some reason I can no longer reproduce this issue, and I'm curious whether you see it in a clean environment, e.g.
```
docker run --rm -w "/go/src/github.com/google/cadvisor" -v "${GOPATH}/src/github.com/google/cadvisor:/go/src/github.com/google/cadvisor" golang:1.6.3 make
```

I think the underlying issue is real, whether or not `go vet` detects it. So lets proceed with this change either way...

manager/manager_test.go
dashpole: The reason go vet was throwing a fit was that we cant do the last line "*handlerMap[failing] = *failingHandler".

PR #1507, opened 2016-10-19
added GetDirFsInfo, which returns filesystem information for the specified directory

dashpole:
This will allow us to, for example, monitor the filesystem that includes /var/lib/kubelet, by calling GetDirFsInfo("/var/lib/kubelet").

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -101,6 +101,9 @@ type Manager interface {
 	// Get version information about different components we depend on.
 	GetVersionInfo() (*info.VersionInfo, error)
 
+	// Get filesystem information for the filesystem that contains the given directory
+	GetDirFsInfo(dir string) (v2.FsInfo, error)
+
 	// Get filesystem information for a given label.
 	// Returns information for all global filesystems if label is empty.
 	GetFsInfo(label string) ([]v2.FsInfo, error)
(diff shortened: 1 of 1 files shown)


dashpole:
@vishh This is the cadvisor change we spoke about.  It allows cadvisor to be queried for filesystem information for the filesystem that includes a given directory.

vishh:
Overall this change LGTM

manager/manager.go
```diff
@@ -657,6 +660,27 @@ func (self *manager) getRequestedContainers(containerName string, options v2.Req
...
+func (self *manager) GetDirFsInfo(dir string) (v2.FsInfo, error) {
+	dirDevice, err := self.fsInfo.GetDirFsDevice(dir)
+	if err != nil {
+		return v2.FsInfo{}, fmt.Errorf("error trying to get filesystem Device for dir %v: err: %v", dir, err)
+	}
+	dirMountpoint, err := self.fsInfo.GetMountpointForDevice(dirDevice.Device)
+	if err != nil {
+		return v2.FsInfo{}, fmt.Errorf("error trying to get MountPoint for Root Device: %v, err: %v", dirDevice, err)
+	}
+	infos, err := self.GetFsInfo("")
```
timstclair: Rather than having this method call `GetFsInfo` (which iterates over all filesystems), move the inner logic from that method to a helper.

timstclair: Spoke offline, decided to stick with this.

Issue #1432, opened 2016-08-22
Release cAdvisor with go1.7


timstclair:
Investigate releasing cAdvisor v0.24 with go 1.7, which has lots of performance and binary size improvements.

PR #1508, opened 2016-10-19
Build & test cAdvisor with go 1.7.1

timstclair:
Fixes https://github.com/google/cadvisor/issues/1432

/cc @dashpole

### DIFF
--- a/build/jenkins_e2e.sh
+++ b/build/jenkins_e2e.sh
@@ -31,8 +31,11 @@ if [[ ! -z "$(git diff --name-only pages)" ]]; then
   exit 1
 fi
 
-make
-go build -tags test github.com/google/cadvisor/integration/runner
+# Build & test with go 1.7
+docker run --rm \
+       -w "/go/src/github.com/google/cadvisor" \
+       -v "${GOPATH}/src/github.com/google/cadvisor:/go/src/github.com/google/cadvisor" \
+       golang:1.7.1 make all test-runner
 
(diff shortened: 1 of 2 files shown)


timstclair:
Oops, I guess 1.7.2 isn't out yet. Let's go with 1.7.1.

build/jenkins_e2e.sh
```diff
@@ -31,8 +31,11 @@ if [[ ! -z "$(git diff --name-only pages)" ]]; then
   exit 1
 fi
 
-make
-go build -tags test github.com/google/cadvisor/integration/runner
+# Build & test with go 1.7
+docker run --rm \
```
vishh: Wouldn't it better to make the local and jenkins build be the same?

timstclair: I'm not sure how to do that without cloning every make rule. Is that worth it? Locally I can control my go version, but with jenkins this is the recommended way.

timstclair:
Now that Kubernetes is requiring go1.7, can we merge this as is? We can update the local builds in a follow up.

build/jenkins_e2e.sh
vishh: One option is to detect if we are in a containerized environment by setting an environment variable. If we are inside a container, then run the make command directly. If not, then invoke make run_container with the actual make command as an argument.

vishh:
@timstclair SGTM

PR #1511, opened 2016-10-20
oomparser: Fix tests

euank:
They had reflect.Equal instead of !reflect.Equal.
They also depended on the current year. Which changed.

### DIFF
--- a/utils/oomparser/oomparser_test.go
+++ b/utils/oomparser/oomparser_test.go
@@ -18,10 +18,11 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	"reflect"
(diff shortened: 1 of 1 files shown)


timstclair:
From your fixes it looks like the test was pretty broken. How come it was passing though?

euank:
@timstclair it was passing because the assertion was `if reflect.DeepEqual(expected, actual) { error("they weren't equal") }`, a missing `!` in the conditional.

They weren't ever equal because of the two small fixes: the fact that `mem3` doesn't appear in the container sample file, but was expected, and the fact that the system error doesn't set victim container but was expected to.

PR #1530, opened 2016-11-10
set envs when converting to v2 spec

danielkrainas:
The `Envs` field was never being set in the v1->v2 container spec conversion.


dashpole:
@k8s-bot ok to test

dashpole:
Apologies for the delay, this looks good!  Please rebase and squash

danielkrainas:
@dashpole rebased.

PR #1531, opened 2016-11-10
Fix a bug in tail.

Random-Liu:
Found this bug when using the `utils/tail` package in node problem detector.

If the reader try to read right after creating `Tail`, `Read` will keep returning `0, nil`.
If it does this too many times, golang will report error [`var ErrNoProgress = errors.New("multiple Read calls return no data or error")`](https://golang.org/pkg/io/).

This PR initializes `readerErr` as `io.EOF`, so that the reader can work properly.

@AdoHe @timstclair

### DIFF
--- a/utils/tail/tail.go
+++ b/utils/tail/tail.go
@@ -45,6 +45,16 @@ const (
 
 // NewTail starts opens the given file and watches it for deletion/rotation
 func NewTail(filename string) (*Tail, error) {
+	t, err := newTail(filename)
+	if err != nil {
+		return nil, err
+	}
+	go t.watchLoop()
+	return t, nil
+}
+
+// newTail creates a Tail object.
+func newTail(filename string) (*Tail, error) {
 	t := &Tail{
 		filename: filename,
 	}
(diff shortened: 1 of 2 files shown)


AdoHe:
@Random-Liu That's why I use `time.Sleep` in node problem detector.  :)

Random-Liu:
@AdoHe Yeah, but even you add sleep:
1) In theory, it can not guarantee that Tail is initialized after sleep.
2) During log rotation, it could still happen.

utils/tail/tail.go
```diff
@@ -71,14 +74,14 @@ func (t *Tail) Read(p []byte) (int, error) {
 var _ io.Reader = &Tail{}
```
timstclair: nit: change to `io.ReadCloser`

Random-Liu: ACK~

Random-Liu:
> Is it possible to add a unit test for this failure case?

I can write a simple one. :)

Random-Liu:
@timstclair Addressed comments. :)

Random-Liu:
@timstclair Are you ok with this PR? :)

timstclair:
LGTM

Issue #1546, opened 2016-11-22
cadvisor do gather but not copy tcpstat of container.


fzu-huang:
When I try to get number of container's tcp CloseWait. I found that cadvisor did not put that into stat data:
`/info/v2/conversion.go`:
```
//func MachineStatsFromV1
if cont.Spec.HasNetwork {
			stat.Network = &NetworkStats{
				// FIXME: Use reflection instead.
				Tcp:        TcpStat(val.Network.Tcp),
				Tcp6:       TcpStat(val.Network.Tcp6),
				Interfaces: val.Network.Interfaces,
			}
		}

//func ContainerStatsFromV1
if spec.HasNetwork {
			// TODO: Handle TcpStats
			stat.Network = &NetworkStats{
				Interfaces: val.Network.Interfaces,
			}
		}
```
I am sure cadvisor will gather tcp stat (only if we set `--disable_metrics=""` ) Why don't cadvisor just copy the data into `v2.ContainerStats.Network.Tcp`  ?  If I set `--disable_metrics=""` , I still can't get any tcpstat when I using cadvisor API:`/api/v2.1/stats/{CID}`

timstclair:
Looks like an oversight. We'd welcome a PR :)

Issue #1523, opened 2016-11-01
partitionRegex can not fit usual device name


fzu-huang:
`var partitionRegex = regexp.MustCompile(`^(?:(?:s|xv)d[a-z]+\d*|dm-\d+)$`)`

```
for scanner.Scan() {
		line := scanner.Text()
		words := strings.Fields(line)
		if !partitionRegex.MatchString(words[2]) {
			continue
		}
                ...
}
```
We can see that cadvisor use this regex rule :`^(?:(?:s|xv)d[a-z]+\d*|dm-\d+)$` to match device name.but it can not match some device name such as `/dev/vda` ,`vda`,`/vda`.

PR #1547, opened 2016-11-23
Fixes partion name, v2 TCP stats

fzu-huang:
1. Fix https://github.com/google/cadvisor/issues/1523 - partitionRegex can not fit usual device name
2. Fix https://github.com/google/cadvisor/issues/1546 - cadvisor do gather but not copy tcpstat of container
3. Handle restarting a Manager instance. This may be necessary if a new disk device is mounted after the cAdvisor manager is started.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -346,7 +346,7 @@ func (self *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, er
 	return filesystems, nil
 }
 
-var partitionRegex = regexp.MustCompile(`^(?:(?:s|xv)d[a-z]+\d*|dm-\d+)$`)
+var partitionRegex = regexp.MustCompile(`^(?:(?:s|v|xv)d[a-z]+\d*|dm-\d+)$`)
 
 func getDiskStatsMap(diskStatsFile string) (map[string]DiskStats, error) {
 	diskStatsMap := make(map[string]DiskStats)
--- a/info/v2/conversion.go
+++ b/info/v2/conversion.go
@@ -119,6 +119,8 @@ func ContainerStatsFromV1(spec *v1.ContainerSpec, stats []*v1.ContainerStats) []
 		if spec.HasNetwork {
 			// TODO: Handle TcpStats
 			stat.Network = &NetworkStats{
+				Tcp:        TcpStat(val.Network.Tcp),
+				Tcp6:       TcpStat(val.Network.Tcp6),
 				Interfaces: val.Network.Interfaces,
 			}
 		}


fzu-huang:
@k8s-bot ok to test

fzu-huang:
@googlebot I signed it!

manager/manager.go
```diff
@@ -321,6 +327,8 @@ func (self *manager) Stop() error {
 		}
 	}
 	self.quitChannels = make([]chan error, 0, 2)
+
+
```
timstclair: nit: remove these extra lines

fs/fs.go
```diff
@@ -346,7 +346,8 @@ func (self *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, er
 	return filesystems, nil
 }
 
-var partitionRegex = regexp.MustCompile(`^(?:(?:s|xv)d[a-z]+\d*|dm-\d+)$`)
+//update by fzu-huang.Update the filter to match disk such as 'vda'.
```
timstclair: Please remove "update by fzu-huang" everywhere. It is recorded in the git history.

manager/manager.go
```diff
@@ -237,6 +237,12 @@ type manager struct {
 
 // Start the container manager.
 func (self *manager) Start() error {
+	//add-by fzu-huang clean the factories which is a external object.
+	//Usually we just run only one manager.So every time we start a manager
+	//we have to clear old handlerfactories, otherwise, new handlerfactories
+	//will be in behand of old factories. and won't be called.
+	container.ClearContainerHandlerFactories()
```
timstclair: If there is another manager running, this will break it. I think it's better to just return an error in this case, something like `"cannot start manager: container factories already registered"`, and leave it up to the caller to call `container.ClearContainerHandlerFactories()` if this was really intentional.

Ideally we would just move away from the global registry.

fzu-huang: I read the code roughly and I found that we init  fs data by `fs/fs.go: NewFsInfo()`.And we put this fsInfo into a new factory when we called manager.Start(). If we run two manager with a global factories array, the handlers of new manager's factory won't be called
```
// Register a ContainerHandlerFactory. These should be registered from least general to most general
// as they will be asked in order whether they can handle a particular container.
func RegisterContainerHandlerFactory(factory ContainerHandlerFactory, watchTypes []watcher.ContainerWatchSource) {
	factoriesLock.Lock()
	defer factoriesLock.Unlock()

	for _, watchType := range watchTypes {
		factories[watchType] = append(factories[watchType], factory)
	}
}
```

```
```
I think that

timstclair:
@k8s-bot ok to test

timstclair:
Please squash your commits, and make sure the email address matches your github account.

PR #1553, opened 2016-11-30
ZFS Filesystem stats support

bakins:
This is a quick and dirty patch to get some ZFS filesystem stats. Without this, I get errors in getFsStats as rootfsStorageDir is not set in container/docker/handler.go 

This is just the bare minimum for usage stats.  I edited the vendored go-zfs to get the referenced field from zfs to get the full container disk usage.  I may be misunderstanding the meaning of that field in `info.FsStats`, however.  If this field - or others - are needed, I can do a separate PR to go-zf. (I worked on the initial implementation of go-zfs, FWIW.)

With ZFS, depending on your dataset, querying the underlying device is not actually useful.  I'm running on a raidz of 3-5 physical drives.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -33,6 +33,7 @@ import (
 	docker "github.com/docker/engine-api/client"
 	dockercontainer "github.com/docker/engine-api/types/container"
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


bakins:
Closing in favor of #1555

PR #1555, opened 2016-12-12
Add watcher for zfs similar to devicemapper

bakins:
After testing, this seems to give the most consistent view of ZFS compared to the other providers. It is a bit more code than I would like, but it is modeled after the devicemapper code.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/cadvisor/machine"
 	"github.com/google/cadvisor/manager/watcher"
 	dockerutil "github.com/google/cadvisor/utils/docker"
+	"github.com/google/cadvisor/zfs"
 
 	docker "github.com/docker/engine-api/client"
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


l8nite:
Would be nice to see some movement on this one.

dashpole:
@k8s-bot ok to test

dashpole:
Apologies for being slow.  The code changes look good, but dont have the context to review this.  @derekwaynecarr, if you could give this a quick look, that would be appreciated.

dashpole:
@bakins, this will eventually need to be squashed and rebased.

bakins:
@dashpole I'll rebase and squash. Happy to answer any questions, as well.  This is based on the devicemapper code in cadvisor but does zfs rather than thin_ls to get disk/filesystem stats.

dashpole:
I've read up on Devicemapper and ZFS, and I think I mostly understand this PR.
How does this differ from the current behavior for ZFS?  Do we currently not support it at all?  Or does it just not work?

bakins:
Squashed.  

BTW - we've been running this for a while by patching Kubernetes.

PR #1561, opened 2016-12-21
mention another prometheus example using Grafana

bamarni:
The GUI part of the blog post example seems not much relevant anymore, there are some nice starter projects from @vegasbrianc that could be interesting to mention here.


bamarni:
I signed it!

timstclair:
Thanks!

PR #1564, opened 2016-12-26
add error log for open log file

andyxning:
Add error log when `tail` can not open a log file. This is useful for debug. :)

PR #1558, opened 2016-12-18
Remove pipe to wc and prevent zombie on err

sykesm:
`GetDirInodeUsage` requires extended functionality provided by the GNU `find` implementation to work correctly. When deployed on a system with the BusyBox `find` implementation, the extra arguments are unrecognized and find terminates. (#1556, kubernetes/minikube#923, kubernetes/kubernetes#38894)

When `find` exits with an error, the wait for the `wc` process never happens and `wc` becomes a zombie. On systems without GNU find, these zombies can accrue quickly.

In lieu of a solution that would work everywhere (ie. not relying on GNU find), this PR simply replaces `wc` with a writer implementation that counts what has been written. This simplifies the code a bit and removes the pipe.

PR #1567, opened 2017-01-04
updated assets to 2017 using make assets

dashpole:
This was causing build failures on git ( #1564, #1558).  We may want to consider fixing this for the future.


timstclair:
Still failing the check. When I rebuild locally it only changes the year, so I'm not sure why `_pagesAssetsHtmlContainersHtml` changed for you.

PR #1568, opened 2017-01-05
Lower verbosity level for spamming log

dashpole:
Addresses [Kubernetes #22586](https://github.com/kubernetes/kubernetes/issues/22586).  The log line in question is designed to alert users that containers cannot span multiple filesystem devices.  However, it is being triggered when gathering information for the "/" (root) filesystem device, which includes all devices on the image.  Images configured to use multiple filesystem devices (coreos, GCI, to name a few) spam the logs with the line: "conversion.go:128 failed to handle multiple devices for container. Skipping Filesystem stats"

For this particular device ("/"), kubernetes does not consume the filesystem stats, making this log message unimportant.  Lowering the verbosity level is appropriate in this case.

cc: @derekwaynecarr (I messaged you about this in december) @timstclair


dashpole:
Now ignores root container, and includes container name in the log message, which will be useful for debugging.

Issue #1572, opened 2017-01-13
cAdvisor is confused by /system.slice/var-lib-docker-containers-...-shm.mount cgroups and may report zero-valued stats for Docker containers


SpComb:
When using cAdvisor to monitor Docker container stats, cAdvisor seems to get confused by any cgroups that happen to contain an existing Docker container ID in their basename. This includes the systemd mount unit for the `/var/lib/docker/containers/*/shm` mountpoint, which seems to randomly result in cAdvisor returning incorrect (all-zero) stats from the `/api/v1.2/docker/...` endpoint.

## Symptoms

The `/api/v1.2/docker` endpoint returns a mixture of  `{"/docker/*: ..., "/system.slice/var-lib-docker-containers-*-shm.mount" ...}` entries. The `/api/v1.2/docker/*` endpoint may occasionally return zero-valued CPU and memory stats.

## Details
With CoreOS 1235.6.0 + systemd 231 + Docker 1.12.3 + Linux 4.7.3-coreos-r2, the cAdvisor Docker driver seems to pick up two cgroups for each running Docker container:

```
I0113 12:22:45.621838       1 factory.go:111] Using factory "docker" for container "/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount"
I0113 12:22:45.626448       1 manager.go:874] Added container: "/system.slice/var-lib-docker-containers-acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536-shm.mount" (aliases: [null-stackless-whoami-1 acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536], namespace: "docker")

I0113 12:22:45.755197       1 factory.go:111] Using factory "docker" for container "/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536"
I0113 12:22:45.756418       1 manager.go:874] Added container: "/docker/acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536" (aliases: [null-stackless-whoami-1 acf76f8a0cf47638f7ab7c7e033872672479017f7447f90d2d6d6d5c39bf7536], namespace: "docker")
```

[...]

Issue #1510, opened 2016-10-20
Cadivsor tries to monitor systemd .mount units via docker


hanikesn:
In my logfiles I see a lot of errors like the following about every minute:

```
Oct 20 11:45:08 ip-10-128-32-121 dockerd[12654]: time="2016-10-20T11:45:08.958795388Z" level=error msg="Handler for GET /containers/012467134e116e692a6218b7534b78b826f0f6ef3dc2d99c99956e43f66277ac/json returned error: No such container: 012467134e116e692a6218b7534b78b826f0f6ef3dc2d99c99956e43f66277ac"
Oct 20 11:45:08 ip-10-128-32-121 dockerd[12654]: time="2016-10-20T11:45:08.959105963Z" level=error msg="Handler for GET /containers/dbf71d333d65340ea108f9dc8778a9e06c70173ba799c6fb38da5885d853bf89/json returned error: No such container: dbf71d333d65340ea108f9dc8778a9e06c70173ba799c6fb38da5885d853bf89"
```

It turns out that these uuids belong to .mount units from systemd:

```
   CGroup: /
           └─system.slice
             ├─var-lib-docker-overlay2-012467134e116e692a6218b7534b78b826f0f6ef3dc2d99c99956e43f66277ac-merged.mount
             ├─var-lib-docker-overlay2-dbf71d333d65340ea108f9dc8778a9e06c70173ba799c6fb38da5885d853bf89-merged.mount
```

It turns out that these queries originate from the cadvisor plugin in kublet.

I'd like to monitor the error rate of API calls to the docker deamon and this makes it impossible and I think cadivsor should be smart enough to know about `.mount` units and monitor them appropriately.

rhuddleston:
Same issue here:

`Nov  1 20:55:44 prom1 dockerd[25267]: time="2016-11-01T20:55:44.822064076Z" level=error msg="Handler for GET /containers/bdd3ca44b528d9fd5704fe5781f5f375d63e5103e0c7c81daa4377bfbb1d68d1/json returned error: No such container: bdd3ca44b528d9fd5704fe5781f5f375d63e5103e0c7c81daa4377bfbb1d68d1"
[...]

PR #1573, opened 2017-01-13
Ignore any .mount cgroup in docker

derekwaynecarr:
Fixes https://github.com/google/cadvisor/issues/1572
Fixes https://github.com/google/cadvisor/issues/1510

/cc @timstclair @ncdc @smarterclayton

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -141,17 +141,21 @@ func ContainerNameToDockerId(name string) string {
 	return id
 }
 
+// isContainerName returns true if the cgroup with associated name
+// corresponds to a docker container.
 func isContainerName(name string) bool {
+	// always ignore .mount cgroup even if associated with docker and delegate to systemd
+	if strings.HasSuffix(name, ".mount") {
+		return false
+	}
 	return dockerCgroupRegexp.MatchString(path.Base(name))
 }
 
 // Docker handles all containers under /docker
 func (self *dockerFactory) CanHandleAndAccept(name string) (bool, bool, error) {
-	// docker factory accepts all containers it can handle.
-	canAccept := true
-
+	// if the container is not associated with docker, we can't handle it or accept it.
 	if !isContainerName(name) {
-		return false, canAccept, fmt.Errorf("invalid container name")
+		return false, false, nil
 	}
 
 	// Check if the container is known to docker and it is active.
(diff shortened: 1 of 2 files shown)


container/docker/factory_test.go
```diff
@@ -49,3 +49,24 @@ func TestEnsureThinLsKernelVersion(t *testing.T) {
...
+			expected: false,
+		},
+		{
+			name:     "/system.slice/docker-72e5a5ff5eef3c4222a6551b992b9360a99122f77d2229783f0ee0946dfd800e.scope",
+			expected: true,
+		},
+	}
+	for _, test := range tests {
+		if actual := isContainerName(test.name); actual != test.expected {
+			t.Fatalf("%s: expected: %v, actual: %v", test.name, test.expected, actual)
```
yujuhong: minor nit: use `t.Errorf` to go through all test cases.

container/docker/factory.go
```diff
@@ -142,6 +142,10 @@ func ContainerNameToDockerId(name string) string {
 }
 
 func isContainerName(name string) bool {
+	// always ignore .mount cgroup even if associated with docker and delegate to systemd
```
yujuhong: cadvisor will log `Error trying to work out if we can handle containerName: error`, which seems misleading. How about changing this so that `CanHandleAndAccept` return `(true, false, nil)` or (false, false, nil)` instead?

I also think the caller of `CanHandleAndAccept` logs way too many errors periodically...but that's another issue.

derekwaynecarr:
@yujuhong -- updated pr so for .mount cgroup, docker will not handle or accept.  in addition, just updated the handler to no longer return an error when a container is not a valid docker name since it caused excessive noise.

at log-level 5, you now see:

```
I0116 11:04:28.508348   29276 factory.go:115] Factory "docker" was unable to handle container "/system.slice/var-lib-docker-overlay-9f086b233ab7c786bf8b40b164680b658a8f00e94323868e288d6ce20bc92193-merged.mount"
I0116 11:04:28.508359   29276 factory.go:108] Factory "systemd" can handle container "/system.slice/var-lib-docker-overlay-9f086b233ab7c786bf8b40b164680b658a8f00e94323868e288d6ce20bc92193-merged.mount", but ignoring.
I0116 11:04:28.508373   29276 manager.go:867] ignoring container "/system.slice/var-lib-docker-overlay-9f086b233ab7c786bf8b40b164680b658a8f00e94323868e288d6ce20bc92193-merged.mount"
```

this should be good to go, PTAL

derekwaynecarr:
@k8s-bot test this

ncdc:
LGTM

derekwaynecarr:
merging...

rata:
@derekwaynecarr just curious (I'm hitting this), will this get to a 1.5.x or 1.6 kubernetes release? (as far as I know, cadvisor is included in the kubelet, so there is no easy way to fix this)

dashpole:
This is in 1.5.6

rata:
@dashpole thanks!

PR #1577, opened 2017-01-18
WIP: cut log spam associated with multiple-fs containers

pmorie:
Attempt at cutting log spam associated with containers with multiple filesystems; posting early to see if it sticks.  It looks at this point like I would have to keep track of the containers in multiple places, since both the http handler and manager are plugged into this code path.

cc @derekwaynecarr

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -457,11 +457,16 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
 
 type version2_1 struct {
 	baseVersion *version2_0
+
+	// loggedContainers is a map that holds which containers a multiple-
+	// devices message has been issued for.
+	loggedContainers map[string]bool
 }
 
 func newVersion2_1(v *version2_0) *version2_1 {
 	return &version2_1{
-		baseVersion: v,
+		baseVersion:      v,
+		loggedContainers: map[string]bool{},
 	}
 }
 
@@ -507,10 +512,15 @@ func (self *version2_1) HandleRequest(requestType string, request []string, m ma
 				// Root cgroup stats should be exposed as machine stats
 				continue
 			}
+
+			firstCall := self.loggedContainers[name]
+
 			contStats[name] = v2.ContainerInfo{
 				Spec:  v2.ContainerSpecFromV1(&cont.Spec, cont.Aliases, cont.Namespace),
-				Stats: v2.ContainerStatsFromV1(name, &cont.Spec, cont.Stats),
+				Stats: v2.ContainerStatsFromV1(name, &cont.Spec, cont.Stats, firstCall),
 			}
+
+			self.loggedContainers[name] = true
 		}
 		return writeResult(contStats, w)
 	default:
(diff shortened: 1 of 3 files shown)


info/v2/conversion.go
```diff
@@ -96,7 +96,7 @@ func MachineStatsFromV1(cont *v1.ContainerInfo) []MachineStats {
 	return stats
 }
 
-func ContainerStatsFromV1(containerName string, spec *v1.ContainerSpec, stats []*v1.ContainerStats) []*ContainerStats {
+func ContainerStatsFromV1(containerName string, spec *v1.ContainerSpec, stats []*v1.ContainerStats, firstCall bool) []*ContainerStats {
```
derekwaynecarr: rather than `firstCall`, maybe just say `log bool` on whether it should log anything or not?  there are two log calls in this existing operation

derekwaynecarr: or an alternative is to update this to return `[]*ContainerStats, []warnings`, and then we can optionally log warnings or not from the caller?

api/versions.go
```diff
@@ -457,11 +457,16 @@ func (self *version2_0) HandleRequest(requestType string, request []string, m ma
...
+	// loggedContainers is a map that holds which containers a multiple-
+	// devices message has been issued for.
+	loggedContainers map[string]bool
 }
 
 func newVersion2_1(v *version2_0) *version2_1 {
 	return &version2_1{
-		baseVersion: v,
+		baseVersion:      v,
+		loggedContainers: map[string]bool{},
```
derekwaynecarr: could this be a lru cache so we don't grow forever?

info/v2/conversion.go
pmorie: I'm not really thrilled about doing any of those, let me think about it.  In general I wish this was something log frameworks were a little smarter about.

api/versions.go
pmorie: totes

derekwaynecarr:
poke

dashpole:
Just curious, are users actually creating containers with multiple filesystem partitions?  When I debugged this months ago, all of the spammy error messages were coming from the "/" partition, which #1568 addresses.

timstclair:
I think at least volumes will show up as a separate device

dashpole:
@k8s-bot test this

dashpole:
I will cut the cadvisor release tomorrow morning if there is still interest in getting this in 1.7

abessifi:
Hi, is there a workaround about this "issue" ?

abessifi:
@derekwaynecarr @pmorie as this fix seems to still en progress, could you please point out a workaround to avoid getting these messages in the kubernetes/openshift logs ?
OpenShift: v1.5
Docker: 1.12.6
CentOS: 7.4

dashpole:
/retest
Seems like this is still something we want

lilic:
Was this fixed already in another PR?

dashpole:
This seems like something we probably still want, but i'm going to close as it probably needs to be re-implemented, or at least rebased.

PR #1583, opened 2017-01-30
Dynamic thin_ls refresh interval

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1405347

thin_ls is currently set at a fixed 15s refresh interval.  For some users, this generates unacceptable levels of I/O load.

thin_ls is disk cache sensitive.  If the cache is lost, due to memory pressure most likely, rebuilding it can take considerable I/O.  This can result in a thrashing situation where thin_ls is running almost continuously.

This PR
- Increases the minimum refresh interval to 1 minute (4x longer)
- Dynamically adjusts the interval so that thin_ls runs for no longer than 10% of the interval

The side effect of longer refresh intervals is slower disk eviction reaction time.  So we want to keep the interval small for nodes that don't have an issue with thin_ls run durations, but also want to avoid continuous I/O for nodes that do have long thin_ls run durations (likely due to pressure on the disk cache).

@ncdc @derekwaynecarr @eparis @pmorie

### DIFF
--- a/devicemapper/thin_pool_watcher.go
+++ b/devicemapper/thin_pool_watcher.go
@@ -22,6 +22,8 @@ import (
 	"github.com/golang/glog"
 )
 
+const minimumPeriod time.Duration = time.Minute
+
 // ThinPoolWatcher maintains a cache of device name -> usage stats for a
 // devicemapper thin-pool using thin_ls.
 type ThinPoolWatcher struct {
@@ -47,7 +49,7 @@ func NewThinPoolWatcher(poolName, metadataDevice string) (*ThinPoolWatcher, erro
 		metadataDevice: metadataDevice,
 		lock:           &sync.RWMutex{},
 		cache:          make(map[string]uint64),
-		period:         15 * time.Second,
+		period:         minimumPeriod,
 		stopChan:       make(chan struct{}),
 		dmsetup:        NewDmsetupClient(),
 		thinLsClient:   thinLsClient,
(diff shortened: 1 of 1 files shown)


ncdc:
FYI the reason thin_ls is so io intensive in this particular case is because we want to determine the exclusive bytes per thin LV. To do that, thin_ls has to read the entire metadata for the thin pool. So if the metadata is 300MB, each invocation of thin_ls has to read 300MB to be able to determine exclusive bytes per thin LV.

derekwaynecarr:
The dynamic period is an improvement.  I would like to verify that this also fixes unintended consequences of thin_ls impacting cpu usage stat collection prior to merging.

derekwaynecarr:
@k8s-bot test this

derekwaynecarr:
per our discussion, once we have some measurement times, we may do a follow-on that requires opt-in to this functionality via an env var of some kind.

derekwaynecarr:
in at least one environment on one cloud provider, we have seen thin_ls never return so period alone is not a sufficient fix for everyone.  for the environments where execution has proven slow, or iops heavy, the period helps.  we need to potentially get a follow-on pr that disables the usage.

derekwaynecarr:
To handle the "command never returns case", what about the following:

* add a RunUntil(period) command in exec.go
* if command fails to complete in time, log a message, and do not call again for life of process

i guess in the case of iops exhaustion, some folks may complain, but i think for now that seems like a relatively safe solution rather than an env var, etc.

thoughts? /cc @vishh @timstclair

vishh:
Ideally, we would run thin_ls in a container :) That is very difficult in the case of k8s. 
We might want to define a new `RunCommand(cmd, timeout, cpu, memory)` interface where we setup sub-containers (cgroups) to run commands. We can kill commands that do not complete by timeout. 

We can such an interface for many binary execs we do across cadvisor and kubelet.

sjenning:
decided to disable thin_ls instead https://github.com/google/cadvisor/pull/1588

PR #1586, opened 2017-02-01
do not look in rootfs for thin_ls binary

sjenning:
xref
https://github.com/openshift/origin/issues/10940
https://github.com/openshift/origin/pull/12753

We should not look in `/rootfs` for the `thin_ls` binary as it is not statically linked and the dependent libraries might not be present in the container.

@derekwaynecarr @ncdc @pmorie @lukegb


ncdc:
@sjenning do we need to add thin-provisioning-tools (I think that's the name?) to https://github.com/google/cadvisor/blob/master/deploy/Dockerfile?

sjenning:
@ncdc yes

PR #1588, opened 2017-02-06
disable thin_ls due to excessive iops

derekwaynecarr:
It has been found in production environments that the usage of thin_ls can result in a large number of iops operations.  In iops restricted environments, this can have an impact on end-user workloads.


derekwaynecarr:
In the future, we need to do any of the following:

1. control thin_ls via blkio
1. increase the period between thin_ls invocations (https://github.com/google/cadvisor/pull/1583)
1. control the execution time of thin_ls via an upper bound
1. possibly only invoke thin_ls based on rate of change reported by lvs

/cc @ncdc @sjenning @smarterclayton

derekwaynecarr:
cc @pmorie

PR #1590, opened 2017-02-08
check docker container environment variables for empty strings

cvlc:
If an environment variable is an empty string, cadvisor panics and fails to start. this change enables cadvisor to start regardless of the content of environment variables.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -252,9 +252,11 @@ func newDockerContainerHandler(
 	// split env vars to get metadata map.
 	for _, exposedEnv := range metadataEnvs {
 		for _, envVar := range ctnr.Config.Env {
-			splits := strings.SplitN(envVar, "=", 2)
-			if splits[0] == exposedEnv {
-				handler.envs[strings.ToLower(exposedEnv)] = splits[1]
+			if envVar != "" {
+				splits := strings.SplitN(envVar, "=", 2)
+				if splits[0] == exposedEnv {
+					handler.envs[strings.ToLower(exposedEnv)] = splits[1]
+				}
 			}
 		}
 	}
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -254,7 +254,7 @@ func newDockerContainerHandler(
 		for _, envVar := range ctnr.Config.Env {
 			if envVar != "" {
 				splits := strings.SplitN(envVar, "=", 2)
-				if splits[0] == exposedEnv {
+				if len(splits) == 2 && splits[0] == exposedEnv {
 					handler.envs[strings.ToLower(exposedEnv)] = splits[1]
 				}
 			}


container/docker/handler.go
```diff
@@ -252,9 +252,11 @@ func newDockerContainerHandler(
 	// split env vars to get metadata map.
 	for _, exposedEnv := range metadataEnvs {
 		for _, envVar := range ctnr.Config.Env {
-			splits := strings.SplitN(envVar, "=", 2)
-			if splits[0] == exposedEnv {
-				handler.envs[strings.ToLower(exposedEnv)] = splits[1]
+			if envVar != "" {
+				splits := strings.SplitN(envVar, "=", 2)
+				if splits[0] == exposedEnv {
```
timstclair: We should check that `len(splits) == 2` too.

timstclair:
@k8s-bot ok to test

timstclair:
Thanks!

container/docker/handler.go
cvlc: I initially thought that too, but I tried removing it and didn't see any issues when passing values with multiple '='. It seems that SplitN only splits on the first occurrence of the delimiter, so omitting the ```len(splits) == 2```provides additional flexibility in that we can have equal signs in the environment variable values if we want to. I couldn't say whether that's useful or not for the particular reason behind this snippet?

timstclair: What if there is no equal sign?

cvlc: Touché! If there's no equal sign, ```splits``` will be a single element list so the comparison with splits[1] would cause the application to fail. An ```len(splits) == 2``` is the only reasonable case, as my previous assumption that ```len(splits)``` could ever be more than 2 is wrong considering that the split happens only on the first equals. I'll update accordingly.

container/docker/handler.go
```diff
@@ -252,9 +252,11 @@ func newDockerContainerHandler(
 	// split env vars to get metadata map.
 	for _, exposedEnv := range metadataEnvs {
 		for _, envVar := range ctnr.Config.Env {
-			splits := strings.SplitN(envVar, "=", 2)
-			if splits[0] == exposedEnv {
-				handler.envs[strings.ToLower(exposedEnv)] = splits[1]
+			if envVar != "" {
+				splits := strings.SplitN(envVar, "=", 2)
+				if len(splits) == 2 && splits[0] == exposedEnv {
```
timstclair: What is the right behavior when `len(splits) == 1`?

1. Fail silently (current approach)
2. Succeed (`handler.envs[strings.ToLower(exposedEnv)] = ""`)
3. Log a warning

I think 2 or 3 would probably be a better approach, but I'm not sure when this could happne, so present your case. If this "should never happen", we should just log a warning.

cvlc: Okay, let's work from the assumption that we'll want to match the behaviour of Docker (with POSIX specs as a fallback since cAdvisor seems to focus on Linux only):

1. If a variable is passed without any equals sign, according to docker/docker#10141 this should be treated as the container inheriting the client environment's value for that variable if such a value exists, otherwise it should not set any value.
2. If a variable is passed with an equals sign but no value, this should set the value of the variable to ```""``` (blank string) as in standard shell convention.
3. If a variable is passed with one or more equals signs, the first equals sign is used to demarcate the first part as key and the second as value. [Escape/null characters and equals signs are not permitted in the variable name as per the POSIX specifications](http://pubs.opengroup.org/onlinepubs/9699919799/).

Assuming that we want to avoid fatal errors and warn of invalid values so the user can correct them in their configuration we should:

1. _Silently ignore_ in the case of a blank variable name (```envVar == ""```) 
i. This is a totally nil element, probably created by a buggy client as most of the Docker clients I've used wouldn't allow such a thing)
ii. I suggest to ignore it since cAdvisor should not be expected to report the existence of a nil element that doesn't affect anything during runtime and shouldn't be responsible for potential client-side bugs that are not in the standard docker clients.
2. _Silently ignore_ ```len(splits) == 1```
i. This is expected behaviour when inheriting the environment from the client when the client has not set any value. In this case, we do not need to do anything as the container's environment will not be affected (the environment variable is neither explicitly nil nor set to a value).
ii. I am not 100% sure of this one but the difficulty is in differentiating between an explicit blank-string variable and an implicit 'my client doesn't know about this, so it has not passed it' variable. In the current case, the behaviour we'll observe is that an explicit blank string will be stored by cAdvisor but a lack of client-side setting won't - otherwise, we'd need some sort of placeholder value or the ability to store a variable without a value but that's more than I'm up for writing right now.
3. Permit variable values to contain as many equal signs as users want to pass, but only demarcate the variable key from value using the first ```=``` regardless of any other characters. I don't think there's any need for other behaviours.

I'm interested in thoughts, especially on 2.

timstclair: Sounds good. Thanks for the thorough response!

PR #1603, opened 2017-02-27
Test cAdvisor with the latest go patch release

timstclair:


dashpole:
Just for my knowledge for future reference: Does not specifying the minor release (e.g. 1.7 vs 1.7.1) cause it to default to the latest release (if, for example, 1.7.2 is out)?

timstclair:
That's the convention, but it's up to the image maintainers to push the updated image to both tags. I checked the golang image on dockerhub, and it looks like 1.7 follows the latest patch version: https://hub.docker.com/_/golang/

PR #1611, opened 2017-03-12
change stdout storage namespace to db name

andyxning:
In statsd storage, we use `ArgDbName` as the namespace. 
In redis storage, we use `ArgDbName` as the redis key. 
I think we need to keep this to stdout storage, too. :)

/ping @timstclair


timstclair:
This is a breaking change, not something to casually change. I don't like that those flags are overloaded for different storage plugins, but fixing it would require a larger overhaul.

PR #1612, opened 2017-03-12
Fix statsd error log

andyxning:
Error message should be logged with `glog.Errorf` instead of `glog.Infof`.

/ping @timstclair

### DIFF
--- a/storage/statsd/client/client.go
+++ b/storage/statsd/client/client.go
@@ -49,8 +49,7 @@ func (self *Client) Send(namespace, containerName, key string, value uint64) err
 	formatted := fmt.Sprintf("%s.%s.%s:%d|g", namespace, containerName, key, value)
 	_, err := fmt.Fprintf(self.conn, formatted)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


storage/statsd/client/client.go
```diff
@@ -49,7 +49,7 @@ func (self *Client) Send(namespace, containerName, key string, value uint64) err
 	formatted := fmt.Sprintf("%s.%s.%s:%d|g", namespace, containerName, key, value)
 	_, err := fmt.Fprintf(self.conn, formatted)
 	if err != nil {
-		glog.V(3).Infof("failed to send data %q: %v", formatted, err)
+		glog.Errorf("failed to send data %q: %v", formatted, err)
```
timstclair: Actually, if errors are returned they should not be logged. Combine this with the next line instead:
```go
return fmt.Errorf("failed to send data %q: %v", formatted, err)
```

andyxning: Will do.

andyxning: @timstclair PTAL.

PR #1613, opened 2017-03-12
fix gce error logging

andyxning:
Error message should be logged with `glog.Errorf` instead of `glog.Infof`.

/ping @timstclair

### DIFF
--- a/utils/cloudinfo/gce.go
+++ b/utils/cloudinfo/gce.go
@@ -32,7 +32,7 @@ const (
 func onGCE() bool {
 	data, err := ioutil.ReadFile(gceProductName)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


timstclair:
This error is expected when running on a non-GCE host. If you would like to add a comment to that effect I would accept that change.

andyxning:
@timstclair 
> This error is expected when running on a non-GCE host.

The error can still be recorded.

> If you would like to add a comment to that effect I would accept that change.

Could you please make it more clear, i can not understand it.  :)

timstclair:
> The error can still be recorded.

It is being recorded, but at the info level. Since this is not an actual error from cAdvisor's perspective, I want to keep it as an Info log.

> Could you please make it more clear

I meant that you could add a comment about why it's expected. Not required.

PR #1619, opened 2017-03-17
Remove out of date support section of running doc

timstclair:
I don't think this section was accurate, and I also don't think that filing an issue is the best way to get support.

/cc @vishh


vishh:
lgtm

PR #1627, opened 2017-03-29
A more strict regexp for docker cgroups

edevil:
[regexp](https://github.com/google/cadvisor/blob/master/container/docker/factory.go#L50) for inspecting a cgroup to see if it's a docker container is too lax. It also matches a lot of other cgroups, namely those created by systemd.

See [here](https://github.com/kubernetes/kubernetes/issues/38979#issuecomment-268340625) and [here](https://github.com/kubernetes/kubernetes/issues/38979#issuecomment-281585690).

@euank suggests anchoring the regexp at the end ```([a-z0-9]{64}$)```


dashpole:
@k8s-bot ok to test

euank:
Looks like there's a test for when docker's using the systemd driver, in which case my claim isn't correct.

Maybe there's a way to determine whether docker's using the default cgroup driver or the systemd one?

supereagle:
@edevil  @euank Any processs about this PR?

edevil:
At this moment I don't know how to solve the problem and I don't have much time to dedicate to this...

PR #1638, opened 2017-04-12
add customize capacity support for cpu&&memory

andyxning:
This PR fix [Kubernetes #44328](https://github.com/kubernetes/kubernetes/issues/44328). 

This PR will add `customize_cpu_capacity` and `customize_memory_capacity` command line flags. This will be useful when we want to customize the actual cpu or memory capacity when doing some test. For example, i just want to make a node with 40cores(hyperthread) registered with 80cores in order to stress the cpu to test its performance.

/ping @timstclair

### DIFF
--- a/machine/info.go
+++ b/machine/info.go
@@ -31,8 +31,14 @@ import (
 	"github.com/golang/glog"
 )
 
+const (
+	defaultIntegerValue = -1
+)
+
 var machineIdFilePath = flag.String("machine_id_file", "/etc/machine-id,/var/lib/dbus/machine-id", "Comma-separated list of files to check for machine-id. Use the first one that exists.")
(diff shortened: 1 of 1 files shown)


dashpole:
Keep in mind that Kubernetes vendors cAdvisor as a library, and calls right into the codebase.  I am not sure how you would be able to use this functionality unless cAdvisor is running in standalone mode.

andyxning:
@dashpole 
> Keep in mind that Kubernetes vendors cAdvisor as a library, and calls right into the codebase. I am not sure how you would be able to use this functionality unless cAdvisor is running in standalone mode.

I have only test this with standalone cadvisor. I will manually update cadvisor vendor and test this with kubelet. However, i think it will work properly just like what has been added to the kubelet comamdn line options from cadvisor `machine_id_file ` and `boot_id_file `

dashpole:
closing via https://github.com/kubernetes/kubernetes/issues/44328#issuecomment-293733002

PR #1639, opened 2017-04-12
Reduce cAdvisor log spam with multiple devices

derekwaynecarr:
This can cause excessive spam in the logs that drowns out useful messages.


ncdc:
👍

Issue #1647, opened 2017-05-04
Docker RestartCount Stats


rikatz:
Docker provides a RestartCount stats.

This stats is used to monitor how many times a container was restarted, and this count increase may represent some kind of problem (like an OOM kill).

It would be great if cAdvisor collects this metric also, so we can put this on Prometheus and alert on some unexpected problem.

Thanks!

dashpole:
How does docker expose this value?  We currently only monitor by using cgroups, so this may be a large change if we need to use docker's api.

rikatz:
Hi @dashpole.

Actually it exposes through API and not through cgroups. I was trying to search were in the code cAdvisor gets its data.

I don't think it's limited to cgroups, because of the 'Labels' and Image retrieval, but I couldn't figure out how to add also the restartCount metrics :)

Something probably specified here: https://github.com/google/cadvisor/blob/d7a44cb1a2c66e1688ccdc5d09e56069eecb659a/info/v1/container.go#L128

Thanks!

rikatz:
@dashpole FIY, made a PR here using DockerHandler and the API :)

Hope this is right!

Thanks!

rikatz:
This was closed in PR #1649

sirlatrom:
Any chance of a follow-up where this would be a separate metric, instead of or in addition to being a metric label?

PR #1649, opened 2017-05-08
Gather information from RestartCount counter

rikatz:
This PR adds a new metric into containers: the restart_count.

This metric can be used in Docker to check how many times a container configured with '--restart=always' has been restarted.

This means some problem like OOM or an application problem (segfault, as an example) and can be used later, with Prometheus to alert about this problem.

This PR uses the Docker API, but can also be extended to other future container engine (or even raw systemd stats, when available).

This is related to issue #1647


dashpole:
@k8s-bot ok to test

container/docker/handler.go
```diff
@@ -248,6 +251,7 @@ func newDockerContainerHandler(
 	handler.image = ctnr.Config.Image
 	handler.networkMode = ctnr.HostConfig.NetworkMode
 	handler.deviceID = ctnr.GraphDriver.Data["DeviceId"]
+	handler.restartCount = ctnr.RestartCount
```
dashpole: If I read this correctly, you are only updating handler.restartCount when a new docker container handler.  Have you verified that this increases when the container restarts?

info/v2/container.go
```diff
@@ -146,6 +146,8 @@ type ContainerStats struct {
 	Filesystem *FilesystemStats `json:"filesystem,omitempty"`
 	// Task load statistics
 	Load *v1.LoadStats `json:"load_stats,omitempty"`
+	// RestartCount contains how many times a container was restarted automatically
+	RestartCount int `json:"restartcount,omitempty"`
```
dashpole: do you convert restartcount from v1 to v2 anywhere?

metrics/prometheus.go
```diff
@@ -135,6 +135,14 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
 					return metricValues{{value: float64(time.Now().Unix())}}
 				},
 			}, {
+				name:      "container_restart_count",
+				help:      "Counter of times a container has been restarted",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					//return metricValues{{value: float64(11111)}}
```
dashpole: remove

container/docker/handler.go
rikatz: @dashpole I've tested this and works fine. The counter is increases each time a container restarts :) This is only a handler of 'how to gather this information', and it's used [here](https://github.com/estaleiro/cadvisor/blob/e42a357488c3b1cb3411b4647d61e73096230bba/container/docker/handler.go#L472)

info/v2/container.go
rikatz: Is that needed? Didn't know about that. Will check and post a new commit wheter this is necessary.

metrics/prometheus.go
rikatz: The idea here is to make available the times a countainer has been restarted to Prometheus. This is why I've created this metric (and also this PR), so we can check and alert here when a container have a big restart count.

Is this really necessary to be removed?

rikatz:
@dashpole Made the comments for you. Only have to check the v2 API.

About Kubernetes engine, it uses an internal counter and you can export those metrics to prometheus using kube-state-metrics.

About rkt, its API  doesn't support a restart count (as in CoreOS, as an example this is controlled through systemd) and maybe this can be removed.

This is also why this 'metric' is handled only by the Docker Handler, and shows a value of '0' in other systems (like raw and RKT).

Maybe using some conditional in Prometheus metrics would make it less noisy, like if the restart count equals to 0 there is no need to export/show it anyway.

Thanks for your time and your review :)

metrics/prometheus.go
dashpole: just meant the extra comment

info/v2/container.go
dashpole: [This is where we convert.  Should be simple to implement](https://github.com/google/cadvisor/blob/master/info/v2/conversion.go).

container/docker/handler.go
dashpole: Yes, I was more making sure that we create a new dockerContainerHandler each time the container is restarted by docker.  Kubernetes doesn't use docker's restart policies, so I am not familiar with how we handle them.

dashpole:
I am on board with exposing this as a prometheus metric.  I agree this is useful for users monitoring pure docker deployments.  Not exporting if it is 0 would also be nice.

My main concern is that this adds a field to the ContainerStats struct, which is exposed via the api/vX.X/containers endpoint.  For users of other runtimes, or using kubernetes, this additional metric would not only not be useful, but would be misleading, since they would expect this number to equal the kubernetes container restart count.  I would prefer at least renaming this to DockerRestartCount to differentiate it from the kubernetes one.  Generally, we don't explicitly add runtime-specific fields like this to the API.

The ContainerStats struct is currently only being used for general resource usage metrics: CPU, memory, disk, network, and diskio.  I don't think restart count fits into this model.  We don't really have a good way to express a runtime-specific orchestration metric like this.

@timstclair any thoughts on the best way to do this?

timstclair:
I agree with David. I think this metric is more runtime specific, which is something we're trying to move away from. This metric makes more sense if cAdvisor is a general container monitoring solution. It makes less sense if we think of it as a resource monitor. I'm not sure we've concretely defined the scope of cAdvisor anywhere.

dashpole:
@k8s-bot ok to test

rikatz:
@dashpole @timstclair OK, understood :)

I though that creating this metric would make easier to monitor some container problems, but I did understand what's the main focus of cAdvisor now :)

No problems, I can even create a standalone golang program to expose only this, or maybe using metricbeat or zabbix to expose this info also would be fine :)

Anyway, I'm going to make the suggested changes (between today and tomorrow), and if You have any suggestion in how (and if we should) to achieve this in cAdvisor, I'm here available for :)

Thanks for your time guys.

container/docker/handler.go
rikatz: Yes, actually the new handler defines how the info is going to be gathered by GetStats function. So I assume this point of review is solved :)

metrics/prometheus.go
rikatz: My bad, removed this :)

rikatz:
@dashpole @timstclair ok, kubernetes exposes this as a label:

``container_cpu_system_seconds_total{[...]container_label_io_kubernetes_container_restartCount="0"[...]} 1.29``

So, if I change the handler from Kubernetes to create this as a label (inside GetStats or something like that), but only for Docker, would this be fine?

The same happens in API v1, but v2 doesn't exposes the Labels, so it have to be checked later as something to be migrated, regardless of being a restart count or not.

This is not the perfect solution for my case (as this is a Label and not a Counter for Prometheus), but as Kubernetes also does this way, I think this is the way to be followed.

May I change some parts of the code to accomplish this way?

Thanks!

rikatz:
@dashpole The changes I've made here insert the restartcount as a label instead of creating a new stats.

This mades easier to implement (also, I don't need to change the API code, and this will be available in APIv2 as someone ports also the labels handler).

Please take a look. I don't think cAdvisor's main feature is to transform some internal Docker Counters into Labels, but I do agree this is probably the best way we can do this instead of changing the stats core.

If, anyway this is not ok, then I'm going to write something else (or maybe fork cAdvisor), as I need this info to Prometheus to check if there is a container in a crash loop or something else (Out of memory, etc), to trigger an alarm to me.

Thanks anyway guys :smile:

container/docker/handler.go
```diff
@@ -31,6 +31,8 @@ import (
 	dockerutil "github.com/google/cadvisor/utils/docker"
 	"github.com/google/cadvisor/zfs"
 
+	"strconv"
```
dashpole: nit: we usually place all of the golang imports (e.g. fmt, time, strconv) together at the top.

dashpole:
@timstclair this looks acceptable to me.  It would never show up in kubernetes, since we don't use docker's restart feature.
WDYT?

container/docker/handler.go
rikatz: Yeah, this is vscode fault. Corrected this :)

rikatz:
@dashpole Anything about this? Thanks :D

rikatz:
@dashpole friendly ping :)

dashpole:
Yeah, sorry.  I plan to merge this once we get our test-infra stuff figured out...  Just hang tight.

dashpole:
would you mind squashing?

rikatz:
@dashpole Sure, will do this.

rikatz:
Did as asked :) Anything else you need I'm here available to help! And thanks for your time :D

rikatz:
Thanks :)

Issue #1436, opened 2016-08-23
Display GPU Info on cAdvisor


jianzhangbjz:
Hi Guys,

I refer to the docs [cAdvisor Remote REST API](https://github.com/google/cadvisor/blob/master/docs/api.md#machine-information), and also read related code [machine.go](https://github.com/google/cadvisor/blob/master/info/v1/machine.go), but didn't find any API about GPU in `MachineInfo` struct. So, if we want to display the GPU info on cAdvisor, how can we do to implement that?

vishh:
Extend machine.go to include GPU related information. PRs are welcome!

andyxning:
@jianzhangbjz @vishh any progress on this? Or any material about GPU. I think we need more info.

jianzhangbjz:
@andyxning I'm so sorry, I have been busy with other work this time. If you want to get this feature ASAP, I think you can start to do this. :)

andyxning:
@jianzhangbjz I have no idea about GPU now. Just be willing to give a try. :)

jianzhangbjz:
@andyxning :+1: Any problems about GPU, please feel free to contact me if you need. :)

cmluciano:
@jianzhangbjz What specific metrics are you looking for? Is this just a count?

pineking:
I think it is just a count,  #GPUs are used by each pods, #GPUs available/total on each nodes, do you have plan to implement this? @cmluciano

liyubobj:
@cmluciano @pineking  No, we want to expose GPU live metrics, such as GPU core and GPU memory utility, to cAdvisor, as what they did for CPU and memory today.

andyxning:
@liyubobj 

> GPU core and GPU memory utility

Could you please explain `GPU memory utility`. IIUC, we should only expose GPU core utility. :)

liyubobj:
@andyxning Sorry, actually I want to say idle/used GPU memory.

cmluciano:
Yes I can submit a PR @pineking . Would this require the use of a nvidia specific tool?

liyubobj:
@cmluciano  Yes, you need either `libnvidia-ml.so` lib or `nvidia-smi` binary to get live metrics. Both of them are in nvidia GPU driver. But if you want to get GPU count only, a hack is to check `/dev/nvidiaX`.

vishh:
I'd prefer avoiding additional executable dependencies to cAdvisor if
[...]

PR #1651, opened 2017-05-11
add devices to supportedSubsystem

WIZARD-CXY:
This pr is fixing https://github.com/google/cadvisor/issues/1652

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -79,6 +79,7 @@ var supportedSubsystems map[string]struct{} = map[string]struct{}{
 	"memory":  {},
 	"cpuset":  {},
 	"blkio":   {},
+	"devices": {},
 }
 
 // Get cgroup and networking stats of the specified container


WIZARD-CXY:
@dashpole PTAL

dashpole:
@k8s-bot ok to test

WIZARD-CXY:
@dashpole PTAL

dashpole:
@WIZARD-CXY, this change looks fine, but I would prefer that we coordinate on the solution we add for monitoring GPUs #1436.
I would like GPU-related PRs to wait until we have come to agreement on how we plan to move forward.

WIZARD-CXY:
@dashpole  I think gpu-related pr and this one are actually different things, this pr is intended to fullfill the functionality of ListProcesses as it is. After all, a seemingly good function doesn't work like it claim to be is very confusing for the user IMHO.

WIZARD-CXY:
I think this function is very basic and could be use elsewhere other than gpu, I think we don't want user use this function and says "ah, it does not work, why???"

dashpole:
@WIZARD-CXY what is the impact for end users?  Where does this information surface?  Sorry, I dont have a ton of bandwidth for looking into this...

WIZARD-CXY:
@dashpole the issue #1652 gives a detail of this

dashpole:
> After all, a seemingly good function doesn't work like it claim to be is very confusing for the user IMHO.

WIZARD-CXY:
"Can you shed some more light on how this impacts users?"

As the issue said, the user can't use ListProcesses method of dockerContainerHandler. When I want to get a container's pid, ListProcesses method always return empty slice. After some digging,I found it uses fsCgroup to get the pids as the code below said https://github.com/google/cadvisor/blob/master/vendor/github.com/opencontainers/runc/libcontainer/cgroups/fs/apply_raw.go#L234. It uses devices subsytem to get pid. But in cadvisor the devices subsytem is not supported https://github.com/google/cadvisor/blob/master/container/libcontainer/helpers.go#L76. So ListProcesses will never work. Adding devices substem works fine, tested ok on my machine. 

This function is never used elsewhere in cadvisor so it explains why nobody has noticed it would never work. Other place like api uses other method (exec a ps command and analyze its output as result) as func (c *containerData) GetProcessList(cadvisorContainer string, inHostNamespace bool) ([]v2.ProcessInfo, error) defined.

WIZARD-CXY:
@dashpole I understand your concern, my intension is just to make the function work, if it is an intented omission now then I will withdraw this pr

dashpole:
Ok, I have had time to dig into this a little bit more, and I would prefer leaving it the way it is, unless we have a more substantial reason to change it.
`GetCgroupSubsystems` is used by the docker, rkt, and raw handlers to get cgroup paths, as well as the raw container watcher.
We currently detect cgroups that belong to containers [using a regex](https://github.com/google/cadvisor/blob/master/container/docker/factory.go#L160), and we have run into issues in the past with some cgroups being mistaken for containers.
I am surprised that ListProcesses exists, if it isnt used.  It looks like it wasnt used even when it was first added...

Issue #1164, opened 2016-03-15
could shortened container id be used in API request ?


stushurik:
In the api docs examples, shortened ids are used.
But when I tried this http://localhost:8080/api/v2.0/stats/a56e78bb4b8e?type=docker  the response is 
unable to find Docker container "a56e78bb4b8e"

But when I use the full id it works as expected. 
The same is for api v.1.3.
Docker version is 1.9.1, cadvisor version is 0.22

I have copy-pasted docker command from the README.

vishh:
It should be possible to look up using partial IDs. Feel free to post a PR
to fix this..

PR #1655, opened 2017-05-18
Allow finding docker containers by short name

Christopher-Bui:
Allow docker containers to be found by a short prefix name to match
the behavior of the docker daemon. This change now matches the examples
on the API docs.

Return an error if the given short, prefix name of a container is not
unique.

---

Resolves https://github.com/google/cadvisor/issues/1164
 
I'm not sure about all the contexts that `DockerContainer` is being called in and am slightly concerned about iterating through all the containers if an exact match is not found.

I'm pretty new to golang so I welcome all feedback.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -566,9 +566,24 @@ func (self *manager) getDockerContainer(containerName string) (*containerData, e
 		Namespace: docker.DockerNamespace,
 		Name:      containerName,
 	}]
+
+	// Look for container by short prefix name if no exact match found.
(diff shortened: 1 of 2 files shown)


Christopher-Bui:
I signed it!

Christopher-Bui:
Do I have to squash the merge commit?

dashpole:
yeah, that would be nice

dashpole:
@k8s-bot ok to test

Issue #1244, opened 2016-04-26
Runtime options documentation out of date


timstclair:
A lot of flags are missing from the [Runtime Options](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md) documentation. This document needs a review pass, and the missing flags should be added. For storage-specific options, we should just refer to the separate storage documents.

PR #1657, opened 2017-05-18
Document runtime options

Christopher-Bui:
Many flags are missing from docs/runtime_options.md. Add all missing
flags that aren't storage driver specific.

Resolves https://github.com/google/cadvisor/issues/1244


dashpole:
@k8s-bot ok to test

PR #1667, opened 2017-06-06
Support tmpfs in processMounts

dims:
Helps with the situation in kubernetes where /var/lib/kubelet is
mounted using tmpfs.


dims:
See https://github.com/kubernetes/kubernetes/issues/44059 for a bit more context

dims:
@dashpole - tested by hand - http://paste.openstack.org/show/611784/ as i don't have access to that centos based environment mentioned in that kubernetes issue

Issue #1670, opened 2017-06-07
Support `docker exec` to collect custom metrics


Colstuwjx:
Hi team,

As the cadvisor is missing some metrics inner container, such as `tcp connections`, `process stats` etc.
It would be better if cadvisor support custom metrics via `docker exec` each container.

Thanks.

dashpole:
tcp metrics are disabled by default.  You can configure this using the `--disable-metrics` flag documented [here](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md#metrics).  We also have already support for [custom application metrics](https://github.com/google/cadvisor/blob/master/docs/application_metrics.md)

Colstuwjx:
Although cadvisor support custom application metrics, app need to expose an endpoint which contain these metrics, but e.g. tcp connections, cpu load, process stats, these are basic system level metrics, and should have a pretty way to achieve, rather than binding into application's code logical.

dashpole:
tcp connections are already supported, cpu load is already supported.  You can see the [api that we expose for containers](https://github.com/google/cadvisor/blob/master/info/v1/container.go#L519).   The current design of cAdvisor is to monitor containers, not processes.

Colstuwjx:
As #1674 merged, I'd like to close this, due to each metrics is included in prom `/metrics` endpoint.

PR #1673, opened 2017-06-08
Add cpu load and tcp/udp usage statistics in prom `/metrics` endpoint

Colstuwjx:
Hi folks,

Related to #1670 , I'd like to take a PR to expose more about `container_cpu_load_average_10s`, `container_network_tcp_usage_total`, `container_network_udp_usage_total` metric within the `/metrics` endpoint.

Currently, tcp/udp usage will be disabled via `disable_metrics` flag by default, thus `/metrics` would show zero values. Due to [ContainerSpec](https://github.com/google/cadvisor/blob/master/info/v2/container.go#L67) has no `ignoreMetrics` flag, I don't know how to handle this...

Thanks.

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -418,6 +418,22 @@ type TcpStat struct {
 	Closing uint64
 }
 
+func (t *TcpStat) TcpStatMap() map[string]uint64 {
+	return map[string]uint64{
+		"tcp_established": t.Established,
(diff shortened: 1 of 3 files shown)


Colstuwjx:
Sorry, since the author config is wrong and cannot take a sign, I'd like to recreate one.

PR #1674, opened 2017-06-08
Add cpu load,tcp\udp usage statistics in prom `/metrics` endpoint.

Colstuwjx:
Hi folks,

Related to #1670  , I'd like to take a PR to expose more about `container_cpu_load_average_10s`, `container_network_tcp_usage_total`, `container_network_udp_usage_total` metric within the `/metrics` endpoint.

Currently, tcp/udp usage will be disabled via `disable_metrics` flag by default, thus `/metrics` would show zero values. Due to [ContainerSpec](https://github.com/google/cadvisor/blob/master/info/v2/container.go#L67) has no `ignoreMetrics` flag, I don't know how to handle this...

Thanks.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -190,6 +190,13 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
 					return metricValues{{value: float64(s.Cpu.CFS.ThrottledTime) / float64(time.Second)}}
 				},
 			}, {
+				name:      "container_cpu_load_average_10s",
+				help:      "Value of container cpu load average over the last 10 seconds.",
+				valueType: prometheus.GaugeValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.LoadAverage)}}
+				},
+			}, {
 				name:      "container_memory_cache",
 				help:      "Number of bytes of page cache memory.",
 				valueType: prometheus.GaugeValue,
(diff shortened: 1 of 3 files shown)


Colstuwjx:
Ping @dashpole  @jimmidyson

dashpole:
@k8s-bot ok to test

dashpole:
> Currently, tcp/udp usage will be disabled via disable_metrics flag by default, thus /metrics would show zero values. Due to ContainerSpec has no ignoreMetrics flag, I don't know how to handle this...

Sorry, why would you want the container spec to have ignoreMetrics?

Colstuwjx:
Yeah, I mean that due to no ignoreMetrics in ContainerSpec, I cannot determine whether the metric is disabled. In the code level, ignoreMetrics should not appeared in that, so, just expose zero values if they are disabled...

Colstuwjx:
@dashpole  Hi, I have made things clear, pls let me know if there is something need to change.

info/v2/container.go
```diff
@@ -262,6 +262,22 @@ type TcpStat struct {
 	Closing     uint64
 }
 
+func (t *TcpStat) TcpStatMap() map[string]uint64 {
```
dashpole: Please remove this.  It isnt used anywhere

info/v1/container.go
```diff
@@ -432,6 +448,15 @@ type UdpStat struct {
 	TxQueued uint64
 }
 
+func (u *UdpStat) UdpStatMap() map[string]uint64 {
```
dashpole: I would prefer not to add functions here.  Do you think you could just create the list within the prometheus function like `container_tasks_state`?  Same goes for the TcpStatMap function

dashpole:
looks good.  please squash your commits

Colstuwjx:
Done.

Colstuwjx:
I've already update to date, let me know if there is anything need to correct, @dashpole .

PR #1675, opened 2017-06-08
Fix race in ignoring non-existent directories

apilloud:
The `github.com/google/cadvisor/container/common` package has some helpers which race on checks for non-existent files. The race is as follows:

1. A call to stat is used to check for existence of the directory.
2. The directory is deleted by something else.
3. A call to open fails because the directory does not exist.

This change modifies the code to eliminate the call to stat and check the error code off the open operation. In practice, this bug causes startup failures in kuberneties kubelet:
```
Failed to start cAdvisor readdirent: no such file or directory
```


dashpole:
@k8s-bot ok to test

Issue #1631, opened 2017-04-04
Connecting cadvisor to TLS enabled docker host via tcp


zette:
dockerd: 17.03.1
cadvisor: 0.24.1
I'm running my docker engine host with management interface available only via tcp TLS enabled socket.

First thing: cadvisor does not use environment variables:
- DOCKER_HOST
- DOCKER_TLS_VERIFY
- DOCKER_CERT_PATH

Second thing: when running cadvisor with command:
`./cadvisor -docker tcp://docker_host:2376`
it starts, but shows only error message, when trying to get container details view:
`Get http://fmb-app03.amg.net.pl:2376/info: malformed HTTP response "\x15\x03\x01\x00\x02\x02".
* Are you trying to connect to a TLS-enabled daemon without TLS?`

It seems that while one can find proper mappings for various system variables DOCKER_*, in the source of client.go, client_unix.go, client_windows.go and client_darwin.go, these are nowhere used.
So setting the environment for TLS enabled (or even only tcp:// socket) docker endpoint via standard variables is not possible.
While unsecure connections are possible with runtime `-docker` parameter, there is no way to provide cadvisor with certificates, keys and other options needed to open TLS connection.

dashpole:
According to #1621, a number of our auth functions are broken right now... @timstclair will probably know more about your specific use case.

timstclair:
This isn't currently supported, but we would welcome a change to add support. Maybe there's an API in the docker client code that we can leverage?

zette:
According to docker docs, the TLS connection does not need any special API calls. You can use curl for call testing:
```
To use curl to make test API requests, you need to use three extra command line flags:

$ curl https://$HOST:2376/images/json \
  --cert ~/.docker/cert.pem \
  --key ~/.docker/key.pem \
  --cacert ~/.docker/ca.pem
```

Source: [https://docs.docker.com/engine/security/https/#daemon-modes](url)

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #1678, opened 2017-06-19
Add runtime options for TLS support

stelzner:
Addresses #1631 by adding runtime options for using TLS to connect to docker. The -docker-tls flag enables TLS, the other three (-docker-tls-cert, -docker-tls-key, -docker-tls-ca) specify the keypair and CAs to use. This solution matches the existing interface in that it uses runtime options (like -docker) over environment variables (like DOCKER_HOST).

If -docker-tls is not set, the behaviour is the same as before.
Let me know if you have any feedback.

### DIFF
--- a/container/docker/client.go
+++ b/container/docker/client.go
@@ -18,9 +18,11 @@
 package docker
 
 import (
+	"net/http"
 	"sync"
 
 	dclient "github.com/docker/engine-api/client"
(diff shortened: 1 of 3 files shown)


dashpole:
please rebase and squash

stelzner:
Done.

PR #1682, opened 2017-06-21
use glog.Warningf when a format string is passed

CaoShuFeng:
ref:
https://godoc.org/github.com/golang/glog#Warningf
https://godoc.org/github.com/golang/glog#Warning


CaoShuFeng:
@timstclair

timstclair:
@k8s-bot ok to test

PR #1694, opened 2017-07-18
add collecting threads func

jianzi123:


dashpole:
@jianzi123 can you let me know what the purpose of this PR is?  These functions are not called anywhere.

jianzi123:
@dashpole e...There is a similar method in the old version, which was not implemented at that time.

dashpole:
closing due to inactivity

PR #1700, opened 2017-07-21
Reduce log spam when unable to get network stats

derekwaynecarr:
I have multiple reports of this spamming logs and causing stability issues.

```
Unable to get network stats from pid 36618: couldn't read network stats: failure opening /proc/36618/net/dev: open /proc/36618/net/dev: no such file or directory
```

suspect the container process has exited but the cgroup still remained.

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -99,7 +99,7 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int, ignoreMetri
 	if !ignoreMetrics.Has(container.NetworkUsageMetrics) {
 		netStats, err := networkStatsFromProc(rootFs, pid)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
@dashpole - the customer reported this was putting their root partition at risk, but in theory the should be able to monitor logs separately.  i want to avoid similar reports in the future, so just knowing this is lgtm from you is fine.  i would like the version of cadvisor we have w/ kube 1.8 has this fix.

dashpole:
oh sorry, you need to rebase

derekwaynecarr:
rebasing...

derekwaynecarr:
@dashpole -- rebased, simple review.

fzu-huang:
@derekwaynecarr @dashpole Do you have any idea about why it happened? I saw so much log but can not find any process match the Pids.

PR #1705, opened 2017-07-26
Machine info introspects hugepages

derekwaynecarr:
I plan to add support for pre-allocated huge pages in Kubernetes 1.8.

In order to add that support, I need the ability to introspect the supported set of huge pages on the node, and the number of them that are pre-allocated.

This adds that support.

### DIFF
--- a/machine/info.go
+++ b/machine/info.go
@@ -17,8 +17,10 @@ package machine
 import (
 	"bytes"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -30,6 +32,8 @@ import (
 
 	"github.com/golang/glog"
 )
+
+const hugepagesDirectory = "/sys/kernel/mm/hugepages/"
 
 var machineIdFilePath = flag.String("machine_id_file", "/etc/machine-id,/var/lib/dbus/machine-id", "Comma-separated list of files to check for machine-id. Use the first one that exists.")
 var bootIdFilePath = flag.String("boot_id_file", "/proc/sys/kernel/random/boot_id", "Comma-separated list of files to check for boot-id. Use the first one that exists.")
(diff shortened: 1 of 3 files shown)


machine/info.go
```diff
@@ -48,6 +50,43 @@ func getInfoFromFiles(filePaths string) string {
...
+			return hugePagesInfo, err
+		}
+
+		numFile := "/sys/kernel/mm/hugepages/" + st.Name() + "/nr_hugepages"
+		val, err := ioutil.ReadFile(numFile)
+		if err != nil {
+			return hugePagesInfo, err
+		}
+		var numPages uint64
+		n, err := fmt.Sscanf(string(val), "%d", &numPages)
```
sjenning: any reason you use `fmt.Sscanf` here and `strconv.ParseUint` on 63?

machine/info.go
```diff
@@ -48,6 +50,43 @@ func getInfoFromFiles(filePaths string) string {
...
+		if err != nil || n != 1 {
+			return hugePagesInfo, fmt.Errorf("could not parse file %v contents %q", numFile, string(val))
+		}
+		if err != nil {
```
sjenning: this block can never be run; shorted by the previous block

machine/info.go
derekwaynecarr: ParseUint got tripped up by the \n in the file.

dashpole:
rebase, and this should pass

derekwaynecarr:
@dashpole -- thanks for updating the ci pieces... rebased, and it appears to pass.

machine/info.go
```diff
@@ -48,6 +50,40 @@ func getInfoFromFiles(filePaths string) string {
...
+		}
+
+		numFile := "/sys/kernel/mm/hugepages/" + st.Name() + "/nr_hugepages"
+		val, err := ioutil.ReadFile(numFile)
+		if err != nil {
+			return hugePagesInfo, err
+		}
+		var numPages uint64
+		n, err := fmt.Sscanf(string(val), "%d", &numPages)
+		if err != nil || n != 1 {
```
dashpole: can you add a comment explaining why we want to return error if n != 1?

machine/info.go
```diff
@@ -48,6 +50,40 @@ func getInfoFromFiles(filePaths string) string {
...
+	}
+	for _, st := range files {
+		nameArray := strings.Split(st.Name(), "-")
+		pageSizeArray := strings.Split(nameArray[1], "kB")
+		pageSize, err := strconv.ParseUint(string(pageSizeArray[0]), 10, 64)
+		if err != nil {
+			return hugePagesInfo, err
+		}
+
+		numFile := "/sys/kernel/mm/hugepages/" + st.Name() + "/nr_hugepages"
```
dashpole: nit: use a constant for the directory

derekwaynecarr:
@dashpole - updated.  i would like us to return a value for each supported page size on the machine.  so in your case, a value of 0 for 2048kb at least lets me know your host supports 2Mi pages.

dashpole:
thanks @derekwaynecarr.  I had misread the code: I thought it was checking the contents of the file, not the number of lines.
LGTM

PR #1544, opened 2016-11-20
oomparser: update to use kmsg based parser

euank:
My comment for PR https://github.com/kubernetes/node-problem-detector/pull/41#issue-189041129 applies here too.

This would fix https://github.com/kubernetes/kubernetes/issues/34965, probably fix #1484, and work-around/fix #645.

Things that should be considered before merging:

1) anywhere kubelet or cadvisor runs in a container now needs access to `/dev/kmsg`. This should be documented as a breaking change and various examples / docs might need changing.
2) this reduces a little how far back we can see (namely after the system reboots, we don't get things before that). This implies that a "kubelet crash -> container OOM -> node reboot" chain of events can lose that container OOM.
3) this works everywhere where the kernel is modern enough to run docker! Fewer dependencies!

Testing note: I removed the log-file formatted files. The parsing of log-files is now the responsibility of the kmsgparser package which lives elsewhere, so we can trust it to do the right thing. We also get to assume the timestamp etc.
I have not done manual testing of this change, but I have done decent manual testing of the kmsgparser.

PR #1706, opened 2017-07-27
oomparser: don't get stuck for certain processes

euank:
The oomparser logic would end up stuck, unable to detect the end of a
given oom trace, for any process witha a name that didn't match \w+.

This includes processes like "python3.4" due to the '.', or
'docker-containerd' due to the '-'.

This fix was included in pr #1544 last year, but since that PR seems
dead it seems like a good idea to break this more important fix out.

I've updated the test such that it would have caught this issue.


dashpole:
rebase, and this should pass

euank:
Rebased.

PR #1707, opened 2017-07-27
use sdjournal instead of journalctl for oom parsing

sjenning:
Addresses kube issue https://github.com/kubernetes/kubernetes/issues/34965

Alternative to #1544

Use `github.com/coreos/go-systemd/sdjournal` instead of exec'ing `journalctl -f -k` and leaking the process.

_NOTE_: Using sdjournal requires the systemd headers be installed on the compiling system.  For EL distros that is `yum install systemd-devel`.

@derekwaynecarr @eparis @euank @dchen1107 @vishh @adohe

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -169,6 +169,11 @@
 			"Rev": "4484981625c1a6a2ecb40a390fcb6a9bcfee76e3"
 		},
 		{
+			"ImportPath": "github.com/coreos/go-systemd/sdjournal",
+			"Comment": "v8-2-g4484981",
+			"Rev": "4484981625c1a6a2ecb40a390fcb6a9bcfee76e3"
+		},
+		{
 			"ImportPath": "github.com/coreos/go-systemd/util",
 			"Comment": "v8-2-g4484981",
 			"Rev": "4484981625c1a6a2ecb40a390fcb6a9bcfee76e3"
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -18,7 +18,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"os/exec"
 	"path"
 	"regexp"
 	"strconv"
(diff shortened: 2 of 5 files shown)


sjenning:
cc @jsafrane

derekwaynecarr:
our queue of cadvisor updates is growing, need to track down e2e problems.

dashpole:
@sjenning please rebase.  The infra issue should be fixed

sjenning:
@dashpole so it hit the new dependency on systemd-devel I mentioned in the PR description.

```
I0801 19:30:09.573] >> building cadvisor
W0801 19:30:11.299] # github.com/google/cadvisor/vendor/github.com/coreos/go-systemd/sdjournal
W0801 19:30:11.300] vendor/github.com/coreos/go-systemd/sdjournal/journal.go:27:33: fatal error: systemd/sd-journal.h: No such file or directory
W0801 19:30:11.300]  // #include <systemd/sd-journal.h>
W0801 19:30:11.300]                                  ^
```

Can you add the distro appropriate package to the builder?

euank:
If we're planning to switch to directly parsing kmsg instead, I don't see why this change is necessary.

sjenning:
@euank at the time I submitted this, that PR hadn't moved since mid-March. I did note that this was an alternative to https://github.com/google/cadvisor/pull/1544

Only one need be accepted, but one _does_ need to be accepted.

sjenning:
It looks like both of these approaches (sdjounal and reading /dev/kmsg) are used in NPD.
https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go

@dchen1107 just mentioned this as a way during a sig-node meeting, so I wrote it up.

sjenning:
closed in favor of https://github.com/google/cadvisor/pull/1729

PR #1710, opened 2017-08-01
update assets to fix cAdvisor builds

dashpole:
fixes https://github.com/kubernetes/test-infra/issues/3419


tallclair:
Hmm, any idea why https://github.com/google/cadvisor/blob/master/build/jenkins_e2e.sh#L25-L32 didn't catch this?

tallclair:
Also, what was the PR that triggered this? It looks like there haven't been any changes to the UI in the past 4 months?

ixdy:
my best guess is changing golang versions caused this (Jenkins went 1.6 -> 1.8).

Issue #1676, opened 2017-06-12
Incorrect container CPU Usage reported with 4.7 kernel


pgeorgiev333:
Original defect submitted in docker space but as per the comments i'm submitting it here:

https://github.com/moby/moby/issues/26591

Environment:

- Docker Engine 17.03.1-ce
- Kernel Version: 4.10.10-200.fc25.x86_64
- Operating System: Fedora 25 (Server Edition)
- cadvisor version: v0.25.0

Description:
caDvisor is started as a container using docker-compose:

```

cadvisor:
  restart: always
  image: google/cadvisor:v0.25.0
  container_name: cadvisor
  hostname: docker
  privileged: true
  environment:
  - TZ=Europe/Sofia
  volumes:
  - /cgroup:/cgroup:ro
  - /:/rootfs:ro
  - /var/run:/var/run:rw
  - /sys:/sys:ro
  - /var/lib/docker/:/var/lib/docker:ro
  command: -docker="tcp://<docker host IP>:2375"
  ports:
  - "7080:8080"
```

I use prometheus to gather the data. The host has 24 CPUS as per /proc/cpuinfo.

When i open the casvisor url with the metrics i see a lot more cpus with 0 usage (http://<docker host IP>:7080/metrics):

container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-tmpfiles-setup-dev.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-tmpfiles-setup.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-udev-trigger.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-udevd.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-update-utmp.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-user-sessions.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/system.slice/systemd-vconsole-setup.service"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/user.slice"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/user.slice/user-0.slice"} 0
container_cpu_usage_seconds_total{cpu="cpu47",id="/user.slice/user-0.slice/session-525520.scope"} 0
[...]

PR #1711, opened 2017-08-03
libcontainer: use real number of CPUs for usage

euank:
As of the 4.7 kernel, the cpustats field returned from libcontainer
contains values for every possible cpu (including nonexistent ones).
The extra values are all 0s.

If we assume that hotplug events won't happen, we can get a more
accurate cpu count by using runtime.NumCPU and then ignoring any values
beyond that.

This was reported as an issue against the CoreOS bug tracker (https://github.com/coreos/bugs/issues/2088).

I haven't tested this, but running it on AWS and looking for extra 0s before/after would apparently be a good way to see.

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -33,6 +33,11 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 )
 
+/*
+#include <unistd.h>
+*/
+import "C"
+
 type CgroupSubsystems struct {
 	// Cgroup subsystem mounts.
 	// e.g.: "/sys/fs/cgroup/cpu" -> ["cpu", "cpuacct"]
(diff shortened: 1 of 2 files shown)


dashpole:
@euank thanks for this.  I haven't had time to get around to this yet.  Issue: #1676
The only remaining item is to test to make sure this works...  I am not sure running on AWS is a possibility for me, but ill ask around and see if I can get my hands on a 4.7 kernel to test.

euank:
Thanks for that reference. I decided to go with the hotplug safe solution after checking how docker did it and seeing this: https://github.com/moby/moby/blob/8b1adf55c2af329a4334f21d9444d6a169000c81/daemon/stats/collector_unix.go#L73

container/libcontainer/helpers.go
```diff
@@ -433,15 +438,40 @@ func DiskStatsCopy(blkio_stats []cgroups.BlkioStatEntry) (stat []info.PerDiskSta
 	return DiskStatsCopy1(disk_stat)
 }
 
+func minUint(x, y uint32) uint32 {
```
rphillips: probably shoudl be minUint32

euank:
@dashpole You don't have to run on aws, you can easily use qemu wit h`-smp cpus=4,maxcpus=10` or any other virtualization platform that lets you twiddle those knobs.

I went ahead an manually tested in that environment.

Before this patch:

```sh
$ uname -a
Linux coreos_production_qemu-1520-0-0 4.13.0-rc7-coreos #1 SMP Tue Aug 29 23:42:52 UTC 2017 x86_64 Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz GenuineIntel GNU/Linux
$ curl -s http://localhost:8080/api/v1.0/containers/ | jq '.stats[0].cpu.usage.per_cpu_usage'
[
  8366979268,
  8594483963,
  7864786060,
  7013534122,
  0,
  0,
  0,
  0,
  0,
  0
]
```

After the patch

```sh
$ curl -s http://localhost:8080/api/v1.0/containers/ | jq '.stats[0].cpu.usage.per_cpu_usage'
[
  10150246524,
  10894847083,
  9936348997,
  9566062369
]
```

PR #1698, opened 2017-07-20
Added findutils at the Dockerfile

lborguetti:
Guys,

This PR fix this issue: https://github.com/google/cadvisor/issues/1641

Thanks!

PR #1712, opened 2017-08-03
findutils-1641

jharshman:
## Summary

In response to issue #1641 and PR #1698 moving too slowly.

## Changes:
add findutils package to `deploy/Dockerfile`


dashpole:
done via #1698

PR #1713, opened 2017-08-04
Cleanup cAdvisor error responses with proper headers & response codes

tallclair:


tallclair:
@dashpole friendly ping

PR #1714, opened 2017-08-08
Release v0.26。1

ZhiqinYang:
how can i get correct container_fs_writes_bytes_total in centos7! I can't  get it! but is correct in Ubuntu! I am so puzzled with it!


dashpole:
Looks like you messed up here.  Feel free to open an issue if you have questions, or would like to help us find a bug

PR #1722, opened 2017-08-17
Skip subcontainer update on v2 calls

sjenning:
Currently `GetInfo()` updates both the spec and the subcontainers.  The subcontainer walk is VERY expensive as it does a stat syscall on basically every directory and file in the cgroupfs tree via `ioutil.Readdir()`.  This uses lots of CPU directly and also allocates a TON of objects that have to be GCed later, consuming even more CPU.

AFAICT, the v2 api does not need to do this subcontainer update at all since the v2 ContainerInfo doesn't have a subcontainers field:
https://github.com/sjenning/cadvisor/blob/master/info/v1/container.go#L128-L139
https://github.com/sjenning/cadvisor/blob/master/info/v2/container.go#L59-L65

The code supports this conclusion as v2 go paths look like this
```
cinfo, err := container.GetInfo(false)
...
result.Spec = self.getV2Spec(cinfo)
```
The `Subcontainers` in the `containerInfo` is discarded.

This is a huge performance win.  In Kubernetes, `GetContainerInfoV2()` is called every 10s from the summaryProvider.  It is currently one of the top ~5 hot paths in steady-state kubelet.

@derekwaynecarr @smarterclayton @dashpole @vishh @tallclair

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -113,16 +113,18 @@ func (c *containerData) allowErrorLogging() bool {
 	return false
 }
 
-func (c *containerData) GetInfo() (*containerInfo, error) {
+func (c *containerData) GetInfo(shouldUpdateSubcontainers bool) (*containerInfo, error) {
 	// Get spec and subcontainers.
 	if time.Since(c.lastUpdatedTime) > 5*time.Second {
 		err := c.updateSpec()
 		if err != nil {
 			return nil, err
 		}
-		err = c.updateSubcontainers()
-		if err != nil {
-			return nil, err
+		if shouldUpdateSubcontainers {
+			err = c.updateSubcontainers()
+			if err != nil {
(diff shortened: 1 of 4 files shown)


smarterclayton:
Nice dig

sjenning:
cc @tallclair

sjenning:
That "Close and comment" button should be red! :confounded:

sjenning:
@dashpole @tallclair if I can get a review on this in short order,  I'll update my kube bump PR https://github.com/kubernetes/kubernetes/pull/50629 to include this change.

dashpole:
ill take a look today later today.

derekwaynecarr:
LGTM, will let @dashpole review before merging.

Awesome investigation.  Seeing the profile before and after is really helpful.

PR #1724, opened 2017-08-22
Add an API to get FsStats from filesystem UUID

yguo0905:
Ref: https://github.com/kubernetes/kubernetes/issues/46984

- Added `GetFsInfoByFsUUID(uuid string) (*v2.FsInfo, error)` in the manager.
- Added timestamp into `v2.FsInfo` so that we will know when the stats are sampled.

Tested manually:

```
$ docker run --volume=/:/rootfs:ro --volume=/var/run:/var/run:rw --volume=/sys:/sys:ro --volume=/var/lib/docker/:/var/lib/docker:ro --volume=/dev/disk:/dev/disk:ro --publish=8080:8080 --detach=true --name=cadvisor cadvisor:9e51697

$ curl http://127.0.0.1:8080/api/v2.0/storage?uuid=9eb3792d-3b9a-4e50-84e0-fb8e6119b9e1
{"timestamp":"2017-08-23T16:11:37.451208174Z","device":"/dev/dm-1","mountpoint":"/rootfs","capacity":58923675648,"available":35672920064,"usage":20234002432,"labels":[],"inodes":3662848,"inodes_free":3277799}

$ curl http://127.0.0.1:8080/api/v2.0/storage?uuid=123
cadvisor: no such device
```

/assign @dashpole 
/cc @yujuhong

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -83,6 +83,8 @@ type RealFsInfo struct {
 	mounts map[string]*mount.Info
 	// devicemapper client
 	dmsetup devicemapper.DmsetupClient
+	// fsUUIDToDeviceName is a map from the filesystem UUID to its device name.
+	fsUUIDToDeviceName map[string]string
 }
 
 type Context struct {
@@ -103,13 +105,19 @@ func NewFsInfo(context Context) (FsInfo, error) {
 		return nil, err
 	}
 
+	fsUUIDToDeviceName, err := getFsUUIDToDeviceNameMap()
+	if err != nil {
+		return nil, err
+	}
+
 	// Avoid devicemapper container mounts - these are tracked by the ThinPoolWatcher
 	excluded := []string{fmt.Sprintf("%s/devicemapper/mnt", context.Docker.Root)}
 	fsInfo := &RealFsInfo{
-		partitions: processMounts(mounts, excluded),
-		labels:     make(map[string]string, 0),
-		mounts:     make(map[string]*mount.Info, 0),
-		dmsetup:    devicemapper.NewDmsetupClient(),
+		partitions:         processMounts(mounts, excluded),
+		labels:             make(map[string]string, 0),
+		mounts:             make(map[string]*mount.Info, 0),
+		dmsetup:            devicemapper.NewDmsetupClient(),
+		fsUUIDToDeviceName: fsUUIDToDeviceName,
 	}
 
 	for _, mount := range mounts {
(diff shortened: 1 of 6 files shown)


manager/manager.go
```diff
@@ -676,24 +681,24 @@ func (self *manager) getRequestedContainers(containerName string, options v2.Req
...
 	if err != nil {
-		return v2.FsInfo{}, fmt.Errorf("error trying to get MountPoint for Root Device: %v, err: %v", dirDevice, err)
+		return nil, fmt.Errorf("failed to get device using filesystem UUID %q: %v", uuid, err)
 	}
-	infos, err := self.GetFsInfo("")
-	if err != nil {
-		return v2.FsInfo{}, err
+	if device == nil {
+		// No such device exists.
+		return nil, nil
```
dashpole: should we return an error here, instead of nil?

fs/fs.go
```diff
@@ -169,17 +216,39 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
...
 
+func (self *RealFsInfo) getFsUUIDFromDeviceName(deviceName string) string {
+	if uuid, found := self.deviceNameToFsUUID[deviceName]; found {
+		return uuid
+	}
+	return ""
+
+}
+
+func (self *RealFsInfo) getDeviceNameFromFsUUID(uuid string) string {
```
dashpole: I dont think this is neccessary.  If we are just going to check to see if device == "" in GetDeviceInfoByFsUUID() anyways, we might as well just do self.fsUUIDToDeviceName[uuid] there.

fs/fs.go
```diff
@@ -433,6 +505,18 @@ func minor(devNumber uint64) uint {
 	return uint((devNumber & 0xff) | ((devNumber >> 12) & 0xfff00))
 }
 
+func (self *RealFsInfo) GetDeviceInfoByFsUUID(uuid string) (*DeviceInfo, error) {
+	deviceName := self.getDeviceNameFromFsUUID(uuid)
+	if deviceName == "" {
+		return nil, nil
```
dashpole: Can we return an error here instead of nil?  This eliminates the need check if *DeviceInfo is nil after calling in manager.GetFsInfoByFsUUID

fs/fs.go
```diff
@@ -71,6 +71,7 @@ type partition struct {
 	minor      uint
 	fsType     string
 	blockSize  uint
+	uuid       string
```
dashpole: sorry if I am missing something, but where is this used?

dashpole:
IIRC from our earlier conversation, this will only work on newer kernels.  What is the behavior for older kernels?  What is the plan for maintaining the current behavior with older kernel versions?

manager/manager.go
yguo0905: This indicates that the filesystem with the UUID does not exist on local machine, which is different from an error on fetching the stats for a known UUID. I think kubelet would need this information -- print an info message at some verbosity when the UUID does not exist, but return or log an error if cadvisor fails to get the stats. WDYT?

fs/fs.go
yguo0905: Good catch. Removed the irreverent code. I originally used this to get the partition from uuid by iterating the partition slice, and later changed to use a separate `fsUUIDToDeviceName` map, but didn't clean this up..

fs/fs.go
yguo0905: Done. Removed the unnecessary function.

fs/fs.go
yguo0905: Done. Removed the unnecessary function.

yguo0905:
The directory (`/dev/disk/by-uuid`) which we use to get the uuid to device name mapping has been available since (at least) 2.26. It's the tool `lsblk` that's too old to print the UUID on my local machine. Also, if this directory does not exist, cadvisor will report `no filesystem with uuid xxx exists` for all uuid queries.

manager/manager.go
dashpole: I would prefer having an error type:
`ErrorNoSuchDevice = errors.New("No Such Device")`
Then we can check for that error in the kubelet if we want to.

yguo0905: Done. PTAL.

dashpole:
can you rebase and squash?  Ill merge after that

yguo0905:
Done. Thanks!

PR #1726, opened 2017-08-22
Bump runc

sjenning:
On the path to fixing https://github.com/openshift/origin/issues/14920

@derekwaynecarr


manager/manager.go
```diff
@@ -136,7 +136,7 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 	}
 
 	// Detect the container we are running on.
-	selfContainer, err := cgroups.GetThisCgroupDir("cpu")
+	selfContainer, err := cgroups.GetOwnCgroupPath("cpu")
```
derekwaynecarr: is this the same literal behavior just a different name?

derekwaynecarr: ok, went back in runc history and see its the same thing just a rename.

PR #1728, opened 2017-08-23
Expose total_rss when hierarchy is enabled

derekwaynecarr:
If hierarchy is enabled, we should return total_rss instead of rss when reporting container stats.

This is a follow-up to previously reported issue in kubernetes here:
https://github.com/kubernetes/kubernetes/pull/43399#issuecomment-287858599

/cc @sjenning @dashpole @vishh @smarterclayton


derekwaynecarr:
I need to test this in a kube context to make sure it works as expected for our summary api.

derekwaynecarr:
ok, i think this is fine.

PR #1731, opened 2017-08-24
Update cadvisor runc godep

mtaufen:
I also had to update anything that had a transitive dependency on Sirupsen/logrus, from before he changed his username to lowercase the first letter, because otherwise the compiler complained about mixed-case imports.

PR #1732, opened 2017-08-25
Update runc godeps

dashpole:
cc @mtaufen 
can you check and see if this accomplishes what you were trying to do in #1731


mtaufen:
Yes, this looks like what I was trying to do - essentially get runc to head so both cadvisor's and kubelet's runc dependency use the same integer types (runc waffled between int64 and uint64 for an API that could accept -1 as a parameter, and eventually settled back on int64).

dashpole:
tim is OOO, so @vishh will have to approve this.

vishh:
Fixed github ACLs. David you can review and merge PRs henceforth.

PR #1734, opened 2017-08-26
Fix app metrics in docker swarm

AlexJakeGreen:
`NetworkSettings.IPAddress` is empty when using overlay networks.
According to https://github.com/docker/docker/issues/18563#issuecomment-163603765 this path is deprecated.

App metrics can't be used in Swarm without this change.


AlexJakeGreen:
I signed it!

PR #1738, opened 2017-08-29
Update docker godeps to be the same as k8s

dashpole:
Updates the docker godeps to be the same as in k8s:
https://github.com/kubernetes/kubernetes/pull/48308


dashpole:
cc @yguo0905 
can I get a second set of eyes on this before I merge it?

yguo0905:
Are we going to change all the call sites to use the new API instead of `docker/engine-api` and eventually remove the old dependency in a separate PR?

yguo0905:
/lgtm
We can do that later since it's not a blocker.

PR #1739, opened 2017-08-29
Add preliminary suport for GPU Info

### LINKED ISSUE #1436: Display GPU Info on cAdvisor

yongtang:
This fix adds preliminary support for GPU (Nvidia) information, as was specified in #1436.

So far Model, Path(mount path), total memory, and used memory has been collected and show up in machine's information.

The information is collected through nvml library by dlopen and dlclose. Thc cgo is needed.

This fix fixes #1436.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

### DIFF
--- a/machine/nvml.h
+++ b/machine/nvml.h
@@ -1,0 +1,4447 @@
+/*
+ * Copyright 1993-2016 NVIDIA Corporation.  All rights reserved.
+ *
+ * NOTICE TO USER:   
+ *
+ * This source code is subject to NVIDIA ownership rights under U.S. and 
+ * international Copyright laws.  Users and possessors of this source code 
+ * are hereby granted a nonexclusive, royalty-free license to use this code 
+ * in individual and commercial software.
+ *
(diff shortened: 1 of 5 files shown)


machine/nvml.h
```diff
@@ -0,0 +1,4447 @@
+/*
+ * Copyright 1993-2016 NVIDIA Corporation.  All rights reserved.
+ *
+ * NOTICE TO USER:   
+ *
+ * This source code is subject to NVIDIA ownership rights under U.S. and
```
mindprince: I am not sure this file can be included in the repo.

PR #1747, opened 2017-09-01
update docker/distribution

dashpole:
somehow missed this earleir


dashpole:
docker distribution commit taken from here: https://github.com/moby/moby/blob/master/vendor.conf

dashpole:
actually from here: https://github.com/moby/moby/blob/8af4db6f002ac907b6ef8610b237879dfcaa5b7a/vendor.conf

Random-Liu:
LGTM

Issue #1691, opened 2017-06-29
zfs watcher.go error (v 0.26.1)


maxramqvist:
We are seeing the following errors in syslog every time cadvisor tries to read stats from zfs.

A few things to note:
Ubuntu 16.10
Docker version 17.05.0-ce, build 89658be
Our location of zfs is /sbin/zfs
cadvisor version: 0.26.1

If we run the command "zfs get -t all -Hp all -r zpool_default/docker" we get nice output.
If we symlink /sbin/zfs to /usr/sbin/zfs and run "/usr/sbin/zfs zfs get -t all -Hp all -r zpool_default/docker" we get syntax error, zfs isnt a sub-command to zfs.

```
Jun 29 11:24:43 vsp535 docker[14047]: E0629 09:24:43.483472       1 watcher.go:103] encountered error getting children of zfs filesystem: zpool_default/docker: exit status 1: "/usr/sbin/zfs zfs get -t all -Hp all -r zpool_default/docker" => cannot open '-r': dataset does not exist
Jun 29 11:24:43 vsp535 docker[14047]: E0629 09:24:43.483542       1 watcher.go:63] encountered error refreshing zfs watcher: exit status 1: "/usr/sbin/zfs zfs get -t all -Hp all -r zpool_default/docker" => cannot open '-r': dataset does not exist 
```

dashpole:
I am not familiar at all with zfs.  cc @bakins who added the zfs watcher.

bakins:
The error printing the command incorrectly is mentioned here: https://github.com/mistifyio/go-zfs/pull/61

I'll take a look at the error message about "-r" - looks like the wrong arguments are being passed but not sure if that's a cadvisor or go-zfs issue.

bakins:
This appears to be an issue with go-zfs related to newer versions of the zfs utilities.  I'll see if I can get a fix merged there. If I can't, I'll see what workaround I can do here.

bakins:
@maxramqvist have you tried master to see if you have the same issue?

maxramqvist:
@bakins coool, nice work! All seems well in the latest 0.28.0 pre-release. It usually shows up right away in the logs - but now nothing.
I'm missing it in the Docker image, cadvisor:canary though. There I got something about **zfs** was missing from $PATH. I forgot to check the version of the cadvisor binary, but should be more recent - right?

bakins:
[...]

PR #1755, opened 2017-09-12
Update go-zfs dependency

bakins:
The issue reported in #1691 is fixed in upstream go-zfs.


dashpole:
thanks for this!  Would you mind squashing your commits?

bakins:
@dashpole I can squash. I thought I only did one commit anyway.

bakins:
@dashpole squashed. Sorry it took so long.

PR #1762, opened 2017-09-27
Add per container GPU metrics

### LINKED ISSUE #1436: Display GPU Info on cAdvisor

mindprince:
Fixes #1436.

The design doc is at: https://docs.google.com/document/d/13O4HNrB7QFpKQcLcJm28R-QBH3Xo0VmJ7w_Pkvmsf68/

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -37,7 +37,6 @@ if [ "${go_version:0:3}" = "1.4" ]; then
 fi
 
 ldflags="
-  -extldflags '-static'
   -X ${repo_path}/version.Version${ldseparator}${version}
   -X ${repo_path}/version.Revision${ldseparator}${revision}
   -X ${repo_path}/version.Branch${ldseparator}${branch}
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -520,6 +520,29 @@ type FsStats struct {
 	WeightedIoTime uint64 `json:"weighted_io_time"`
 }
 
+type AcceleratorStats struct {
+	// Make of the accelerator (nvidia, amd, google etc.)
+	Make string `json:"make"`
+
+	// Model of the accelerator (tesla-p100, tesla-k80 etc.)
+	Model string `json:"model"`
+
+	// ID of the accelerator.
+	ID string `json:"id"`
+
+	// Total accelerator memory.
+	// unit: bytes
+	MemoryTotal uint64 `json:"memory_total"`
+
+	// Total accelerator memory allocated.
+	// unit: bytes
+	MemoryUsed uint64 `json:"memory_used"`
+
+	// Percent of time over the past sample period during which
+	// the accelerator was actively processing.
+	DutyCycle uint64 `json:"duty_cycle"`
+}
+
 type ContainerStats struct {
 	// The time of this stat point.
 	Timestamp time.Time    `json:"timestamp"`
@@ -533,6 +556,9 @@ type ContainerStats struct {
 
 	// Task load stats
 	TaskStats LoadStats `json:"task_stats,omitempty"`
+
+	// Metrics for Accelerators. Each Accelerator corresponds to one element in the array.
+	Accelerators []AcceleratorStats `json:"accelerators,omitempty"`
 
 	// Custom metrics from all collectors
 	CustomMetrics map[string][]MetricVal `json:"custom_metrics,omitempty"`
--- a/manager/container.go
+++ b/manager/container.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/cadvisor/accelerators"
 	"github.com/google/cadvisor/cache/memory"
 	"github.com/google/cadvisor/collector"
 	"github.com/google/cadvisor/container"
@@ -78,6 +79,9 @@ type containerData struct {
 
 	// Runs custom metric collectors.
 	collectorManager collector.CollectorManager
+
+	// nvidiaCollector updates stats for Nvidia GPUs attached to the container.
+	nvidiaCollector accelerators.AcceleratorCollector
 }
 
 // jitter returns a time.Duration between duration and duration + maxFactor * duration,
@@ -557,6 +561,12 @@ func (c *containerData) updateStats() error {
 		}
 	}
 
+	var nvidiaStatsErr error
+	if c.nvidiaCollector != nil {
+		// This updates the Accelerators field of the stats struct
+		nvidiaStatsErr = c.nvidiaCollector.UpdateStats(stats)
+	}
+
 	ref, err := c.handler.ContainerReference()
 	if err != nil {
 		// Ignore errors if the container is dead.
@@ -571,6 +581,9 @@ func (c *containerData) updateStats() error {
 	}
 	if statsErr != nil {
 		return statsErr
+	}
+	if nvidiaStatsErr != nil {
+		return nvidiaStatsErr
 	}
 	return customStatsErr
 }
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -18,6 +18,7 @@ package manager
 import (
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
 	"strconv"
@@ -25,6 +26,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/cadvisor/accelerators"
 	"github.com/google/cadvisor/cache/memory"
 	"github.com/google/cadvisor/collector"
 	"github.com/google/cadvisor/container"
@@ -44,8 +46,6 @@ import (
 	"github.com/google/cadvisor/utils/oomparser"
 	"github.com/google/cadvisor/utils/sysfs"
 	"github.com/google/cadvisor/version"
-
-	"net/http"
 
 	"github.com/golang/glog"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
@@ -212,6 +212,7 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
 		containerWatchers:        []watcher.ContainerWatcher{},
 		eventsChannel:            eventsChannel,
 		collectorHttpClient:      collectorHttpClient,
+		nvidiaManager:            &accelerators.NvidiaManager{},
 	}
 
 	machineInfo, err := machine.Info(sysfs, fsInfo, inHostNamespace)
@@ -257,6 +258,7 @@ type manager struct {
 	containerWatchers        []watcher.ContainerWatcher
 	eventsChannel            chan watcher.ContainerEvent
 	collectorHttpClient      *http.Client
+	nvidiaManager            accelerators.AcceleratorManager
 }
 
 // Start the container manager.
@@ -309,6 +311,9 @@ func (self *manager) Start() error {
 		return nil
 	}
 
+	// Setup collection of nvidia GPU metrics if any of them are attached to the machine.
+	self.nvidiaManager.Setup()
+
 	// Create root and then recover all containers.
 	err = self.createContainer("/", watcher.Raw)
 	if err != nil {
@@ -338,6 +343,7 @@ func (self *manager) Start() error {
 }
 
 func (self *manager) Stop() error {
+	defer self.nvidiaManager.Destroy()
 	// Stop and wait on all quit channels.
 	for i, c := range self.quitChannels {
 		// Send the exit signal and wait on the thread to exit (by closing the channel).
@@ -916,6 +922,15 @@ func (m *manager) createContainerLocked(containerName string, watchSource watche
 	cont, err := newContainerData(containerName, m.memoryCache, handler, logUsage, collectorManager, m.maxHousekeepingInterval, m.allowDynamicHousekeeping)
 	if err != nil {
 		return err
+	}
+	devicesCgroupPath, err := handler.GetCgroupPath("devices")
+	if err != nil {
+		glog.Infof("Error getting devices cgroup path: %v", err)
+	} else {
+		cont.nvidiaCollector, err = m.nvidiaManager.GetCollector(devicesCgroupPath)
+		if err != nil {
+			glog.Infof("GPU metrics may be unavailable/incomplete for container %q: %v", cont.info.Name, err)
+		}
 	}
 
 	// Add collectors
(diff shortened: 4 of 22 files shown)


mindprince:
/cc @vishh @jiayingz @dashpole 
Looking for early feedback.

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
 	WeightedIoTime uint64 `json:"weighted_io_time"`
 }
 
+// NvidiaGpuStats are measured using NVML.
+type NvidiaGpuStats struct {
```
dashpole: Do we want this to be nvidia-specific?  Would it be better to have a GpuStats that we could identify as an nvidia GPU?  I guess it comes down to whether all GPUs generally have the same metrics, or if Nvidia's metrics are different than AMD's GPU metrics, for example

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
...
+	// Minor number of the GPU device (/dev/nvidia[minor_number])
+	MinorNumber uint64 `json:"minor_number"`
```
dashpole: How do we plan to identify a GPU in the device plugin model?  It would be nice if the minimum set of identifying information was the same across cAdivsor and kubernetes (e.g. maker, model, minor_number)?

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
...
+	// unit: bytes
+	MemoryUsed uint64 `json:"memory_used"`
+
+	// Percent of time over the past sample period (of nvml) during which
+	// one or more kernels was executing on the GPU.
+	Busy uint64 `json:"busy"`
+
+	// Power usage for this GPU and its associated circuitry.
+	// unit: milliwatts
+	PowerUsage uint64 `json:"power_usage"`
```
dashpole: Is this cumulative?  if so, we should specify.  If not, what is the period of time?

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
...
+type NvidiaGpuStats struct {
+	// Minor number of the GPU device (/dev/nvidia[minor_number])
+	MinorNumber uint64 `json:"minor_number"`
+
+	// Product name of the GPU
+	Name string `json:"name"`
+
+	// Total installed GPU memory.
+	// unit: bytes
+	MemoryTotal uint64 `json:"memory_total"`
```
dashpole: This should be a constant, right?  If so, does it belong in the monitoring pipeline?  Would it be better to surface it as a "spec" somewhere?

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
...
+
+	// Total GPU memory allocated.
+	// unit: bytes
+	MemoryUsed uint64 `json:"memory_used"`
+
+	// Percent of time over the past sample period (of nvml) during which
+	// one or more kernels was executing on the GPU.
+	Busy uint64 `json:"busy"`
```
dashpole: I am not familiar at all with nvml's capabilities, but I think it is generally preferable to use cumulative usage (in seconds), or another cumulative metric.  This allows metric aggregators to get average usage over short (for "instantaneous") or long (for average) intervals.  If we cant get that, then I would think it would be difficult to measure average usage over longer periods of time.  Do you know if the "sample period" is a constant?  The added c file below says "Each sample period may be between 1 second and 1/6 second, depending on the product being queried".

info/v1/container.go
mindprince: I prefixed this with nvidia because I wasn't really able to find what metrics other GPU vendors expose (AMD/Intel etc.) and what I did find seemed very non-standard.

Not sure what the best way here is. I would prefer to not add a generic API without knowing that it's really generic enough to fit all (most) use cases.

info/v1/container.go
mindprince: Good point. It depends on the device plugin implementation (which is outside the kubernetes/kubernetes tree). In the current implementation, when you ask for `nvidia.com/gpu` resource, the GPU IDs are `nvidia0`, `nvidia1` etc. So, just the minor numbers. I think there are plans to request different models of GPUs through annotations possibly but nothing concrete has happened in that regard yet.

cc - @jiayingz

info/v1/container.go
mindprince: It's constant for a particular GPU, but there can be multiple GPUs with different total memory each. If I move it to spec I would have to have a way to link the total memory of a GPU with its usage. I felt using two different time series while plotting would be inconvenient for the user.

info/v1/container.go
mindprince: Unfortunately, nvml doesn't expose cumulative usage. It only exposes this point-in-time gauge. The sample period is dependent upon the GPU type as you found out.

info/v1/container.go
mindprince: The unit is milliwatts. The  [documentation](http://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g7ef7dff0ff14238d08a19ad7fb23fc87) doesn't say how long they sample the energy consumption to calculate the power usage. I would assume it's another point-in-time gauge.

info/v1/container.go
dashpole: SGTM

info/v1/container.go
dashpole: Ok, I think that is reasonable.

dashpole:
@mindprince Ill take another look at the implementation once it is close to being finished.  Let me know when it is ready

build/build.sh
```diff
@@ -37,7 +37,6 @@ if [ "${go_version:0:3}" = "1.4" ]; then
 fi
 
 ldflags="
-  -extldflags '-static'
```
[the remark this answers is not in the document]
flx42: @tianshapjq this shouldn't be needed if the container runtime does it for you (e.g. with `nvidia-docker` or a device plugin).

info/v1/container.go
```diff
@@ -516,6 +516,31 @@ type FsStats struct {
...
 
+// NvidiaGpuStats are measured using NVML.
+type NvidiaGpuStats struct {
+	// Minor number of the GPU device (/dev/nvidia[minor_number])
+	MinorNumber uint64 `json:"minor_number"`
+
+	// Product name of the GPU
+	Name string `json:"name"`
+
+	// Total installed GPU memory.
```
flx42: You don't "install" GPU memory like you do for CPU memory.

manager/container.go
```diff
@@ -572,9 +579,116 @@ func (c *containerData) updateStats() error {
...
+
+		// NVIDIA graphics devices are character devices with major number 195.
+		// https://github.com/torvalds/linux/blob/v4.13/Documentation/admin-guide/devices.txt#L2583
+		if fields[0] == "c" && majorMinor[0] == "195" {
+			minorNumber, err := strconv.Atoi(majorMinor[1])
+			if err == nil {
+				nvidiaMinorNumbers = append(nvidiaMinorNumbers, minorNumber)
+			}
+			// We are ignoring the "195:*" case
+			// where the container has access to all NVIDIA devices on the machine.
```
flx42: This comment slightly confused me, a container can have access to all GPUs on the machine but not have `195:*`, for instance if it enumerates each device:
```
c 195:3 rwm
c 195:2 rwm
c 195:1 rwm
c 195:0 rwm
```
It depends on what the container runtime is doing.

What is the rationale behind skipping this use case?

info/v1/container.go
```diff
@@ -530,6 +555,9 @@ type ContainerStats struct {
 	// Task load stats
 	TaskStats LoadStats `json:"task_stats,omitempty"`
 
+	// Metrics for NVIDIA GPUs. Each GPU corresponds to one element in the array.
+	NvidiaGpu []NvidiaGpuStats `json:"nvidia_gpu,omitempty"`
```
flx42: This assumes that one GPU is used by only one container, right?
Does that mean you won't be able to see the status of GPUs that are not being used by any container? It might still be useful (e.g., they might be overheating, or be in a unhealthy state).

My feeling is that the current `NvidiaGpuStats` struct should be per-node (for all GPUs) and not per-container.
Per-container, you could still try to detect the assigned GPUs, or even better: determine what each container is really using. This could be done by using the NVML API to list the PIDs that have an active context on the GPU and then finding which container these PIDs belong to. This would be similar to what you see when you do `nvidia-smi`.

build/build.sh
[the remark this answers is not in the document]
tianshapjq: @flx42 that means I have to install nvidia-docker in prior if we want to use this functionality? That we may have to note in the ReadMe as well IIUC.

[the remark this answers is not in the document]
mindprince: I tested this by directly running it using:
```
sudo ./cadvisor -alsologtostderr
```
Kubelet also runs directly on the host, so it would work there as well.

You are right, we need to document that to run inside a container and still get GPU metrics, you will need to either use `nvidia-docker` or mount the libraries and setup some environment variables like what device-plugins do.

manager/manager.go
```diff
@@ -347,6 +359,35 @@ func (self *manager) Stop() error {
...
+		return
+	}
+	glog.Infof("NVML initialized. Number of nvidia devices: %v", numDevices)
+	nvidiaDevices = make(map[int]gonvml.Device, numDevices)
+	for i := 0; i < int(numDevices); i++ {
+		device, err := gonvml.DeviceHandleByIndex(uint(i))
+		if err != nil {
+			glog.Warningf("Failed to get nvidia device handle %d: %v", i, err)
+			continue
+		}
```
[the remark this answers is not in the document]
mindprince: This method is called only after NVML is initialized and we know the device count, I don't see how this can be a transient error after those two conditions are satisfied. The only failure scenario after that is if the user doesn't have permission to access the device ([documentation](http://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g52677ecb45f937c5005124608780a3f4)). In that case, there's no point in retrying.

Currently, the code requires the GPUs to be already attached and NVML installed before cadvisor is started. I am thinking of making it so that it rechecks for NVML/GPU presense every so often.

info/v1/container.go
flx42: Nevermind then, let's be consistent.

manager/manager.go
```diff
@@ -363,6 +363,27 @@ func (self *manager) Stop() error {
...
+
+	go func() {
+		glog.Info("Starting goroutine to initialize NVML")
+		initializeNVML()
+		if nvmlInitialized {
+			return
+		}
+		for range time.Tick(5 * time.Minute) {
+			initializeNVML()
+			if nvmlInitialized {
```
tianshapjq: the nvidiaDevices attribute is updated in the initializeNVML(), since it's also called somewhere else, do we need to sync lock it up when updating the nvidiaDevices?

manager/container.go
```diff
@@ -572,9 +579,116 @@ func (c *containerData) updateStats() error {
...
 }
 
+// updateNvidiaStats updates the stats for NVIDIA GPUs (if any) attached to the container.
+func (c *containerData) updateNvidiaStats(stats *info.ContainerStats) error {
+	if !nvmlInitialized || len(nvidiaDevices) == 0 {
+		return nil
+	}
+	nvidiaMinorNumbers := c.getNvidiaMinorNumbers()
+	for _, minor := range nvidiaMinorNumbers {
+		device := nvidiaDevices[minor]
```
tianshapjq: as your new commit, the nvidiaDevices will be updated every 5 minutes. should we sync to lock it up here as well?

manager/manager.go
mindprince: I don't think a lock is needed. `nvidiaDevices` is only ever written by this goroutine. Once we return from here, it's never updated. Everywhere else we are only reading the value. In the worst case, if the reads happen while `initializeNVML()` is in progress, the read will see incomplete devices. That will be corrected in the next housekeeping.

tianshapjq: SGTM

mindprince:
@vishh @dashpole This is ready for review.

info/v1/container.go
```diff
@@ -520,6 +520,33 @@ type FsStats struct {
...
+	// unit: bytes
+	MemoryTotal uint64 `json:"memory_total"`
+
+	// Total accelerator memory allocated.
+	// unit: bytes
+	MemoryUsed uint64 `json:"memory_used"`
+
+	// Percent of time over the past sample period during which
+	// the accelerator was actively processing.
+	DutyCycle uint64 `json:"duty_cycle"`
```
dashpole: Is this common jargon for GPU utilization?  It doesn't seem intuitive to me, but I also am also not very familiar with GPUs...

dashpole: To clarify: Is "DutyCycle" something most people will know is the utilization of their GPU?

mindprince: Yeah, duty cycle is commonly used to refer percent of time in which a system was active.

manager/container.go
```diff
@@ -572,9 +580,123 @@ func (c *containerData) updateStats() error {
...
+
+		// Split the second field to find out major:minor numbers
+		majorMinor := strings.Split(fields[1], ":")
+		if len(majorMinor) != 2 {
+			return nvidiaMinorNumbers, fmt.Errorf("invalid devices cgroup entry %q: second field should have one colon", text)
+		}
+
+		// NVIDIA graphics devices are character devices with major number 195.
+		// https://github.com/torvalds/linux/blob/v4.13/Documentation/admin-guide/devices.txt#L2583
+		if fields[0] == "c" && majorMinor[0] == "195" {
```
vishh: @flx42 Are these APIs reliable enough for at-least a year?

flx42: It's not an API, it's an implementation detail that happens to work with Docker.
There are other ways to block GPUs besides groups, for instance you can bind mount the devices and then block `mknod`, for instance when using user namespaces.

That being said, for the Docker use case, I'm not aware of any upcoming breaking change, but my vision doesn't go up to one year forward.

mindprince:
@vishh @dashpole Abstracted out the nvidia metrics collection. Please take a look. If it looks good now, let's get this merged.

vishh:
@dashpole this PR LGTM. If you are happy with it, let's get it merged.

PR #1766, opened 2017-10-05
adaptive longOp for du operation

sjenning:
When running pod density tests on the order of 500 pods, the du operation in the fsHandler can take longer than a second, even if everything about the filesystem is cached.  This is a persistent condition that causes the message to continually print.

This PR makes `longOp` adaptive so if the long du duration is a persistent condition, the message won't continue in the logs forever.

https://bugzilla.redhat.com/show_bug.cgi?id=1498632

@dashpole @vishh @derekwaynecarr @ravisantoshgudimetla

### DIFF
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -51,7 +51,6 @@ type realFsHandler struct {
 }
 
 const (
-	longOp           = time.Second
 	timeout          = 2 * time.Minute
 	maxBackoffFactor = 20
 )
@@ -111,6 +110,7 @@ func (fh *realFsHandler) update() error {
 
 func (fh *realFsHandler) trackUsage() {
 	fh.update()
+	longOp := time.Second
 	for {
 		select {
 		case <-fh.stopChan:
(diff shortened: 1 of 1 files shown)


ravisantoshgudimetla:
@sjenning Sometimes, I have noticed that du operation is taking more than 30 seconds, which means this line will be printed 300 times after which it should stop logging, which I believe should work fine. Any specific reason for increasing log level from 2 to higher value?

container/common/fsHandler.go
```diff
@@ -60,6 +59,8 @@ const DefaultPeriod = time.Minute
 
 var _ FsHandler = &realFsHandler{}
 
+var longOp = time.Second
```
ravisantoshgudimetla: Can we move this variable to trackUsage fn?

sjenning:
@ravisantoshgudimetla at first I did just decrease the log level.  But then someone running at the higher logging level would have the same complaint; that being continually seeing a message warning me about a persistent condition that I can't do anything about because someone decided what "too long" was.

ravisantoshgudimetla:
> at first I did just decrease the log level. But then someone running at the higher logging level would have the same complaint

Makes sense.

container/common/fsHandler.go
```diff
@@ -129,6 +129,10 @@ func (fh *realFsHandler) trackUsage() {
 			duration := time.Since(start)
 			if duration > longOp {
 				glog.V(2).Infof("du and find on following dirs took %v: %v", duration, []string{fh.rootfs, fh.extraDir})
+				// adapt longOp time so that message doesn't continue to print
```
derekwaynecarr: log something like the following:

"du and find on following dirs took %v: %v; will log in future if exceeds %v"

derekwaynecarr:
LGTM

PR #1768, opened 2017-10-10
Container stats expose memory max usage

derekwaynecarr:
Expose `memory.max_usage_in_bytes` per container from memory cgroup.

Value is visible from REST API and Prometheus (container_memory_max_usage_bytes)

This is useful for understanding peak usage for a container over its life.

/cc @dashpole @sjenning @smarterclayton @dchen1107 @vishh


smarterclayton:
Very happy - short running jobs now can be effectively estimated by prometheus based on peak memory usage in bursty scenarios.

dashpole:
FYI, this change should not add any additional overhead because we were already reading this cgroup file.
We previously threw away the value, as we did not expose it in any API.

Issue #1765, opened 2017-10-05
PercpuUsage errors in logs


dada941:
After upgrading to Kubernetes 1.8, the kubelet logs is flooded with the following error :

`helpers.go:468] PercpuUsage had 0 cpus, but the actual number is 2; ignoring extra CPUs`

Looks like it's related to https://github.com/google/cadvisor/pull/1711. Should those extras 0 filter log an error ?

We're running CoreOS beta (**1520.4.0**) in AWS :
`Linux ip-10-150-16-116.ap-southeast-2.compute.internal 4.13.3-coreos #1 SMP Wed Sep 20 22:17:11 UTC 2017 x86_64 Intel(R) Xeon(R) CPU E5-2676 v3 @ 2.40GHz GenuineIntel GNU/Linux`

unixwitch:
Also seeing this on GCE with CentOS 7.4 and Kubernetes 1.8:

```
Oct 10 05:29:39 staging-head-2 kubelet[1513]: E1010 05:29:39.001380    1513 helpers.go:468] PercpuUsage had 0 cpus, but the actual number is 1; ignoring extra CPUs
```

```
Linux staging-head-2 3.10.0-693.2.2.el7.x86_64 #1 SMP Tue Sep 12 22:26:13 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

I understood from #1711 that this should only affect 4.7+, so I'm not entirely sure it's the same issue, but RedHat backports a lot of changes from newer kernels.

dashpole:
cc @euank

euank:
@dada941 How are you running the kubelet? Can you also post the output of `cat /sys/fs/cgroup/cpu,cpuacct/cpuacct.usage_percpu` (or cpu or cpuacct depending on which exist)?

We should probably treat '0' as a special value and shortcut out early since if `len(s.CpuStats.CpuUsage.PercpuUsage)` that could indicate an error condition I think, such as cgroups not mounted or that cgroup feature not being available.

After I understand why this is happening I can make a patch for that.

dada941:
Thanks for looking into this @euank 

Yes, the Kubelet runs in a RKT container in CoreOS, but I've tried out of a container with the binary, and get the same result.

Here's the output of the command, run from the RKT container :

`# cat /sys/fs/cgroup/cpu,cpuacct/cpuacct.usage_percpu`
`87983471343 95801030681 0 0 0 0 0 0 0 0 0 0 0 0 0`

[...]

PR #1769, opened 2017-10-11
libcontainer: ignore nil cpustats

euank:
Cadvisor can inotify watch for new cgroups. This leads to it racing
fairly tightly with cgroup creation... So tightly, that sometimes
cpustats are nil.

The runc library code we call
(https://github.com/opencontainers/runc/blob/v1.0.0-rc4/libcontainer/cgroups/fs/apply_raw.go#L179-L182)
doesn't actaully consider this an error, so we have to handle that
scenario ourselves.

This fixes https://github.com/google/cadvisor/issues/1765

I've verified that applying this patch to the kubelet fixes the instance of the above issue I could observe.

As far as I can tell, this patch is cosmetic in that it fixes an error log, but no actually erroneous behavior.

### DIFF
--- a/container/libcontainer/helpers.go
+++ b/container/libcontainer/helpers.go
@@ -452,6 +452,17 @@ var numCpusFunc = getNumberOnlineCPUs
 func setCpuStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	ret.Cpu.Usage.User = s.CpuStats.CpuUsage.UsageInUsermode
 	ret.Cpu.Usage.System = s.CpuStats.CpuUsage.UsageInKernelmode
+	ret.Cpu.Usage.Total = 0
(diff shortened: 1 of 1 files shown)


dashpole:
please rebase, and then Ill merge it!

euank:
Rebased. 

Only one commit to this repository has happened since I made my PR and that change didn't touch any related files.

Does this repository have the "Allow merge commits" setting off? Is there any reason? The commit history does look nice and linear as a result, but this rebase was 100% unnecessary. There's no conflicts, no reason it would conflict, and a total waste of developer time.

If the worry is that it isn't being tested identically to how it's merged, well, that's what the submit queue solves for k8s proper, and what homu solves for other problems... and those solutions don't require your contributors to do utterly pointless busy work.

PR #1770, opened 2017-10-11
Monitor diff directory for overlay2

dashpole:
Fixes https://github.com/kubernetes/kubernetes/issues/52336

The directory structure of overlay2 is different from overlay, and we should monitor the contents of /var/lib/docker/overlay2/CONTAINER_ID/diff/ instead of just /var/lib/docker/overlay2/CONTAINER_ID/

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -43,7 +43,8 @@ import (
 
 const (
 	// The read write layers exist here.
-	aufsRWLayer = "diff"
+	aufsRWLayer     = "diff"
+	overlay2RWLayer = "diff"
 
 	// Path to the directory where docker stores log files if the json logging driver is enabled.
 	pathToContainersDir = "containers"
@@ -195,8 +196,10 @@ func newDockerContainerHandler(
 	switch storageDriver {
 	case aufsStorageDriver:
 		rootfsStorageDir = path.Join(storageDir, string(aufsStorageDriver), aufsRWLayer, rwLayerID)
-	case overlayStorageDriver, overlay2StorageDriver:
+	case overlayStorageDriver:
 		rootfsStorageDir = path.Join(storageDir, string(storageDriver), rwLayerID)
+	case overlay2StorageDriver:
+		rootfsStorageDir = path.Join(storageDir, string(storageDriver), rwLayerID, overlay2RWLayer)
 	case zfsStorageDriver:
 		status, err := Status()
 		if err != nil {


dashpole:
we should cherrypick this onto the v0.27 and v0.26 branch

dashpole:
cc @tallclair

dashpole:
actually, lets wait on this until I have done some more digging.

derekwaynecarr:
fyi @runcom @mrunalp @sjenning -- input here?

also do we have the same or similar issue w/ crio?

mrunalp:
@derekwaynecarr Yeah, we are returning the path to container rootfs. So, we need to add "diff" here to do the equivalent for cri-o.

runcom:
> @derekwaynecarr Yeah, we are returning the path to container rootfs. So, we need to add "diff" here to do the equivalent for cri-o - https://github.com/google/cadvisor/blob/master/container/crio/handler.go#L144

during my testing overlay2 didn't show this issue, I'm not sure...

dashpole:
I tested the directory setup by doing the following:
On a node running ubuntu xenial, docker 1.12.6.
Change storage driver to aufs, restart docker, run `docker run -it busybox sh`, then `touch myfile`.
Repeat for overlay and overlay2.

`/var/lib/docker# find . | grep myfile`
`./overlay2/25bee5c56e3f26d2aaaf228a71ffa947970329fc9692e85ae68a5418dcc6387b/diff/myfile`
`./aufs/diff/6d1dbb63ff495abe6389cdadc9bc49b7621f03c5d98f136e30316f3d0bf8984f/myfile`
`./overlay/b037efcadbc8129c861ec5e613803697fd09c0c368c02bd48db0e576d8400690/upper/myfile`

So we may also have a bug with overlay as well...

dashpole:
Based on the [docker overlay documentation](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/#how-the-overlay-driver-works), we may be running du and find on the image layers in overlay as well... 

Looking at `/var/lib/docker/overlay/`, I can see some directories structured like this: `ID/root`, and others with `ID/work`, `ID/upper`, `ID/merged`.  According to the docs, the one with `ID/root` is an image, and the one with work, upper, and merged directories is a container...

derekwaynecarr:
@mrunalp @runcom can you do same exercise as @dashpole on crio?

mrunalp:
If we want to get to the upper layer of the container in crio, it is as I pointed out earlier https://github.com/google/cadvisor/blob/master/container/crio/handler.go#L144 + /diff.

e.g. /var/lib/containers/storage/overlay/1462529fa84a113ec84ef5406c6db96812d792e632c66555e1032404c2431486/diff

dashpole:
Based on the docker docs, and my own experiments, I think this fixes the overlay2 incorrect root directory issue.  @derekwaynecarr would you mind providing a review?

runcom:
@mrunalp yes, I'm working on a cri-o fix

timothysc:
I think this is cherry-pick worthy fwiw.

derekwaynecarr:
@dashpole I will review

timothysc:
@dashpole Can we get this all the way onto 1.8.2?  

/cc @jpbetz

warmchang:
👍

lf1029698952:
LGTM 👍

Issue #1247, opened 2016-04-27
Ability to get root container stats at separate housekeeping interval


derekwaynecarr:
In order to support out of resource monitoring in Kubernetes, I want to be able to get information about the root container at a separate interval than the information I gather on containers associated with pods.  For example, I would set housekeeping for containers associated with pods at 10s, but root container at 100ms.

A potential option is to add a flag:

``` shell
-root_housekeeping_interval duration 
   if specified, perform housekeeping on the root container at the specified interval 
   rather than default housekeeping interval
```

/cc @pmorie @ncdc @vishh - Thoughts?

ncdc:
SGTM

derekwaynecarr:
From what I can gather, fsInfo is computed on demand, so no separate housekeeping interval is needed there, but @pmorie has informed me that thin_ls data per container is cached.  Either way for out of resource killing, we care more about rootfs available bytes and imagefs available bytes.

pmorie:
Yep, thin_ls data is cached, but my WIP hasn't established at what interval it is refreshed

vishh:
My understanding is that kubelet mainly needs higher resolution for machine level stats and not for container stats. `thin_ls` matters only to containers and not the machine itself.
@timstclair suggested adding a query param `maxAge` to `/api/v2.1/machineStats` endpoint that will help us get more recent stats on demand.

timstclair:
+1 for on demand stats. I'd also like to avoid adding more flags if we can.

derekwaynecarr:
@vishh @timstclair  
My understanding is that machine stats are derived by the root container stats:

https://github.com/google/cadvisor/blob/master/api/versions.go#L483

I will throw another wrinkle in here, and broaden the request.

I suspect when we get a little further into the future, we will want to get the stats for certain special containers at a higher frequency.  For example, the `kubelet` container, the `docker` container, and the `container that parents end-user pods` are all special, and should reasonably be able to ask for higher fidelity housekeeping intervals.

The `container that parents end-user pods` is probably the container that we will want to drive eviction on when we move to pod-level cgroups world.

So I want to come back and re-phrase my request, I want to be able to tell cadvisor a set of special containers that have a shorter housekeeping interval.  I am fine not exposing it as a flag to the binary, but I would like to be able to specify it for how Kubernetes starts its internal cadvisor.

Thoughts?

timstclair:
Can't we address that with on-demand scraping as well?

[...]

PR #1779, opened 2017-10-23
On-Demand container metrics

dashpole:
This is my take on implementing https://github.com/google/cadvisor/issues/1247.
This PR does the following:

1. Adds an UpdateStats field to v2 request options.  Setting this to true causes cAdvisor to update stats for that container before returning.
2. Adds an OnDemandHousekeeping function to the containerData object, which allows housekeeping to be triggered before the interval.
3. This removes the check with instantaneous CPU usage if the difference is too small.  This check could result in not getting instantaneous CPU usage if OnDemandHousekeeping is triggered right after the periodic collection.

There is still only a single thread which calls housekeepingTick().
The UpdateStats field is part of the v2 API in cAdvisor, but will be used only for node-level metrics in kubernetes.  See https://github.com/kubernetes/kubernetes/compare/master...dashpole:on_demand_metrics for the way this would be enabled in kubernetes.


dashpole:
cc @derekwaynecarr

manager/container.go
```diff
@@ -407,24 +424,22 @@ func (c *containerData) housekeeping() {
...
-				glog.V(3).Infof("[%s] Housekeeping took %s", c.info.Name, duration)
-			}
-		}
-
+		case finishedChan := <-c.onDemandChan:
+			c.housekeepingTick(longHousekeeping)
+			// notify the calling function that housekeeping has completed
+			close(finishedChan)
+		case <-houseKeepingTimer.C:
+			c.housekeepingTick(longHousekeeping)
```
tallclair: drain the `finishedChan` after this step. I.e.
```go
for {
  select finishedChan := <-c.onDemandChan:
    close(finishedChan)
  default:
    break
}
```

info/v2/container.go
```diff
@@ -234,6 +234,8 @@ type RequestOptions struct {
 	Count int `json:"count"`
 	// Whether to include stats for child subcontainers.
 	Recursive bool `json:"recursive"`
+	// Whether to update stats before returning
+	UpdateStats bool `json:"update_stats"`
```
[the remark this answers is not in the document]
dashpole: I like that it is clearer.  UpdateStats seemed vague, but I couldn't think of anything clearer.  I have no plans to do anything other than trigger an update, mainly for responding when memcg thresholds are crossed.  Do you think it is still worth doing MaxAge if we wont use it in k8s?

manager/container.go
dashpole: done, although I dont think we want a for loop around it.

manager/container.go
```diff
@@ -407,24 +424,22 @@ func (c *containerData) housekeeping() {
 
 	// Housekeep every second.
 	glog.V(3).Infof("Start housekeeping for container %q\n", c.info.Name)
-	lastHousekeeping := time.Now()
+	c.housekeepingTick(longHousekeeping)
 	for {
+		houseKeepingTimer := time.NewTimer(c.nextHousekeepingInterval())
```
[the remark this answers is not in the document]
dashpole: done

info/v2/container.go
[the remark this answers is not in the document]
tallclair: Yeah, I think it's still preferable. It's definitely useful outside of k8s, but even for the k8s usecase, there is probably some max age you could set for the memcg trigger (e.g. 10ms). It means that if a burst of requests came in simultaneously (for whatever reason), they can reuse the cached results.

manager/container.go
tallclair: why not? If there are multiple routines waiting for new stats and we just computed new stats, we should unblock them?

dashpole: done

info/v2/container.go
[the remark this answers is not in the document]
dashpole: Ok, I changed it to MaxAgeMS *int32.  Although I was wondering if just using time.Duration directly would be better?

dashpole:
FYI, I plan to cut cadvisor v0.28.1, which will correspond to k8s v1.9.0 this Friday, 11/17.

info/v2/container.go
```diff
@@ -236,6 +236,9 @@ type RequestOptions struct {
 	Count int `json:"count"`
 	// Whether to include stats for child subcontainers.
 	Recursive bool `json:"recursive"`
+	// Update stats if they are older than MaxAge
+	// nil indicates no update, and 0 will always trigger an update.
+	MaxAge *time.Duration `json:"max_age"`
```
dashpole: Duration is just a wrapper around int64, so this should be safe for the api
https://golang.org/src/time/time.go?s=21095:21114#L610

manager/container.go
```diff
@@ -117,6 +120,34 @@ func (c *containerData) allowErrorLogging() bool {
...
+		c.onDemandChan <- housekeepingFinishedChan
+		select {
+		case <-c.stop:
+			close(housekeepingFinishedChan)
+		case <-housekeepingFinishedChan:
+		}
+	}
+}
+
+// notifyOnDemand notifies all calls to OnDemandHousekeeping
```
tallclair: nit: add "that housekeeping is finished"

dashpole: done

dashpole:
@tallclair unit test added

manager/container_test.go
```diff
@@ -226,3 +226,23 @@ func TestUpdateNvidiaStats(t *testing.T) {
...
+		nil,
+	)
+	go func() {
+		cd.housekeeping()
+	}()
+	cd.OnDemandHousekeeping(0 * time.Second)
+	// expect one update stats call from the initial housekeeping loop, then a second from on-demand
+	checkNumStats(t, memoryCache, 2)
+	mockHandler.AssertExpectations(t)
+	cd.Stop()
```
tallclair: nit: make this a defer

manager/container_test.go
```diff
@@ -226,3 +226,23 @@ func TestUpdateNvidiaStats(t *testing.T) {
...
 	assert.Equal(t, info.ContainerStats{}, stats)
 }
+
+func TestOnDemandHousekeeping(t *testing.T) {
+	cd, _, _ := newTestContainerData(t)
+	statsList := itest.GenerateRandomStats(1, 4, 1*time.Second)
+	stats := statsList[0]
+
+	cd, mockHandler, memoryCache := newTestContainerData(t)
+	mockHandler.On("GetStats").Return(
```
tallclair: Use Run instead of Return to ensure this is called, and grab the current time.

manager/container_test.go
```diff
@@ -226,3 +226,23 @@ func TestUpdateNvidiaStats(t *testing.T) {
...
+		stats,
+		nil,
+	)
+	go func() {
+		cd.housekeeping()
+	}()
+	cd.OnDemandHousekeeping(0 * time.Second)
```
tallclair: I would like to see a few other cases: non-zero call get's stats if necessary, but not if unnecessary, nil call does not get new stats, and multiple simultaneous calls only get stats once, and all return. You might need to plumb a fakeclock through to make this non-flaky.

dashpole:
Unit testing is complete.  I had to plum through a clock.  It should be fine to use apimachinery, right?

manager/container_test.go
```diff
@@ -226,3 +231,69 @@ func TestUpdateNvidiaStats(t *testing.T) {
...
+			housekeepingFinishedChan := make(chan struct{})
+			c.onDemandChan <- housekeepingFinishedChan
+			waitForQueue.Done()
+			select {
+			case <-c.stop:
+			case <-housekeepingFinishedChan:
+			}
+			waitForHousekeeping.Done()
+		}
+	}
```
tallclair: Correct me if I'm wrong, but I think you can get away with not mocking this:

```go
onDemandCache := make(chan chan struct{}, numConcurrentCalls)
for i := 0; i < numConcurrentCalls; i++ {
	go func() {
		c.OnDemandHousekeeping(0*time.Second)
		waitForHousekeeping.Done()
	}
	// Wait for work to be queued
	onDemandCache = append(onDemandCache, <-c.onDemandChan)
}
// Requeue work:
for _, ch := range onDemandCache {
	c.OnDemandChan <- ch
}

**Comment by dashpole** [comment]
I tested this in kubernetes/kubernetes and verified that it passes the summary API test as well

**Comment by tallclair** [comment]
Thanks for the tests. We've tried to avoid dependencies on Kuberneties to avoid the circular dependency, but I think it's ok in this case since it's a small utility package, and only used for testing.

File: manager/container_test.go (hunk 9 above)
Comment (dashpole) [reply manager/container_test.go]: I can switch to that.

Comment (dashpole) [reply manager/container_test.go]: done

**Comment by dashpole** [comment]
Ok, it should be ready

File: manager/container_test.go [hunk 10]
```diff
@@ -226,3 +231,66 @@ func TestUpdateNvidiaStats(t *testing.T) {
...
+	var waitForHousekeeping sync.WaitGroup
+	waitForHousekeeping.Add(numConcurrentCalls)
+	onDemandCache := make(chan chan struct{}, numConcurrentCalls)
+	for i := 0; i < numConcurrentCalls; i++ {
+		go func() {
+			cd.OnDemandHousekeeping(0 * time.Second)
+			waitForHousekeeping.Done()
+		}()
+		// Wait for work to be queued
+		onDemandCache <- <-cd.onDemandChan
```
tallclair: nit: I'd prefer a slice for the cache. It makes the requeue more straightforward.

dashpole: yeah, but the double arrow looks really cool.  Done

derekwaynecarr:
@dashpole --- this is really awesome.

derekwaynecarr:
fyi @sjenning xmas came early!

PR #1782, opened 2017-10-26
Change timeout handling for long running disk operations

majst01:
Currently the timeout for `GetDirDiskUsage` and `GetDirInodeUsage` is set to a fixed time of 2 Minutes, regardless how long the fh.period is set.
If disks are slow and fh.period is way shorter than the duration of execution of above methods, this will generate a lot of go routines waiting to be killed after 2min.

With this PR we set the overall timeout of all disk functions to no longer than fh.period that after period all disk activities should be finished, or canceled at least.

Furthermore there was a typo in the handling of errors, fixed that.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -416,8 +416,8 @@
 		},
 		{
 			"ImportPath": "github.com/mistifyio/go-zfs",
-			"Comment": "v2.1.1-5-g1b4ae6f",
-			"Rev": "1b4ae6fb4e77b095934d4430860ff202060169f8"
+			"Comment": "v2.1.1-31-g166dd29",
+			"Rev": "166dd29edf054bff262cdaa5ffcaddb50ced9e31"
 		},
 		{
 			"ImportPath": "github.com/mrunalp/fileutils",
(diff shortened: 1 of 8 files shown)


container/common/fsHandler.go
```diff
@@ -78,27 +77,31 @@ func (fh *realFsHandler) update() error {
...
 	fh.lastUpdate = time.Now()
-	if rootDiskErr == nil && fh.rootfs != "" {
+	if rootInodeErr == nil && fh.rootfs != "" {
 		fh.usage.InodeUsage = inodeUsage
 	}
-	if rootInodeErr == nil && fh.rootfs != "" {
+	if rootDiskErr == nil && fh.rootfs != "" {
 		fh.usage.TotalUsageBytes = baseUsage + extraDirUsage
-	}
-	if extraDiskErr == nil && fh.extraDir != "" {
```
dashpole: why remove this error checking?

container/common/fsHandler.go
```diff
@@ -78,27 +77,31 @@ func (fh *realFsHandler) update() error {
 		rootDiskErr, rootInodeErr, extraDiskErr error
 	)
 	// TODO(vishh): Add support for external mounts.
+	methodTimeout := fh.period
```
dashpole: so the default period is 1 minute, and we are trying to do 3 operations within that time, whereas the old timeout was 2 minutes for each operation.  Seems like a huge change...

majst01:
Got your point and will make 2 distinct new PR. Thanks for review.

PR #1789, opened 2017-11-03
Calculate disk usage only for same mount

majst01:
Do not add mounted volumes outside of container volume to diskusage, this makes the behavior of du equivalent to find in GetDirInodeUsage and reduces wait io on large host volumes.


dashpole:
Looks like the unit test uses a separate filesystem to test

majst01:
OK, will check

majst01:
Testing locally works, might be a flaky test ?

/retest

majst01:
Hi David, i have gone through this test code and i have no clue why it does not work in the testing infrastructure, it works locally and on several other machines as well.

majst01:
Hi @dashpole, thanks for letting tests run again, i am unsure how to proceed here. I think my PR is correct and solves big io issues here, but tests are failing on the google testing infra. I cannot reproduce this behavior by running all tests on different machines here.

Any idea whats the reason why this tests are failing ?

If no idea pops up we should close this PR, what do you think ?
Greetings
Stefan

dashpole:
FAIL: TestDirDiskUsage (0.06s)
	Error Trace:	fs_test.go:106
	Error:      	Should be true
	Messages:   	expected dir size to be at-least 102400; got size: 4096

Looks like your unit tests are failing.  Try running `make` locally

dashpole:
I think this would be great to add.  Can you rebase this, and fix the failing test?

majst01:
Hi @dashpole ,

two years later, i checked the actual implementation and all has changed, now the disk usage is calculated with syscalls, which is good, but my small improvement cannot be applied anymore.

I will check if the same idea can be applied, but unsure, will come back with my findings.

dashpole:
oh, right.  Thanks for reminding me.

PR #1792, opened 2017-11-06
 libcontainer: Use first cgroup subsystem found 

euank:
Typically the first cgroup found is going to be the right one (the
/sys/fs/cgroup mount on most normal machines).

This is only a heuristic to make cadvisor more right more of the time,
but it should be correct in practice to the point where it's okay.

See kubernetes/kubernetes#49835


euank:
Note that I have not manually tested that this resolves the issue I mentioned in reality, but I have a strong suspicion this will.

dashpole:
Would it make sense to log something (at low verbosity) when we skip a subsystem?

euank:
Sure, I added logging for mounts that get skipped.

PR #1795, opened 2017-11-07
Containerd cadvisor integration

abhi:
This commit includes : 
1) changes to integrate containerd runtime to cadvisor to collect container stats.
2) test cases.

cc @Random-Liu


abhi:
I signed it!

Random-Liu:
@dashpole Could you help take a look when you have time? Thanks!

Godeps/Godeps.json
```diff
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/google/cadvisor",
-	"GoVersion": "go1.8",
+	"GoVersion": "go1.9",
```
Random-Liu: Can you use go1.8?

Godeps/Godeps.json
```diff
@@ -610,36 +687,40 @@
 			"Rev": "1e2e08e8a2dcdacaae3f14ac44c5cfa31361f270"
 		},
 		{
-			"ImportPath": "golang.org/x/crypto/ssh/terminal",
-			"Rev": "eb71ad9bd329b5ac0fd0148dd99bd62e8be8e035"
-		},
-		{
 			"ImportPath": "golang.org/x/net/context",
-			"Rev": "5627bad10b821d907ac4dc81fb3535ec7dd81c2b"
+			"Rev": "7dcfb8076726a3fdd9353b6b8a1f1b6be6811bd6"
```
Random-Liu: Why do we need to update `golang.org/x/net`?

Random-Liu:
@abhi 
```
W1108 17:12:00.427] vendor/github.com/sirupsen/logrus/text_formatter.go:13:2: cannot find package "golang.org/x/crypto/ssh/terminal" in any of:
W1108 17:12:00.428] 	/go/src/github.com/google/cadvisor/vendor/golang.org/x/crypto/ssh/terminal (vendor tree)
W1108 17:12:00.428] 	/usr/local/go/src/golang.org/x/crypto/ssh/terminal (from $GOROOT)
W1108 17:12:00.428] 	/go/src/golang.org/x/crypto/ssh/terminal (from $GOPATH)
```

Not all dependencies are included.

Random-Liu:
@abhi The godep change is merged. Please rebase your PR.

The test is passing. I think ready for a review now. :)

abhi:
@Random-Liu Done. ready for review.

container/containerd/client_test.go
```diff
@@ -0,0 +1,53 @@
...
+	returnErr error
+}
+
+func (c *containerdClientMock) LoadContainer(ctx context.Context, id string) (*containers.Container, error) {
+	if c.returnErr != nil {
+		return nil, c.returnErr
+	}
+	cntr, ok := c.cntrs[id]
+	if !ok {
+		return nil, fmt.Errorf("unable to find container %s", id)
```
Random-Liu: nit: s/%s/%q

container/containerd/factory.go
```diff
@@ -0,0 +1,151 @@
...
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/manager/watcher"
+)
+
+var ArgContainerdEndpoint = flag.String("containerd", "unix:///var/run/containerd.sock", "containerd endpoint")
+
+// The namespace under which containerd aliases are unique.
+const (
+	ContainerdNamespace = "containerd"
+	RootFSStorageDir    = "/var/run/containerd/io.containerd.runtime.v1.linux/k8s.io/"
```
Random-Liu: nit: Use `k8sContainerdNamespace`?

container/containerd/grpc.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return invoker(ctx, method, req, reply, cc, opts...)
+}
+
+func (ni namespaceInterceptor) stream(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	_, ok := namespaces.Namespace(ctx)
+	if !ok {
+		ctx = namespaces.WithNamespace(ctx, ni.namespace)
+	}
+
```
Random-Liu: nit: unnecessary empty line.

container/containerd/factory.go
```diff
@@ -0,0 +1,151 @@
...
+		return false, false, nil
+	}
+	// Check if the container is known to containerd and it is running.
+	id := ContainerNameToContainerdID(name)
+	// If container and task lookup in containerd fails then we assume
+	// that the container state is not known to containerd
+	ctx := context.Background()
+	_, err := self.client.LoadContainer(ctx, id)
+	if err != nil {
+		return false, true, fmt.Errorf("failed to load container: %v", err)
```
Random-Liu: Why `canHandle=false` and `canAccept=true`? Don't quite understand. :p 
And should we load task instead, and check whether task is running?

container/containerd/factory_test.go
```diff
@@ -0,0 +1,75 @@
...
+// limitations under the License.
+
+package containerd
+
+import (
+	"testing"
+
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/typeurl"
+	containerlibcontainer "github.com/google/cadvisor/container/libcontainer"
```
Random-Liu: nit: Separate group for `cadvisor`.

integration/runner/runner.go
```diff
@@ -44,7 +44,7 @@ const (
 	testTimeout    = 15 * time.Minute
 )
 
-var cadvisorTimeout = flag.Duration("cadvisor_timeout", 15*time.Second, "Time to wait for cAdvisor to come up on the remote host")
+var cadvisorTimeout = flag.Duration("cadvisor_timeout", 30*time.Second, "Time to wait for cAdvisor to come up on the remote host")
```
Random-Liu: I still a bit concern. Have you tried without containerd running? How long does it take to try connecting containerd?

fs/fs.go
```diff
@@ -90,9 +90,10 @@ type RealFsInfo struct {
 
 type Context struct {
 	// docker root directory.
-	Docker  DockerContext
-	RktPath string
-	Crio    CrioContext
+	Docker     DockerContext
+	RktPath    string
+	Crio       CrioContext
+	Containerd ContainerdContext
```
Random-Liu: Do we need this change? It seems that we are not using it.

container/containerd/handler.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+
+	taskPid, err := client.TaskPid(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	rootfs := "/"
+	if !inHostNamespace {
+		rootfs = RootFSStorageDir + cntr.ID + spec.Root.Path
```
Random-Liu: This doesn't seem right to me. What rootfs do we need here? It seems that we are using this rootfs to get network stats.
However,
1) Why docker/cri-o are using `/rootfs`?
2) There may not be `proc` inside container, right? e.g. an image built from scratch.

container/containerd/handler.go
```diff
@@ -0,0 +1,244 @@
...
+	return container.ContainerTypeContainerd
+}
+
+func (self *containerdContainerHandler) Start() {
+}
+
+func (self *containerdContainerHandler) Cleanup() {
+}
+
+func (self *containerdContainerHandler) GetContainerIPAddress() string {
```
Random-Liu: This is for custom metrics. We need to figure out whether it's ok to not support it.
@dashpole Are we still using custom metrics?

container/containerd/handler.go
```diff
@@ -0,0 +1,244 @@
...
+	// label
+	if !self.ignoreMetrics.Has(container.NetworkUsageMetrics) {
+		//TODO change it to exported cri-containerd constants
+		return self.labels["io.cri-containerd.kind"] == "sandbox"
+	}
+	return false
+}
+
+func (self *containerdContainerHandler) GetSpec() (info.ContainerSpec, error) {
+	hasFilesystem := !self.ignoreMetrics.Has(container.DiskUsageMetrics)
```
Random-Liu: What does this mean? I saw both cri-o and docker have this.
However, because we don't collect image filesystem stats, so just make sure.

container/containerd/handler.go
dashpole: I believe we still support it...  IIRC from my last conversation with the warsaw folks, we probably wont support this model forever, but I believe we still support it ATM

dashpole:
FYI, I plan to cut cadvisor v0.28.1, which will correspond to k8s v1.9.0 this Friday, 11/17.

container/containerd/handler.go
abhi: if its in non host namespace then proc is located at ```/var/run/containerd/io.containerd.runtime.v1.linux/k8s.io/<containerid>/rootfs```. This from what I verified during coding. Am I wrong here ?

integration/runner/runner.go
abhi: I am going to get back on this. Running on a single machine doesnt seem to be an issue . Its only when the CI runs ssh and runs it that the timeout needs to be increased. Will come back with a valid explaination

abhi:
@dashpole thanks for the heads up.

container/containerd/handler.go
Random-Liu: @dashpole I don't think we could get ip of docker container in cadvisor, when we are using kubenet. So do we really support this in GKE?

container/containerd/handler.go
Random-Liu: TBH, I don't know.
The pid we get is host pid, right? Why don't we just access `/proc/pid/net/ns`?

Random-Liu: I believe the `inHostNameSpace` here, means whether cadvisor is running in host namespace or inside container.

When cadvisor is running inside a container, we mount host root filesystem to `/rootfs`.

Random-Liu: So we should have same code with docker, and remove the container rootfs related logic.

Random-Liu: See [here](https://github.com/google/cadvisor/blob/master/manager/manager.go#L191-L196).

container/containerd/client.go
```diff
@@ -0,0 +1,123 @@
...
+	return containerFromProto(r.Container), nil
+}
+
+func (c *client) TaskPid(ctx context.Context, id string) (uint32, error) {
+	response, err := c.taskService.Get(ctx, &tasksapi.GetRequest{
+		ContainerID: id,
+	})
+	if err != nil {
+		err = errdefs.FromGRPC(err)
+		if errdefs.IsNotFound(err) {
```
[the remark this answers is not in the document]
abhi: Done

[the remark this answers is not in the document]
Random-Liu: Still. Don't understand this one.

container/containerd/factory.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/google/cadvisor/container/libcontainer"
+	"github.com/google/cadvisor/fs"
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/manager/watcher"
+)
+
+var ArgContainerdEndpoint = flag.String("containerd", "unix:///var/run/containerd.sock", "containerd endpoint")
+
+// The namespace under which containerd aliases are unique.
+const (
```
Random-Liu: Remove brackets.

container/containerd/factory.go
```diff
@@ -0,0 +1,150 @@
...
+	// if the container is not associated with containerd, we can't handle it or accept it.
+	if !isContainerName(name) {
+		return false, false, nil
+	}
+	// Check if the container is known to containerd and it is running.
+	id := ContainerNameToContainerdID(name)
+	// If container and task lookup in containerd fails then we assume
+	// that the container state is not known to containerd
+	ctx := context.Background()
+	_, err := self.client.LoadContainer(ctx, id)
```
Random-Liu: Check running?

fs/fs.go
```diff
@@ -105,6 +105,10 @@ type CrioContext struct {
 	Root string
 }
 
+type ContainerdContext struct {
```
Random-Liu: Never used?

Random-Liu:
@abhi LGTM other than `hasFilesystem`. Because I'm not quite sure what that means. :)

container/containerd/handler.go
Random-Liu: Thanks @dashpole @DirectXMan12!
Confirmed that this doesn't work anymore. We don't need to support it.

container/containerd/handler.go
dashpole: do you collect container disk stats?  If not, you should just set this to false.  I believe this is meant to indicate whether or not this container has disk stats.

dashpole:
It doesnt need to happen now, but eventually it would be nice to have disk stats for containerd through cAdvisor.  k8s would disable disk stats, but it would mean that standalone cAdvisor could still be used to monitor docker after docker rebases on top of containerd.

This PR lgtm other than the one nit.

abhi:
@dashpole @Random-Liu thanks for reviewing. I have addressed the comments and rebased. 

@dashpole : also current docker master already uses containerd 1.0 beta. But we would still use the docker apis to collect stats for standalone cadvisor case ?

container/containerd/client.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Client creates a containerd client
+func Client() (containerdClient, error) {
+	gopts := []grpc.DialOption{
+		grpc.WithInsecure(),
+		grpc.FailOnNonTempDialError(true),
+		grpc.WithDialer(dialer.Dialer),
+		grpc.WithBlock(),
+		grpc.WithTimeout(2 * time.Second),
+		grpc.WithBackoffMaxDelay(3 * time.Second),
```
Random-Liu: Why do you add this back?

Random-Liu:
> It doesnt need to happen now, but eventually it would be nice to have disk stats for containerd through cAdvisor. k8s would disable disk stats, but it would mean that standalone cAdvisor could still be used to monitor docker after docker rebases on top of containerd.

@dashpole When docker move to containerd 1.0 image management, I guess the image fs stats in the current docker integration won't work. We need to either change cadvisor to understand the new disk layout, or let cadvisor get fs stats from underlying containerd.

@abhi I see that we hard code containerd unix socket path now. In the future, after docker moves to the new containerd, if it is using the same socket path, current cadvisor code might get confused.
The current containerd integration is mainly for `k8s.io` containerd namespace, so the containerd namespace may help us avoid trouble. However, we should keep an eye on that.

container/containerd/client.go
abhi: to avoid the retries by the dialer. So I set the timeout and use withblock and backoffdelay.

abhi:
@Random-Liu yes thats right. We are currently covered by the namespace seperation. We can revisit for docker+containerd support

PR #1799, opened 2017-11-09
Updating grpc version to v1.3.0

abhi:
This commit includes godep change to update the grpc version and also updates rkt version to v1.25.0 in sync with kubernetes/kubernetes. A minor change has been made in the client based on how rkt client is used in kubernetes/kubernetes.

cc @Random-Liu 
Signed-off-by: abhi <abhi@docker.com>


dashpole:
Can you explain the change in the rkt handler a bit more?  I would prefer just removing the code instead of commenting it out.

abhi:
@dashpole yes the intention is to remove the code. It appears that the newer version of rkt doesnt require the tcp dialer.  I was relying on the way rkt client is being used in kubernetes/kubernetes.  I was having trouble running the test-runner locally and had to rely on CI to confirm the change. If the rest of the PR looks good , I can update the PR

Random-Liu:
@dashpole @abhi has removed the rkt comments.

PR #1801, opened 2017-11-10
Update gonvml to workaround bazelbuild/rules_go#1003.

mindprince:


mindprince:
Without this bazel tests were failing in https://github.com/kubernetes/kubernetes/pull/55188.

/cc @dashpole

dashpole:
Would you like me to merge this, or do you want another set of eyes for the nvml changes?

mindprince:
I think they are good to be merged. But cc'ing @vishh and @jiayingz just in case.

PR #1642, opened 2017-04-19
Report container FS metrics into prometheus /metrics

smarterclayton:
Fixes #1403

PerDiskStats reported from cgroups were not being surfaced into
prometheus. In order to properly correlate the metrics, we need to
assign a device label to each metric (which is the FS or device path).
Since blkio cgroup tracks devices, we create a synthetic device
`/dev/NAME` for the metric.

Assign a Device label to each PerDiskStat for the handlers up front, and
then surface the PerDiskStat values into the prometheus metrics. Report
two new metrics - total bytes read and total bytes written.

Open questions:

- Does this add too much extra data (are there metrics we would drop?)
- Do I have the correct mapping between blkio and the Filesystem stats (at casual reading yes, but not positive)
- Do we want to add the `_bytes_total` metrics?
- Is there an equivalent to weighted_time in blkio (I don't think so, but could be wrong)
- Did this regress performance for cadvisor gathering (device assignment is fairly cheap, but...)

Issue #1450, opened 2016-09-07
Can't get any diskio data from cadvisor API if docker use overlayFS 


workhardcc:
Seems if docker use overlayFS for its storage cadvisor can not get diskio data from the API:
`"diskio": {},`
I found that there is no data in /sys/fs/cgroup/blkio/docker/$CONTAINER_ID/blkio.throttle.io_service_bytes either.

fzu-huang:
use `dd` to make some disk IO.Then check it again.    I met this situation before..

workhardcc:
@fzu-huang You're right. `dd if=/dev/zero of=/home/cc bs=1M count=3000` makes it work. 
I have a another question: why I `dd` some data in container's disk  but there is no change in blkio.throttle.io_service_bytes

derekwaynecarr:
@workhardcc -- i cant tell if you resolved your own issue?

workhardcc:
@derekwaynecarr sorry for late reply. I didn't resolve this problem, now I'm almost give up get container's diskio metrics.

dashpole:
closing via #1642.

PR #1805, opened 2017-11-15
marshal device name to json output

andyxning:
This PR will make `device` available when json marshal disk stats as we will use device name as a label in heapster when adding disk io metrics in [heapster #1450](https://github.com/kubernetes/heapster/pull/1450)

Since @smarterclayton added `device` to cadivsor in #1642 , so could you please take a look at this.

/cc @dashpole @derekwaynecarr

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -307,7 +307,7 @@ type CpuStats struct {
 }
 
 type PerDiskStats struct {
-	Device string            `json:"-"`
+	Device string            `json:"device"`
 	Major  uint64            `json:"major"`
 	Minor  uint64            `json:"minor"`
 	Stats  map[string]uint64 `json:"stats"`


dashpole:
heapster collects disk-io stats?  Does it use cadvisor directly?

andyxning:
@dashpole Actually, because kubelet has added cAdvisor as a lib, heapster can retrieve cAdvisor `stats` endpoint to access cpu/memory/disk io data and emit metrics to different backends.

andyxning:
Even kubelet and heapster can use kubelet `summary` endpoint, the data for `summary` endpoint, is also retrieved from cAdvisor.

dashpole:
right, but does heapster consume metrics from cAdvisor's APIs?  I was under the impression that it only consumed metrics from the summary API

andyxning:
@dashpole Yes. Even summary api need this device name when adding disk io metrics.

andyxning:
@dashpole FYI: https://github.com/kubernetes/heapster/blob/master/metrics/sources/kubelet/kubelet_client.go#L134-L148

dashpole:
IIUC, this wont affect the summary API, as we dont marshal/unmarshal when the kubelet consumes this. 
So this change is purely to allow heapster to consume this, which is fine.

andyxning:
@dashpole Thanks for pointing this and correct my wrong understanding. :)

PR #1807, opened 2017-11-17
Revert "fix #1708; move from inotify to fsnotify"

### LINKED ISSUE #1708: Migrate off the deprecated package `golang.org/x/exp/inotify`
`golang.org/x/exp/inotify` has been deleted (see https://github.com/golang/exp/tree/master/inotify)
cadvisor should switch to using fsnotify, or something equivalent.

xref: https://github.com/kubernetes/kubernetes/issues/40123

dashpole:
This reverts commit e6b6a1ac5796804cd1a42270b2f0b0f1a0aa8a79.
issue: https://github.com/kubernetes/kubernetes/issues/55909


tallclair:
/lgtm

What's the issue? Are we going to be able to fix it and switch back to fsnotiy?

dashpole:
Not actually sure.  Presumably we can fix it, but I'm not sure I have time to fix it in time for the release

khenidak:
@dashpole I think https://github.com/kubernetes/kubernetes/issues/58081 is caused by `Close()` on golang's inotify is not closing the watch fd. Any chance that we move back to fsnotify? which seems to correctly close it https://github.com/fsnotify/fsnotify/blob/master/inotify.go#L183

dashpole:
@khenidak, I would love to move back to fsnotify, but I havent found time to do so.  If you know what the issue was, and can fix it, I would be happy to review a PR.

khenidak:
Thanks :-) didn't mean to "FIX IT!" - i was just saying.

More than happy to help. Do you want to move back to fsnotify or fix the existing inotify deprecated package?

dashpole:
Yes, the goal is to move to fsnotify.

Issue #1810, opened 2017-11-20
There is a Race between Setup and GetCollector in nvidia.go


choury:
`GetCollector` will failed when  `nm.nvmlInitialized` are `true` but `initializeNVML` hasn't finished.
I can open a PR to fix it if you think this is a bug.

dashpole:
cc @mindprince

mindprince:
@choury You are right. I sent a quick PR to fix this: https://github.com/google/cadvisor/pull/1814

Thanks for finding this out.

@dashpole Do you think this fix can make the vendoring cut for 1.9?

PR #1814, opened 2017-11-21
Avoid race in accessing nvidiaDevices between Setup() and GetCollector()

mindprince:
Fixes https://github.com/google/cadvisor/issues/1810

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	info "github.com/google/cadvisor/info/v1"
@@ -30,6 +31,8 @@ import (
 )
 
 type NvidiaManager struct {
+	sync.RWMutex
+
 	// true if the NVML library (libnvidia-ml.so.1) was loaded successfully
 	nvmlInitialized bool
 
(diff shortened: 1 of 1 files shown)


mindprince:
This PR is just fixing the case where we could have a concurrent map access.

The time taken by `initializeNVML()` on  my desktop is in ms but on some cloud environments it can take up to 5s. That's why I didn't put it in the critical path. We can see how big of an issue this is and change that in the future. My guess is that it's unlikely that there are containers with GPUs attached when the kubelet starts up but I maybe wrong and we can change that in the future.

choury:
> it's unlikely that there are containers with GPUs attached when the kubelet starts up

We should not require to restart all containers if we just want to restart or upgrade kubelet.
May be call  GetCollector after NVML initialized or wait it in GetCollector?

mindprince:
> We should not require to restart all containers if we just want to restart or upgrade kubelet.

Agreed. Good point.

> May be call GetCollector after NVML initialized or wait it in GetCollector?

Not sure what you mean exactly. Can you elaborate?

dashpole:
> My guess is that it's unlikely that there are containers with GPUs attached when the kubelet starts up but I maybe wrong and we can change that in the future.

will this work across kubelet restarts then?

vishh:
LGTM

mindprince:
Discussed with @vishh and made changes to work better in case of kubelet restarts.

vishh:
LGTM again. @dashpole WDYT?

accelerators/nvidia.go
```diff
@@ -129,7 +139,9 @@ func (nm *NvidiaManager) Destroy() {
 // present in the devices.list file in the given devicesCgroupPath.
 func (nm *NvidiaManager) GetCollector(devicesCgroupPath string) (AcceleratorCollector, error) {
 	nc := &NvidiaCollector{}
+	nm.RLock()
 	if !nm.nvmlInitialized || len(nm.nvidiaDevices) == 0 {
+		nm.RUnlock()
```
dashpole: do we want to defer unlock here?  What if we do not enter into this if statement?

mindprince: Updated. Thanks for noticing this! (facepalm)

PR #1817, opened 2017-11-21
Switch from apimachinery clock to k8s.io/utils/clock

dashpole:
this is to allow vendoring of cadvisor into kubernetes/kubernetes

cc @tallclair


dashpole:
going to rebase this on top of https://github.com/kubernetes/utils/pull/25

dashpole:
rebase finished

PR #1827, opened 2017-11-28
Clean up cAdvisor logging

tallclair:
I got annoyed by the misleading "failed to connect to rkt" type warnings, and that lead me down a rathole of cleaning up all of cAdvisors logging, using the guidelines [here](https://github.com/kubernetes/community/blob/master/contributors/devel/logging.md).


dashpole:
great work!

PR #1830, opened 2017-11-30
Add timeouts for docker queries

jsravn:
As these can otherwise block indefinitely due to docker issues.

It would be better if these methods took a context, so the client could
specify timeouts. To preserve API compatibility, I've just added an
internal timeout of 5s to all the calls. In my testing, this is plenty
of time even for slower queries (like image lists, which should take <1s
even with thousands of images).

This is to fix https://github.com/kubernetes/kubernetes/issues/53207,
where kubelet relies on cadvisor for gathering docker information as
part of its periodic node status update.

### DIFF
--- a/container/docker/docker.go
+++ b/container/docker/docker.go
@@ -23,26 +23,33 @@ import (
 	dockertypes "github.com/docker/docker/api/types"
 	"golang.org/x/net/context"
 
+	"time"
+
 	"github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/machine"
 )
+
+const defaultTimeout = time.Second * 5
+
+func defaultContext() context.Context {
+	ctx, _ := context.WithTimeout(context.Background(), defaultTimeout)
+	return ctx
+}
 
 func Status() (v1.DockerStatus, error) {
 	client, err := Client()
 	if err != nil {
 		return v1.DockerStatus{}, fmt.Errorf("unable to communicate with docker daemon: %v", err)
 	}
-	dockerInfo, err := client.Info(context.Background())
+	dockerInfo, err := client.Info(defaultContext())
 	if err != nil {
 		return v1.DockerStatus{}, err
 	}
-	return StatusFromDockerInfo(dockerInfo), nil
+	return StatusFromDockerInfo(dockerInfo)
 }
 
-func StatusFromDockerInfo(dockerInfo dockertypes.Info) v1.DockerStatus {
+func StatusFromDockerInfo(dockerInfo dockertypes.Info) (v1.DockerStatus, error) {
 	out := v1.DockerStatus{}
-	out.Version = VersionString()
-	out.APIVersion = APIVersionString()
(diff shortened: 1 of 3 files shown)


dashpole:
Just to make sure I understand correctly, cadivisor would return "Unknown" for anything that it times out, right?
Can we make "Unknown" an exported constant so that we can handle it in the kubelet?  It may make sense for the kubelet to use the previous value for something if it recieves an "Unknown" response.

jsravn:
@dashpole That's right - which is why I added the warn log. We can export it, sure. Even better would be to return the error.

jsravn:
I've updated it to return the `error`, then client code can handle it appropriately.

container/docker/factory.go
```diff
@@ -340,7 +340,10 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo, ignoreMetrics c
 			glog.Errorf("devicemapper filesystem stats will not be reported: %v", err)
 		}
 
-		status := StatusFromDockerInfo(*dockerInfo)
+		status, err := StatusFromDockerInfo(*dockerInfo)
```
dashpole: This actually seems concerning, as we only call `Register` once.  If it times out the first time, then you get no disk stats on DM.

dashpole: Although I suppose it was still possible for it to fail before.  Now it is just more likely since we have added a timeout.

container/docker/docker.go
```diff
@@ -23,26 +23,43 @@ import (
...
 }
 
-func StatusFromDockerInfo(dockerInfo dockertypes.Info) v1.DockerStatus {
+func StatusFromDockerInfo(dockerInfo dockertypes.Info) (v1.DockerStatus, error) {
 	out := v1.DockerStatus{}
-	out.Version = VersionString()
-	out.APIVersion = APIVersionString()
+	ver, err := VersionString()
+	if err != nil {
+		return out, err
```
dashpole: can we move these to the end?  Some callers may wish for partial results, in which case we wouldn't want to skip the assignments below.

container/docker/factory.go
dashpole: nvm.  Can you change this to ignore the error?  It should never fail to populate `status.DriverStatus` if you reorder the operations above as I requested.

jsravn:
@dashpole Okay, addressed your comments.

dashpole:
lgtm.  Squash, and Ill get this merged.

jsravn:
@dashpole Squashed.

jsravn:
ping @dashpole

jsravn:
@dashpole If you cut a release then I can try to get this updated in Kubernetes. Thanks!

dashpole:
Ill cut another release

jsravn:
Can you reopen https://github.com/kubernetes/kubernetes/issues/53207? Github closed it.. doh.

dashpole:
done

PR #1832, opened 2017-12-01
Add context to docker API to timeout requests

jsravn:
Docker requests can hang sometimes. Add a context so clients of the
docker API can timeout the requests and retry, or some other action.

To fix https://github.com/kubernetes/kubernetes/issues/53207


jsravn:
As alternative to https://github.com/google/cadvisor/pull/1830. Will require updating k8s for the api changes.

jsravn:
Hmm, I think it's better to just return the error, to keep the api simpler. I've done that in my prior PR https://github.com/google/cadvisor/pull/1830.

PR #1847, opened 2017-12-27
Skip getVfsStats when file does not exist

dims:
There are a lot of spurious exceptions in the kubernetes kubelet logs
like:
E1018 21:03:09.616581   22780 fs.go:332] Stat fs failed. Error: no such
file or directory

Since we know that calling syscall.Statfs will just fail when the path
does not exist, we should just skip making the call.

Fix for https://github.com/kubernetes/kubernetes/issues/35062

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -35,6 +35,7 @@ import (
 	"github.com/docker/docker/pkg/mount"
 	"github.com/golang/glog"
 	"github.com/google/cadvisor/devicemapper"
+	"github.com/google/cadvisor/utils"
 	dockerutil "github.com/google/cadvisor/utils/docker"
 	zfs "github.com/mistifyio/go-zfs"
 )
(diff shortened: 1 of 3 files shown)


pineking:
@dims when this is merged, which version of k8s can be fixed, v1.9.1 or master branch of k8s?

dashpole:
Can we add a `glog.V(4).Infof` message specifying which mountpoint was not found if `utils.FileExists` returns false?  This will allow us to find issues during testing.

dims:
@dashpole yep, i can do that

dims:
@dashpole something wrong with the build? local make looks file (http://paste.openstack.org/show/636661/)

dashpole:
looks like it has been failing for a few days: https://k8s-testgrid.appspot.com/sig-node-cadvisor#cadvisor-kubelet.  Not sure why.

dims:
Thanks @dashpole

dashpole:
Can you regenerate the assets?  I think the change from 2017->2018 made the tests fail (happened in k/k as well :)
Running `./build/jenkins_e2e.sh` seems to do the trick

dashpole:
Classic new year's bugs

dims:
:)

PR #1851, opened 2018-01-04
auto refresh fsInfo when capture mount event

fzu-huang:
when we attach a disk and mount to a dir during cadvisor running, it cannot check out and update it self, so we cannot get the stats of new disk (such as `/dev/vdc` )  when we called `api/v2.1/machine` or `api/v2.1/machinestats`.  this pr is going to solve it.


fzu-huang:
@dashpole  it blocked here, what should I do for this pr?

dashpole:
I can take a look.  From my first glance, I would rather not add a dependency to do the detection of mount changes.  For something that simple, we should just add it to cAdvisor.  Also it looks like you didnt use Godep to add the dependency.

PR #1852, opened 2018-01-05
 Add Prometheus total CPU metric and allow per-cpu stats to be disabled

### LINKED ISSUE #1513: allow to disable "container_cpu_usage_seconds_total" metric
I'm running cAdvisor on a 128 CPU core machine with nearly 200 containers. This metric is killing cAdvisor and prometheus, which collects the data. Some statistics about the "/metrics" endpoint:

```
$ ls -sh metrics.txt
19M metrics.txt
```

```
$ wc -l metrics.txt 
64635 metrics.txt
```

```
$ grep "container_cpu_usage_seconds_total" metrics.txt| wc -l
47362
```

So this single metric takes over 70% of the 19MB response. I think it must be possible to disable it.

bboreham:
Fixes #1513 

I'm a little uneasy that disabling the per-cpu stats means we use a different source for the data.
I could have it still go through the loop, but is that necessary?

cc @dashpole

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -150,10 +150,18 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
 				},
 			}, {
 				name:        "container_cpu_usage_seconds_total",
-				help:        "Cumulative cpu time consumed per cpu in seconds.",
+				help:        "Cumulative cpu time consumed in seconds.",
 				valueType:   prometheus.CounterValue,
 				extraLabels: []string{"cpu"},
 				getValues: func(s *info.ContainerStats) metricValues {
+					if len(s.Cpu.Usage.PerCpu) == 0 {
+						if s.Cpu.Usage.Total > 0 {
+							return metricValues{{
+								value:  float64(s.Cpu.Usage.Total) / float64(time.Second),
+								labels: []string{"total"},
+							}}
+						}
+					}
 					values := make(metricValues, 0, len(s.Cpu.Usage.PerCpu))
 					for i, value := range s.Cpu.Usage.PerCpu {
 						if value > 0 {
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -72,6 +72,7 @@ var (
 		container.NetworkUsageMetrics:    struct{}{},
 		container.NetworkTcpUsageMetrics: struct{}{},
 		container.NetworkUdpUsageMetrics: struct{}{},
+		container.PerCpuUsageMetrics:     struct{}{},
 	}
 )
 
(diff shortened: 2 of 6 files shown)


dashpole:
I think we should collect the total usage the same way regardless of whether we are ignoring per-cpu stats.  If we are ignoring them, just skip that portion.

I would be interested to see how the CpuStats.Total compares with summing the per-cpu stats.

dashpole:
looking at the libcontainer implementation, it reads from `cpuacct.usage` to populate `stats.CpuStats.CpuUsage.TotalUsage`.  It reads from `cpuacct.usage_percpu` to populate `stats.CpuStats.CpuUsage.PercpuUsage`.  I think using CpuStats.Total to populate our total cpu usage makes more sense.  We should use that regardless of whether or not we are collecting per-cpu usage.

bboreham:
oops, just changed it the other way.  hold on...

dashpole:
Can you do a sanity check to see if it gives approximately correct values?  You can use the [stress container](https://github.com/vishh/stress), by using image `k8s.gcr.io/stress:v1` to test it, and run it with args: `--cpus=X`, and see if it uses the correct quantity of CPU.

bboreham:
Sure, will get to that later.

It appears there is a test which checks that CPU usage is computed by adding.

bboreham:
I have fixed the failing test, and changed my mind about the metric name.

I believe most people are currently using queries like

    sum (rate(container_cpu_usage_seconds_total{}[5m])) by (some_criteria)

If we change the name then everybody has to go round and edit all those queries.
Instead, if we keep the same name and just switch from `cpu01`, `cpu02`, etc., to `total` under the `cpu` label if `percpu` is disabled, then everything will Just Work.

Still haven't done the sanity check.

metrics/prometheus.go
```diff
@@ -150,10 +150,16 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
 				},
 			}, {
 				name:        "container_cpu_usage_seconds_total",
-				help:        "Cumulative cpu time consumed per cpu in seconds.",
+				help:        "Cumulative cpu time consumed in seconds.",
```
dashpole: IIUC, users either get total OR per-cpu?  Why not just add a new metric for total?

bboreham: See previous comment.

dashpole: Ok, I agree with this.

dashpole:
Oh, and please also squash your commits.

dashpole:
@bboreham have you done your sanity check?  I would like to get this in the next release.

bboreham:
Thanks for the reminder - I got sidetracked onto 50 other things. Will squash and re-check tomorrow.

bboreham:
OK, I have run the `k8s.gcr.io/stress:v1` container a few times with different parameters, while running my build of cAdvisor from this branch, and checked the CPU metrics tallied with expectations.

I had forgotten to check `>0` before returning a CPU value so added that, for the percpu-disabled case.

And squashed commits as requested.

PR #1859, opened 2018-01-11
reduce labels for container info

andyxning:
Currently there are two `labels` fields in `ContainerInfo`. One is in `ContainerReference` and the other is in `ContainerSpec`. By leaving only one `labels`, it is should be good to reduce the output when requesting container info api.

/cc @dashpole @tallclair

### DIFF
--- a/cache/memory/memory.go
+++ b/cache/memory/memory.go
@@ -70,16 +70,16 @@ type InMemoryCache struct {
 	backend           storage.StorageDriver
 }
 
-func (self *InMemoryCache) AddStats(ref info.ContainerReference, stats *info.ContainerStats) error {
+func (self *InMemoryCache) AddStats(cInfo *info.ContainerInfo, stats *info.ContainerStats) error {
 	var cstore *containerCache
 	var ok bool
 
 	func() {
 		self.lock.Lock()
 		defer self.lock.Unlock()
-		if cstore, ok = self.containerCacheMap[ref.Name]; !ok {
-			cstore = newContainerStore(ref, self.maxAge)
-			self.containerCacheMap[ref.Name] = cstore
+		if cstore, ok = self.containerCacheMap[cInfo.ContainerReference.Name]; !ok {
+			cstore = newContainerStore(cInfo.ContainerReference, self.maxAge)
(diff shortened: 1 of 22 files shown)


dashpole:
@tallclair do you think this is a reasonable change?  It is technically backwards-incompatible, but just requires users to use a different field.
@andyxning I think labels make more sense as part of the reference, rather than the spec.  In kubernetes, [labels are part of metadata](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L210), rather than spec.

andyxning:
@dashpole 

> I think labels make more sense as part of the reference, rather than the spec. In kubernetes, labels are part of metadata, rather than spec.

Yes. I have thought about this before making this PR. And it is true that this is an backwards-incompatible change. What i prefer spec instead of reference is that heapster actually use spec to access labels. This is my selfishness just not to change heapster and labels in spec has been first added and it maybe used widely than the labels in reference.

dashpole:
@andyxning that makes sense to me.  We can keep it in the spec for now.  If we ever create a v3 api, which is unlikely to be soon, we can include labels in the reference.
I am still hoping to get a second opinion from @tallclair on whether the backwards-incompatible change is acceptable.

andyxning:
@dashpole Thanks for the review. @tallclair PTAL.

andyxning:
Ping @tallclair in case you are swallowed in bunch of notifications. :)

andyxning:
@dashpole Could we merge this as seems @tallclair is not available recently. :) 
Or should we ping someone else to give this another review?

dashpole:
I talked with @tallclair, and we think the best path is to put a deprecation notice in for this coming release (which isnt too far off).  Then, after that, we can merge this and the release after that will not have the field anymore.

andyxning:
> we think the best path is to put a deprecation notice in for this coming release (which isnt too far off)

This seems good to me. BTW, can you please add a release label or milestone label for this PR in case we forget it. :) I have no auth to do so for now.

dashpole:
added a milestone so we do not forget.

andyxning:
@dashpole Can we merge this for now as the v0.29.0 release for kubernetes v1.10 has been released, we can move on for next release of cadvisor. :)

PR #1862, opened 2018-01-12
Drop container vm from testing.

dashpole:
We are migrating the vms from the jenkins project to a new project, since jenkins is not being used anymore.
To make this slightly easier, remove container vm, which we aren't testing against anymore.

cc @tallclair


dashpole:
The failure doesnt look related to RHEL.

dashpole:
superceeded by #1868

PR #1870, opened 2018-01-18
auto fresh fsInfo when some storage device mount or umount

fzu-huang:


dashpole:
/ok-to-test

fzu-huang:
@dashpole  3q~  have rebased.

Issue #1866, opened 2018-01-16
Missing aliases from spec section


sabras75:
On some of our computer (of course, the production ones :( ), cadvisor does not returns a full spec of the container: it is missing the aliases, namespace and labels. 

For instance, when calling the cAdvisor api 
> http GET http://10.132.48.8:8080/api/v2.1/stats/docker/e75fc4a8077cff7c1060f92a32827be5dda25fc793f15566bedfa7a2c82a5bf2)

it returns :
`{
  "/docker/e75fc4a8077cff7c1060f92a32827be5dda25fc793f15566bedfa7a2c82a5bf2": {
    "spec": {
      "creation_time": "2018-01-16T10:18:03.546818282Z",
      "has_cpu": true,
      "cpu": {
        "limit": 1024,
        "max_limit": 0,
        "mask": "0-15"
      },
      "has_memory": true,
      "memory": {
        "limit": 9223372036854772000,
        "reservation": 9223372036854772000
      },
      "has_custom_metrics": false,
      "has_network": false,
      "has_filesystem": false,
      "has_diskio": true
    },
    "stats": [
      {
...
`

Calling docker inspect on the same container 
> docker inspect e75fc4a8077cff7c1060f92a32827be5dda25fc793f15566bedfa7a2c82a5bf2 

returns (see following attachment) : [docker-inspect.txt](https://github.com/google/cadvisor/files/1634835/docker-inspect.txt)

I was expecting to have a namespace 'docker' and at least an alias equivalent to the name returned by docker inspect (`"Name": "/lifyqalmomsgr9ngzj_cadvisor_1"`) as this is the behaviour found on our development computers.

Interestingly, on the UI when reaching the route "http://localhost:8080/docker/", it returns "failed to get docker info: context deadline exceeded", so we suspect some kind of timeout but are not able to diagnose those with precision and hopefully find a solution.

We would greatly appreciate any help in solving this issue.

thanks,

-Sebastien

> Docker version 17.12.0-ce
> cadvisor runnign in docker with image: google/cadvisor:v0.28.3

dashpole:
Does this issue happen for short periods of time, or is it permanent?  Can you check the cAdvisor logs to see if it is logging any error messages related to connecting to docker?

sabras75:
This is permanent. 
I'll check the logs and post it in a future comment.

sabras75:
here are the logs. I raise the verbosity to get something that I hope to be interesting (--v=42)
[...]

PR #1868, opened 2018-01-17
Migrate to prow

dashpole:
Implements changes in cAdvisor to use:
https://github.com/kubernetes/test-infra/pull/6302

cc @krzyzacy

PR #1871, opened 2018-01-18
Connect to docker with no timeout on startup

jsravn:
For https://github.com/google/cadvisor/issues/1866.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -50,6 +50,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
+	"golang.org/x/net/context"
 	"k8s.io/utils/clock"
 )
 
@@ -58,6 +59,8 @@ var logCadvisorUsage = flag.Bool("log_cadvisor_usage", false, "Whether to log th
 var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h", "Max length of time for which to store events (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is a duration. Default is applied to all non-specified event types")
 var eventStorageEventLimit = flag.String("event_storage_event_limit", "default=100000", "Max number of events to store (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is an integer. Default is applied to all non-specified event types")
 var applicationMetricsCountLimit = flag.Int("application_metrics_count_limit", 100, "Max number of application metrics to store (per container)")
+
+const dockerClientTimeout = 10 * time.Second
 
 // The Manager interface defines operations for starting a manager and getting
 // container and machine information.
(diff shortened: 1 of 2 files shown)


dashpole:
cadvisor tests are not working right now.  We are in the process of migrating them off of jenkins.

tn-osimis:
I could be wrong but this might mean that if the engine never replies for some reason cadvisor will hang indefinitely without giving a clue as to what it's trying to do in the logs.

It's a very welcome fix for #1866 and should probably be merged; though as a long-term approach I suggest a higher default timeout at startup (e.g. 30s or 60s) as well as the addition of a setting for configuring a non-default value.

(Otherwise, maybe at least add log output *before* attempting to connect so that ops can more easily figure out what's blocking the startup process in this scenario.)

WDYT?

jsravn:
@tn-osimis yes, I agree. It is a more systemic problem in cadvisor though, it doesn't use timeouts for anything. So it should really be fixed throughout. I'd prefer using a larger timeout than indefinite timeout on startup.

dashpole:
@jsravn I agree with @tn-osimis that just using a longer timeout would be preferable.  What I meant in the issue by "infinite retry", is actually using a call with a long timeout (e.g. 30 sec), but retrying the call if the error is a timeout.  This way, as soon as docker recovers, cAdvisor starts correctly.  If docker is permanently hung, we have bigger problems than missing metrics.  WDYT?

tn-osimis:
If there is a generic library implementing an exponential backoff algorithm (or whatever else) somewhere in your stack already, that's probably what you want (and the logical conclusion to your thought). Otherwise indeed a basic fixed-interval retry approach usually works better than just waiting indefinitely for cases where the state machine of the peer (i.e. Docker engine) is a little messed up.

FWIW (opinion of a random guy on the Internet) I agree with retrying indefinitely (as long as attempts are logged).

jsravn:
Okay, I can make that change 👍

dashpole:
This will have to wait for #1868 before tests will pass.

jsravn:
Okay, have a look.

manager/manager.go
```diff
@@ -58,6 +59,7 @@ var logCadvisorUsage = flag.Bool("log_cadvisor_usage", false, "Whether to log th
 var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h", "Max length of time for which to store events (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is a duration. Default is applied to all non-specified event types")
 var eventStorageEventLimit = flag.String("event_storage_event_limit", "default=100000", "Max number of events to store (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is an integer. Default is applied to all non-specified event types")
 var applicationMetricsCountLimit = flag.Int("application_metrics_count_limit", 100, "Max number of application metrics to store (per container)")
+var dockerClientTimeout = flag.Duration("docker_client_timeout", 10*time.Second, "Timeout to the docker daemon")
```
dashpole: I would rather avoid adding a flag right now.  I think the retry mechanism is enough to fix the issue, and I don't think there is a meaningful decision for users to make here.

manager/manager.go
```diff
@@ -234,6 +235,31 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, maxHousekeepingIn
...
+	for {
+		ctx, _ := context.WithTimeout(context.Background(), startupTimeout)
+		dockerStatus, err := docker.StatusWithContext(ctx)
+		if err != nil {
+			return dockerStatus
+		}
+
+		switch err {
+		case context.DeadlineExceeded:
+			glog.Warningf("Timeout trying to communicate with docker, will retry")
```
dashpole: nit: add "during initialization" so we know this is delaying startup

jsravn:
@dashpole Addressed your comments.

dashpole:
/ok-to-test

jsravn:
@dashpole Okay, rebased.

Issue #1865, opened 2018-01-15
Adding /proc/<pid>/schedstat


nielsole:
I was looking into measuring whether a container is running into CPU performance issues on its host.
While the `cpu.stats.throttled_time`/`container_cpu_cfs_throttled_seconds_total` is helpful, it does only reflect when the scheduler has explicitly throttled the process. What I was looking for was a metric that shows, whether a process was waiting for CPU time or not (kind of blocked by CPU). 

The kernel measures this metric, but cAdvisor does not collect it. It apparently was already included back in 2014 https://github.com/google/cadvisor/commit/324874411d0d4ad6f0b76fa46af8e53696fed866 

Is there a reason this is not included in cAdvisor?

From the kernel docs: https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt

    /proc/<pid>/schedstat
    ----------------
    schedstats also adds a new /proc/<pid>/schedstat file to include some of
    the same information on a per-process level.  There are three fields in
    this file correlating for that process to:
     1) time spent on the cpu
     2) time spent waiting on a runqueue
     3) # of timeslices run on this cpu

    A program could be easily written to make use of these extra fields to
    report on how well a particular process or set of processes is faring
    under the scheduler's policies.  A simple version of such a program is
    available at
    http://eaglet.rain.com/rick/linux/schedstat/v12/latency.c

hanikesn:
Doing some further investigation it looks like https://github.com/google/cadvisor/blob/master/info/v1/container.go#L256 could give us cpu saturation information, though it won't be as accurate as blocked time is not accumulated like with `time spent waiting on a runqueue`.

dashpole:
Yeah, it looks like it was exposed once upon a time: https://github.com/google/cadvisor/pull/142, but then removed as https://github.com/google/cadvisor/issues/1865#issuecomment-357943893 points out.  According to https://github.com/google/cadvisor/pull/431, it looks like some of the metrics were not accurate.

nielsole:
I read #431 that `sched_debug` (not `schedstat`) is not suitable for measuring load. It doesn't mention `schedstat` but just removes it. I could well imagine that `schedstat` is not suitable for measuring load on a system, but that is not what we are attempting to do.

dashpole:
I would be willing to review a PR which adds those metrics.  I request that they be disabled by default (we have an `ignoreMetrics` mechanism you can use).  They should also be metrics associated with containers, rather than processes in the API.

[...]

PR #1872, opened 2018-01-22
Adding /proc/<pid>/schedstat

nielsole:
Adding scheduler statistics of all processes in a container as discussed in #1865 . 

I will try this out in the coming days see how useful these are in reality and whether anything else pops up.

@dashpole

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -297,7 +297,6 @@ type CpuCFS struct {
 type CpuSchedstat struct {
 	// https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt
 
-	// These numbers are summed up over all processes in a cgroup and can therefore not capture processes after they are gone.
 	// time spent on the cpu
 	RunTime uint64 `json:"run_time"`
 	// time spent waiting on a runqueue
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -293,10 +293,24 @@ type CpuCFS struct {
 	ThrottledTime uint64 `json:"throttled_time"`
 }
 
+// Cpu Aggregated scheduler statistics
+type CpuSchedstat struct {
+	// https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt
+
+	// These numbers are summed up over all processes in a cgroup and can therefore not capture processes after they are gone.
+	// time spent on the cpu
+	RunTime uint64 `json:"run_time"`
+	// time spent waiting on a runqueue
+	RunqueueTime uint64 `json:"runqueue_time"`
+	// # of timeslices run on this cpu
+	RunPeriods uint64 `json:"run_periods"`
+}
+
 // All CPU usage metrics are cumulative from the creation of the container
 type CpuStats struct {
-	Usage CpuUsage `json:"usage"`
-	CFS   CpuCFS   `json:"cfs"`
+	Usage     CpuUsage     `json:"usage"`
+	CFS       CpuCFS       `json:"cfs"`
+	Schedstat CpuSchedstat `json:"schedstat"`
 	// Smoothed average of number of runnable threads x 1000.
 	// We multiply by thousand to avoid using floats, but preserving precision.
 	// Load is smoothed over the last 10 seconds. Instantaneous value can be read
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -198,6 +198,27 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
 					return metricValues{{value: float64(s.Cpu.CFS.ThrottledTime) / float64(time.Second)}}
 				},
 			}, {
+				name:      "container_cpu_schedstat_run_seconds_total",
+				help:      "Time duration the processes of the container have run on the CPU.",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.Schedstat.RunTime) / float64(time.Second)}}
+				},
+			}, {
+				name:      "container_cpu_schedstat_runqueue_seconds_total",
+				help:      "Time duration processes of the container have been waiting on a runqueue.",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.Schedstat.RunqueueTime) / float64(time.Second)}}
+				},
+			}, {
+				name:      "container_cpu_schedstat_run_periods_total",
+				help:      "Number of times processes of the cgroup have run on the cpu",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.Schedstat.RunPeriods)}}
+				},
+			}, {
 				name:      "container_cpu_load_average_10s",
 				help:      "Value of container cpu load average over the last 10 seconds.",
 				valueType: prometheus.GaugeValue,
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -64,6 +64,7 @@ var (
 	ignoreMetrics metricSetValue = metricSetValue{container.MetricSet{
 		container.NetworkTcpUsageMetrics: struct{}{},
 		container.NetworkUdpUsageMetrics: struct{}{},
+		container.ProcessSchedulerMetrics: struct{}{},
 	}}
 
 	// List of metrics that can be ignored.
@@ -73,6 +74,7 @@ var (
 		container.NetworkTcpUsageMetrics: struct{}{},
 		container.NetworkUdpUsageMetrics: struct{}{},
 		container.PerCpuUsageMetrics:     struct{}{},
+		container.ProcessSchedulerMetrics: struct{}{},
 	}
 )
 
--- a/container/factory.go
+++ b/container/factory.go
@@ -41,16 +41,17 @@ type ContainerHandlerFactory interface {
 type MetricKind string
 
 const (
-	CpuUsageMetrics        MetricKind = "cpu"
-	PerCpuUsageMetrics     MetricKind = "percpu"
-	MemoryUsageMetrics     MetricKind = "memory"
-	CpuLoadMetrics         MetricKind = "cpuLoad"
-	DiskIOMetrics          MetricKind = "diskIO"
-	DiskUsageMetrics       MetricKind = "disk"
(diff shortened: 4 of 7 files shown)


dashpole:
I am going to wait for https://github.com/google/cadvisor/pull/1868 to be resolved before testing this.  We are currently in the process of migrating our testing infrastructure

info/v1/container.go
```diff
@@ -295,10 +295,24 @@ type CpuCFS struct {
 	ThrottledTime uint64 `json:"throttled_time"`
 }
 
+// Cpu Aggregated scheduler statistics
+type CpuSchedstat struct {
+	// https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt
+
+	// These numbers are summed up over all processes in a cgroup and can therefore not capture processes after they are gone.
+	// time spent on the cpu
+	RunTime uint64 `json:"run_time"`
```
dashpole: Is this different from CPU usage?

metrics/prometheus.go
```diff
@@ -189,6 +189,27 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc) *PrometheusCo
...
+				},
+			}, {
+				name:      "container_cpu_schedstat_runqueue_seconds_total",
+				help:      "Time duration processes of the container have been waiting on a runqueue.",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.Schedstat.RunqueueTime) / float64(time.Second)}}
+				},
+			}, {
+				name:      "container_cpu_schedstat_run_number_of_times_total",
```
dashpole: do you think `container_cpu_schedstat_run_periods_total` would be better?

container/libcontainer/helpers.go
```diff
@@ -165,6 +176,43 @@ func GetStats(cgroupManager cgroups.Manager, rootFs string, pid int, ignoreMetri
...
+		contents, err := ioutil.ReadAll(f)
+		if err != nil {
+			return info.CpuSchedstat{}, fmt.Errorf("couldn't read scheduler statistics for process %d: %v", pid, err)
+		}
+		rawMetrics := bytes.Split(contents, []byte(" "))
+		if len(rawMetrics) != 3 {
+			return info.CpuSchedstat{}, fmt.Errorf("unexpected number of metrics in schedstat file for process %d", pid)
+		}
+
+		runtime, err := strconv.ParseUint(string(rawMetrics[1]), 10, 64)
```
nielsole: Note to self: The indices need to be adapted.

metrics/prometheus.go
nielsole: 👍

nielsole:
I signed it!

nielsole:
There is an issue where the metrics are only collected for some containers. E.g. it works for all docker containers, except when they are kubelet containers. I suspect `GetAllPids()` to not do what I expect it to do, but haven't had the time to dig in.

info/v1/container.go
nielsole: I am not so sure what the difference is, but they look quite distinct:
screenshot from 2018-01-31 15-25-40

nielsole:
https://github.com/google/cadvisor/pull/1872#issuecomment-360165511 was resolved. I was running it within docker and had not set `--pid=host`

dashpole:
/ok-to-test

nielsole:
`googleapi: Error 403: Quota 'IN_USE_ADDRESSES' exceeded. Limit: 8.0 in region us-central1.`
/retest

info/v1/container.go
dashpole: Since CPU and RunTime are the same, we should omit it from the API.

info/v1/container.go
```diff
@@ -295,10 +295,24 @@ type CpuCFS struct {
...
+	// time spent waiting on a runqueue
+	RunqueueTime uint64 `json:"runqueue_time"`
+	// # of timeslices run on this cpu
+	RunPeriods uint64 `json:"run_periods"`
+}
+
+// CFS, Usage and LoadAverage metrics are cumulative from the creation of the container, Schedstat is the sum across the cumulative metrics of running processes
```
dashpole: How feasible would it be to maintain a cache of PID->schedstat metrics, update the PIDs that are active, and then sum over all PIDs ever seen?  This way we would get a cumulative metric, and we get the invariant that our reported schedstat metrics never decrease.

info/v1/container.go
hanikesn: @nielsole I don't think those graphs are comparable, when summing up those container stats there's some double accounting going on. I'd recommend doing it for a single container.

info/v1/container.go
nielsole: Made a little something here: https://github.com/flix-tech/cadvisor/pull/1 So it is feasible, but there are of course still lost metrics between your housekeeping intervals.
I do not really like the code changes this requires.

info/v1/container.go
dashpole: The two graphs here are not comparable, but he says here https://github.com/google/cadvisor/pull/1872#issuecomment-361949155 that he figured it out and they are the same.

info/v1/container.go
dashpole: What in particular don't you like about the code changes?  I guess my main concern is that it is difficult to make sense of a number that decreases when a process dies.  I would rather have a slightly more complex implementation than a metric that isn't useful.  Most monitoring systems can't make sense of a total value that can decrease.

info/v1/container.go
nielsole: Actually I did not figure it out. #1872 was referring to a problem I had with running cAdvisor.

info/v1/container.go
nielsole: I share your gripe that ContainerHandler contains so many fields it should not care about. Also I did not like the `seen`, but luckily that is gone.

nielsole:
We had some doubts about the unit used in the /schedstat file because the documentation just said "time".

Here is the list of places that you can follow to find out the units:

run_delay
defined in https://github.com/torvalds/linux/blob/61f14c015f5be9151ba25e638d349f4d40cb7cd4/include/linux/sched.h#L250-L250
Indirectly uses rq_clock: https://github.com/torvalds/linux/blob/ab2d92ad881da11331280aedf612d82e61cb6d41/kernel/sched/sched.h#L925
Which is updated here: https://github.com/torvalds/linux/blob/a2e5790d841658485d642196dbb0927303d6c22f/kernel/sched/core.c#L232
https://github.com/torvalds/linux/blob/8e9a2dba8686187d8c8179e5b86640e653963889/kernel/sched/clock.c#L353 
https://github.com/torvalds/linux/blob/8e9a2dba8686187d8c8179e5b86640e653963889/kernel/sched/clock.c#L18

nielsole:
I just rebased onto the changes from https://github.com/google/cadvisor/pull/1887. 
What do you think @dashpole ?

info/v1/container.go
```diff
@@ -293,10 +293,24 @@ type CpuCFS struct {
 	ThrottledTime uint64 `json:"throttled_time"`
 }
 
+// Cpu Aggregated scheduler statistics
+type CpuSchedstat struct {
+	// https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt
+
+	// These numbers are summed up over all processes in a cgroup and can therefore not capture processes after they are gone.
```
dashpole: this is no longer the case, right?

info/v1/container.go
dashpole: What is the resolution here @nielsole?  Is this the same as cpu usage, or are they different metrics?

dashpole:
Can you add this to the v2 api and implement conversion between them?  It should be pretty straightforward.

dashpole:
otherwise, it lgtm

info/v1/container.go
nielsole: You are right. 
Processes spawned and exiting between housekeeping intervals would still be lost, but I guess that is not comment-worthy. Took it out.

nielsole:
As it is part of the CpuStats, it is already part of the V2 API:

/api/v2.0/stats
screenshot from 2018-03-08 10-40-09

/api/v1.3/containers
screenshot from 2018-03-08 10-46-18

Could you elaborate what still needs to be done?

dashpole:
oh, ok.  Didn't look closely enough.

bobrik:
These metrics are supposed to be counters, but if a PID disappears from a cgroup, it takes its delay stats with it. This causes counter to go backward. If counter goes backwards, prometheus detects reset and next rate will be calculated not from increment, but from the whole counter value.

Is there any way to live with this?

nielsole:
IIRC we cache the values, so the counter never goes backwards when a process dies.

nielsole:
Now that I think about it: We have this cache by PID so if a process dies and (after some time) another process comes up with the same PID, it would overwrite the cache and only then would the counter go backwards. I have no idea how frequently the same PID will be handed out again, so this might become an issue for some workloads.

bobrik:
I definitely see counter go backwards, especially for `cron.service` where PIDs have often limited lifespan. There's no way I'm wrapping around PID space in a few minutes.

bobrik:
I created a ticket and pinged cgroups mailing list about this:

* https://bugzilla.kernel.org/show_bug.cgi?id=200593

aktau:
From the commit (https://github.com/google/cadvisor/pull/1872/commits/4ccd4581da78bd6beb16efc178018173fde7a20a), it looks like for each pid in the cgroup, `/proc/pid/schedstat` is read. But the [kernel code](https://github.com/torvalds/linux/blob/28eb75e178d389d325f1666e422bc13bbbb9804c/fs/proc/base.c#L517-L520) for producing this proc file makes it look like it applies to a single task (thread).

Here's some numbers I got from a process:

```
==> /proc/2176633/schedstat <==
26704171776 327075 64

==> /proc/2176633/task/2176633/schedstat <==
26704171776 327075 64

==> /proc/2176633/task/2176634/schedstat <==
82469732 58599 2713

==> /proc/2176633/task/2176635/schedstat <==
24788546225 1753893 201

==> /proc/2176633/task/2176636/schedstat <==
16431851 0 110

==> /proc/2176633/task/2176637/schedstat <==
26699696446 3407562 135
...
```

It does not look like `/proc/2176633/schedstat` is the sum of all the child-schedstats. Is this intended? Am I wrong?  Thanks!

nielsole:
> It does not look like /proc/2176633/schedstat is the sum of all the child-schedstats. Is this intended?
I wasn't aware of this. I assumed it'd be the sum of it's child tasks.
> Am I wrong?
I don't know, the code you shared suggests you are right.

PR #1876, opened 2018-01-26
let the user who call this method to handle the error

### LINKED ISSUE #1875: [ feature ][ could make cadvisor client to be alone package? ]
### Version
- git clone the latest code today use this command:
```bash
go get github.com/google/cadvisor
```

### What happen?
Today I try to use cadvisor, and import "github.com/google/cadvisor/client", and follow the Readme.md, add this code in my project
```golang
        key := fmt.Sprintf("http://%s:%d", masters[hostip].HostIP, 4194)
	cClient, err := client.NewClient(key)
	if err != nil {
		log.Printf("get cadvisor client by %s, but failed: %v", key, err)
		continue
	}

	// 获取机器信息
	m, err := cClient.MachineInfo()
	if err != nil {
		log.Printf("get machine info failed: %v", err)
		continue
	}
```
but get error when compile
```bash
/tmp/___go_build_main_go flag redefined: log_dir
panic: /tmp/___go_build_main_go flag redefined: log_dir

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc42005e120, 0x1a97aa0, 0xc4203f50a0, 0x12bb7b2, 0x7, 0x12e6c08, 0x2f)
	/software/golang/go-v1.9.2/src/flag/flag.go:793 +0x5e1
flag.(*FlagSet).StringVar(0xc42005e120, 0xc4203f50a0, 0x12bb7b2, 0x7, 0x0, 0x0, 0x12e6c08, 0x2f)
	/software/golang/go-v1.9.2/src/flag/flag.go:696 +0x8b
flag.(*FlagSet).String(0xc42005e120, 0x12bb7b2, 0x7, 0x0, 0x0, 0x12e6c08, 0x2f, 0xc4203adf20)
	/software/golang/go-v1.9.2/src/flag/flag.go:709 +0x8b
flag.String(0x12bb7b2, 0x7, 0x0, 0x0, 0x12e6c08, 0x2f, 0xed84f9)
	/software/golang/go-v1.9.2/src/flag/flag.go:716 +0x69
github.com/google/cadvisor/vendor/github.com/golang/glog.init()
	/home/gogeof/work/gopath/src/github.com/google/cadvisor/vendor/github.com/golang/glog/glog_file.go:41 +0x14a
github.com/google/cadvisor/client.init()
	<autogenerated>:1 +0x71
ufleet/go-cluster-daemon/do.init()
	<autogenerated>:1 +0x70
main.init()
	<autogenerated>:1 +0x44
```

### What expected?
- could use cadvisor client alone

gogeof:
## let the user to handle the problem when decode event stream data fail
- here should not be fatal if some error occur, because it will cause the program to exit. And if the error just return, after catch that, we can try again or do something else to deal with that.
- if cadvisor client is just the client, maybe you can choose which log package to use as you want. The whole client package, except test code, just here use the glog package, so, I want return the error instead.
- if just return the error, #1875 fix.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -30,7 +30,6 @@ import (
 
 	"github.com/google/cadvisor/info/v1"
 
-	"github.com/golang/glog"
 	"time"
 )
 
@@ -229,8 +228,8 @@ func (self *Client) getEventStreamingData(url string, einfo chan *v1.Event) erro
 			if err == io.EOF {
 				break
 			}
-			// if called without &stream=true will not be able to parse event and will trigger fatal
-			glog.Fatalf("Received error %v", err)
+			// if called without &stream=true will not be able to parse event and let the user who call this method to handle this error
+			return fmt.Errorf("received error %v", err)
 		}
 		einfo <- m
 	}


gogeof:
I signed it!

dashpole:
This doesnt seem related to #1875.  Can you explain how this solves that issue?

gogeof:
@dashpole thanks for response. ok, let me try:
1. I think here should not be fatal if some error occur, because it will cause my program to exit. And if the error just return, after catch that, I can try again or do something else. 
2. yes, I have set the log_dir in my project. But I think that if cadvisor client is just the client, maybe you can choose which log package to use as you want. The whole client package, except test code, just here use the glog package, so, I want return the error instead. 
3. if just return the error, #1876 fix.

dashpole:
/ok-to-test

PR #1877, opened 2018-01-27
cadvisor: build: fix broken build with Makefile

bhuvanchandra:
When using `make build` to compile the source,
build fails with error/s. The reason is because
of GO_CMD variable is set to 'install' as default,
which is not valid when running `make build` or
`make`.

Depending on the valid git tag, release.sh will
set the GO_CMD variable. If valid git tag is
available e.g.: vx.y.z.beta1, then GO_CMD is
set to `build`.

If valid git tag is not available, release.sh
will any way bails out serving the puspose.
So drop using GO_CMD, to fix the default build.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -17,7 +17,6 @@
 set -e
 
 GO_FLAGS=${GO_FLAGS:-"-tags netgo"}    # Extra go flags to use in the build.
-GO_CMD=${GO_CMD:-"install"}
 BUILD_USER=${BUILD_USER:-"${USER}@${HOSTNAME}"}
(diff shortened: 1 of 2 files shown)


dashpole:
/ok-to-test

bhuvanchandra:
@dashpole, rebased on top of latest master.

dashpole:
Just curious, what errors does it fail with?

bhuvanchandra:
Here is the error log:
```
$ make
>> vetting code
>> checking go formatting
>> checking file boilerplate
>> building assets
>> building binaries
>> building cadvisor
go install net: open /usr/lib/go/pkg/linux_amd64/net.a: permission denied
make: *** [Makefile:43: build] Error 1
```

PR #1878, opened 2018-01-28
Fix Warning->Warningf for better logging

dims:
Looks like we forgot to use the method that uses formatting arguments


dashpole:
sorry, tests are not working right now.  We are migrating to prow.  It should be done in the next few days

dims:
ack @dashpole

dashpole:
@dims you will have to rebase before tests will pass

dims:
@dashpole rebased! cross my fingers

PR #1885, opened 2018-02-07
Do not error out when /dev/disk/by-uuid is not found.

Random-Liu:
We are switching off UUID in CRI. But I think it's still good to keep this feature.

However, Cadvisor shouldn't fail if `/dev/disk/by-uuid` is not found. Because `/dev/disk/by-uuid` is not always available on different OS distributions.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -114,7 +114,9 @@ func NewFsInfo(context Context) (FsInfo, error) {
 
 	fsUUIDToDeviceName, err := getFsUUIDToDeviceNameMap()
 	if err != nil {
-		return nil, err
+		// UUID is not always avaiable across different OS distributions.
+		// Do not fail if there is an error.
+		glog.Warningf("Failed to get disk UUID mapping, getting disk info by uuid will not work: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


dashpole:
Just curious, why keep the feature?  Just in-case we need it later?

Random-Liu:
> Just curious, why keep the feature? Just in-case we need it later?

@dashpole For backward compatibility. There might be users using it, right? It is part of cadvisor api.
And also, maybe we'll also need it someday.

If you think it is fine to remove the feature, and it won't break users. I can do that :)

dashpole:
> @dashpole For backward compatibility. There might be users using it, right? It is part of cadvisor api.

It is part of the manager API, which doesnt have any stability guarantees.  A quick search of github doesnt show anyone using it outside of k8s forks.

Random-Liu:
@dashpole See this. https://github.com/google/cadvisor/pull/1724/files#diff-b833d5a6501040e03694b19aa7e8ae7bR424

dashpole:
ah, ok.  We can leave it in

PR #1887, opened 2018-02-07
Refactor handlers to move common code to libcontainer handler

dashpole:
This should make implementing https://github.com/google/cadvisor/pull/1872 a bit cleaner
cc @nielsole


container/libcontainer/helpers.go
```diff
@@ -398,8 +414,8 @@ func scanUdpStats(r io.Reader) (info.UdpStat, error) {
 	return stats, nil
 }
 
-func GetProcesses(cgroupManager cgroups.Manager) ([]int, error) {
-	pids, err := cgroupManager.GetPids()
+func (l *Handler) GetProcesses() ([]int, error) {
```
nielsole: Maybe move all methods of the Handler struct together? Right now they are so scattered throughout the file. Maybe even move them out of `helpers.go`?

dashpole: done

dashpole:
@nielsole does this look good to you?

nielsole:
I will take a closer look at it tomorrow. On first glance it lgtm

nielsole:
:+1:  lgtm

container/containerd/handler.go
```diff
@@ -36,31 +34,21 @@ import (
...
-	pid int
+	reference info.ContainerReference
+	envs      map[string]string
+	labels    map[string]string
 	// Image name used for this container.
 	image string
-	// The host root FS to read
-	rootFs string
 	// Filesystem handler.
 	ignoreMetrics container.MetricSet
```
nielsole: We missed removing the `ignoreMetrics` from all handlers

nielsole: We missed removing the `ignoreMetrics` from all handlers

nielsole: wrong line actually, aimed for 48-49

dashpole: you are welcome to do this in your PR.  IIRC ignoreMetrics was still used by the other handlers, but I may be mistaken...

PR #1894, opened 2018-02-23
Use backoff to tolerant race condition.

Random-Liu:
I keep seeing this error in kubelet log:
```
Feb 23 09:31:05 workstation kubelet[11956]: I0223 09:31:05.336135   11956 factory.go:105] Error trying to work out if we can handle /kubepods/burstable/pod445bc55c-187c-11e8-bb75-42010af00002/de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373: error inspecting container: Error: No such container: de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373
Feb 23 09:31:05 workstation kubelet[11956]: I0223 09:31:05.336156   11956 factory.go:116] Factory "docker" was unable to handle container "/kubepods/burstable/pod445bc55c-187c-11e8-bb75-42010af00002/de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373"
Feb 23 09:31:05 workstation kubelet[11956]: I0223 09:31:05.336733   11956 factory.go:112] Using factory "containerd" for container "/kubepods/burstable/pod445bc55c-187c-11e8-bb75-42010af00002/de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373"
Feb 23 09:31:05 workstation kubelet[11956]: W0223 09:31:05.337858   11956 manager.go:1178] Failed to process watch event {EventType:0 Name:/kubepods/burstable/pod445bc55c-187c-11e8-bb75-42010af00002/de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373 WatchSource:0}: task de9b277dbb62d6e2bc2e372f190c84e227986acd40f6e2f2af8a115a51061373 not found: not found
```

The reason is that container cgroup is created in the middle of task creation. And there is a race condition that cadvisor see the cgroup, but the corresponding task hasn't been fully created yet in containerd.

There is no such problem for docker, because docker has some internal lock to make sure `Inspect` only returns after container start is finished. CRI-Containerd and I believe cri-o are doing the same thing. Actually cadvisor is relying on some container runtime internal implementation details here.

However, here we are talking with containerd directly, which still has this race condition.

This PR added a retry to avoid this problem for now, and we should come up with a better fix next release.

I've validated this PR, and it works for me.

/cc @abhi @dashpole 

### DIFF
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"path"
 	"strings"
+	"time"
 
+	"github.com/containerd/containerd/errdefs"
 	cgroupfs "github.com/opencontainers/runc/libcontainer/cgroups/fs"
 	libcontainerconfigs "github.com/opencontainers/runc/libcontainer/configs"
 	"golang.org/x/net/context"
@@ -91,10 +93,28 @@ func newContainerdContainerHandler(
 		return nil, err
 	}
 
-	taskPid, err := client.TaskPid(ctx, id)
-	if err != nil {
-		return nil, err
+	// Cgroup is created during task creation. When cadvisor sees the cgroup,
+	// task may not be fully created yet. Use a retry+backoff to tolerant the
+	// race condition.
+	// TODO(random-liu): Use cri-containerd client to talk with cri-containerd
+	// instead. cri-containerd has some internal synchronization to make sure
+	// `ContainerStatus` only returns result after `StartContainer` finishes.
+	var taskPid uint32
+	backoff := 100 * time.Millisecond
+	retry := 5
+	for {
+		taskPid, err = client.TaskPid(ctx, id)
+		if err == nil {
+			break
+		}
+		retry--
+		if !errdefs.IsNotFound(err) || retry == 0 {
+			return nil, err
+		}
+		time.Sleep(backoff)
+		backoff *= 2
 	}
+
 	rootfs := "/"
 	if !inHostNamespace {
 		rootfs = "/rootfs"


container/containerd/handler.go
```diff
@@ -91,10 +93,26 @@ func newContainerdContainerHandler(
...
-		return nil, err
+	// Cgroup is created during task creation. When cadvisor sees the cgroup,
+	// task may not be fully created yet. Use a retry+backoff to tolerant the
+	// race condition.
+	// TODO(random-liu): Use cri-containerd client to talk with cri-containerd
+	// instead. cri-containerd has some internal synchronization to make sure
+	// `ContainerStatus` only returns result after `StartContainer` finishes.
+	var taskPid uint32
+	backoff := 100 * time.Millisecond
+	for retry := 5; retry > 0; retry-- {
```
dashpole: so after 5 retries we continue on?  Will we not get any metrics in this case?  Would it be better just to return an error so it doesnt fail silently?

Random-Liu: ... my bad

Random-Liu: Done

container/containerd/handler.go
```diff
@@ -91,10 +93,28 @@ func newContainerdContainerHandler(
...
+	backoff := 100 * time.Millisecond
+	retry := 5
+	for {
+		taskPid, err = client.TaskPid(ctx, id)
+		if err == nil {
+			break
+		}
+		retry--
+		if !errdefs.IsNotFound(err) || retry == 0 {
+			return nil, err
```
dashpole: If retry == 0 case is hit, we will return err = nil.  This will probably result in a nil pointer somewhere down the road.  Can we return a new error for this case?

Random-Liu: No, we check `err == nil` before this, right?

dashpole: ah, right

Issue #1899, opened 2018-03-10
InfluxDB as backend storage, to add memory_rss measurement.


alicek106:
When InfluxDB is set as backend storage, memory related data is collected in two measurements.

(1) memory_usage
(2) memory_working_set

According to #638 , 
memory_usage = rss + cache
memory_working_set = memory_usage - inactive (inactive file + inactive anon)

But in many cases, memory except cache (only rss) will necessary because memory rss is consistent with docker stats API, so I propose to add a feature to push memory_rss measurement into influxdb backend storage.

Actually, I already forked cadvisor and using my modified version well. :D It was very easy.

https://github.com/alicek106/cadvisor/commit/ed3d9c40ae83ac377891061a7dc22f92ccd53f8c

I hope to update this feature the latest version of cadvisor. thank you.

dashpole:
sounds good to me! Feel free to post a PR.

alicek106:
Close because https://github.com/google/cadvisor/pull/1908 has been merged.

Issue #638, opened 2015-04-08
cadvisor memory legend


cowboy-cod3r:
When looking at the total memory usage of a running container, the legend shows "hot" and "total".  Is there any documentation that explains the definition "hot" and "total" and how that is calculated?

vishh:
"hot" refers to the *working set*, which is held in RAM, and "total" refers
to all the memory used by a process, a part of which may reside on a swap
device.

+1 to your suggestion of documenting it. AFAIK this isn't documented
anywhere.

rjnagal:
Hot is the working set - pages that has been recently touched as calculated
by the kernel.
Total includes hot + cold memory - where cold are the pages that have not
been touched in a while and can be reclaimed if there was global memory
pressure.

I don't think the total on cadvisor page includes swap, but we should
probably report that too.

vmarmol:
We used to have a bug open for this, but we closed it with some API comments. We should document this in the UI as well.

humbargs:
Thanks for the info!

ArtworkAD:
What I find confusing when using cAdvisor the first time is that it says 16Gib of a total of 24 Gib memory is used.

Via ```free -m```I found out that I actually have 22 Gib free memory and that 14 Gib are just cached by linux.

vmarmol:
Memory accounting is interesting because different parts of the kernel do it differently. I am not sure what `free -m` uses. I'm curious what the hot vs. cold memory we were reporting in that case

mboussaa:
in the same topic, data dumped into influxdb as "memory_usage" refers to which definition of memory usage ?

vishh:
AFAIK it does not account for swap as of now.

mboussaa:
Working set memory == RSS ?

jimmidyson:
Working set = RSS + cache (no swap as @vishh said above). We have #975 open for reporting RSS by itself.

We delegate to libcontainer to retrieve stats. We could implement this ourselves but it's kind of nice to have consistency.

mboussaa:
the hot or total memory is sent  to backend storage like influxdb ?

jimmidyson:
Working set & total are sent - see https://github.com/google/cadvisor/blob/master/storage/influxdb/influxdb.go#L119-L124

mboussaa:
So memory_usage (total) = RAM + Swap ? because @vishh said that cAdvisor is memory+swap accounting as of now here https://github.com/google/cadvisor/issues/913

[...]

PR #1901, opened 2018-03-13
Add memory_rss in influxdb.go

alicek106:
Writing again #1899 ...

When InfluxDB is set as backend storage, memory related data is collected in two measurements.
(1) memory_usage
(2) memory_working_set

According to #638 ,
memory_usage = rss + cache
memory_working_set = memory_usage - inactive (inactive file + inactive anon)

But in many cases, memory except cache (only rss) will necessary because memory rss is consistent with docker stats API, so I propose to add a feature to push memory_rss measurement into influxdb backend storage.

alicek106/cadvisor@d620d74

It can be resolved by just adding two simple line in influxdb.go :D

### DIFF
--- a/storage/influxdb/influxdb.go
+++ b/storage/influxdb/influxdb.go
@@ -60,6 +60,8 @@ const (
 	serMemoryUsage string = "memory_usage"
 	// Working set size
 	serMemoryWorkingSet string = "memory_working_set"
+        // Memory-rss usage
+        serMemoryRss string = "memory_rss"
 	// Cumulative count of bytes received.
 	serRxBytes string = "rx_bytes"
 	// Cumulative count of receive errors encountered.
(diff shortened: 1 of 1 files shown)


dashpole:
Can you update the unit tests to test serMemoryRss?

storage/influxdb/influxdb.go
```diff
@@ -60,6 +60,8 @@ const (
 	serMemoryUsage string = "memory_usage"
 	// Working set size
 	serMemoryWorkingSet string = "memory_working_set"
+        // Memory-rss usage
```
dashpole: indenting is off everywhere

dashpole:
`checking go formatting`
`I0313 16:05:18.659] The following files are not properly formatted:`
`I0313 16:05:18.659] storage/influxdb/influxdb.go storage/influxdb/influxdb.go`

dashpole:
closing as inactive.  Feel free to reopen if you are still working on this.

PR #1903, opened 2018-03-14
fix #1902 bug with retryDockerStatus

### LINKED ISSUE #1902: Should be `err == nil` in function retryDockerStatus()
image

dashpole:
cc @jsravn


derekwaynecarr:
@sjenning ack.

@dashpole thanks for the fix.

PR #1908, opened 2018-03-20
Add memory rss to statsd storage

IvanVlasic:
Solves same issue as https://github.com/google/cadvisor/issues/1899 for statsd storage.


IvanVlasic:
I signed it.

PR #1916, opened 2018-03-27
Switch from Inotify to Fsnotify

### LINKED ISSUE #1708: Migrate off the deprecated package `golang.org/x/exp/inotify`

dashpole:
This is a revert of https://github.com/google/cadvisor/pull/1807, which is a revert of #1760.  The issue, which I wasn't able to debug at the time, was that occasionally we would not wait for the pods to restart before querying for metrics.  I was able to confirm that even during these failures, Fsnotify correctly identifies the creation and deletion of containers.

/assign @tallclair 

fixes #1708


dashpole:
Ok, the reason the e2e tests are failing is because of https://github.com/google/cadvisor/pull/1917.  For some reason, this race condition is more likely with fsnotify.  This will need to be rebased on https://github.com/google/cadvisor/pull/1917 to pass

dashpole:
The only thing I changed relative to #1807 was renaming variables and comments with inotify -> fsnotify.  It is the same as the other PR otherwise.

PR #1917, opened 2018-03-28
Upon discovering a subdirectory, add a creation event for it

dashpole:
This fixes a bug in cAdvisor where it could miss container discovery.  The following sequence of events would cause this to occurr:
1. Start cAdvisor
2. cAdvisor completes recovery of all cgroups
3. create two levels of cgroups at once (e.g. /docker/CONTAINER_ID)
4. cAdvisor will generate a ContainerAdd event for docker, and start watching the docker/ directory and the /docker/CONTAINER_ID directory.

However, cAdvisor will never generate a ContainerAdd event for the /docker/CONTAINER_ID directory, as it already exists when we start watching the /docker directory.

This fixes the issue by adding a ContainerAdd event when preexisting subdirectories are discovered.

/assign @tallclair

### DIFF
--- a/manager/watcher/raw/raw.go
+++ b/manager/watcher/raw/raw.go
@@ -71,7 +71,7 @@ func NewRawContainerWatcher() (watcher.ContainerWatcher, error) {
 func (self *rawContainerWatcher) Start(events chan watcher.ContainerEvent) error {
 	// Watch this container (all its cgroups) and all subdirectories.
 	for _, cgroupPath := range self.cgroupPaths {
-		_, err := self.watchDirectory(cgroupPath, "/")
+		_, err := self.watchDirectory(events, cgroupPath, "/")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


manager/watcher/raw/raw.go
```diff
@@ -135,7 +135,8 @@ func (self *rawContainerWatcher) watchDirectory(dir string, containerName string
 	for _, entry := range entries {
 		if entry.IsDir() {
 			entryPath := path.Join(dir, entry.Name())
-			_, err = self.watchDirectory(entryPath, path.Join(containerName, entry.Name()))
+			subcontainerName := path.Join(containerName, entry.Name())
+			alreadyWatchingSubDir, err := self.watchDirectory(events, entryPath, subcontainerName)
```
tallclair: Does this correctly handle the case when there are more than 2 layers created simultaneously? I'm not sure what entryPath & containerName would be in that case.

dashpole: Yes, the recursive call to watchDirectory will run through that directory's subdirectories and end up adding the container.

PR #1919, opened 2018-03-30
Fix containerd connection wait.

Random-Liu:
Avoid cadvisor 2 seconds start up latency.

/cc @abhi @miaoyq 

Without this PR cadvisor will always wait for 2 seconds to confirm containerd is not there. It means that kubelet will also finish initialization 2 seconds slower. This is a regression in Kubernetes 1.9 and 1.10. We may want to cut a cherry-pick.

@mtaufen Do we still care about kubelet start up latency now?

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -16,6 +16,8 @@ package containerd
 
 import (
 	"context"
+	"fmt"
+	"net"
 	"sync"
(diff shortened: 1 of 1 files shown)


dashpole:
Thanks for this @Random-Liu!

PR #1923, opened 2018-04-09
add accelerator info to container spec

tianshapjq:
Since accelerators have been merged into cadvisor, we need to add some spec of accelerators into the container spec, such as the memory_total of GPU.
PTAL if this pr is appropriate. I will commit the test cases after the first round :)

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -159,6 +159,21 @@ func (nm *NvidiaManager) GetCollector(devicesCgroupPath string) (AcceleratorColl
 	return nc, nil
 }
 
+// GetCollector returns a collector that can fetch all nvidia gpu metrics for nvidia devices
+func (nm *NvidiaManager) GetAllDevicesCollector() (AcceleratorCollector, error) {
+	nc := &NvidiaCollector{}
+	nm.RLock()
+	if !nm.nvmlInitialized || len(nm.nvidiaDevices) == 0 {
+		nm.RUnlock()
+		return nil, nil
+	}
+	nm.RUnlock()
+	for _, device := range nm.nvidiaDevices {
+		nc.Devices = append(nc.Devices, device)
+	}
+	return nc, nil
+}
+
 // parseDevicesCgroup parses the devices cgroup devices.list file for the container
 // and returns a list of minor numbers corresponding to NVIDIA GPU devices that the
 // container is allowed to access. In cases where the container has access to all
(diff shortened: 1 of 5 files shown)


tianshapjq:
@mindprince @vishh PTAL at your convenience ~

mindprince:
Why do you want to duplicate this information? It's available through the stats struct.

tianshapjq:
Because I can't get these info from the _/spec_ api. I would like to obtain all the gpu info attached to the node, and I think gpu info is just similar to cpu and memory, we should be able to get these hardware info of the node through _/spec_.
Another thing, the nvidia gpu has some kind of bug in the production environment, that is, some cards just disappeared for some reasons we didn't know, and I have to figure out which one got disappeared through comparing the gpus in spec and the gpus in stats. So, in fact, what I need most in spec is the ID and memory of the gpu.

mindprince:
The reason spec doesn't have accelerator info is because cAdvisor only supports container level accelerator metrics, i.e., metrics are only shown when there are containers running that have GPUs attached. We don't support machine level accelerator metrics and so spec was not needed. I am not sure if we want to add support for machine level accelerator metrics.

Heapster already supports accelerator metrics: https://github.com/kubernetes/heapster/blob/v1.5.2/metrics/core/metrics.go#L795-L801

tianshapjq:
To my understanding, the containerName == "/" means the info of the node, which is collected by cAdvisor at present, so it does collect the info of node. Right?
The metric "_MetricAcceleratorMemoryTotal_ " in hepaster is just used for summary api, and will not show if I want the stats data, such as the total_memory or average_memory of gpus  the container used.

mindprince:
> To my understanding, the containerName == "/" means the info of the node, which is collected by cAdvisor at present, so it does collect the info of node. Right?

No, at `/`, the NVIDIA GPU devices are not listed individually in the devices cgroup and so they are not collected (https://github.com/google/cadvisor/blob/v0.29.1/accelerators/nvidia.go#L210).

tianshapjq:
I got that it's not listed in the devices cgroup, so I coded like 
`if containerName == "/" {
 	cont.nvidiaCollector, err = m.nvidiaManager.GetAllDevicesCollector()
 } else {
 	cont.nvidiaCollector, err = m.nvidiaManager.GetCollector(devicesCgroupPath)
 }`
to get the gpu spec of containers and node.
This is useful in our production, and I would like to know if this does for others.

dashpole:
If I read this correctly, we will be collecting metrics on each GPU twice?
I would like to hear more about how you plan to use this... It isn't going to be exposed in the summary API, and the kubelet cadvisor port is deprecated and will be removed in an upcoming version.  How will you access this information?
/ok-to-test

tianshapjq:
yes, this will collect the metrics twice. I could refactor to collect just once with a cache or something like that, but I need to be sure this feature is acceptable. 
The reason for why I did this is mentioned above:

> Because I can't get these info from the /spec api. I would like to obtain all the gpu info attached to the node, and I think gpu info is just similar to cpu and memory, we should be able to get these hardware info of the node through /spec.
Another thing, the nvidia gpu has some kind of bug in the production environment, that is, some cards just disappeared for some reasons we didn't know, and I have to figure out which one got disappeared through comparing the gpus in spec and the gpus in stats. So, in fact, what I need most in spec is the ID and memory of the gpu.

I will deploy by DaemonSet If cAdvisor is removed from in-tree Kubernetes.

bobbypage:
Closing since there has been no update on this PR since 2018.

PR #1926, opened 2018-04-10
add docker_only_prefix_whitelist

andyxning:
Partially fix https://github.com/kubernetes/kubernetes/issues/61994

This PR adds `docker_only_prefix_whitelist` to cadvisor to specify `A comma-separated list of cgroup path prefix that needs to be collected even docker-only is specified`.

/cc @dashpole @tallclair

### DIFF
--- a/container/raw/factory.go
+++ b/container/raw/factory.go
@@ -17,6 +17,7 @@ package raw
 import (
 	"flag"
 	"fmt"
+	"strings"
 
 	"github.com/google/cadvisor/container"
 	"github.com/google/cadvisor/container/common"
@@ -45,6 +46,9 @@ type rawFactory struct {
 
 	// List of metrics to be ignored.
 	ignoreMetrics map[container.MetricKind]struct{}
+
+	// List of raw container cgroup path prefix whitelist.
+	rawPrefixWhiteList []string
 }
 
 func (self *rawFactory) String() string {
@@ -62,6 +66,13 @@ func (self *rawFactory) NewContainerHandler(name string, inHostNamespace bool) (
 // The raw factory can handle any container. If --docker_only is set to false, non-docker containers are ignored.
 func (self *rawFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	accept := name == "/" || !*dockerOnly
+
+	for _, prefix := range self.rawPrefixWhiteList {
+		if strings.HasPrefix(name, prefix) {
+			accept = true
+			break
+		}
+	}
 	return true, accept, nil
 }
 
@@ -69,7 +80,7 @@ func (self *rawFactory) DebugInfo() map[string][]string {
 	return common.DebugInfo(self.watcher.GetWatches())
 }
 
-func Register(machineInfoFactory info.MachineInfoFactory, fsInfo fs.FsInfo, ignoreMetrics map[container.MetricKind]struct{}) error {
+func Register(machineInfoFactory info.MachineInfoFactory, fsInfo fs.FsInfo, ignoreMetrics map[container.MetricKind]struct{}, rawPrefixWhiteList []string) error {
 	cgroupSubsystems, err := libcontainer.GetCgroupSubsystems()
 	if err != nil {
 		return fmt.Errorf("failed to get cgroup subsystems: %v", err)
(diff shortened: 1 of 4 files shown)


container/raw/factory.go
```diff
@@ -29,6 +30,7 @@ import (
 )
 
 var dockerOnly = flag.Bool("docker_only", false, "Only report docker containers in addition to root stats")
+var dockerOnlyPrefixWhitelist = flag.String("docker_only_prefix_whitelist", "", "A comma-separated list of cgroup path prefix that needs to be collected when docker-only is specified")
```
dashpole: can we call this rawPrefixWhitelist?  I am hoping we can move away from `dockerOnly`, and instead use a whitelist of handlers in the future.

dashpole:
lgtm other than naming nit

dashpole:
Actually, how would you feel about making this an argument to manager.New()?  The desire to whitelist particular cgroup trees seems fairly kubernetes specific, and I would rather not add another flag if we can avoid it.  I think this would be a win for everyone in the community, as everyone would see better performance.  It would also give us flexibility to change this interface in the future if we want a different model without having to deprecate flags and such.

The only downside is that if people are using cadvisor metrics for particular cgroups that are outside of what we plan to monitor, those would no longer be accessible.

andyxning:
> how would you feel about making this an argument to manager.New()? 

I have also thought about this since `rawPrefixWhitelist` is a inter package variable. What i think about integrate this cAdvisor functionality into Kubernetes is to only pass this raw prefix whitelist when `docker_only` is specified automatically. Adding `rawPrefixWhitelist` should help us to easily integrate this functionality into Kubernetes.

> The only downside is that if people are using cadvisor metrics for particular cgroups that are outside of what we plan to monitor, those would no longer be accessible.

Yep. That is the truth when make `rawPrefixWhitelist` an argument to manager.New(). IMHO, it should be fine as no users have complained about lacking the ability to customize the whitelist when `docker_only` has been specified. Since we have no clear plan about how to evolve the related settings, leaving customizing whitelist from command line seems fine. If we need this later, it should also not difficult to add the support.

andyxning:
@dashpole @tallclair PTAL. I have moved the configuration from command line to `manager` field. This should be fine for us to integrate cadvisor into Kubernetes.

container/raw/factory.go
```diff
@@ -62,14 +66,24 @@ func (self *rawFactory) NewContainerHandler(name string, inHostNamespace bool) (
 // The raw factory can handle any container. If --docker_only is set to false, non-docker containers are ignored.
 func (self *rawFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 	accept := name == "/" || !*dockerOnly
+
+	if *dockerOnly {
```
dashpole: instead of having this only in effect when `--docker_only` is specified, can we make it always take effect, but have the default be `[ "/" ]`, so all raw cgroups are collected?  We can make setting --docker_only set this to `[ ]` during creation of the raw factory.

andyxning: Sounds good to me. Will do.

andyxning: > We can make setting --docker_only set this to [ ] during creation of the raw factory.

We should not do this, imo. What we want is to respect with whitelist settings even `docker_only` is set.

andyxning: What i think is about this: 
* by default setting `rawPrefixWhiteList` to `[]string{"/"}` to allow all raw cgroups to be collected or be compatible with backwards
* always filter whitelist raw cgroups and respect with the whitelist settings.

andyxning: PTAL.

andyxning:
Ping @dashpole @tallclair 

BTW, after this PR is merged, i will prepare an PR to bump cAdvisor dep to Kubernetes to utilizing this new change.

dashpole:
This looks good to me, although I would really like to hear @tallclair 's opinion on this.
@andyxning have you brought this up at sig-instrumentation at all?  It may be valuable to hear more opinions on limiting the cgroups that are scraped.

andyxning:
@dashpole Seems @tallclair has temporarily no response. Do anyone else can help us to review
 this?

andyxning:
Ping @dashpole @tallclair

andyxning:
Ping @dashpole @tallclair in case the notification emails are lost in your inbox. :)

andyxning:
Ping @dashpole @tallclair

andyxning:
@dashpole @tallclair PTAL. 😆

andyxning:
Will make a PR to Kubernetes once the dependency for cAdvisor is updated to make use of the new functionality. :)

serathius:
@andyxning is this integrated into k8s?

andyxning:
@serathius No. Will work on this later this week.

PR #1927, opened 2018-04-14
Update docs - go 1.7 needed for building

mavidser:
The project uses the built-in `context` package introduced in go 1.7 - update the docs to reflect that.


mavidser:
I signed it!

PR #1930, opened 2018-04-17
Enable kafka logs

mariusbld:
Logs are disabled by default in Shopify/sarama library.
* enabled Shopify/sarama logs
* fix small log string bug


mariusbld:
I signed it!

PR #1934, opened 2018-04-20
Fix os release parsing

sjenning:
The current `/etc/os-release` parsing code is not robust and assumes that the value of `PRETTY_NAME` is in double quotes when bash env var rules do not require it.  In such cases, the parsing fails.

Docker has a function for this already that does it in a more robust way.  This PR converts to using that method.  The new rules are double quotes and single quotes work and required if the value has spaces.  If no spaces, not quotes are required.

Notes: other options include fixing the code in place or copy-pasting that function from docker.

xref https://bugzilla.redhat.com/show_bug.cgi?id=1521087

@dashpole @derekwaynecarr

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -347,6 +347,11 @@
 			"Rev": "4f3616fb1c112e206b88cb7a9922bf49067a7756"
 		},
 		{
+			"ImportPath": "github.com/docker/docker/pkg/parsers/operatingsystem",
+			"Comment": "docs-v1.12.0-rc4-2016-07-15-7401-g4f3616fb1",
+			"Rev": "4f3616fb1c112e206b88cb7a9922bf49067a7756"
+		},
+		{
 			"ImportPath": "github.com/docker/docker/pkg/system",
 			"Comment": "docs-v1.12.0-rc4-2016-07-15-7401-g4f3616fb1",
 			"Rev": "4f3616fb1c112e206b88cb7a9922bf49067a7756"
@@ -493,6 +498,11 @@
 		{
 			"ImportPath": "github.com/kr/text",
 			"Rev": "6807e777504f54ad073ecef66747de158294b639"
+		},
+		{
+			"ImportPath": "github.com/mattn/go-shellwords",
+			"Comment": "v1.0.3-18-g39dbbfa",
+			"Rev": "39dbbfa24bbc39559b61cae9b20b0e8db0e55525"
 		},
 		{
 			"ImportPath": "github.com/matttproud/golang_protobuf_extensions/pbutil",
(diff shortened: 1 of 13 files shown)


Godeps/Godeps.json
```diff
@@ -904,10 +880,6 @@
 		{
 			"ImportPath": "k8s.io/utils/clock",
 			"Rev": "aedf551cdb8b0119df3a19c65fde413a13b34997"
-		},
-		{
-			"ImportPath": "k8s.io/utils/clock/testing",
```
dashpole: looks like this is removed from the godeps, but the packages in `vendor/` are still there.  I think some of the other packages are in the same boat

dashpole:
Opened https://github.com/google/cadvisor/issues/1935 for verifying godeps.

sjenning:
@dashpole Ah yes, godeps removed these, but I didn't want to change more than was needed and forgot to do a git add interactive on Godeps.json like I normally do.  I can add a commit to this that does the cleanup if you like?

dashpole:
That is fine with me

dashpole:
did you do `godep save ./...` (with `./...` at the end)?  I sometimes have to run save more than once as well...

machine/info.go
```diff
@@ -173,20 +174,11 @@ func Info(sysFs sysfs.SysFs, fsInfo fs.FsInfo, inHostNamespace bool) (*info.Mach
...
-		for _, line := range strings.Split(string(os_release), "\n") {
-			parsed := strings.Split(line, "\"")
-			if len(parsed) == 3 && parsed[0] == "PRETTY_NAME=" {
-				container_os = parsed[1]
-				break
-			}
-		}
+	os, err := operatingsystem.GetOperatingSystem()
+	if err != nil {
+		os = "Unknown"
```
dashpole: do we want to log the error V(4)?  Just in-case we get support questions about unknown OS.

sjenning: We didn't before so I just maintained what the old code did before i.e. returning "Unknown" if it couldn't be parsed.

My opinion is that we shouldn't log.  The "Unknown" in the `kubectl get nodes` output is evidence enough that the parsing failed.

dashpole: ok, sgtm.  Thanks!

sjenning:
@dashpole are we looking to bump the kube dep in 1.11?  is there anything else you'd like to get in before a bump?

dashpole:
Looks like freeze is may 29th, so ill set the next milestone for may 20th.  Who knows what other bugs we will find in the meantime :)

dashpole:
Of course, if you want extra testing time on something, feel free to bump the godeps sooner.

sjenning:
no rush, just wanted to make sure we bumped for 1.11 at some point. thanks!

PR #1938, opened 2018-04-30
Update partition regex to match NVME disks

rsevilla87:
Partition regexp should match nvme disks which are presented to the OS as e.g:  /dev/nvme0n1


rsevilla87:
I signed it!

dashpole:
ping this PR when you have the CLA resolved

PR #1939, opened 2018-05-08
Run du with ionice

dims:
Let us run du with idle io priority

Will this help? mentioned by @meeee in https://github.com/kubernetes/kubernetes/issues/61999#issuecomment-387367763

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -552,7 +552,7 @@ func GetDirDiskUsage(dir string, timeout time.Duration) (uint64, error) {
 	if dir == "" {
 		return 0, fmt.Errorf("invalid directory")
 	}
-	cmd := exec.Command("nice", "-n", "19", "du", "-s", dir)
+	cmd := exec.Command("ionice", "-c3", "nice", "-n", "19", "du", "-s", dir)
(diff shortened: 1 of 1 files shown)


dashpole:
Should we do the same for `find`?  Does this require any additional packages in our [Dockerfile](https://github.com/google/cadvisor/blob/master/deploy/Dockerfile)?

dims:
@dashpole i tried a few public latest images and they all seemed to have `ionice` 

```
docker run -it ubuntu bash
docker run -it busybox sh
docker run -it fedora sh
docker run -it debian sh
```

Also tried our hyperkube

```
docker run -it k8s.gcr.io/hyperkube-amd64:v1.10.2 sh
```

dims:
while looking at the various issues filed, i did not see find causing issue, so may be we should wait to do that?

dashpole:
ok, great.

dashpole:
They are usually lumped together.  I did a performance analysis at one point and found that `find` used about 1/2 the CPU of `du`, but i'm not sure about the disk io usage of find.

dims:
aha! so we could do at least "nice" on the `find` command

PR #1940, opened 2018-05-08
Run find command with nice to reduce cpu load

dims:
Reduce impact of running the "find" command when system is loaded. We
already use "nice" with the "du" command.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -601,7 +601,7 @@ func GetDirInodeUsage(dir string, timeout time.Duration) (uint64, error) {
 	}
 	var counter byteCounter
 	var stderr bytes.Buffer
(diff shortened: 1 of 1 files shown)


dims:
we recently added "ionice" to the "du" command, but deliberately leaving that out here. when we have evidence of io wait problems with find, we can add that

dashpole:
Is there a downside to using ionice?  Inode metrics are generally less important than disk metrics for containers.  It would be odd to prioritize the find command over the du command in that respect.  I am in favor of adding ionice here as well.

dashpole:
Also thanks for opening this

dims:
cool. Done! (added ionice)

PR #1947, opened 2018-05-24
Initial setup for building aarch64

niedbalski:
This patch modifies the build scripts to allow other
architectures such as aarch64 to be build.

Particularly, for the aarch64 case, the following changes are included:

- Modified the Makefiles and build/*.sh scripts to allow the ARCH variable to be passed.

- gccgo-aarch64-linux-gnu: is required for building aarch64 binaries , added as the default CC for arm64.
(sudo apt-get install gccgo-aarch64-linux-gnu).

- Added a Dockerfile under deploy/Dockerfile.arm64 that uses multiarch/alpine:aarch64-edge as the base image.

** What we are missing:

- A travis meta-build system integration for running the build jobs for each architecture.
- Integration on the github releases process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ifeq ($(ARCH),)
+ARCH := amd64 #amd64 as the default architecture.
+endif
+
+ifeq ($(REV),)
+REV := $(shell git rev-parse --short HEAD)
+endif
+
+
 GO := go
 pkgs  = $(shell $(GO) list ./... | grep -v vendor)
+
 
 all: presubmit build test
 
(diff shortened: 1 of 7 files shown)


niedbalski:
@dashpole any comments on this?

build/build.sh
```diff
@@ -43,12 +43,24 @@ ldflags="
...
+if [ "${ARCH}" == "arm64" ]; then
+  ldflags="${ldflags} -extldflags '-static'"
+fi
 
 if [ -n "$VERBOSE" ]; then
   echo "Building with -ldflags $ldflags"
 fi
 
+if [ "${ARCH}" == "arm64" ]; then
+   CGO_ENABLED=1 CC=aarch64-linux-gnu-gccgo GOARCH=$ARCH GOBIN=$PWD go build ${GO_FLAGS} -o cadvisor.${ARCH} -ldflags "${ldflags}" "${repo_path}"
```
suihkulokki: CC is C compiler. The correct line is CC=aarch64-linux-gnu-gcc

Makefile
```diff
@@ -12,20 +12,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ifeq ($(ARCH),)
+ARCH := amd64 #amd64 as the default architecture.
```
suihkulokki: you can use export ARCH after this line to avoid the ARCH=$ARCH lines below

deploy/Dockerfile.arm64
```diff
@@ -0,0 +1,10 @@
+FROM multiarch/alpine:aarch64-edge
```
tianon: Wouldn't it be more consistent to use `arm64v8/alpine:3.7` here so this matches `Dockerfile.amd64`?

(If you visit https://hub.docker.com/_/alpine/ and click on `arm64v8` under "Supported architectures:", it takes you to https://hub.docker.com/r/arm64v8/alpine/, which is the architecture-specific push location for official images and shows that 3.7 is supported there.)

dashpole:
Closing this as inactive.  Feel free to reopen if you are still working on this.

PR #1949, opened 2018-05-29
Update Dockerfile and dependencies

ElfoLiNk:
Update base image to alpine:3.7
Update glibc version to 2.27-r0


deploy/build
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Copyright 2015 Google Inc. All rights reserved.
```
dashpole: why is this file added?

ElfoLiNk: Wrongly added. Just removed

PR #1950, opened 2018-05-30
Dockerfile Healthcheck

ElfoLiNk:
The HEALTHCHECK feature was added in Docker 1.12.
I don't know what are the current supported compatibility versions by cadvisor or if this is only a build requirement instead of a runtime requirement.


ElfoLiNk:
@dashpole when i merged back in master i wrongly removed curl needed by the health check 😞 Opened https://github.com/google/cadvisor/pull/2120

PR #1956, opened 2018-06-08
resetting max memory usage after reading it

gvessere:
Hi,
tracking max_memory_usage would be potentially much more useful if we were resetting it every time we read it. This gives a better view of the max memory usage between each point we collect the data while at the same time not doing so for the entire lifetime of the app / container (the original time series can still be recovered with a sliding max operator).

What do you guys think?


gvessere:
I signed it!

dashpole:
You can't edit vendor files directly.  This is a change you will have to suggest to the runc folks, and then vendor into cAdvisor.  

While I agree it could be useful in some cases, it has somewhat confusing behavior and would interfere with other monitoring agents on the node.  I might suggest running a separate container which periodically resets these values instead.

PR #1963, opened 2018-06-13
container: fix concurrent map acccess

runcom:
GetSpec() can be called concurrently in manager/container.go.updateSpec()
results into a concurrent map access on the labels map because we're
directly updating the map inside GetSpec(). The labels map from the
container handler is not a copy of the map itself, just a reference,
that's why we're getting the concurrent map access.
Fix this by moving the label update with restartcount to the handler's
initialization method which is not called concurrently.

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -65,9 +65,6 @@ type crioContainerHandler struct {
 
 	ignoreMetrics container.MetricSet
 
-	// container restart count
-	restartCount int
-
 	reference info.ContainerReference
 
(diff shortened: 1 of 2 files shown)


runcom:
@derekwaynecarr @sjenning @mrunalp PTAL

dashpole:
Just curious, what behavior did you see that lead you to this fix?

runcom:
> Just curious, what behavior did you see that lead you to this fix?

we got a panic about a concurrent map access :)

PR #1964, opened 2018-06-13
Dont watch "*.mount" cgroups

dashpole:
We already [don't collect metrics on .mount cgroups](https://github.com/google/cadvisor/blob/master/container/systemd/factory.go#L39).  But watching each of the mount cgroups for sub-cgroups can use up all of the inotify watches on the node.  This change now makes the watcher ignore .mount cgroups.

### DIFF
--- a/manager/watcher/raw/raw.go
+++ b/manager/watcher/raw/raw.go
@@ -110,6 +110,11 @@ func (self *rawContainerWatcher) Stop() error {
 // Watches the specified directory and all subdirectories. Returns whether the path was
 // already being watched and an error (if any).
 func (self *rawContainerWatcher) watchDirectory(events chan watcher.ContainerEvent, dir string, containerName string) (bool, error) {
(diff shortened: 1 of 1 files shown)


yujuhong:
I think this is the right thing to do since watching those cgroups exhausts node resources easily.
/lgtm 

/cc @derekwaynecarr

manager/watcher/raw/raw.go
```diff
@@ -110,6 +110,9 @@ func (self *rawContainerWatcher) Stop() error {
 // Watches the specified directory and all subdirectories. Returns whether the path was
 // already being watched and an error (if any).
 func (self *rawContainerWatcher) watchDirectory(events chan watcher.ContainerEvent, dir string, containerName string) (bool, error) {
+	if strings.HasSuffix(containerName, ".mount") {
```
yujuhong: Need a comment explaining why :-)

Issue #1944, opened 2018-05-23
Add mesos containerizer support to cadvisor


sashankreddya:
Cadvisor could also support collecting [Mesos containerizer](http://mesos.apache.org/documentation/latest/mesos-containerizer/) based container metrics

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #1965, opened 2018-06-14
Add mesos containerizer support

sashankreddya:
This commit includes support for collecting container stats
launched by[ mesos containerizer](http://mesos.apache.org/documentation/latest/mesos-containerizer/).

This refers to the issue [1944](https://github.com/google/cadvisor/issues/1944)


container/mesos/client_test.go
```diff
@@ -0,0 +1,40 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mesos
+
+import "fmt"
+
+type mesosAgentClientMock struct {
```
dashpole: nit: We use "mock" to refer to a testing struct for which function calls and return values are explicitly specified in the test.  See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/testing/cadvisor_mock.go#L23, as an example.  Most use the github.com/stretchr/testify/mock library.

For testing structs like this one, which is a "fake" version of the real client, we refer to them as `fake`.

sashankreddya: I referred to the one in "https://github.com/google/cadvisor/blob/master/container/containerd/client_test.go#L24". Isnt it the preferred way ?

sashankreddya: Addressed it

container/mesos/handler.go
```diff
@@ -0,0 +1,219 @@
...
+	return handler, nil
+}
+
+func (self *mesosContainerHandler) ContainerReference() (info.ContainerReference, error) {
+	// We only know the container by its one name.
+	return self.reference, nil
+}
+
+func (self *mesosContainerHandler) Start() {
+	if self.fsHandler != nil {
```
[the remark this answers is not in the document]
sashankreddya: I added it as we might add in upcoming PR's. Let me know if you want me to add it later.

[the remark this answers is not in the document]
dashpole: Yeah, just add it later.

container/mesos/mesos_agent.go
```diff
@@ -0,0 +1,140 @@
...
+
+package mesos
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// Mesos agent state json structure
+type (
```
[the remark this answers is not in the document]
sashankreddya: These are defined now in the mesos-go library. We don't need to look for multiple versions especially with the info needed with state and container endpoints of mesos. 

Also as discussed, once the wrappers defined here are available in opensource mesos go client library, then I can remove these from here.

container/mesos/mesos_agent.go
```diff
@@ -0,0 +1,113 @@
...
+	t, err := s.GetTask(exID)
+	if err != nil {
+		return err
+	}
+
+	// Identify revocability. Can be removed once we have a proper label
+	for _, resource := range t.Resources {
+		if resource.Name == "cpus" {
+			if resource.Revocable != nil {
+				labels["scheduler_sla"] = "revocable"
```
dashpole: lets make these label names package constants, and then use them in tests as well.  This includes other label names defined in this file (framework, source, scheduler_sla).

container/mesos/mesos_agent_test.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, ts := range []testCase{
+		{
+			frameworkID: "fw-id1",
+			executorID:  "exec-id1",
+			agentState: &agent.Response_GetState{
+				GetFrameworks: PopulateFrameworks("fw-id1"),
+				GetExecutors:  PopulateExecutors("exec-id1"),
+				GetTasks:      PopulateTasks(),
+			},
+			expectedError: nil,
```
dashpole: lets add some test cases that get errors here to ensure we handle them correctly

container/mesos/factory.go
```diff
@@ -0,0 +1,153 @@
...
+		&self.cgroupSubsystems,
+		self.machineInfoFactory,
+		self.fsInfo,
+		self.ignoreMetrics,
+		inHostNamespace,
+		client,
+	)
+}
+
+// Returns the Mesos ID from the full container name.
```
dashpole: nit: comment should include exported function name.

container/mesos/factory_test.go
```diff
@@ -0,0 +1,74 @@
...
+	testContainers["04e20821-67d3-4bf7-96b4-7d4495f50b28"] = testContainer
+
+	f := &mesosFactory{
+		machineInfoFactory: nil,
+		cgroupSubsystems:   containerlibcontainer.CgroupSubsystems{},
+		fsInfo:             nil,
+		ignoreMetrics:      nil,
+		client:             fakeMesosAgentClient(testContainers, nil),
+	}
+	for k, v := range map[string]bool{
```
dashpole: nit: prefer iterating over a list of predefined structs (like the above function does) so we understand what the test-case's inputs are, and so that we can easily add inputs in the future if we change the function.

dashpole:
Just a few minor nits, and this is good to go.  I checked the godeps and licenses, and they look good (2 apache, 1 MIT license).

sashankreddya:
Sure will address the nits. Thanks David

container/mesos/factory_test.go
sashankreddya: done

container/mesos/factory.go
sashankreddya: done

container/mesos/handler_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+func PopulateContainer() *mContainer {
+	var pid uint32 = 123
+	cntr := &mContainer{
+		ContainerStatus: &mesos.ContainerStatus{ExecutorPID: &pid},
+	}
+	return cntr
+}
+
+func TestHandler(t *testing.T) {
```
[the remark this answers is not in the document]
sashankreddya: Done

container/mesos/mesos_agent_test.go
sashankreddya: done

container/mesos/mesos_agent.go
sashankreddya: done

PR #1967, opened 2018-06-14
vendor: update docker import

jessfraz:
whats up :)

So fun story, I was trying to update the docker API types for https://github.com/kubernetes/kubernetes/pull/64283 but since pkg/tlsconfig no longer exists in the updated repo, and cadvisor depends on the client package which depends on pkg/tlsconfig I could not... so this is also updating the docker dep here

cc @vishh


dashpole:
Is this needed for 1.11?  Trying to determing urgency :)

jessfraz:
it can wait :)

jessfraz:
Thanks!!

dashpole:
You can update the cAdvisor godeps in master without a release once the 1.11 release has been cut.  We just need to make sure k/k releases have a versioned cAdvisor release, but master can be updated to a specific commit.

jessfraz:
awesome, thanks!

PR #1969, opened 2018-06-16
Initialize NVML on demand.

mindprince:
Earlier if the NVIDIA driver was not installed when cAdvisor was started
we would start a goroutine to try to initialize NVML every minute.

This resulted in a race. We can have a situation where:
- goroutine tries to initialize NVML but fails. So, it sleeps for a minute.
- the driver is installed.
- a container that uses NVIDIA devices is started.
This container would not get GPU stats because a minute has not passed
since the last failed initialization attempt and so NVML is not
initialized.

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -31,7 +31,10 @@ import (
 )
 
 type NvidiaManager struct {
-	sync.RWMutex
+	sync.Mutex
+
+	// true if there are NVIDIA devices present on the node
+	devicesPresent bool
 
 	// true if the NVML library (libnvidia-ml.so.1) was loaded successfully
 	nvmlInitialized bool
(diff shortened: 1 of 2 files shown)


accelerators/nvidia.go
```diff
@@ -139,6 +125,20 @@ func (nm *NvidiaManager) Destroy() {
...
+	if !nm.devicesPresent {
+		return nc, nil
+	}
+	// Makes sure that we don't call initializeNVML() concurrently and
+	// that we only call initializeNVML() when it's not initialized.
+	nm.Lock()
+	if !nm.nvmlInitialized {
+		initializeNVML(nm)
+		nm.Unlock()
+	} else {
```
dashpole: you dont need the `else` here.  Just unlock outside the if statement.

mindprince: Done.

mindprince: Actually this made me realize that I can simplify the taking locks multiple times to taking lock just once. PTAL.

PR #1971, opened 2018-06-19
fix kubernetes issue #65204

MaximilianMeister:
https://github.com/kubernetes/kubernetes/issues/65204

Signed-off-by: Maximilian Meister <mmeister@suse.de>


MaximilianMeister:
@googlebot I signed it!

dashpole:
This looks reasonable to me, although it makes the brtfs-specific code even more complex.
cc @flavio @dmrub who have made improvements to this in the past in-case (and likely are more familiar with brtfs than I am), in-case they have any comments

flavio:
@dashpole the change looks good to me. I worked with @MaximilianMeister on that ;)

PR #1978, opened 2018-06-21
Update docker godeps to match kubernetes

dashpole:
This is to allow updating cadvisor godeps here: https://github.com/kubernetes/kubernetes/pull/65334, which is currently failing with a compile error.


dashpole:
whoops, just need to update docker in k/k

PR #1979, opened 2018-06-26
Add cadvisor daemonset using kustomize

dashpole:
Adds the base cAdvisor daemonset in the base directory, and includes examples on how to mark it "critical", how to limit the metrics cadvisor collects, and how to add a prom-to-sd sidecar container to put metrics in stackdriver

### DIFF
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -1,0 +1,36 @@
+# cAdvisor Kubernetes Daemonset
+
+cAdvisor uses [Kustomize](https://github.com/kubernetes-sigs/kustomize) to manage kubernetes yaml files.  See the [Kustomize](https://github.com/kubernetes-sigs/kustomize) readme for installation instructions, and for a description of how it works.
+
+## Usage
+
+From the cadvisor/ directory, to generate the base daemonset:
+```
+kustomize build deploy/kubernetes/base
+```
+
+To apply the base daemonset to your cluster:
+```
+kustomize build deploy/kubernetes/base | kubectl apply -f -
+```
+
+To generate the daemonset with example patches applied:
+```
+kustomize build deploy/kubernetes/overlays/examples
+```
+
+To apply the daemonset to your cluster with example patches applied:
+```
+kustomize build deploy/kubernetes/overlays/examples | kubectl apply -f -
+```
+
+## Kustomization
+
+On your own fork of cAdvisor, create your own overlay directoy with your patches.  Copy patches from the example folder if you intend to use them, but don't modify the originals.  Commit your changes in your local branch, and use git to manage them the same way you would any other piece of code.
+
+To run the daemonset with your patches applied:
+```
+kustomize build deploy/kubernetes/overlays/<my_custom_overlays> | kubectl apply -f -
+```
+
+To get changes made to the upstream cAdvisor daemonset, simply rebase your fork of cAdvisor on top of upstream.  Since you didn't make changes to the originals, you won't have any conflicts.
--- a/deploy/kubernetes/base/daemonset.yaml
+++ b/deploy/kubernetes/base/daemonset.yaml
@@ -1,0 +1,51 @@
+apiVersion: apps/v1 # for Kubernetes versions before 1.9.0 use apps/v1beta2
+kind: DaemonSet
+metadata:
+  name: cadvisor
+  namespace: cadvisor
+  annotations:
+      seccomp.security.alpha.kubernetes.io/pod: 'docker/default'
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
+        image: k8s.gcr.io/cadvisor:v0.30.2
+        resources:
+          requests:
+            memory: 200Mi
+            cpu: 150m
+          limits:
+            cpu: 300m
+        volumeMounts:
+        - name: rootfs
+          mountPath: /rootfs
+          readOnly: true
+        - name: var-run
+          mountPath: /var/run
+          readOnly: true
+        - name: sys
+          mountPath: /sys
+          readOnly: true
+        - name: docker
+          mountPath: /var/lib/docker
+          readOnly: true
+        ports:
+          - name: http
+            containerPort: 8080
+            protocol: TCP
+      automountServiceAccountToken: false
+      terminationGracePeriodSeconds: 30
+      volumes:
+      - name: rootfs
+        hostPath:
+          path: /
+      - name: var-run
+        hostPath:
+          path: /var/run
+      - name: sys
+        hostPath:
+          path: /sys
+      - name: docker
+        hostPath:
+          path: /var/lib/docker
--- a/deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
+++ b/deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
@@ -1,0 +1,33 @@
+# This patch adds a sidecar which pushes all metrics to stackdriver
+apiVersion: apps/v1 # for Kubernetes versions before 1.9.0 use apps/v1beta2
+kind: DaemonSet
+metadata:
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: prometheus-to-sd
+        image: gcr.io/google-containers/prometheus-to-sd:v0.2.6
+        ports:
+          - name: profiler
+            containerPort: 6061
+        command:
+          - /monitor
+          - --stackdriver-prefix=custom.googleapis.com
+          - --source=cadvisor:http://localhost:8080
+          - --pod-id=$(POD_NAME)
+          - --namespace-id=$(POD_NAMESPACE)
+        env:
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+        securityContext:
+          runAsNonRoot: true
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ cAdvisor is now running (in the background) on `http://localhost:8080`. The setu
 
 We have detailed [instructions](docs/running.md#standalone) on running cAdvisor standalone outside of Docker. cAdvisor [running options](docs/runtime_options.md) may also be interesting for advanced usecases. If you want to build your own cAdvisor Docker image see our [deployment](docs/deploy.md) page.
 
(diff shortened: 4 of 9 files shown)


dashpole:
Should we put cAdvisor in the kube-system daemonset?

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1beta2 # for versions before 1.8.0 use extensions/v1beta1 [apps/v1beta2]
```
tallclair: for *Kubernetes* versions ...

same below.

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
...
+kind: DaemonSet
+metadata:
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
+        image: gcr.io/google-containers/cadvisor:v0.30.2
```
tallclair: We want to make sure this is kept up-to-date, but I'm not sure what the best way to do that is... We could have a test that parses the release notes headers and ensures this is kept in sync?

tallclair: Let's use `k8s.gcr.io/cadvisor` instead. It should point to the same place.

deploy/kubernetes/overlays/examples/filter-metrics.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: DaemonSet
+metadata:
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
+        command:
+          - /usr/bin/cadvisor
```
tallclair: nit: prefer
```yaml
command:
  - /usr/bin/cadvisor
  - --logtostderr
arguments:
- --disable_metrics=percpu,disk,network,tcp,udp
- --docker_only
```

Alternatively, move the command to the base template

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1beta2 # for versions before 1.8.0 use extensions/v1beta1 [apps/v1beta2]
+kind: DaemonSet
+metadata:
+  name: cadvisor
```
tallclair: let's add a cadvisor namespace

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
...
+          path: /
+      - name: var-run
+        hostPath:
+          path: /var/run
+      - name: sys
+        hostPath:
+          path: /sys
+      - name: docker
+        hostPath:
+          path: /var/lib/docker
```
tallclair: nit: add `automountServiceAccountToken: false`

deploy/kubernetes/overlays/examples/filter-metrics.yaml
```diff
@@ -0,0 +1,14 @@
...
+          - --disable_metrics=percpu,disk,network,tcp,udp
+          - --docker_only
+          - --logtostderr
```
tallclair: I think we should either add an optional overlay with the kubernetes default arguments:
```yaml
            - --housekeeping_interval=10s
```

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1beta2 # for versions before 1.8.0 use extensions/v1beta1 [apps/v1beta2]
+kind: DaemonSet
+metadata:
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
```
tallclair: I'm surprised cAdvisor doesn't need to run as privileged. I guess that's just because it reads everything from the mounted hostPath volumes?

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
...
+            cpu: 150m
+          limits:
+            cpu: 300m
+        volumeMounts:
+        - name: rootfs
+          mountPath: /rootfs
+          readOnly: true
+        - name: var-run
+          mountPath: /var/run
+          readOnly: false
```
tallclair: It would be great if we could make this `readOnly`... do you know why cAdvisor requires this to be writeable?

deploy/kubernetes/base/daemonset.yaml
tallclair: As of k8s 1.9, there's apps/v1

deploy/kubernetes/base/daemonset.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1beta2 # for versions before 1.8.0 use extensions/v1beta1 [apps/v1beta2]
+kind: DaemonSet
+metadata:
```
tallclair: Does it still work if you run with seccomp?
```
    annotations:
      seccomp.security.alpha.kubernetes.io/pod: 'docker/default'
```

deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
```diff
@@ -0,0 +1,28 @@
...
+          - --namespace-id=$(POD_NAMESPACE)
+        env:
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
```
tallclair: Add:
```yaml
securityContext:
  runAsNonRoot: true
  readOnlyRootFilesystem: true  # I think this should work?
  allowPrivilegeEscalation: false
```

deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
```diff
@@ -0,0 +1,28 @@
...
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: prometheus-to-sd
+        image: gcr.io/google-containers/prometheus-to-sd:v0.2.6
+        ports:
+          - name: profiler
+            containerPort: 6060
```
tallclair: 6061? https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/main.go#L62

deploy/kubernetes/README.md
```diff
@@ -0,0 +1,36 @@
+# cAdvisor Kubernetes Daemonset
```
tallclair: Consider adding a link to this from the main cadvisor readme

deploy/kubernetes/base/daemonset.yaml
dashpole: not sure.  I tried running it with `readOnly: true`, and it seemed to work fine

deploy/kubernetes/base/daemonset.yaml
dashpole: still works

deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
dashpole: `readOnlyRootFilesystem: true` works

deploy/kubernetes/README.md
dashpole: done

deploy/kubernetes/base/daemonset.yaml
dashpole: updated

deploy/kubernetes/base/daemonset.yaml
dashpole: yeah, you should know better than me, but I thought privileged only mattered for making syscalls.  cAdvisor just reads lots of stuff in sys/fs/cgroup

deploy/kubernetes/overlays/examples/filter-metrics.yaml
dashpole: done

deploy/kubernetes/base/daemonset.yaml
dashpole: done

deploy/kubernetes/base/daemonset.yaml
dashpole: done

deploy/kubernetes/overlays/examples/filter-metrics.yaml
dashpole: removed `command` entirely, as those are the default args specified by the Dockerfile for cadvisor

deploy/kubernetes/base/daemonset.yaml
dashpole: done

deploy/kubernetes/base/daemonset.yaml
dashpole: done

deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml
dashpole: I was following https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/kubernetes/prometheus-to-sd-kube-state-metrics.yaml#L30, but maybe they got it wrong

dashpole: yeah, they got it wrong

deploy/kubernetes/base/daemonset.yaml
dashpole: Ideally, we want it to track whatever we consider the "latest version" version to be, right?  So maybe add it to the "updating latest" part of the release?

dashpole:
It is also worth noting that the stackdriver sidecar doesn't work in current form.  cAdvisor, because it adds labels for each docker container label, and produces more labels than the stackdriver sidecar is willing to accept.
I plan to add a way to limit the set of labels attached to prometheus, and to exclude metrics specified in `--disable_metrics` from metrics in the prometheus endpoint.

dashpole:
(but I tested it with those changes, and it works out of the box)

dashpole:
It can connect to the docker client even with `/var/run` readonly.

tallclair:
Hmm, you need the write permission to connect to a socket, but maybe the volume doesn't need to be writeable?

dashpole:
Talked with Filipe, and it turns out that to "write" to a socket, you actually just need to open the socket file.  After that, interactions with the socket are not governed by r/w permissions on the socket file.  So we can make /var/run read-only

Issue #1925, opened 2018-04-09
Prometheus metrics for ignored metrics are still exported.


mightyguava:
cadvisor supports a `disable_metrics` flag, but this flag doesn't get propagated to the prometheus collector. It looks like cadvisor still exports all of these metrics to prometheus, even if they are disabled. While the cost is low for cadvisor, it adds unnecessary burden on prometheus to index a ton of metrics that don't have any values.

PR #1980, opened 2018-06-27
Don't emit prometheus metrics for metrics that are disabled

dashpole:
Not all of the metrics can actually be ignored, but I found it helpful to sort them by the MetricKind.
Fixes https://github.com/google/cadvisor/issues/1925

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -83,7 +83,7 @@ type dockerContainerHandler struct {
 	// The IP address of the container
 	ipAddress string
 
-	ignoreMetrics container.MetricSet
+	includedMetrics container.MetricSet
 
 	// the devicemapper poolname
 	poolName string
@@ -128,7 +128,7 @@ func newDockerContainerHandler(
 	inHostNamespace bool,
 	metadataEnvs []string,
 	dockerVersion []int,
-	ignoreMetrics container.MetricSet,
+	includedMetrics container.MetricSet,
 	thinPoolName string,
 	thinPoolWatcher *devicemapper.ThinPoolWatcher,
 	zfsWatcher *zfs.ZfsWatcher,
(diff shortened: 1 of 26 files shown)


tallclair:
Should we make this a whitelist instead of a blacklist?

dashpole:
We can change the flag to a whitelist from a blacklist in a separate PR if we want.   This PR should make prometheus more use-able for those who are ignoring metrics, and I would like to add this soon.

tallclair:
I meant that maybe the `NewPrometheusCollector` function should take a whitelist. If the plan is to call this from Kubernetes, I think that makes more sense (so we don't need to update k8s for metrics it's not using). WDYT?

dashpole:
We should probably do that for both `NewPrometheusCollector` and [manager.New](https://github.com/google/cadvisor/blob/master/manager/manager.go#L145).  Ill can change this PR to do that.

tallclair:
test failure looks legit? Other than that, lgtm.

container/docker/handler.go
```diff
@@ -345,14 +345,14 @@ func (self *dockerContainerHandler) ContainerReference() (info.ContainerReferenc
...
-	if !self.ignoreMetrics.Has(container.NetworkUsageMetrics) {
+	if self.includedMetrics.Has(container.NetworkUsageMetrics) {
 		return !self.networkMode.IsContainer()
 	}
 	return false
 }
 
 func (self *dockerContainerHandler) GetSpec() (info.ContainerSpec, error) {
-	hasFilesystem := !self.ignoreMetrics.Has(container.DiskUsageMetrics)
+	hasFilesystem := !self.includedMetrics.Has(container.DiskUsageMetrics)
```
dashpole: Test failed because I forgot to remove the `!` here.

bboreham:
Do you need someone to port this into kubelet?

Issue #1730, opened 2017-08-24
Whitelist labels


cirocosta:
Hey,

I'm new to using `cadvisor` together with `InfluxDB`  but from what I understood we can't specify a set of labels to be retrieved from containers and sent to influx - all of them are sent. 

In my use case that's a problem as we don't control the full set of labels that a user can attach to his/her container (whilst we only need one or two labels that our platform defines). 

I noticed that for Prometheus there's a way of programmatically specifying the set of labels to be retrieved (https://github.com/google/cadvisor/pull/1429) but I think something more like `-docker_env_metadata_whitelist string` could be useful for not only `InfluxDB` users but also those using other storage drivers which are sending possibly more labels than the desired. 

I couldn't find another issue about this so I think it's not a duplicate (please let me know if this is incorrect!)

I'd be willing to implement this if desired 👍  

Wdyt?

Thx!

dashpole:
That seems like a reasonable use case.  Feel free to post a PR.  My main request is that it defaults to the current behavior.

dashpole:
fixed by #1984

Issue #1951, opened 2018-05-30
cadvisor assigns labels to prometheus metrics with empty values


jaloren:
when cadvisor, exports metrics for docker containers, there is a root cgroup (/) and cgroup for a docker container (/docker/uuid). If docker container has a label on it, then this label is applied to all containers including the root container. Because some containers don't have that label, the label will have an empty value.

For example, lets say I have a python flask app in a docker container and I have labelled it in docker with the label python_version and the value 3.6. The root container will also get this metric label except it will be python_version=""

It looks like the cause of this is the code here. It loops through all the containers, getting labels for each container and then combines all of those labels into a single slice.

https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L853

I can't think of any valid usecase in which a user would want labels with no values, so I am wondering if its a bug that a label with an empty value is being applied to the metric.

dashpole:
prometheus does not allow sending a metric with the same name, but different labels.  So we had to "fake" it by adding empty label values based on the set of all labels for a given metric.

dashpole:
By "in the metric" I presume you mean in the metric name?  I think this goes against the prometheus model for metrics.
Labels are a big reason why prometheus became popular.   Older metric systems used to make the naming of the metric contain information about where it came from: e.g. you would have a metric named: `cluster_a.namespace_b.pod_c.container_d.memory_usage`.  You end up with a tree of metrics.  But this confines you to a single hierarchy.  Someone who comes along and wants an infrastructure based tree would want `cluster_a.node_b.pod_c.container_d.memory_usage` instead, so they can aggregate by node, rather than namespace.  But you can only get one of the two with any tree.  Labels allow arbitrary groupings, and can meet many different needs without conforming to any hierarchy.

jaloren:
Fair point. Okay, let me take this from a different angle. 

Is there a way to filter what docker labels are turned into prometheus labels or alternatively simply turn "convert docker labels int prom labels" off? I use a large number of docker labels that vary a lot across images so the set of labels will be enormous, where most of the labels will be empty. 

In my case, labels provide a useful metadata set for a variety of reasons outside of prometheus and many of them are of questionable value as a prometheus label.

In other words, could this func get some filtering options?

```golang
func DefaultContainerLabels(container *info.ContainerInfo) map[string]string {}
```

dashpole:
Sorry for the slow response, but this is definitely something we should tackle.

dashpole:
[...]

PR #1981, opened 2018-06-27
Add prometheus label whitelisting

dashpole:
[Prometheus best practices state](https://prometheus.io/docs/practices/naming/#labels) `Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.`.  We should encourage users to follow this by allowing label whitelisting.  This is initially only for prometheus, but flags are intentionally non-specific to allow for other storage plugins to make use of this as well.  The default behavior is unchanged, as `--enforce_storage_label_whitelist` is disabled by default.

This PR adds two new flags:
`enforce_storage_label_whitelist`, which turns on the whitelisting, and
`label_whitelist`, which is the set of labels allowed.

Related: #1730  
The daemonset (https://github.com/google/cadvisor/pull/1979) makes use of these flags so prometheus-to-sd doesn't exceed its label limit.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -29,6 +29,7 @@ import (
 	"github.com/google/cadvisor/container"
 	cadvisorhttp "github.com/google/cadvisor/http"
 	"github.com/google/cadvisor/manager"
+	"github.com/google/cadvisor/metrics"
 	"github.com/google/cadvisor/utils/sysfs"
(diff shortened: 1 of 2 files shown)


dashpole:
also related #1951

dashpole:
closing in favor of https://github.com/google/cadvisor/pull/1984

PR #1984, opened 2018-06-28
Control whether container labels are exported as prometheus metrics.

jaloren:
See #1951 for context.

when cadvisor exports metrics for docker containers, there is a root cgroup (/) and cgroup for a docker container (/docker/uuid).
If docker container has a label on it, then this label is applied to all containers including the root container. Because some containers don't have that label, the label will have an empty value. The reason for this is that Prometheus does not allow sending a metric with the same name, but different labels, so cadvisor uses empty label values based on the set of all labels for a given metric. This can result in many docker containers getting a large number of empty labels because another container has that label.

If large number of docker labels vary a lot across images, then the set of labels will be enormous, where most of the labels will be empty and have no value as prometheus metrics. To avoid this problem, a flag is provided that allows a user to disable exporting docker labels as metrics.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -35,6 +35,7 @@ import (
 	"crypto/tls"
 
 	"github.com/golang/glog"
+	"github.com/google/cadvisor/metrics"
 )
 
 var argIp = flag.String("listen_ip", "", "IP to listen on, defaults to all IPs")
@@ -57,6 +58,8 @@ var enableProfiling = flag.Bool("profiling", false, "Enable profiling via web in
 
 var collectorCert = flag.String("collector_cert", "", "Collector's certificate, exposed to endpoints for certificate based authentication.")
 var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
+
+var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
 
 var (
 	// Metrics to be ignored.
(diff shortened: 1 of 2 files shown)


jaloren:
I signed it!

cadvisor.go
```diff
@@ -58,6 +59,8 @@ var enableProfiling = flag.Bool("profiling", false, "Enable profiling via web in
 var collectorCert = flag.String("collector_cert", "", "Collector's certificate, exposed to endpoints for certificate based authentication.")
 var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
 
+var dockerLabels = flag.Bool("docker_labels", true, "convert docker labels and environment variables into labels on prometheus metrics for each docker container. If flag set to false, then only metrics exported are container name, first alias, and image name")
```
dashpole: This isn't limited to docker labels (as it works for other container runtimes that expose labels), lets rename this to `store_container_labels` or something similar.

cadvisor.go
```diff
@@ -58,6 +59,8 @@ var enableProfiling = flag.Bool("profiling", false, "Enable profiling via web in
 var collectorCert = flag.String("collector_cert", "", "Collector's certificate, exposed to endpoints for certificate based authentication.")
 var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
 
+var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
```
jaloren: @dashpole flag name store_container_labels seems unnecessarily verbose for someone executing this on a command line, especially since its a boolean. But to be consistent with the other flags, I followed the "flag name should match variable name but Pascal case instead of underscores"

Issue #1970, opened 2018-06-19
AllDockerContainers failed as cache not synced


lichuqiang:
Our upper components call cadvisor to get all container information regularly.
And upon container deletion, we found it has a chance to fail with a status code of 500.

Looking into the code, in `AllDockerContainers`:
```
func (self *manager) AllDockerContainers(query *info.ContainerInfoRequest) (map[string]info.ContainerInfo, error) {
	containers := self.getAllDockerContainers()

	output := make(map[string]info.ContainerInfo, len(containers))
	for name, cont := range containers {
		inf, err := self.containerDataToContainerInfo(cont, query)
		if err != nil {
			return nil, err
		}
		output[name] = *inf
	}
	return output, nil
}
```

we get the  containers from manager.containers, and in `containerDataToContainerInfo`, we query stats from its memoryCache. there is a chance we get the containers list before a container deleted, and when querying `containerDataToContainerInfo`, a container has already been removed from the memoryCache. as a result, we got error.

**What expected to happen**:
Avoid the cache not synced issue.
Or at least avoid returning with a status code 500 roughly, maybe log the error, and return the data of available containers.

dashpole:
What is the entry-point of your upper components?  Do they call `AllDockerContainers` directly, or are they making an API call?

lichuqiang:
> What is the entry-point of your upper components? Do they call AllDockerContainers directly, or are they making an API call?

api call: http://127.0.0.1:4194/api/v1.2/docker
i.e: https://github.com/google/cadvisor/blob/master/api/versions.go#L198

dashpole:
Sounds like it would be better to log a warning, and return a best-effort response.  Feel free to send me a PR, and ill review it.

PR #1985, opened 2018-06-30
Return best-effort result when AllDockerContainers hits the cache not synced issue

lichuqiang:
Fix: https://github.com/google/cadvisor/issues/1970

/assign @dashpole

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -628,6 +628,11 @@ func (self *manager) AllDockerContainers(query *info.ContainerInfoRequest) (map[
 	for name, cont := range containers {
 		inf, err := self.containerDataToContainerInfo(cont, query)
 		if err != nil {
+			// Ignore the error because of race condition and return best-effort result.
+			if err == memory.ErrDataNotFound {
+				glog.Warningf("Error getting data for container %s because of race condition", name)
+				continue
+			}
 			return nil, err
 		}
 		output[name] = *inf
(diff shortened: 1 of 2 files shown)


lichuqiang:
I signed it!

api/versions.go
```diff
@@ -216,7 +216,10 @@ func (self *version1_2) HandleRequest(requestType string, request []string, m ma
 			// Get all Docker containers.
 			containers, err = m.AllDockerContainers(query)
 			if err != nil {
-				return fmt.Errorf("failed to get all Docker containers with error: %v", err)
+				if len(containers) == 0 {
```
dashpole: i'm not sure I understand why we return an error here.  Are we trying to distinguish between when docker is entirely down vs when we miss a container because of a race?  Is there a way to do that without relying on the number of containers?  If we only had 1 container on the node, then we still get 500s when we shouldn't.

lichuqiang: Rollback the change here, and validate error content in `AllDockerContainers` explicitly

manager/manager.go
```diff
@@ -628,6 +628,11 @@ func (self *manager) AllDockerContainers(query *info.ContainerInfoRequest) (map[
 	for name, cont := range containers {
 		inf, err := self.containerDataToContainerInfo(cont, query)
 		if err != nil {
+			// Ignore the error because of race condition and return best-effort result.
+			if strings.Contains(err.Error(), "unable to find data in memory cache") {
```
dashpole: Define this error as a new type of error, and then use type assertion here, rather than string parsing.

lichuqiang: done

PR #1987, opened 2018-07-03
Fix build errors for mesos

dashpole:
This fixes errors when building with `godep go build .`
I'm not sure why we didn't catch this in #1965, which added the initial implementation.  Is it possible 

@sashankreddya, can you review this?

### DIFF
--- a/container/mesos/client.go
+++ b/container/mesos/client.go
@@ -100,7 +100,8 @@ func (self *client) getContainer(id string) (*mContainer, error) {
 	// Check if there is a container with given id and return the container
 	for _, c := range cntrs.Containers {
 		if c.ContainerID.Value == id {
-			return &c, nil
+			container := mContainer(c)
+			return &container, nil
 		}
 	}
 	return nil, fmt.Errorf("can't locate container %s", id)
--- a/container/mesos/mesos_agent.go
+++ b/container/mesos/mesos_agent.go
@@ -30,7 +30,7 @@ const (
 )
 
 type mContainers *agent.Response_GetContainers
-type mContainer = agent.Response_GetContainers_Container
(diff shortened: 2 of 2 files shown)


dashpole:
I found this in the failure for https://github.com/google/cadvisor/pull/1979

container/mesos/client.go
```diff
@@ -100,7 +100,8 @@ func (self *client) getContainer(id string) (*mContainer, error) {
 	// Check if there is a container with given id and return the container
 	for _, c := range cntrs.Containers {
 		if c.ContainerID.Value == id {
-			return &c, nil
+			container := mContainer(c)
```
sashankreddya: make doesn't complain any error with fresh checkout of the repo. Not sure why this is failing

dashpole:
The failure was on RHEL here: https://148.100.110.186/jenkins/job/cAdvisor_RHEL7_s390x_GHPRB/255/console

dashpole:
+ godep go build .
# github.com/google/cadvisor/container/mesos
container/mesos/mesos_agent.go:33: syntax error: unexpected = in type declaration
godep: go exit status 2

dashpole:
Apparently type aliases don't exist in go versions before go 1.9.  We need to update the jenkins builds to use a more recent go version.

sashankreddya:
type alias is introduced in go.1.9(https://github.com/golang/proposal/blob/master/design/18130-type-alias.md)

sashankreddya:
@dashpole : So this means that we don't need to commit this change, but update the jenkins go version ?

dashpole:
closing in favor of https://github.com/google/cadvisor/issues/1988

PR #2000, opened 2018-07-13
Allow on-demand container metrics in prometheus endpoint

### LINKED ISSUE #1989: Allow on-demand houskeeping for Prometheus scraping
We use cAdvisor with Prometheus. In this scenario you usually only care about metrics at exact moment of scraping and don't bother with intermediate results.

If you scrape your metrics every 60s, cAdvisor still pulls metrics from `sysfs` and whatnot every 1s, that's 60x overhead in the worst case (dynamic housekeeping may help a bit). With increased houskeeping interval we run into issues when prometheus scrapes can be misaligned with housekeeping and produce sawtooth.

For example, if your housekeeping interval is 30s, you can end up with either 1 or 2 updates per scrape depending on your luck. Something I found in the old tickets: `one` and `two` are containers that constantly burn a single CPU each, `USER` is from `node_exporter`.

image

I suggest to add a mode where cAdvisor only does on-demand housekeeping when asked for Prometheus metrics. This way we can lose short-lived containers, but that's something we can accept. For machines running lots of long-lived containers this should provide substantial efficiency benefits.

cc @dqminh

dqminh:
Fix #1989 

This does the following
- adds `max_age` parsing to `getRequestOptions`.
- expose `GetRequestOptions` so we can use it in prometheus handler.
- update prometheus `infoProvider` interface to use `GetRequestedContainersInfo`
so we can leverage `v2.RequestOptions`, namely `MaxAge` property to signal that we
are interested in on-demand collection of container metrics.

Signed-off-by: Daniel Dao <dqminh89@gmail.com>

### DIFF
--- a/http/handlers.go
+++ b/http/handlers.go
@@ -96,13 +96,22 @@ func RegisterHandlers(mux httpmux.Mux, containerManager manager.Manager, httpAut
 // the provided HTTP mux to handle the given Prometheus endpoint.
 func RegisterPrometheusHandler(mux httpmux.Mux, containerManager manager.Manager, prometheusEndpoint string,
 	f metrics.ContainerLabelsFunc, includedMetrics container.MetricSet) {
-	r := prometheus.NewRegistry()
-	r.MustRegister(
-		metrics.NewPrometheusCollector(containerManager, f, includedMetrics),
-		prometheus.NewGoCollector(),
-		prometheus.NewProcessCollector(os.Getpid(), ""),
-	)
-	mux.Handle(prometheusEndpoint, promhttp.HandlerFor(r, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}))
+	mux.Handle(prometheusEndpoint, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		opt, err := api.GetRequestOptions(req)
+		if err != nil {
+			http.Error(w, "No metrics gathered, last error:\n\n"+err.Error(), http.StatusInternalServerError)
+			return
+		}
+		opt.Count = 1        // we only want the latest datapoint
+		opt.Recursive = true // get all child containers
+		r := prometheus.NewRegistry()
+		r.MustRegister(
+			metrics.NewPrometheusCollector(containerManager, f, includedMetrics, opt),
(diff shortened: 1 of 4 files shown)


dqminh:
ping @dashpole

http/handlers.go
```diff
@@ -96,13 +96,22 @@ func RegisterHandlers(mux httpmux.Mux, containerManager manager.Manager, httpAut
...
 func RegisterPrometheusHandler(mux httpmux.Mux, containerManager manager.Manager, prometheusEndpoint string,
 	f metrics.ContainerLabelsFunc, includedMetrics container.MetricSet) {
-	r := prometheus.NewRegistry()
-	r.MustRegister(
-		metrics.NewPrometheusCollector(containerManager, f, includedMetrics),
-		prometheus.NewGoCollector(),
-		prometheus.NewProcessCollector(os.Getpid(), ""),
-	)
-	mux.Handle(prometheusEndpoint, promhttp.HandlerFor(r, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}))
+	mux.Handle(prometheusEndpoint, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
dashpole: Does this recreate the PrometheusCollector each time a request comes in?  Can we avoid this?

dqminh: Sorry for the my own delay :cry: !
Yes it does. In general i think its not costly, but i think we can also just initialize the collector outside of the handler, register with the registry, and then customize the option ( perhaps with new exported option like `func (c *PrometheusCollector) SetRequestOptions(opt)` before running `promhttp.HandlerFor` which triggers `Collect` for each registered collector.

dashpole: I would prefer a solution that does something like that to the current solution

rbastic:
Any update on merging this?

dashpole:
I'm going to close this as inactive.  Feel free to reopen if you are still working on this.

xjewer:
cc @dqminh are u going to finish it?

dqminh:
I'm going to spend some cycle to resurrect this. @dashpole is this still something you want to see merged ?

dashpole:
I think this is reasonable to have.

PR #2004, opened 2018-07-19
Check the length before using container.Stats

wgliang:
Issue detail is https://github.com/kubernetes/kubernetes/issues/66366

I found the only code is,

```
// Now for the actual metrics
stats := container.Stats[0]
```


wgliang:
Who can take a look?  thanks!

dashpole:
I can take a look

PR #2010, opened 2018-07-27
Ensure that labels for mesos tasks launched via MesosCommandExecutor are collected

sashankreddya:
This is an add on to issue 1965

### DIFF
--- a/container/mesos/factory.go
+++ b/container/mesos/factory.go
@@ -107,14 +107,9 @@ func (self *mesosFactory) CanHandleAndAccept(name string) (handle bool, accept b
 	// Check if the container is known to mesos and it is active.
 	id := ContainerNameToMesosId(name)
 
-	c, err := self.client.ContainerInfo(id)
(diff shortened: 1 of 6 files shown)


dashpole:
let me know when you would like this merged (or if you are waiting for review from anyone else).

sashankreddya:
sure David. Will let you know by EOD tomorrow.

container/mesos/factory.go
```diff
@@ -107,16 +107,11 @@ func (self *mesosFactory) CanHandleAndAccept(name string) (handle bool, accept b
 	// Check if the container is known to mesos and it is active.
 	id := ContainerNameToMesosId(name)
 
-	c, err := self.client.ContainerInfo(id)
+	_, err = self.client.ContainerInfo(id)
```
dashpole: do we still need to make this call?

sashankreddya: Yes as part of factory, we need to see if the container info object  can be created with the given id.

Issue #2008, opened 2018-07-24
cAdvisor reports incorrect `container_memory_cache` metrics


vboginskey:
### Environment
OS: CoreOS 1745.7.0
Kernel: 4.14.48-coreos-r2
systemd: 238
cAdvisor: `google/cadvisor:latest`

### Summary
cAdvisor seems to report incorrect metrics for `container_memory_cache` with nested cgroups. `container_memory_rss` is rolled up from child cgroups, whereas `container_memory_cache` is not. This succinctly illustrates the problem:

```
container_memory_cache{container_name="",id="/podruntime.slice",image="",name="",namespace="",pod_name=""} 0
container_memory_cache{container_name="",id="/podruntime.slice/podruntime-kubelet.slice",image="",name="",namespace="",pod_name=""} 0
container_memory_cache{container_name="",id="/podruntime.slice/podruntime-kubelet.slice/kubelet.service",image="",name="",namespace="",pod_name=""} 1.3787136e+09
container_memory_cache{container_name="",id="/podruntime.slice/podruntime-runtime.slice",image="",name="",namespace="",pod_name=""} 0
container_memory_cache{container_name="",id="/podruntime.slice/podruntime-runtime.slice/containerd.service",image="",name="",namespace="",pod_name=""} 3.231744e+06
container_memory_cache{container_name="",id="/podruntime.slice/podruntime-runtime.slice/docker.service",image="",name="",namespace="",pod_name=""} 1.7538719744e+10
```

```
container_memory_rss{container_name="",id="/podruntime.slice",image="",name="",namespace="",pod_name=""} 2.53140992e+08
container_memory_rss{container_name="",id="/podruntime.slice/podruntime-kubelet.slice",image="",name="",namespace="",pod_name=""} 9.3331456e+07
container_memory_rss{container_name="",id="/podruntime.slice/podruntime-kubelet.slice/kubelet.service",image="",name="",namespace="",pod_name=""} 9.3331456e+07
container_memory_rss{container_name="",id="/podruntime.slice/podruntime-runtime.slice",image="",name="",namespace="",pod_name=""} 1.59809536e+08
container_memory_rss{container_name="",id="/podruntime.slice/podruntime-runtime.slice/containerd.service",image="",name="",namespace="",pod_name=""} 6.6609152e+07
container_memory_rss{container_name="",id="/podruntime.slice/podruntime-runtime.slice/docker.service",image="",name="",namespace="",pod_name=""} 9.3134848e+07
```

[...]

PR #2014, opened 2018-08-04
Fix cache reporting with cgroup hierarchy

vboginskey:
This makes the behavior of `container_memory_cache` consistent with `container_memory_rss` for hierarchical cgroups. See https://github.com/google/cadvisor/issues/2008 for more detail.


vboginskey:
I signed the CLA.

Issue #2020, opened 2018-08-15
prometheusCollector if condition error


zhangjianweibj:
![image](https://user-images.githubusercontent.com/18585129/44135546-0db04ba2-a09c-11e8-8be2-26df3dd2755c.png)

PrometheusCollector if condition error.This function prepare to judge the meric type is networkUdpMetrics or not .

PR #2021, opened 2018-08-15
networkUdpMetrics error if condition

zhangjianweibj:
https://github.com/google/cadvisor/issues/2020

prometheusCollector if condition error


zhangjianweibj:
I signed it!

PR #2028, opened 2018-08-21
add start and end time range to prometheus metrics api

ringtail:
In kubernetes,prometheus will scrapy metrics from `/api/v1/nodes/${1}/proxy/metrics/cadvisor`, cadvisor metrics api collect SubcontainersInfo without time range currently. If the metrics has not been  collected for some reasons, the metrics api will return the old metrics values then prometheus will get some more older metrics, it will cause some troubles when aggregating data especially for cpu_usage_rate. we always get higher `cpu_usage_rate` because of older metrics values.

ladpbbcc1raiwqxm180cja_548_215 jpg_620x10000q90g
lalpbbcc1rkn8tvnalvnb1e_1873_603 png_620x10000q90g

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -26,6 +26,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	defaultValidTimeRange = 3 * 10 * time.Second // 3 times of defaultHousekeepingInterval
+)
+
 // infoProvider will usually be manager.Manager, but can be swapped out for testing.
 type infoProvider interface {
 	// SubcontainersInfo provides information about all subcontainers of the
@@ -907,7 +911,17 @@ func BaseContainerLabels(container *info.ContainerInfo) map[string]string {
 }
 
 func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric) {
-	containers, err := c.infoProvider.SubcontainersInfo("/", &info.ContainerInfoRequest{NumStats: 1})
+
+	now := time.Now()
+	start := now.Truncate(defaultValidTimeRange)
+	end := start.Add(defaultValidTimeRange)
+
+	containers, err := c.infoProvider.SubcontainersInfo("/", &info.ContainerInfoRequest{
+		Start:    start,
+		NumStats: 1,
+		End:      end,
+	})
+
 	if err != nil {
 		c.errors.Set(1)
 		glog.Warningf("Couldn't get containers: %s", err)


ringtail:
Here is the result after the fix.

e9bce8a6-5559-410e-9c1a-d7d657fb4aba

dashpole:
That seems like a bug, rather than something we should work around.  Any idea why it chooses a data point further back in time, rather than just the most recent data point?

ringtail:
@dashpole real-time data point may not be a good idea because of it's time cost. In heapster,It scrapy metrics with a specific time range based on `metric_resolution` like this.
lalpby0v5ewdgjvnabhnazo_826_433 png_620x10000q90g

Because the most of aggregating methods will not be aware of the metrics size.If the metrics need divide some number (core number) then it will aggregate a much more confused value(such as: over 100% cpu_usage_rate on each core).

The metrics api should not return the out of date metrics. Three times of `defaultHousekeepingInterval` is reasonable because the official scraping interval is 30s in prometheus. Less than 30s is strongly not recommended. So,the prometheus api will guarantee the real-time and accuracy of metrics.

dashpole:
> The metrics api should not return the out of date metrics.

@ringtail I agree.  The question is: _Why_ is cAdvisor is returning out-of-date metrics?  The current query should return the most recent metric for each container.  That should never go backwards in time for any reason.  While your change makes the problem go away, it isn't clear (at least to me) why it fixes the issue. It means it is likely to happen again in the future.

Before we move forward with this change, I would like to figure out why metrics returned by the prometheus endpoint go backwards in time.

ringtail:
@dashpole I agree with you and I'll help to find the reason of out-of-date metrics.

ringtail:
@dashpole #2059 I found another reasonable way to avoid this problem.

Betulle:
@ringtail @dashpole  I don't think this PR is not a proper solution for your problem. 

Actually, it's due to inconsistency between `Prometheus scraped timestamp from /metrics` with `cadvisor scraped timestamp from cgroup`. The problem is, **CAdvisor do not expose /metrics properly.**

**CAdvisor should expose data in /metrics with timestamp** as [documentation](1).

**Expected:**
`container_cpu_system_seconds_total{container_name="",id="/kubepods",image="",name="",namespace="",pod_name=""} 6.21201942e+06 1395066363000`
**Current:**
`container_cpu_system_seconds_total{container_name="",id="/kubepods",image="",name="",namespace="",pod_name=""} 6.21201942e+06`

Precise timestamp is essential for `type Counter` metrics.
 
Supposed that 
- CAdvisor scraped `6.21201942e+06` from cgroup at `1395066363000 2014/3/17 22:26:3`
- Prometheus scraped the same data `6.21201942e+06` from CAdvsior /metrics at `1395066395000 2014/3/17 22:26:35`

  an inconsistency exist if without exposing metrics  timestamp.
 


[1]: https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-example

ringtail:
@Betulle I fixed it in prometheus node exporter. prometheus collect metrics from kubelet(cadvisor) with timestamp. So that the rate calculation is not correct when cadvisor can not provide latest metrics in time. The code is here(https://github.com/ringtail/node_exporter/blob/feature/fix_usage_rate/collector/container_rate_stats.go).

Betula-L:
@ringtail i mean, the root cause is the gap of timestamp between cadvisor and prometheus. "Only add rate metrics" is not proper way to fix this bug.

If exposed exact cadvisor timestamp for prometheus as  [documentation](https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-example) , to fix this gap, we do not need add this rate metrics.

ringtail:
@Betula-L Yes，this pr may not be the solution and I find any way to fix it. But I don't know how to find the root cause and fix the problem in cadvisor

dashpole:
This is fixed by https://github.com/google/cadvisor/pull/2124

PR #2030, opened 2018-08-23
enhance numa-aware topology to show node/core/thread/cache/memory/pci layout

lubinszARM:
The purpose of this patch is to display all information related to numa.
Just like the output of command ‘lstopo'. 
With this patch, I can enhance the numa feature in k8s. Such as, I can configure the right numa for gpu/sriov/cpuset resources.
Please see the output of 'lstopo' as reference:
image


lubinszARM:
@vishh 
Can you check it?
Thanks.

vishh:
If this change is primarily meant to make cAdvisor expose system topology for introspection purposes, then I can go ahead and review this PR.
If this PR is meant to tackle HW topology in k8s, then I'd suggest holding off this PR until we get to finalize the overall design on k8s side. 
For example, we currently expect PCI device topology to be exposed relative to CPU sockets by device plugins. So I'm not sure for cAdvisor data will help.
Exposing adequate CPU and memory topology from cAdvisor will be useful though if the current information isn't adequate.

lubinszARM:
hi @vishh 
The current data of topology in cAdvisor is not consistent with the output of 'lstopo'.
This is the main purpose of my patch. I want to fix it in this PR.

Also, the difference between the contents of ’/proc/cpuinfo' on x86/arm is very large.
On Arm platform, there are no 'core id' and 'physical id' in '/proc/cpuinfo'. So it means that the topology of cadvisor is not working on Arm platform.
I will fix it by checking '/sys/bus' to get the whole topology in next PR: the same method in 'lstopo'.
For your reference, we can get the correct topology to show Node/Core/Thread/Pci layout by checking following sysfs:
const nodeBusPath = "/sys/bus/node/devices/"
const cpuBusPath   = "/sys/bus/cpu/devices/"
const pciBusPath     = "/sys/bus/pci/devices/"
Also , please see a typical '/proc/cpuinfo' on Arm platform:
...
processor       : 222
BogoMIPS        : 400.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics cpuid asimdrdm
CPU implementer : 0x43
CPU architecture: 8
CPU variant     : 0x1
CPU part        : 0x0af
CPU revision    : 0

processor       : 223
BogoMIPS        : 400.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics cpuid asimdrdm
CPU implementer : 0x43
CPU architecture: 8
CPU variant     : 0x1
CPU part        : 0x0af
CPU revision    : 0

info/v1/machine.go
```diff
@@ -41,6 +41,7 @@ type Node struct {
 	Memory uint64  `json:"memory"`
 	Cores  []Core  `json:"cores"`
 	Caches []Cache `json:"caches"`
+	Pcis   []string   `json:"pcis"`
```
vishh: What is this field supposed to contain? List of PCI IDs of devices? 
Also add to v2 API as well.

vishh:
I see you are adding a list of PCI device IDs per NUMA node. Is that adequate or should each device hold more metadata than just it's PCI ID?

lubinsz:
Hi @vishh 
I think the information is enough.
If we want to know which node the devices(net,gpu,fpga) belongs to, we can  find the pci device address of this device firstly, then check the list of PCI device IDs per NUMA node  which was provided by cAdvisor.

Such as:
step1, get the 'PCI_SLOT_NAME' of the network deivce.

root@entos-thunderx2-02:/sys/class/net/enp11s0f1# cat /sys/class/net/enp11s0f1/device/uevent
DRIVER=qede
PCI_CLASS=20000
PCI_ID=1077:8070
PCI_SUBSYS_ID=1458:0011
PCI_SLOT_NAME=0000:0b:00.1
MODALIAS=pci:v00001077d00008070sv00001458sd00000011bc02sc00i00

step2, check the pci list, and get the value of 'numa_node'

root@entos-thunderx2-02:~# cat /sys/bus/pci/devices/0000\:0b\:00.1/numa_node
1

lubinsz:
@dashpole @vishh 
Any comments on this?
Thanks.

info/v1/machine.go
lubinsz: Yes. This field contains a list of PCI IDs.
Such as: [ "0000:00:00.0", "0000:00:00.1", "0000:00:01.0"].
You can see the test cases in "machine/topology_test.go" as reference.

The field of 'Pcis' was belong to struct "Topology".
The struct of 'Topology' was already added into v2 API (.info/v2/machine.go).
I think this field can be obtained by accessing the topology in v2 API.
So 'Pcis' should not be added inot v2 API.
I don't know if my understanding is wrong.
Thanks.

lubinsz:
ping @vishh

PR #2031, opened 2018-08-24
Add kubernetes metadata as prometheus labels

dashpole:
This adds three new labels to prometheus metrics for kubernetes-managed containers:
 * pod_name
 * pod_id
 * pod_namespace

While this technically relies on a kubelet implementation detail (the format of the container label used to store this information), this information would be very useful for those using cAdvisor with kubernetes.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -17,6 +17,7 @@ package metrics
 import (
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/google/cadvisor/container"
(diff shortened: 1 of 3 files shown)


brancz:
How about we make this a runtime configuration? As in we can pass via flags mappings of labels, like:

```
--prometheus-label-mapping=io.kubernetes.pod.name=pod_name
```

At least that's how I read Kubernetes essentially modifying labels: https://github.com/kubernetes/kubernetes/blob/ae8a046985d1e768dc589509d584b6fef13e900a/pkg/kubelet/cadvisor/cadvisor_linux.go#L75-L101

The way I understand it (and I may very well be wrong) the `io.kubernetes.pod.name` label is a generic docker label.

dashpole:
> The way I understand it (and I may very well be wrong) the io.kubernetes.pod.name label is a generic docker label.

It is a generic docker label.  But it is added by the kubelet for the purpose of generating pod status after a pod restart (so it can identify the pod associated with a container).  That is why we consider it an internal implementation detail.  We could add the mechanism you suggest, although that would just be encouraging users to rely on the internal implementation detail.

I am going to close this for now while I try and come up with a better way to do this.

brancz:
Happy to discuss a way forward. :slightly_smiling_face:

Issue #2016, opened 2018-08-09
MemoryStats does not expose mapped_file stat


stevenjm:
Hi there!

According to the [Linux kernel documentation](https://github.com/torvalds/linux/blob/v4.9/Documentation/cgroup-v1/memory.txt#L523-L530):

```
    Only anonymous and swap cache memory is listed as part of 'rss' stat.
    This should not be confused with the true 'resident set size' or the
    amount of physical memory used by the cgroup.
    'rss + file_mapped" will give you resident set size of cgroup.
    (Note: file and shmem may be shared among other cgroups. In that case,
     file_mapped is accounted only when the memory cgroup is owner of page
     cache.)
```

(The relevant field seems to actually be named `mapped_file`, not `file_mapped`.)

However, [setMemoryStats](https://github.com/google/cadvisor/blob/18afaa921b8dd8bc3bf5e6f95dd27a4fee7cf6bb/container/libcontainer/handler.go#L497-L528) does not expose the `mapped_file` stat, and there is no attribute for this in the [MemoryStats](https://github.com/google/cadvisor/blob/18afaa921b8dd8bc3bf5e6f95dd27a4fee7cf6bb/info/v1/container.go#L338-L370) type. If I understand correctly, this means that it is impossible to accurately calculate the resident set size of a cgroup based on the stats returned by cadvisor.

Please correct me if I'm wrong and there is a better way to do this. Otherwise, I'm happy to work on a PR if this is a desired feature.

PR #2034, opened 2018-08-28
Adds mapped_file metric

gwkunze:
This is a simple fix for #2016.


dashpole:
Looks good.  Please add this to the prometheus table: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

PR #2043, opened 2018-09-13
elasticsearch storage driver: update to v6

mannychang:
update elasticsearch lib to v6


dashpole:
I don't believe the v2 and v6 storage drivers are compatible with each other.  This would be a backwards incompatible change.

Issue #1483, opened 2016-09-27
Make container providers opt-in


timstclair:
The manager should take a list of container handlers to install. The `--docker-only` flag should be deprecated.

EDIT: Repurposed issue. Was:
> We've seen a few issues around the current mechanism that attempts to connect to the rkt service unconditionally (e.g. https://github.com/google/cadvisor/issues/1481), and it also produces some log noise.

> Is there a better way to detect whether rkt is installed on the host before attempting to connect? Maybe checking for a specific process, or looking for the binary?

euank:
Checking for the binary won't work if it's not in your path.

Is there a reason not to make rkt integration opt-in and have kubelet instantiate cadvisor with or without rkt support depending on the runtime it's configured with?

timstclair:
opt-in SGTM. Add a list of handlers to install to the manager interface, and maybe replace the `--docker-only` flag with a flag that takes a list of handlers (raw, docker, rkt for now).

amcrn:
any update on this? we're still stuck running v0.23.8 because of it. thanks!

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2055, opened 2018-09-26
Avoid too many mesos errors in logs when mesos is not present

dims:
We see a lot of logs in k/k CI as follows:
"Factory "mesos" was unable to handle container "/system.slice/home-kubernetes-containerized_mounter.mount"

Example, please see:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/69003/pull-kubernetes-node-e2e/105485/artifacts/tmp-node-e2e-9f0e4e26-cos-stable-63-10032-71-0/kubelet.log

It would be better if we do some sanity check for mesos running before
we try to use it.

Change-Id: I5f6ebcd44fdd4f8d724b85857edf1600473ef1ab

### DIFF
--- a/container/mesos/client.go
+++ b/container/mesos/client.go
@@ -70,6 +70,11 @@ func Client() (mesosAgentClient, error) {
 			),
 		}
 	})
+
+	_, err := mesosClient.getVersion()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get version")
+	}
 	return mesosClient, nil
 }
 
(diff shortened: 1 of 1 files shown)


dims:
cc @sashankreddya @dashpole

dims:
@dashpole yep, just filing it for next round of updates or if someone complains :)

sashankreddya:
> @dims thanks for catching this.
> 
> Ill let @sashankreddya review before I merge. I don't think this is worth a cherrypick, as it is mostly to reduce testing log volume...

Ideally we need to have and opt-in for every container type as mentioned in https://github.com/google/cadvisor/issues/1483 in the longterm

sashankreddya:
I am fine with the short-term solution too.

PR #2056, opened 2018-09-26
Use go1.10 for canary builds

dashpole:
The canary builds run with the latest go version, which has been failing since it was updated to 1.11:
https://k8s-testgrid.appspot.com/sig-node-cadvisor#cadvisor-push

Rather than making cadvisor work with 1.11, this changes the canary version to 1.10 (the same development version as k/k).  We can change it back to the `latest` tag after we move to 1.11, assuming 1.11 and 1.12 are compatible.

### DIFF
--- a/deploy/canary/Dockerfile
+++ b/deploy/canary/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM golang:1.10
 MAINTAINER vmarmol@google.com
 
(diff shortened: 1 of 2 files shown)


dashpole:
heh, our e2e tests and canary use different go versions...

dashpole:
I tested this using 
`docker build -t google/cadvisor:canary --no-cache=true --pull=true --file=deploy/canary/Dockerfile .`
to ensure the canary builds will succeed now.

PR #2059, opened 2018-09-28
add rate metrics to prometheus api 

ringtail:
prometheus use cadvisor to aggregate cpu_uage_rate with the metric `container_cpu_usage_seconds_total`(cumulative cpu time consumed in seconds) but cadvisor don't provide timestamp. So the cpu usage rate will  be not accuracy because of the deviation of the calculation of time range especially when data collection delayed.Here is a demo to show the impact.   
07f9cf92-e2fa-44d7-97ae-b37a446f8055
411bebf8-d29e-407c-aa77-6127630ad0f9
The first image is using  the metric `container_cpu_usage_seconds_total` and the second image is using the new metric `container_cpu_user_usage_rate`.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -87,12 +87,13 @@ func ioValues(ioStats []info.PerDiskStats, ioType string, ioValueFn func(uint64)
 // containerMetric describes a multi-dimensional metric used for exposing a
 // certain type of container statistic.
 type containerMetric struct {
-	name        string
-	help        string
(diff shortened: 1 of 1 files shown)


kongbo1987:
I  have test this patch,  It works well on my kubernetes cluster, Thanks!

dashpole:
I thought we agreed in the issue to find the root cause?  This seems like another workaround.  I prefer the original workaround to this one, as this one adds new metrics.

PR #2085, opened 2018-10-21
improve ListDirectories performance

orisano:
used github.com/karrick/godirwalk to reduces syscall.
```
name               old time/op    new time/op    delta
ListDirectories-2    17.4ms ± 9%     3.7ms ±15%  -78.88%  (p=0.000 n=9+9)

name               old alloc/op   new alloc/op   delta
ListDirectories-2    3.55MB ± 0%    0.48MB ± 0%  -86.36%  (p=0.000 n=9+10)

name               old allocs/op  new allocs/op  delta
ListDirectories-2     20.8k ± 0%     11.6k ± 0%  -44.13%  (p=0.000 n=10+10)
```


orisano:
What is this PR's status now?

orisano:
rebased

dashpole:
`The following files are not properly formatted: container/common/endian_little.go container/common/helpers_unix.go container/common/endian_little.go container/common/helpers_unix.go`

orisano:
sorry, my mistake.

orisano:
passed e2e

dashpole:
Is there a reason why you added a new repo and vendored it in here?  That might make it harder for us to make changes in the future.  Are there any other more commonly used implementations of this?

I'm also not sure if we can use BSD licenses or not.  Let me check...

orisano:
for licensing problem. my code based on golang/go private functions.
golang/go license is [here](https://github.com/golang/go/blob/master/LICENSE).

When I try to specify license on code, boilerplate check fails.
what will i do?

dashpole:
Is there a golang function we can use that isn't private?  Yeah, we keep all of the code in the repo that we don't own.  If it is code from golang, then the proper thing to do is keep it under their license and vendor it in.  If you can actually add a commit which is the exact copy of the original golang code, and then your seperate changes on top of that, that would be easier for me to review.  Also link to the original golang code if possible.

dashpole:
sorry for all the hassle :/

orisano:
## References
https://github.com/golang/go/blob/master/src/syscall/syscall_linux.go#L862-L880
https://github.com/golang/go/blob/master/src/os/dir_unix.go#L47-L88
https://github.com/golang/go/blob/master/src/syscall/dirent.go
https://github.com/golang/go/blob/master/src/syscall/endian_big.go
https://github.com/golang/go/blob/master/src/syscall/endian_little.go

vendor/github.com/orisano/fls/list_linux.go
```diff
@@ -0,0 +1,206 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func parseDirent(buf []byte, dirs []string) (consumed int, newdirs []string) {
```
dashpole: it looks like we could just use the golang implementation directly, since it is public? https://github.com/golang/go/blob/master/src/syscall/dirent.go#L64

vendor/github.com/orisano/fls/endian_big.go
```diff
@@ -0,0 +1,34 @@
...
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//    * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//    * Redistributions in binary form must reproduce the above
```
dashpole: I don't think you need to repaste the license into each file.  Having it as the license is good enough

vendor/github.com/orisano/fls/list_linux.go
orisano: parseDirent is edited for handle Dirent.Type.
original implementation can't handle Dirent.Type.

vendor/github.com/orisano/fls/endian_big.go
orisano: sorry, I don't know the right way of redistributing the licensed software.

dashpole:
So the proper way to do this is:

- make your own fork of the original go repo (the entire thing)
- make you changes to that fork
- vendor that fork into cAdvisor

You leave their license and boilerplate (header of the files), just as it is.
I'm not sure how licenses work if you are just copying files, so please use the above approach.

orisano:
to used github.com/karrick/godirwalk to resolve license problem.

orisano:
ping @dashpole

sashankreddya:
@orisano / @dashpole : Looks like after this commit I am unable to recognize "mesos" containers in the directory path "/sys/fs/cgroup/<cgroup-name>/mesos". I have verified without including this commit things work fine. 
I am debugging it, you guys could share any inputs you have.

orisano:
Is using symlink in "/sys/fs/cgroup/mesos" ?
currently, maybe unhandle symlink type.

by the way, Is "/sys/fs/cgroup//mesos" equals "/sys/fs/cgroup/mesos" ?

sashankreddya:
no symlinks are used and yes it should be the cgroups path is "/sys/fs/cgroup/<cpu,memory>/mesos"
I could see the error thrown by 
```
dirents, err := godirwalk.ReadDirents(dirpath, buf)
```
in listDirectories when "updateSubcontainers" tries to fetch the subcontainers.


Error
```
error(*github.com/google/cadvisor/vendor/github.com/pkg/errors.withStack) *{
	error: error(*github.com/google/cadvisor/vendor/github.com/pkg/errors.withMessage) *{
		cause: error(*os.PathError) ...,
		msg: "cannot Open",},
	stack: *github.com/google/cadvisor/vendor/github.com/pkg/errors.stack len: 15, cap: 32, [11136028,11135472,11164142,11163830,11167113,18958881,19150681,19113714,19174192,19183722,19178323,19514576,19512011,19484179,4585041],}
(dlv) n
```

orisano:
sorry, please show me the results of "err.Error()".

orisano:
@sashankreddya 
sorry, my mistake. fix by #2094 . please review

PR #2091, opened 2018-11-04
[WIP] Move from glog to klog

dims:
Moving to the new forked repository : k8s.io/klog


dims:
related to https://github.com/kubernetes/kubernetes/issues/70264

dims:
cc @dashpole @sigma

dashpole:
Note that this should wait until after 1.32.0 is cut

cadvisor.go
```diff
@@ -34,8 +34,8 @@ import (
 
 	"crypto/tls"
 
-	"github.com/golang/glog"
-	"github.com/google/cadvisor/metrics"
+		"github.com/google/cadvisor/metrics"
```
dashpole: this looks strange.  I am surprised it passes go lint?  If you put a new line between these, does it go away?

PR #2069, opened 2018-10-17
Emit number of processes and file descriptors of a container

sashankreddya:
#1995 #1994

PR #2097, opened 2018-11-09
Remove unused interface method

dashpole:
The interface was added in https://github.com/google/cadvisor/pull/2069, but then we took a different approach in https://github.com/google/cadvisor/pull/2093.  This is causing issues for the cAdvisor godependency update: https://github.com/kubernetes/kubernetes/pull/70829.

cc @dims @sashankreddya


dims:
+1 yes please let's do this.

dashpole:
merging based on https://github.com/google/cadvisor/pull/2097#issuecomment-437439429

PR #2101, opened 2018-11-16
Add example kustomization which gives cAdvisor access to NVML

dashpole:
I have gotten a bunch of questions about how to use containerized cAdvisor to monitor GPUs.  This should give _something_ for me to point at, even if it isn't perfect.

cc @mindprince 
/assign @tallclair

### DIFF
--- a/deploy/kubernetes/overlays/examples/gpu-privilages.yaml
+++ b/deploy/kubernetes/overlays/examples/gpu-privilages.yaml
@@ -1,0 +1,29 @@
+# This patch adds the required hostpath mounts and privilages to initialize NVML, as well as to access devices.
+# This patch meets the requirements described in the accellerator docs: 
+# https://github.com/google/cadvisor/blob/master/docs/running.md#hardware-accelerator-monitoring
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
+        volumeMounts:
+        - name: dev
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/overlays/examples/gpu-privilages.yaml
```diff
@@ -0,0 +1,27 @@
+# This patch adds the required hostpath mounts and privilages to initialize NVML, as well as to access devices.
+apiVersion: apps/v1 # for Kubernetes versions before 1.9.0 use apps/v1beta2
```
mindprince: I would skip the comment here.

deploy/kubernetes/overlays/examples/gpu-privilages.yaml
```diff
@@ -0,0 +1,27 @@
+# This patch adds the required hostpath mounts and privilages to initialize NVML, as well as to access devices.
```
mindprince: Maybe link to https://github.com/google/cadvisor/blob/master/docs/running.md#hardware-accelerator-monitoring from here?

deploy/kubernetes/overlays/examples/gpu-privilages.yaml
dashpole: done

deploy/kubernetes/overlays/examples/gpu-privilages.yaml
dashpole: done

Issue #2045, opened 2018-09-18
Reducing CAdvisor memory usage


SamSaffron:
We have noticed that resource usage (both CPU and memory) is less than ideal on cadvisor.

In particular on a machine with 40 or so containers we see cadvisor ramp up 100MB RSS fairly quickly. CPU is also highish.

From memory dumps I isolated the about 50% of memory usage is due to container disk IO stats:

```
(pprof) top50
Showing nodes accounting for 34021.71kB, 100% of 34021.71kB total
Showing top 50 nodes out of 73
      flat  flat%   sum%        cum   cum%
10754.13kB 31.61% 31.61% 14850.68kB 43.65%  github.com/google/cadvisor/container/libcontainer.DiskStatsCopy
 7680.15kB 22.57% 54.18%  7680.15kB 22.57%  bufio.(*Scanner).Text (inline)
 3074.62kB  9.04% 63.22% 17925.30kB 52.69%  github.com/google/cadvisor/container/libcontainer.newContainerStats
 2048.45kB  6.02% 69.24%  2048.45kB  6.02%  github.com/google/cadvisor/container/libcontainer.DiskStatsCopy1
 2048.09kB  6.02% 75.26%  2048.09kB  6.02%  github.com/google/cadvisor/container/libcontainer.DiskStatsCopy0 (inline)
 1536.23kB  4.52% 79.78%  1536.23kB  4.52%  github.com/google/cadvisor/vendor/github.com/aws/aws-sdk-go/aws/endpoints.init
 1184.27kB  3.48% 83.26%  1184.27kB  3.48%  bytes.makeSlice
 1030.64kB  3.03% 86.29%  1030.64kB  3.03%  regexp/syntax.(*compiler).inst (inline)
  561.50kB  1.65% 87.94%   561.50kB  1.65%  html.init
  516.01kB  1.52% 89.45%   516.01kB  1.52%  github.com/google/cadvisor/vendor/golang.org/x/net/trace.init
  514.63kB  1.51% 90.97%   514.63kB  1.51%  math/rand.NewSource
  512.56kB  1.51% 92.47%   512.56kB  1.51%  compress/flate.newHuffmanEncoder (inline)
  512.31kB  1.51% 93.98%  1073.81kB  3.16%  html/template.init
  512.05kB  1.51% 95.49%   512.05kB  1.51%  github.com/google/cadvisor/vendor/golang.org/x/exp/inotify.(*Watcher).readEvents
[...]

PR #2103, opened 2018-11-22
Add support to disable diskIO metrics

WanLinghao:
Add support to disable diskIO metrics
ref:https://github.com/google/cadvisor/issues/2045

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -66,8 +66,7 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
 	libcontainerStats := &libcontainer.Stats{
 		CgroupStats: cgroupStats,
 	}
-	withPerCPU := h.includedMetrics.Has(container.PerCpuUsageMetrics)
-	stats := newContainerStats(libcontainerStats, withPerCPU)
+	stats := newContainerStats(libcontainerStats, h.includedMetrics)
 
 	if h.includedMetrics.Has(container.ProcessSchedulerMetrics) {
 		pids, err := h.cgroupManager.GetAllPids()
@@ -599,14 +598,16 @@ func setNetworkStats(libcontainerStats *libcontainer.Stats, ret *info.ContainerS
 	}
 }
 
-func newContainerStats(libcontainerStats *libcontainer.Stats, withPerCPU bool) *info.ContainerStats {
+func newContainerStats(libcontainerStats *libcontainer.Stats, includedMetrics container.MetricSet) *info.ContainerStats {
 	ret := &info.ContainerStats{
 		Timestamp: time.Now(),
 	}
 
 	if s := libcontainerStats.CgroupStats; s != nil {
-		setCpuStats(s, ret, withPerCPU)
-		setDiskIoStats(s, ret)
+		setCpuStats(s, ret, includedMetrics.Has(container.PerCpuUsageMetrics))
+		if includedMetrics.Has(container.DiskIOMetrics) {
+			setDiskIoStats(s, ret)
+		}
 		setMemoryStats(s, ret)
 	}
 	if len(libcontainerStats.Interfaces) > 0 {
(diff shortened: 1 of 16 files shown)


WanLinghao:
@dashpole @SamSaffron  PTAL

WanLinghao:
I signed it!

WanLinghao:
I signed it!

WanLinghao:
@dashpole  CLA cert has been fixed, PTAL

container/common/helpers.go
```diff
@@ -258,6 +258,13 @@ func assignDeviceNamesToPerDiskStats(namer DeviceNamer, diskStats ...[]info.PerD
 	}
 }
 
+// FilterCgroupPaths delete cgroup paths which we do't need base on includeMetrics
+func FilterCgroupPaths(cgroups map[string]string, includedMetrics container.MetricSet) {
```
dashpole: rather than filtering out afterwards, I think it would be cleaner to do this when we first get the list of cgroup subsystems: https://github.com/google/cadvisor/blob/master/container/libcontainer/helpers.go#L37.

container/common/helpers.go
```diff
@@ -258,6 +258,13 @@ func assignDeviceNamesToPerDiskStats(namer DeviceNamer, diskStats ...[]info.PerD
...
+	if !includedMetrics.Has(container.DiskIOMetrics) {
```
dashpole: we don't currently do this for other ignored metrics.  What is the benefit of removing the blkio cgroup from the list of cgroups?

container/libcontainer/handler.go
```diff
@@ -599,14 +600,16 @@ func setNetworkStats(libcontainerStats *libcontainer.Stats, ret *info.ContainerS
 	}
 }
 
-func newContainerStats(libcontainerStats *libcontainer.Stats, withPerCPU bool) *info.ContainerStats {
+func newContainerStats(libcontainerStats *libcontainer.Stats, withPerCPU, withDiskIO bool) *info.ContainerStats {
```
dashpole: Now that we are passing in two of these, I would rather just pass ignoreMetrics to the function.

container/raw/handler.go
```diff
@@ -117,7 +120,8 @@ func (self *rawContainerHandler) Cleanup() {}
 
 func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	const hasNetwork = false
-	hasFilesystem := isRootCgroup(self.name) || len(self.externalMounts) > 0
+	hasFilesystem := (isRootCgroup(self.name) || len(self.externalMounts) > 0) &&
```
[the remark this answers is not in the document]
WanLinghao: It isn't related to diskIO. It is a DiskUsageMetrics bug that people will see `has_filesystem` be true even they disable `disk` metrics. I will extract it and launch a new patch @dashpole

container/common/helpers.go
WanLinghao: As far as I can see, if we don't remove the blkio cgroup, it would keep extracting data from /sys/fs/cgroup/blkio/(ref:https://github.com/google/cadvisor/blob/45d3b774d690cf0221f06fc799e62e5cd8715e58/vendor/github.com/opencontainers/runc/libcontainer/cgroups/fs/blkio.go#L115). And the data it extracted would be just dropped cause we have disabled the diskIO metrics. So here I am trying to do is avoiding useless IO request when diskIO was disabled.

WanLinghao:
@dashpole All comments addressed, PTAL

container/raw/handler.go
```diff
@@ -117,7 +119,7 @@ func (self *rawContainerHandler) Cleanup() {}
 
 func (self *rawContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	const hasNetwork = false
-	hasFilesystem := isRootCgroup(self.name) || len(self.externalMounts) > 0
+	hasFilesystem := (isRootCgroup(self.name) || len(self.externalMounts) > 0)
```
dashpole: nit: remove extra parenthesis

WanLinghao:
comments addressed @dashpole

PR #2104, opened 2018-11-23
Add white list cgroup path list

chenchun:
We only have a path prefix args previously, so we can't collect specific paths and blacklist the others under a prefix.

ref: https://github.com/kubernetes/kubernetes/issues/71367

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -141,7 +141,7 @@ func main() {
 
 	collectorHttpClient := createCollectorHttpClient(*collectorCert, *collectorKey)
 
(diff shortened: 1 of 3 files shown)


chenchun:
@dashpole I'd like to collect cgroup stats only from `/`  and `/kubepods`, but not `/system.slice/NetworkManager.service`,  how can I achieve this if cadvisor only supports`rawPrefixWhitelist`?
If I add `/` to `rawPrefixWhitelist`, it'll match every cgroup path.

chenchun:
Sorry, I missed `accept := name == "/" || !*dockerOnly`

PR #2108, opened 2018-11-29
Fix bug: getTopology will be failed on Arm platform if no 'core id' o…

lubinszARM:
Fix bug: getTopology will be failed on Arm platform if no 'core id' or 'physical id' in /proc/cpuinfoq

So we should search sysfs cpu path directly to get the data of 'thread_id' &'core_id' & 'node_id'. 
Such as: 


lubinszARM:
> Once you've signed (or fixed any issues), please

I signed it!

lubinszARM:
@vishh 
Hi, can you help to check it?
Thanks.

lubinszARM:
I signed it!

PR #2110, opened 2018-11-30
Fix bug:getTopology will be failed on Arm platform

lubinsz:
So we should search sysfs cpu path directly to get the data of
'thread_id' &'core_id' & 'node_id'.
Such as:

### DIFF
--- a/machine/machine.go
+++ b/machine/machine.go
@@ -16,8 +16,10 @@
 package machine
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


lubinsz:
@dashpole  @vishh     any comments on this?

dashpole:
Can you fix the CLA?  It looks like you are pushing a commit by someone else?

Issue #2109, opened 2018-11-29
Improve alpine container to use native libc6-compat


tiagoapimenta:
https://github.com/google/cadvisor/blob/cf21050b5906cce83527f5c531f0cd7d279a3da3/deploy/Dockerfile#L10

Instead of downloading a binary out of nowhere it would be better to use native package libc6-compat:

    apk add --no-cache libc6-compat

As well as removing `wget`, `certs` and `findutils` and other unnecessary tools after the download, or use native `wget` without https.

Other tip would be use multistage-build:
https://docs.docker.com/develop/develop-images/multistage-build/

So it wouldn't be necessary to remove `wget` and `certs` tools, it would just use another base container and copy the downloaded files from the first one.

dashpole:
Feel free to open a PR with these improvements.  If we use multistage build, we should add documentation about the docker version required to build.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2112, opened 2018-11-30
Removing external glibc dependency

tiagoapimenta:
It simplifies the `Dockerfile` as issued here https://github.com/google/cadvisor/issues/2109

Note that I changed `zfs` due to not working from `edge` repository, but IMO it could be even more simplified.


tiagoapimenta:
I signed it!

PR #2113, opened 2018-12-01
Add flag to white list container labels for prometheus metrics

mikkeloscar:
This adds a flag `white_listed_container_labels` which allows specifying
a comma separated list of container labels that are allowed to be
attached to prometheus metrics. This helps limit the label space of
each data point and limits the memory used by cadvisor compared to just
add all labels from all containers which is the default.

This is somewhat similar to the `store_container_labels` flag but a bit
more flexible because it allows white listing some common labels e.g.
`io.kubernetes.pod.name`, `io.kubernetes.pod.namespace` etc.

Without this white listing in place we saw the memory footprint of
cadvisor explode in our production environment because users add a lot
of data in pod labels, data which is not useful for metrics but may be
useful for the individual applications.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -61,7 +61,7 @@ var collectorCert = flag.String("collector_cert", "", "Collector's certificate, 
 var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
 
 var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
-var whiteListedContainerLabels = flag.String("white_listed_container_labels", "", "comma separated list of container labels to be converted to labels on prometheus metrics for each container.")
+var whitelistedContainerLabels = flag.String("whitelisted_container_labels", "", "comma separated list of container labels to be converted to labels on prometheus metrics for each container. store_container_labels must be set to false for this to take effect.")
 
 var (
 	// Metrics to be ignored.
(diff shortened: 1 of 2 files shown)


metrics/prometheus.go
```diff
@@ -902,6 +902,32 @@ const (
 	LabelImage = "image"
 )
 
+// WhiteListedContainerLabels returns a ContainerLabelsFunc that exports the
+// container name, first alias, image name as well as white listed label
+// values.
+func WhiteListedContainerLabels(whiteList []string) func(container *info.ContainerInfo) map[string]string {
```
dashpole: I don't think we need seperate `WhiteListedContainerLabels` and `BaseContainerLabels`.  BaseContainerLabels is just the case where `WhitelistedContainerLabels` is passed an empty list.

cadvisor.go
```diff
@@ -162,7 +163,11 @@ func main() {
 	}
 
 	containerLabelFunc := metrics.DefaultContainerLabels
-	if !*storeContainerLabels {
+
+	whiteListedLabels := strings.Split(*whiteListedContainerLabels, ",")
+	if len(whiteListedLabels) > 0 {
+		containerLabelFunc = metrics.WhiteListedContainerLabels(whiteListedLabels)
```
dashpole: See my comment about not separating `WhitelistedContainerLabels` and `BaseContainerLables`.  That simplifies this section as well, as you can just do a single assignment here.

cadvisor.go
```diff
@@ -61,6 +61,7 @@ var collectorCert = flag.String("collector_cert", "", "Collector's certificate,
 var collectorKey = flag.String("collector_key", "", "Key for the collector's certificate")
 
 var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
+var whiteListedContainerLabels = flag.String("white_listed_container_labels", "", "comma separated list of container labels to be converted to labels on prometheus metrics for each container.")
```
dashpole: I think we want to explicitly say that you need to specify `store_container_labels=false` to use the whitelistedContainerLables.

dashpole: whitelisted is one word, not two (even though my browser disagrees).  The flag should be `whitelisted_container_labels`, and variables should be `whitelistedFoo`.

PR #2114, opened 2018-12-03
Fix bug:getTopology will be failed on Arm platform

lubinszARM:
So we should search sysfs cpu path directly to get the data of
'thread_id' &'core_id' & 'node_id'.
Such as:


lubinszARM:
@dashpole 
Hi, I have fixed the CLA issue.
Any comments on this?

machine/machine.go
```diff
@@ -127,6 +130,60 @@ func parseCapacity(b []byte, r *regexp.Regexp) (uint64, error) {
...
 
+/* look for sysfs cpu path containing core_id */
+/* such as: sys/bus/cpu/devices/cpu0/topology/core_id */
+func getCoreIdFromCpuBus(cpuBusPath string, threadId int) (int, error) {
+	path := filepath.Join(cpuBusPath, fmt.Sprintf("cpu%d/topology", threadId))
+	file := filepath.Join(path, "core_id")
+
+	num, err := ioutil.ReadFile(file)
+	if err != nil {
+		// report threadId if no NUMA
```
dashpole: does this failure always indicate there is no NUMA?  Maybe only return threadId if the error is `os.IsNotExist(err)`.

machine/machine.go
```diff
@@ -127,6 +130,53 @@ func parseCapacity(b []byte, r *regexp.Regexp) (uint64, error) {
...
+
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return 0, err
+	}
+
+	nodeId := 0
+	for _, file := range files {
+		filename := file.Name()
+		if !strings.HasPrefix(filename, "node") {
```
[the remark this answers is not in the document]
dashpole: As in, can we use the regexp to do the matching? I.E. `regexp.MatchString`

machine/machine.go
lubinsz: Done. I have removed it.

machine/machine.go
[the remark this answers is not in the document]
lubinsz: regexp.MatchString was added for it.

machine/machine.go
dashpole: I was suggesting 
```
if os.IsNotExist(err) {
    // report threadId if no NUMA
    return threadId, nil
} else if err != nil {
    return 0, err
}
```
I don't have enough context on the change to know if that is _actually_ the right thing to do here, but whatever you think the correct behavior is works for me.

dashpole:
let me know about the `os.IsNotExist(err)`; otherwise lgtm

Issue #2018, opened 2018-08-14
Cadvisor behind reverse proxy does not work due to initial redirection to "/containers/"


sbakir:
I checked issue history and it seems that when reverse proxying via Host:Port/CadvisorAppName/ initial redirection replace "CadvisorAppName" with "/containers/" because it is given as absolute path . Is there any information regarding this topic?

sbakir:
Thanks @tghosgor  for the fix.

davidnewhall:
I'm leaving a comment here since this is the first hit when googling "cadvisor reverse proxy." See this Issue: https://github.com/google/cadvisor/issues/1587

PR #2117, opened 2018-12-03
added --uri_base_prefix option (#2018)

tghosgor:
Some reverse proxies do not support overriding 'Location' header
in 3xx responses. This commit adds a --uri_base_prefix option
that adds a prefix to all URLs of cAdvisor so that
redirects go to the correct places.

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -62,6 +62,8 @@ var collectorKey = flag.String("collector_key", "", "Key for the collector's cer
 
 var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
 
+var urlBasePrefix = flag.String("url_base_prefix", "", "prefix path that will be prepended to all paths to support some reverse proxies")
+
 var (
 	// Metrics to be ignored.
 	// Tcp metrics are ignored by default.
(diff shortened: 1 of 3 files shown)


cadvisor.go
```diff
@@ -62,6 +62,8 @@ var collectorKey = flag.String("collector_key", "", "Key for the collector's cer
 
 var storeContainerLabels = flag.Bool("store_container_labels", true, "convert container labels and environment variables into labels on prometheus metrics for each container. If flag set to false, then only metrics exported are container name, first alias, and image name")
 
+var basePrefix = flag.String("base_prefix", "", "prefix path that will be prepended to all paths to support some reverse proxies")
```
dashpole: can we make this less generic-sounding?  Maybe `url_base_prefix`?

tghosgor: Good idea.

dsandersAzure:
Hi All, is there any timeline when this will make it into a Docker build? We're waiting for this feature to run cAdvisor behind Traefik.

Thanks, David.

dashpole:
You are welcome to run the canary build.  We will have a release within the next month that corresponds with the kubernetes 1.14 release

dsandersAzure:
@dashpole Thanks for the guidance - much appreciated :)

aparedero:
Hi. I'm using last version pushed in dockerhub but option --uri_base_prefix is not current available

PR #2127, opened 2018-12-14
Bug fix if no NUMA support in kernel

lubinszARM:
After testing for edge devices with minimal kernel configuration(without numa),
I think we should report coreId = threadId/nodeId=0 if found error.

### DIFF
--- a/machine/machine.go
+++ b/machine/machine.go
@@ -130,8 +130,8 @@ func parseCapacity(b []byte, r *regexp.Regexp) (uint64, error) {
 	return m * 1024, err
 }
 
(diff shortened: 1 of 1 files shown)


dashpole:
/ok-to-test
What is the effect of reporting `coreId = threadId/nodeId=0`?  Can you expand on that a little bit?

lubinszARM:
> /ok-to-test
> What is the effect of reporting `coreId = threadId/nodeId=0`? Can you expand on that a little bit?

Just same with the topology of 'no numa'.

On a non-numa board, coreId = threadId = physical-cpu-id. And only contains node 0.

PR #2132, opened 2018-12-20
Provide free inodes and disk space with container filesystem stats

RobertKrawitz:
cadvisor currently does not provide free bytes and inodes for container volumes; this information is useful for e. g. eviction decisions based on inode pressure in Kubernetes/OpenShift.

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -257,6 +257,8 @@ func (self *crioContainerHandler) getFsStats(stats *info.ContainerStats) error {
 	fsStat.BaseUsage = usage.BaseUsageBytes
 	fsStat.Usage = usage.TotalUsageBytes
 	fsStat.Inodes = usage.InodeUsage
+	fsStat.Available = usage.FreeBytes
+	fsStat.InodesFree = usage.InodesFree
 
 	stats.Filesystem = append(stats.Filesystem, fsStat)
 
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -34,7 +34,9 @@ type FsHandler interface {
 type FsUsage struct {
 	BaseUsageBytes  uint64
 	TotalUsageBytes uint64
+	FreeBytes       uint64
 	InodeUsage      uint64
+	InodesFree      uint64
 }
 
 type realFsHandler struct {
@@ -73,12 +75,13 @@ func NewFsHandler(period time.Duration, rootfs, extraDir string, fsInfo fs.FsInf
 
 func (fh *realFsHandler) update() error {
 	var (
-		rootUsage, extraUsage fs.UsageInfo
-		rootErr, extraErr     error
+		rootUsage, extraUsage, freeSpace fs.UsageInfo
+		rootErr, extraErr, vfsStatsErr   error
 	)
 	// TODO(vishh): Add support for external mounts.
 	if fh.rootfs != "" {
 		rootUsage, rootErr = fh.fsInfo.GetDirUsage(fh.rootfs)
+		_, freeSpace.Bytes, _, _, freeSpace.Inodes, vfsStatsErr = fh.fsInfo.GetVfsStats(fh.rootfs)
 	}
 
 	if fh.extraDir != "" {
@@ -96,6 +99,10 @@ func (fh *realFsHandler) update() error {
 	}
 	if fh.extraDir != "" && extraErr == nil {
 		fh.usage.BaseUsageBytes = rootUsage.Bytes
+	}
+	if fh.rootfs != "" && vfsStatsErr == nil {
+		fh.usage.FreeBytes = freeSpace.Bytes
+		fh.usage.InodesFree = freeSpace.Inodes
 	}
 
 	// Combine errors into a single error to return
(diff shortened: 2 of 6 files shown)


dashpole:
> cadvisor currently does not provide free bytes and inodes for container volumes; 

By container volumes, do you mean PVs?  Or are you including "ephemeral local storage" ones as well (empty-dir backed volumes)?

> this information is useful for e. g. eviction decisions based on inode pressure in Kubernetes/OpenShift.

Can you expand on this?  We currently only take usage into account for inode pressure.  We wouldn't want to take PV usage into account, as those are on their own filesystems and have a separate pool of inodes.  We already take empty-dir usage into account, but those metrics are provided by the kubelet.

dashpole:
> Ephemeral also (including writable layers).

So this is really just for container writable layers, right?  Because cAdvisor doesn't monitor empty-dir backed volumes.

> The issue is that the container_fs_inodes_free metric is returning 0 for all volumes, so we can't alert on pods that are close to running out of inodes on one or more of their filesystems.

container_fs_inodes_free for a container should be the same as that metric for the node, right?  Can we just alert on that metric?  Does that work for devicemapper, or does that have a separate filesystem for each writable layer?

dashpole:
> I don't have a devicemapper cluster set up to test that (and the underlying issue referred to use of an overlay2 filesystem); with overlay2, I've verified that the free inodes is reported correctly with
>
> container_fs_inodes_free{container_name=""}

I was just trying to see if this was to fix something for dm specifically.  I think my point still stands that you can use the node-level inodes free to monitor this.  In fact, you would end up with an identical alert per-container if you monitored this per-container.

IIRC, the initial intent of leaving capacity-related (I consider `free` to be capacity related) out of container metrics is that we expect usage to be compared to requests or limits, not the actual capacity of the node.  For inode metrics, I can see your point, as we don't expose the capacity anywhere, or set requests/limits for inodes.  But I still think the principle of emitting only usage metrics for containers from cAdvisor is a good one.

On the other hand, if this is required to monitor dm-based containers, then this is a change worth making.

RobertKrawitz:
If we decide not to do this, then perhaps we shouldn't be reporting the -free information for containers at all, to eliminate a source of confusion.

dashpole:
> If we decide not to do this, then perhaps we shouldn't be reporting the -free information for containers at all, to eliminate a source of confusion.

Completely agree

RobertKrawitz:
It does appear that when using devicemapper that each container's root filesystem shows a different number of free inodes.

I created one pod that created 18 files under the local /tmp, and another that created only 3.  df -i on the pod with 18 files showed

Filesystem              Inodes      Used Available Use% Mounted on
/dev/mapper/docker-259:2-11537289-8d1bdb5a474934e969166e66a573a53584de9698d4f9bbc544589f31a1f6f324
                       5242880        63   5242817   0% /

On the one with only three files under the local /tmp, however, I got this:

Filesystem              Inodes      Used Available Use% Mounted on
/dev/mapper/docker-259:2-11537289-f1b5d44a0eccafa0ed8333428696e068c84998e8680970669591eb26506db05e
                       5242880        47   5242833   0% /

So there may be value in doing this, after all.

dashpole:
i'll try and take a look today or tomorrow.

derekwaynecarr:
i am not sure i understand the relationship of devicemapper and inodes.  i would not expect container fs for devicemapper to have any inodes reported.

RobertKrawitz:
It has been a while since I ran the test, and I'll have to reproduce what I did, but it does appear that the container's file system does report a fixed number of inodes, and that a pod could consume some (or potentially all) of those inodes.

dashpole:
I'm also not 100% sure what the vfs capacity for devicemapper means...  But if it is useful, this change is fine with me.

Where does this actually get used?  I don't think we use container inode capacity for eviction, and I think we overwrite the value in the summary API using the image fs capacity...

RobertKrawitz:
This came up a while back in the context of using metrics to detect overloaded nodes.  IIRC a container was running out of inodes and that was blowing somebody's CI job out of the water, so they wanted to monitor free inodes and were a little surprised (and dismayed) to find that they were always getting 0 reported back.

dashpole:
Do you remember if inode capacity and used still show up?

container/crio/handler.go
```diff
@@ -257,6 +257,8 @@ func (self *crioContainerHandler) getFsStats(stats *info.ContainerStats) error {
 	fsStat.BaseUsage = usage.BaseUsageBytes
 	fsStat.Usage = usage.TotalUsageBytes
 	fsStat.Inodes = usage.InodeUsage
+	fsStat.Available = usage.FreeBytes
```
dashpole: The limit is set up above based on the filesystem capacity.  Will that mean that Used + Free != Capacity?

dashpole:
I took another pass at this, and I think it would be better to record the capacity of the filesystem on startup rather than poll the free amount at collection time.  We might also want to add inode capacity to FsStats.  We also override some of these values [in the kubelet](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/helper.go#L315), so we will need to change that for these changes to be visible in the summary API.

Issue #1704, opened 2017-07-25
Inconsistent container metrics in prometheus route


zeisss:
Our cadvisor reports different containers each we time I query the /metrics route. The problems are consistent across various environments and VMs. I initially found #1635 and thought this to be the same, but the linked #1572 explains that cadvisor seems to pickup two systemd slices for the container, which is not the case according to my logs. Thus a separate issue, just to be sure.

```
17:50 $ curl -s http://docker-012.<domain>:8701/metrics | fgrep container_cpu_usage_seconds_total| wc -l
      98
17:51 $ curl -s http://docker-012.<domain>:8701/metrics | fgrep container_cpu_usage_seconds_total| wc -l
      18
17:51 $ curl -s http://docker-012.<domain>:8701/metrics | fgrep container_cpu_usage_seconds_total| wc -l
      98
17:51 $ curl -s http://docker-012.<domain>:8701/metrics | fgrep container_cpu_usage_seconds_total| wc -l
```

`:8701` is started as follows: `$ sudo /opt/cadvisor/bin/cadvisor -port 8701 -logtostderr  -v=10`

Neither dockerd nor cadvisor print any logs during those requests.

# Startup Logs

```
I0725 17:02:09.462596  109834 storagedriver.go:50] Caching stats in memory for 2m0s
I0725 17:02:09.462727  109834 manager.go:143] cAdvisor running in container: "/"
W0725 17:02:09.496040  109834 manager.go:151] unable to connect to Rkt api service: rkt: cannot tcp Dial rkt api service: dial tcp 127.0.0.1:15441: getsockopt: connection refused
I0725 17:02:09.531430  109834 fs.go:117] Filesystem partitions: map[/dev/dm-0:{mountpoint:/ major:254 minor:0 fsType:ext4 blockSize:0} /dev/mapper/rs--pre--docker--012--vg-var:{mountpoint:/var/lib/docker/aufs major:254 minor:2 fsType:ext4 blockSize:0} /dev/mapper/rs--pre--docker--012--vg-varlog:{mountpoint:/var/log major:254 minor:3 fsType:ext4 blockSize:0}]
[...]

PR #2136, opened 2019-01-03
Only add non-empty labels to metrics

mikkeloscar:
Instead of adding all labels to all exposed metrics, only add labels to a metric if the label also has a value. This reduces the memory footprint used by cadvisor at no additional downside, as empty label is the same as not providing the label in the first place.
This reduces the data exposed on the `/metrics` endpoint by several megabytes (depending on the amount of pods) because it doesn't include a lot of labels with empty value for each metric. 

This change is done in an effort to run cadvisor in production Kubernetes clusters without incurring too much of an overhead in terms of resources used for cluster monitoring.

The effect of this can be seen in the following graphs showing CPU and Memory usage for cadvisor pods running in one of our clusters. The impact is not incredible big, but visible:

image

(switch to run with this patch was done around 15:48 on the graph).

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1153,8 +1153,10 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		labels := make([]string, 0, len(rawLabels))
 		containerLabels := c.containerLabelsFunc(cont)
 		for l := range rawLabels {
-			labels = append(labels, sanitizeLabelName(l))
-			values = append(values, containerLabels[l])
+			if v, ok := containerLabels[l]; ok && len(v) > 0 {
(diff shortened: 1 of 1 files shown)


dashpole:
Happy 2019! Please rebase on top of #2137, and that should fix it.

Ill try and take a look today or tomorrow

mikkeloscar:
@dashpole Thanks! I rebased!

dashpole:
I'm surprised this doesn't reintroduce https://github.com/google/cadvisor/issues/1704.  I thought all metrics in prometheus had to have the same set of labels?

mikkeloscar:
@dashpole Arg, you are right. I didn't caught this when I tested it, but after looking at #1704 I clearly see the same behavior.

I'll try to save memory in another way :) Thanks for the review!

PR #2160, opened 2019-02-01
change default oom victim container name

gaorong:
# backgroud

In kubernetes, kubelet subscribs to cadvisor and follow the OOM event, the request is as below, (kubelet's related code is [here](https://github.com/kubernetes/kubernetes/blob/927fd1f78501a91b04a20ca10661a897ec626562/pkg/kubelet/oom_watcher.go#L52-L67))
```go
	request := events.Request{
		EventType: map[cadvisorapi.EventType]bool{
			cadvisorapi.EventOom: true,
		},
		ContainerName:        "/",
		IncludeSubcontainers: false,
	}
```
then cadvisor will return all system oom event against this request.  

# what happen to me
However, when I used this request as kubernetes with the only difference is the eventType, which is `EventOomKill` in mine, it seems not work to me, which I expect that cadvisor return all system oom kill event. (aside: The difference of `EventOomKill` and `EventOom` is described [here](https://github.com/google/cadvisor/pull/648#issuecomment-93560251)). 

# cadvisor's behavior
When some process indeedly been killed because of oom, cadvisor check if event satisfies requet. In situation described as above, when cadvisor check satisfication and invork `checkIfIsSubcontainer`, it always fail, which shouldn't happen. 

this PR fixs this bug by initialize OomInstance.VictimContainerName as '/', which is same as what OomInstance.ContainerName does. when oom parser parse kmsg's log and doesn't find a killed container, it will stay this name, otherwise, replace this name. and then if someone request as above, it will match this oom kill event and return.


gaorong:
the CI is green now

Issue #2129, opened 2018-12-17
exclude system.slice metrics from being exposed as prometheus metrics


viberan:
we are using cAdvisor to collect docker container metrics with Prometheus. There are tons of cgroup related metrics being exposed that are not relevant and we want cAdvisor to filter them out
--docker_only flag does not change anything for some reason, we continue to see all /system.slice/* stuff being exposed. After some googling, I came across rawPrefixWhiteList = "/system.slice/kubelet.service, /system.slice/docker.service.." setting #2048 #1926 
But I can not figure out how to use that setting and how it relates to docker_only flag? Could anyone explain it in more details how to use rawPrefixWhiteList? Seems it's not documented anywhere.

dashpole:
`docker_only` makes cAdvisor only collect docker container metrics, and machine level metrics.  We don't actually expose rawPrefisWhiteList as a flag.  We may want to consider consuming this in the kubelet to get that behavior.

viberan:
thanks for your quick response. 
does it mean we should NOT see  /system.slice/* metrics when using docker_only flag? What do you refer to as machine level metrics?

I'm just looking for a way to tell cAdvisor not to collect/expose /system.slice/* metrics as there is noticeable performance impact. We don't use kubernetes

dashpole:
with `docker_only`, you only get cgroup metrics for cgroups that belong to docker containers, e.g. `/kubepods/pod<pod_uid>/<container_id>` will be collected, but `/kubepods/pod<pod_uid>` will not.

Machine metrics are metrics from the `/` cgroup, as well as, for example, disk metrics from filesystems on the node.  We collect these metrics when `--docker_only` is specified even though it isn't a docker container cgroup.

viberan:
so, is there a way not to collect system.slice metrics?

lixianyang:
We don't use kubernetes too, but want to use cadvisor to collect docker container metrics. It collect all cgroup metrics in / , prometheus curl these metrics take a long time.

viberan:
I wanted to elaborate more on our use case for cAdvisor and Prometheus.
We're running fleets of spot instances that come and go many times a day with peak number of 1500-2000 instances at a moment. Each spot instance hosts a dozen of docker containers which are not static either. All in all it is pretty dynamic environment which in turn produces millions of different time series and leads to extremely high cardinality metrics. 
Basically we need just a handful of metrics to monitor CPU and memory usage of every docker container at resolution of every few seconds.
[...]

PR #2164, opened 2019-02-03
add docker_only_prefix_whitelist flag and fix issue #2129

viberan:


cadvisor.go
```diff
@@ -65,6 +65,8 @@ var whitelistedContainerLabels = flag.String("whitelisted_container_labels", "",
 
 var urlBasePrefix = flag.String("url_base_prefix", "", "prefix path that will be prepended to all paths to support some reverse proxies")
 
+var dockerOnlyPrefixWhiteList = flag.String("docker_only_prefix_whitelist", "", "A comma-separated list of cgroup path prefix that needs to be collected even when docker-only is specified")
```
dashpole: i'm not a particular fan of the naming for the `docker_only` flag, since it is runtime-specific.  Can we call it `raw_cgroup_roots` or something?  We can make it clear in the comment that setting it causes those cgroups to be collected even when --docker-only is specified since comments don't have backwards compatibility guarantees

viberan: @dashpole would raw_cgroup_prefix_whitelist be a better fit?

dashpole: that would be fine as well

dashpole:
looks good, please squash

viberan:
@dashpole done

PR #2167, opened 2019-02-07
Change log level to warning instead of error when cgroups file read f…

nekop:
…ailed.

Because cgroups file may not exist at the timing of the read, it's better not to use error level.


nekop:
I signed it!

dashpole:
> Because cgroups file may not exist at the timing of the read

Can you describe such a case?

nekop:
> Can you describe such a case?

When a container and the cgroups removed before read the file. This error log appears multiple times every day in healthy k8s clusters.

PR #2184, opened 2019-02-28
Fix for influxdb container labels and network stats

### LINKED ISSUE #1737: InfluxDB Storage driver sends network stats of only the first interface found
Hey,

I was comparing some metrics from both `influxdb` and `prometheus` and it looked weird to me that `influx` wasn't reporting per-interface statistics.

https://github.com/google/cadvisor/blob/2835c35bde4983eee1cc76aefc74d778776cca80/storage/influxdb/influxdb.go#L206-L209

This seems wrong as that data is coming from a single iface:

https://github.com/google/cadvisor/blob/c2f8191ba9550e302379473faf41d98759d1eff1/container/libcontainer/helpers.go#L138-L141

Am I missing something?

If that's really incorrect, how should we approach this? Should we add a new metric like `rx_bytes_per_iface` to not change expectations of those that already rely on `rx_bytes` or do you think it's ok to just add an extra tag on `rx_bytes` with the `interface` field (with values for each interface)?

Thanks!

ddtmachado:
This fixes the lack of labels when using the influxdb storage.

The driver was referencing an attribute from `ContainerInfo` that was not initialized, resulting in no container labels being stored even when the user set the flags for exporting labels on metrics: 
`store_container_labels / whitelisted_container_labels`. Complements #2113 

In addition to that I added a new flag `storage_driver_influxdb_whitelisted_labels` only for influxdb, so it is controlled separately but on the same format of #1984

Also `Network.Interfaces` was not being used so it only returned stats for the first network interface inside the container (the old behavior). I changed that to collect stats for each network interface adding a tag/label to the metric with the interface name. Fixes #1737 

What do you think of this approach?

Let me know if I missed something


dashpole:
can we just re-use the whitelisted_container_labels flag?

manager/manager.go
```diff
@@ -40,7 +40,7 @@ import (
 	"github.com/google/cadvisor/events"
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
-	"github.com/google/cadvisor/info/v2"
+	v2 "github.com/google/cadvisor/info/v2"
```
dashpole: I don't think this does anything?

dashpole: same with container.go

storage/influxdb/influxdb.go
```diff
@@ -299,14 +319,26 @@ func newStorage(
 		return nil, err
 	}
 
+	//Reuse cadvisor flag for whitelisted container labels
+	var argWhitelistedLabels = flag.Lookup("whitelisted_container_labels").Value.(flag.Getter).Get().(string)
```
dashpole: I would rather explicitly plumb this through, and do all of the flag parsing when cAdvisor starts if possible.

manager/manager.go
ddtmachado: yes, it does nothing. Sorry for that it was just my editor autoformat on save.

storage/influxdb/influxdb.go
ddtmachado: It can be done, but I'm not so familiar with the patterns and preferences here yet.

Are you suggesting to pass it down on the `storage` constructor or maybe add it to the `StorageDriver.AddStats` interface? 

I was trying to avoid modifying anything on other storage drivers as I have no context on then but it makes sense that they also respect the labels when storing if possible.

PR #2189, opened 2019-03-06
Fix stats in LXD with ZFS storage pool

ktsakalozos:
When LXD containers are backed up by a ZFS storage pool they do not have the `/dev/zfs` device. Trying to get the file system stats from that device will fail (va the zfs utility).

With this PR, in the above setup (ZFS inside and LXD container), we default to getting the file system stats from VFS.

Here is the error you get when `/dev/zfs` is not present:
```
Feb 19 09:38:54 definite-cricket microk8s.daemon-kubelet[3118]: I0219 09:38:54.595436    3118 fs.go:425] Stat fs failed. Error: exit status 1: "/snap/microk8s/383/sbin/zfs zfs get -Hp all zfsdefault/containers/definite-cricket" => /dev/zfs and /proc/self/mounts are required.
```

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -404,20 +404,19 @@ func (self *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, er
 				err error
 				fs  Fs
 			)
-			var fsType = partition.fsType
-			if fsType == ZFS.String() {
-				if _, devZfs := os.Stat("/dev/zfs"); os.IsNotExist(devZfs) {
-					fsType = "ZFS_no_dev"
-				}
-			}
-			switch fsType {
+			switch partition.fsType {
 			case DeviceMapper.String():
 				fs.Capacity, fs.Free, fs.Available, err = getDMStats(device, partition.blockSize)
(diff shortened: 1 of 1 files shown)


dashpole:
You will need to rebase on https://github.com/google/cadvisor/pull/2191, since the update to go 1.12 made our go-lint fail

fs/fs.go
```diff
@@ -404,7 +404,13 @@ func (self *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, er
 				err error
 				fs  Fs
 			)
-			switch partition.fsType {
+			var fsType = partition.fsType
+			if fsType == ZFS.String() {
+				if _, devZfs := os.Stat("/dev/zfs"); os.IsNotExist(devZfs) {
```
dashpole: I'm not a fan of defining a new fsType.  Maybe put this inside of `case ZFS.String()`, and use `goto default` to revert to the default behavior when this is the case?

ktsakalozos: Thank you for the review @dashpole. I moved the `/dev/zfs` check inside the ZFS case. Are you firmly set on using goto with an extra label (I believe the case label default does not qualify as a goto target)? I went with `fallthrough` what do you think?

dashpole: oh, if you can't goto default, then I don't know if that will work...  Let me take another look

PR #2191, opened 2019-03-07
Always collect disk stats for rootfs.

Random-Liu:
Always collect disk stats for root.

Kubelet needs it for all runtimes.

@dashpole Can you help me double check whether this is sufficient for CRI runtimes?

Based on my understanding and testing, this seems sufficient.


dashpole:
`container/common/container_hints_test.go:30:3: Errorf format %s has arg cHints of wrong type github.com/google/cadvisor/container/common.containerHints`
`container/common/container_hints_test.go:47:4: Errorf format %s has arg cHints of wrong type github.com/google/cadvisor/container/common.containerHints`

I think maybe we updated the go version or something recently.  Can you fix the go lint errors?

Your change looks correct, and should work for CRI runtimes.

Random-Liu:
Fixed the test.

Issue #2183, opened 2019-02-28
kubelet cAdvisor eventually losing kubernetes metadata labels


k0nstantinv:
I use:

kubernetes 1.8.11
prometheus 2.6.1
docker 17.03.1-ce

I have some kubernetes nodes (many) where the kubelet (http://NODE_IP:4194/metrics) eventually stops exporting a huge part of the metrics, such as **container_*** . To be more precise - it does, but it starts losing all the kubernetes metadata labels. For example:

"good" node
```
container_cpu_usage_seconds_total{container_name="POD",cpu="cpu00",id="/kubepods/burstable/pod2823c623-3b4c-11e9-9b8a-d09466092ef4/a50b87d8e7483e6c717490b43a19aa1acd254a15ffad4b1d5ff8eac3ed70f566",image="gcr.io/google_containers/pause-amd64:3.0",instance="srv0005",job="kubernetes-cadvisors",name="k8s_POD_frontend-86cc98b5fc-kqjfc_frontend_2823c623-3b4c-11e9-9b8a-d09466092ef4_0",namespace="frontend",pod_name="frontend-86cc98b5fc-kqjfc"} | 0.000023848
```
"bad" node
```
container_cpu_usage_seconds_total{cpu="cpu00",id="/kubepods/burstable/pod32a657f3-1b2e-11e9-9b8a-d09466092ef4/113d6bf59bebc9042ac5eeac1641577cf97bcc9d23fff726b07796ec19fb6b09",instance="srv0014",job="kubernetes-cadvisors"}
```
As you can see there's no lablels like: _container_name, namespace, pod_name_ etc. The ploblem is -  there's no metric queries with that kind of labels at all.

The nodes configured exactly the same way. The only thing I found interesting is the /validate endpoint outputs. For the "good" node it returns:
```
...
Managed containers:
	/kubepods/burstable/pod5f1000a0-3068-11e9-9b8a-d09466092ef4/9d350c41379f1073bf86f5362dc03d128118289c1bd6bd7deed437da9d6fa377
		Namespace: docker
		Aliases:
			k8s_some-name-b79bf94ff-4ng6m_some-name_5f1000a0-3068-11e9-9b8a-d09466092ef4_0
			9d350c41379f1073bf86f5362dc03d128118289c1bd6bd7deed437da9d6fa377
...
```
while for the "bad" node:
```
...
Managed containers:
[...]

PR #2194, opened 2019-03-07
Avoid dupe labels in prom metrics

### LINKED ISSUE #2181: Panic in v0.33.0
Kubernetes 1.11.5 (GKE)

Startup options

```
       '--housekeeping_interval=10s',
       '--max_housekeeping_interval=15s',
       '--event_storage_event_limit=default=0',
       '--event_storage_age_limit=default=0',
       '--disable_metrics=tcp,udp',
       '--docker_only=true',
       '--raw_cgroup_prefix_whitelist=kubepods',
```

panics immediately:
```
cadvisor-mf2vd cadvisor panic: duplicate label names
cadvisor-mf2vd cadvisor
cadvisor-mf2vd cadvisor goroutine 171 [running]:
cadvisor-mf2vd cadvisor github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus.MustNewConstMetric(0xc0003bb420, 0x2, 0x41d7176cf6000000, 0xc0001e4c00, 0x40, 0x40, 0x40, 0x0)
cadvisor-mf2vd cadvisor 	/usr/local/google/home/dashpole/go/src/github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus/value.go:102 +0xb0
cadvisor-mf2vd cadvisor github.com/google/cadvisor/metrics.(*PrometheusCollector).collectContainersInfo(0xc00018c370, 0xc000a1c060)
cadvisor-mf2vd cadvisor 	/usr/local/google/home/dashpole/go/src/github.com/google/cadvisor/metrics/prometheus.go:1165 +0x581
cadvisor-mf2vd cadvisor github.com/google/cadvisor/metrics.(*PrometheusCollector).Collect(0xc00018c370, 0xc000a1c060)
cadvisor-mf2vd cadvisor 	/usr/local/google/home/dashpole/go/src/github.com/google/cadvisor/metrics/prometheus.go:1078 +0x89
cadvisor-mf2vd cadvisor github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
cadvisor-mf2vd cadvisor 	/usr/local/google/home/dashpole/go/src/github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus/registry.go:434 +0x193
cadvisor-mf2vd cadvisor created by github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus.(*Registry).Gather
cadvisor-mf2vd cadvisor 	/usr/local/google/home/dashpole/go/src/github.com/google/cadvisor/vendor/github.com/prometheus/client_golang/prometheus/registry.go:5
```

blakebarnett:
Fixes #2181

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1155,11 +1155,10 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		values := make([]string, 0, len(rawLabels))
 		labels := make([]string, 0, len(rawLabels))
 		containerLabels := c.containerLabelsFunc(cont)
-		duplicate := false
 		for l := range rawLabels {
+			duplicate := false
 			sl := sanitizeLabelName(l)
 			for _, x := range labels {
-				duplicate = false
 				if sl == x {
 					duplicate = true
 					break
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1160,7 +1160,7 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 			sl := sanitizeLabelName(l)
 			for _, x := range labels {
 				duplicate = false
-				if sl != x {
+				if sl == x {
 					duplicate = true
 					break
 				}
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1166,7 +1166,7 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 			}
 			if !duplicate {
 				labels = append(labels, sl)
-			   values = append(values, containerLabels[l])
+				values = append(values, containerLabels[l])
 			}
 		}
 
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/google/cadvisor/container"
 	info "github.com/google/cadvisor/info/v1"
-	"github.com/google/cadvisor/utils"
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog"
 )
@@ -1156,10 +1155,17 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		values := make([]string, 0, len(rawLabels))
 		labels := make([]string, 0, len(rawLabels))
 		containerLabels := c.containerLabelsFunc(cont)
+		duplicate := false
 		for l := range rawLabels {
-			s := sanitizeLabelName(l)
-			if utils.SliceContains(labels, s) == false {
-				labels = append(labels, s)
+			sl := sanitizeLabelName(l)
+			for _, x := range labels {
+			   if sl != x {
+				   duplicate = true
+					break
+			   }
+			}
+			if duplicate != true {
+			   labels = append(labels, sl)
 			}
 			values = append(values, containerLabels[l])
 		}
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1159,13 +1159,13 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		for l := range rawLabels {
 			sl := sanitizeLabelName(l)
 			for _, x := range labels {
-			   if sl != x {
-				   duplicate = true
+				if sl != x {
+					duplicate = true
 					break
-			   }
+				}
 			}
 			if duplicate != true {
-			   labels = append(labels, sl)
+				labels = append(labels, sl)
 			}
 			values = append(values, containerLabels[l])
 		}
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1166,8 +1166,8 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 			}
 			if !duplicate {
 				labels = append(labels, sl)
+			   values = append(values, containerLabels[l])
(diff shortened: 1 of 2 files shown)


blakebarnett:
I signed it!

dashpole:
metrics/prometheus.go

metrics/prometheus.go
```diff
@@ -1155,8 +1155,18 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		values := make([]string, 0, len(rawLabels))
 		labels := make([]string, 0, len(rawLabels))
 		containerLabels := c.containerLabelsFunc(cont)
+		duplicate := false
 		for l := range rawLabels {
-			labels = append(labels, sanitizeLabelName(l))
+			sl := sanitizeLabelName(l)
+			for _, x := range labels {
+			   if sl != x {
+				   duplicate = true
```
dashpole: I think this will stay permanently true, and we will skip all subsequent labels

blakebarnett: oops, fixing

metrics/prometheus.go
```diff
@@ -1155,8 +1155,18 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
...
 		for l := range rawLabels {
-			labels = append(labels, sanitizeLabelName(l))
+			sl := sanitizeLabelName(l)
+			for _, x := range labels {
+				if sl != x {
+					duplicate = true
+					break
+				}
+			}
+			if duplicate != true {
```
dashpole: s/duplicate != true/!duplicate

metrics/prometheus.go
```diff
@@ -1155,8 +1155,19 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		values := make([]string, 0, len(rawLabels))
 		labels := make([]string, 0, len(rawLabels))
 		containerLabels := c.containerLabelsFunc(cont)
+		duplicate := false
```
dashpole: maybe just declare `duplicate` inside the `for l := range rawLabels` block?  Then you don't need to reset it to false each loop.

blakebarnett: Yeah, confused myself by doing this a different way before submitting the PR. Thanks :)

metrics/prometheus.go
```diff
@@ -1156,7 +1156,17 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
...
+			duplicate := false
+			sl := sanitizeLabelName(l)
+			for _, x := range labels {
+				if sl == x {
+					duplicate = true
+					break
+				}
+			}
+			if !duplicate {
+				labels = append(labels, sl)
```
dashpole: can we end up with fewer labels than values?  Should we move the `values = append(values, containerLabels[l])` statement inside here as well?

blakebarnett: I was wondering that after the test failure, but this shouldn't change that behavior right?  It will only exclude something if it a label if one already exists and the value will still get set for that label.

dashpole: I would be surprised if prometheus didn't yell at us if we tried to, for example, use a description with 3 labels, but then provide 4 label values when creating the metric.  What you are implicitly doing here is just using the first occurrence of a given sanitized label, and ignoring subsequent ones.

blakebarnett: Hmm, yeah I see. It could happen if someone provides multiple permutations of a label that all normalize to the same thing. Should we just throw them out in that case? I can't think of a great default behavior there.

blakebarnett: ^ on the same container, it should hopefully be a rare edge-case...

dashpole: wasn't that exactly what your bug was?

dashpole: I think just picking the first is a fine behavior.

blakebarnett: No #2181 was node-wide. 2 separate containers (separate pods) with annotations that normalize to the same thing caused the panic.

dashpole: Ok, I get the difference now.  I think we would still always get more values than labels, since if the label isn't present, we still add it with the value "".

blakebarnett: That's true, in fact I noticed when running without `--store_container_labels=false` all labels that were present on _any_ of the containers on the host showed up with empty values for all container metrics in prometheus, that was what made me look into the cgroup whitelisting initially and then I noticed this crash behavior.

dashpole: Yeah... Prometheus requires that all metric streams in a given scrape have the same set of labels.  So our workaround is just to add empty values for all labels we don't have.

blakebarnett:
We noticed a something similar to this: https://github.com/google/cadvisor/issues/2183 with `--store_container_labels=false`, the first container in the cgroup has all the kubernetes-provided label values, the rest don't. I can't figure if this is a regression of #1704 or something new...

dashpole:
Feel free to close if you are no longer working on this.  IIRC, this re-introduces #1704 in its current form.  Let me know if that is not the case.

blakebarnett:
Sorry, lost track of this one.  This should be fine to go in. The issue I mentioned above about `store_cotnainer_labels=false` seems to be the bigger problem for us, I'll open another issue/PR for it.

nightah:
@dashpole is this likely to make it in a release for Docker anytime soon?
I am experiencing the same issue that is exhibited in #2181 and have rolled back to v0.32.0 in the interim.

PR #2172, opened 2019-02-15
Changing the refresh period for the web UI

nzoueidi:
This is to change to the refreshing period from 1 second to 10 second. It should close #2141

Issue #2141, opened 2019-01-09
Change refresh period for web UI


dick-twocows:
Have the ability to change how often the web UI is auto refreshed.

dick-twocows:
(apologies, hit enter too soon)

It would be useful if the web UI could be configured to auto refresh at a set frequency, i.e. 10 seconds.

dashpole:
It looks like we refresh stats every second: https://github.com/google/cadvisor/blob/master/pages/assets/js/containers.js#L1101.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2196, opened 2019-03-14
Added flag to manually set machine info refresh rate

ivenk:
Solves #2141. 

Hi,
I found ticket #2141 and wanted to create a more generic solution than provided in #2172. 

Currently the machine info displayed on the web page gets updated every 1000ms. I implemented a new flag which replaces the old hard coded value. The flag has the default value 1000 in order to not change the default behavior. However the flag can be used to set a customized refresh rate (in ms). The flag is named machine_info_refresh_rate. 

On my system it passes all test.


ivenk:
I signed it!

dashpole:
are we using this flag for anything in cAdvisor itself?  Or just as a way to pass this value to the UI?

ivenk:
@dashpole just to get the value to the ui.

ivenk:
I'm going to look into the test results in the next few days.

ivenk:
Currently I don't have the time to fix this.

Issue #2082, opened 2018-10-19
[feature request]: make containerd namespace configurable


Starnop:
As known, Contaierd offers a fully namespaced API .However, we create a containerd client with default namespace "k8s.io" which is not configurable. It’s not make sense. How about add a flag to make it configurable.  ^_^

dashpole:
cc @Random-Liu

0902horn:
There has been a variable [ArgContainerdEndpoint](https://github.com/google/cadvisor/blob/master/container/containerd/factory.go#L34) for flag which is unused. I think we could add a 'ArgContainerdNamespace'  and use both to configure the containerd factory.

Maybe looks like:

```
var ArgContainerdEndpoint = flag.String("containerd", "/var/run/containerd.sock", "containerd endpoint")
var ArgContainerdNamespace = flag.String("containerd-namespace", "k8s.io", "containerd namespace")
```

```
client, err := Client(*ArgContainerdEndpoint, *ArgContainerdNamespace)
```

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2203, opened 2019-03-22
Make the containerd factory configurable

0902horn:
1. Change the default value of command line flag `--containerd` to '/run/containerd/containerd.sock'.
2. Add a command line flag `--containerd-namespace` defaulting to 'k8s.io'.
3. Create containerd client with values from `--containerd` and `--containerd-namespace`.

### DIFF
--- a/container/containerd/factory.go
+++ b/container/containerd/factory.go
@@ -31,7 +31,8 @@ import (
 	"github.com/google/cadvisor/manager/watcher"
 )
 
-var ArgContainerdEndpoint = flag.String("containerd", "unix:///var/run/containerd.sock", "containerd endpoint")
+var ArgContainerdEndpoint = flag.String("containerd", "/run/containerd/containerd.sock", "containerd endpoint")
+var ArgContainerdNamespace = flag.String("containerd-namespace", "k8s.io", "containerd namespace")
 
 // The namespace under which containerd aliases are unique.
 const k8sContainerdNamespace = "containerd"
@@ -56,7 +57,7 @@ func (self *containerdFactory) String() string {
 }
 
 func (self *containerdFactory) NewContainerHandler(name string, inHostNamespace bool) (handler container.ContainerHandler, err error) {
-	client, err := Client()
+	client, err := Client(*ArgContainerdEndpoint, *ArgContainerdNamespace)
 	if err != nil {
 		return
 	}
@@ -118,7 +119,7 @@ func (self *containerdFactory) DebugInfo() map[string][]string {
 
 // Register root container before running this function!
 func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo, includedMetrics container.MetricSet) error {
-	client, err := Client()
+	client, err := Client(*ArgContainerdEndpoint, *ArgContainerdNamespace)
 	if err != nil {
 		return fmt.Errorf("unable to create containerd client: %v", err)
 	}
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -31,11 +31,6 @@ import (
 	"google.golang.org/grpc"
 )
 
-const (
-	// k8sNamespace is the namespace we use to connect containerd.
-	k8sNamespace = "k8s.io"
-)
-
 type client struct {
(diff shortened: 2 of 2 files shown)


0902horn:
This PR fixes #2082 .

dashpole:
/ok-to-test

cc @Random-Liu

dashpole:
is it possible to get containers from all namespaces?  Just curious

0902horn:
> is it possible to get containers from all namespaces? Just curious

According to [comments](https://github.com/containerd/containerd/blob/master/namespaces/validate.go#L20) of containerd namespaces package, I think getting containers across namespaces could confuse cAdvisor or other applications that rely on metrics provided by cAdvisor, since identifiers in different containerd namespaces might be the same and these namespaces are not reflected by metrics.

dashpole:
@0902horn that makes sense.  I was just curious if there was a way (e.g. empty string) to get everything if I wanted to.  Since cAdvisor is a cgroup monitor, it will end up collecting metrics for those containers anyways.  It just wouldn't attach metadata (e.g. container name) for those.  I am fine with the change without that, but I wouldn't be surprised if we get a feature request for being able to do that eventually.

0902horn:
@dashpole Thanks for your replies.
It seems reasonable to get such a feature. So I did a little further work on this. The metric tag named 'id' is a cgroup path actually, so I guess cAdvisor would be fine if it gets containers from all namespaces. However, I found that 'ctr' requires a namespace (default to 'default').
```
#ctr --namespace= container ls
ctr: namespace is required: failed precondition
```
I confirmed that this is by design by reading the [doc](https://github.com/containerd/containerd/blob/1ac546b3c4a3331a9997427052d1cb9888a2f3ef/docs/namespaces.md). 
When I specified '--containerd-namespace=' on cAdvisor built from this PR, I got this error message:
```
I0326 06:35:34.698090       1 factory.go:109] Error trying to work out if we can handle /kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod149154bb_4c7b_11e9_beec_00163e041f4b.slice/a9d2267854302a842e9dd52b70f658c5ebcbab1e338f8e88e22d7c0ff462a89f: failed to load container: namespace is required: failed precondition
```
So a possible way is listing all namespaces and creating multiple clients with each to query containers.
If the feature is requested, I would like to contribute another PR.

dashpole:
Thanks a bunch for the investigation.  In that case, just supporting a single namespace seems like the correct behavior.  I'll do a review in the next few days

dashpole:
FYI @Random-Liu

container/containerd/factory.go
```diff
@@ -31,7 +31,8 @@ import (
 	"github.com/google/cadvisor/manager/watcher"
 )
 
-var ArgContainerdEndpoint = flag.String("containerd", "unix:///var/run/containerd.sock", "containerd endpoint")
+var ArgContainerdEndpoint = flag.String("containerd", "/run/containerd/containerd.sock", "containerd endpoint")
```
dashpole: wait, does this still work if it isn't a unix socket?

0902horn: It has to be a unix socket because the original implementation of 'func Client' has following code ([see](https://github.com/0902horn/cadvisor/blob/150c78bd124181c983f86b423c6877fab1451ff7/container/containerd/client.go#L58)):
```
tryConn, err := net.DialTimeout("unix", address, connectionTimeout)
```
I think it is by design too. See this containerd [issue](https://github.com/containerd/containerd/issues/1324)

0902horn:
I'm happily seeing this PR is merged. Thanks @dashpole

container/containerd/factory.go
dashpole: I meant to ask if this will still connect to the unix socket if you remove the `unix://` prefix

0902horn: Oh...I thought you were asking TCP or something.
ArgContainerdEndpoint was not used before, so don't worry about changing its value's format. Now it is used and requires such format without `unix://` prefix.

PR #2209, opened 2019-03-29
Reorganize code to allow conditional enablement of runtimes

dims:
Change-Id: I76583736d7ad39190a1a2bca820d4e957caadc84


dashpole:
Let me know when this is ready for review

dims:
@dashpole this is ready now (just rebased)

dims:
cc @tallclair

tallclair:
This is awesome (I haven't thoroughly reviewed it yet though)

What is the plan for integrating this into Kubernetes though?

BTW, there are still a bunch of docker dependencies, but those probably deserve to be addressed independently.
image

container/mesos/init.go
```diff
@@ -0,0 +1,36 @@
...
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/watcher"
+	"k8s.io/klog"
+)
+
+func init() {
+	err := container.RegisterPlugin("mesos", func(factory info.MachineInfoFactory, fsInfo fs.FsInfo, includedMetrics container.MetricSet) (watcher.ContainerWatcher, error) {
+		err := Register(factory, fsInfo, includedMetrics)
+		if err != nil {
+			klog.V(5).Infof("Registration of the mesos container factory failed: %v.", err)
```
dashpole: Aren't all of these logs a duplicate of the one in InitializePlugins?  I think we could just `return nil, err` here, and in other factories.

dims: Fixed. thanks!

dims:
@tallclair yep! wanted to peel off some of the leaves first (like mesos)

In kubernetes, we will import only the items corresponding to the CRI(s) that we support.

container/containerd/init.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/google/cadvisor/fs"
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/watcher"
+	"k8s.io/klog"
+)
+
+func init() {
+	err := container.RegisterPlugin("containerd", func(factory info.MachineInfoFactory, fsInfo fs.FsInfo, includedMetrics container.MetricSet) (watcher.ContainerWatcher, error) {
+		err := Register(factory, fsInfo, includedMetrics)
+		if err != nil {
```
dashpole: We don't need to check err here or in other plugins that don't register any watchers.  Just `return nil, err`

dims: Done!

PR #2213, opened 2019-04-04
Remove CloudProvider SDK dependencies

tallclair:
Change CloudInfo to use a registration pattern, making the dependencies on the cloud provider SDKs optional. The CloudProviders are registered in the main cadvisor binary, so for non-library usages there are no changes.

For Kubernetes, I verified that the info provided by CloudInfo is unused by removing it and confirming everything still builds: https://github.com/tallclair/kubernetes/commit/9274b43f7c2a270b464e8ff5a7a5034d4618d562 (the only reference in Kubernetes was in the cadvisor fake)

### DIFF
--- a/cadvisor.go
+++ b/cadvisor.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


tallclair:
For https://github.com/kubernetes/kubernetes/issues/69585

dashpole:
This will mean that the cadvisor v1 machine info returned by `/spec/` won't have this information anymore.  From cAdvisor's perspective, this is fine, but we should at least add that to the release notes in kubernetes, assuming the loss of this information from that endpoint is desired.

PR #2217, opened 2019-04-10
move plugin-specific init code into plugin hooks

liggitt:
* restores `manager.New` signature from before #2209 was merged
* moves plugin install import packages to standalone packages (so that importing the docker package, for example, which is done to satisfy the `DockerImages()` and `DockerInfo()` methods, doesn't automatically run the docker collector)
* adds a `github.com/cadvisor/container/install` package that installs all optional container collectors
* installs all optional plugins in manager test and in cadvisor#main()
* switches plugin registration to be an actual interface, instead of a bare method
* moves plugin-specific init code into plugin hooks so callers don't have to recreate it

/cc @dims


liggitt:
/cc @dashpole 

still testing this, but wanted to get it opened for early review

liggitt:
the first commit simply restores the manager.New signature and logic... the remaining commits make it easy to see the imports getting whittled away

liggitt:
I'll open a WIP PR in kubernetes as well to exercise vendoring and CI there

dims:
thanks for taking care of this @liggitt. glad we were able to go back to the old signature!

liggitt:
kube PR to exercise CI is at https://github.com/kubernetes/kubernetes/pull/76355

derekwaynecarr:
I like this :-)

liggitt:
yay for CI:

> ERROR(darwin/386,darwin/amd64,windows/386,windows/amd64) cmd/genkubedocs/gen_kube_docs.go:33:13: could not import k8s.io/kubernetes/cmd/kubelet/app (type-checking package "k8s.io/kubernetes/cmd/kubelet/app" failed (/home/prow/go/src/k8s.io/kubernetes/cmd/kubelet/app/auth.go:34:2: could not import k8s.io/kubernetes/pkg/kubelet/server (type-checking package "k8s.io/kubernetes/pkg/kubelet/server" failed (/home/prow/go/src/k8s.io/kubernetes/pkg/kubelet/server/server.go:36:18: could not import github.com/google/cadvisor/container (type-checking package "k8s.io/kubernetes/vendor/github.com/google/cadvisor/container" failed (/home/prow/go/src/k8s.io/kubernetes/vendor/github.com/google/cadvisor/container/factory.go:83:32: Context not declared by package fs))))))
ERROR(darwin/386,darwin/amd64,windows/386,windows/amd64) cmd/genman/gen_kube_man.go:37:13: could not import k8s.io/kubernetes/cmd/kubelet/app (type-checking package "k8s.io/kubernetes/cmd/kubelet/app" failed (/home/prow/go/src/k8s.io/kubernetes/cmd/kubelet/app/auth.go:34:2: could not import k8s.io/kubernetes/pkg/kubelet/server (type-checking package "k8s.io/kubernetes/pkg/kubelet/server" failed (/home/prow/go/src/k8s.io/kubernetes/pkg/kubelet/server/server.go:36:18: could not import github.com/google/cadvisor/container (type-checking package "k8s.io/kubernetes/vendor/github.com/google/cadvisor/container" failed (/home/prow/go/src/k8s.io/kubernetes/vendor/github.com/google/cadvisor/container/factory.go:83:32: Context not declared by package fs))))))

since the Context type was used in the plugin interface, moved it to the os-neutral fs/types.go file

dashpole:
I'm out for a couple weeks.  If this needs attention soon, @derekwaynecarr or @tallclair will need to review it.  Otherwise, I'll take a look when I get back.

liggitt:
thanks for the heads up. yeah, would like to avoid letting the signature change linger in cadvisor master

liggitt:
kube CI was green with this

derekwaynecarr:
I am fine with this change.

container/docker/install/install.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/klog"
+)
+
+const dockerClientTimeout = 10 * time.Second
+
+func init() {
+	err := container.RegisterPlugin("docker", container.PluginFn{
+		InitializeContextFn: func(context *fs.Context) error {
+			docker.SetTimeout(dockerClientTimeout)
+			// Try to connect to docker indefinitely on startup.
```
jaypipes: Is the Docker plugin the only one that retries the initialization because of the design of Docker itself? or were you planning on adding this connection retry logic for the other plugins at a later time? I'm just curious why the docker plugin is different from the others... thanks in advance for your answer!

liggitt: this is simply moving the existing behavior, I'll defer to the cadvisor maintainers on the reason for the current behavior

jaypipes: Ah, cool. Sorry, didn't realize it was the existing behaviour...

container/rkt/install/install.go
```diff
@@ -26,6 +26,14 @@ import (
 
 func init() {
 	err := container.RegisterPlugin("rkt", container.PluginFn{
```
tallclair: (optional) I have a slight preference for implementing the interface with an empty struct, since it reduces the amount of indentation here, although it does add a bit of boilerplate. Your call.

container/factory.go
```diff
@@ -78,15 +78,28 @@ var pluginsLock sync.Mutex
 var plugins = make(map[string]Plugin)
 
 type Plugin interface {
+	// InitializeContext is invoked when populating an fs.Context object for a new manager.
+	// A returned error here is fatal.
+	InitializeContext(context *fs.Context) error
```
tallclair: Can we call this `InitializeFSContext`?

container/docker/install/install.go
tallclair: historical context: https://github.com/google/cadvisor/pull/1871

It was fixing a specific bug with docker. Not sure if the same bug could exist in the other runtimes...

container/rkt/install/install.go
liggitt: done, implemented in empty structs in the main provider package, accessible via `NewPlugin()`

container/factory.go
liggitt: done

liggitt:
comments addressed... diff visible at https://github.com/google/cadvisor/compare/655e1c19b6a3e679c0ecff2cb40b57fc2716f09d..f8a73e018b9690a8c6f0c5ce05c367d6bb59bb66

dims:
/lgtm

@tallclair can we please merge this? would be good to try to get a full weekends worth of CI testing in k/k

PR #2219, opened 2019-04-10
Adding ability to handle OverlayFS mounts

### LINKED ISSUE #2218: Lacking support for OverlayFS in the mount processing
When I try to start Kubernetes it fails with the following error:

failed to get device for dir "/var/lib/kubelet": could not find device with major: 0, minor: 22 in cached partitions map

The missing device is described in the following way in my mountinfo file (/proc/self/mountinfo):

29 15 0:22 / /var/lib rw,relatime shared:18 - overlay overlay rw,lowerdir=/var/lib,upperdir=/mnt/flash/var/lib,workdir=/mnt/flash/.work-var-lib

I need support for overlayfs in the processMounts function (fs/fs.go) to make Kubernetes work.

jaerik:
Fixes #2218
Signed-off-by: Erik Jansson <erikja@axis.com>

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -184,10 +184,11 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
 
 	supportedFsType := map[string]bool{
 		// all ext systems are checked through prefix.
-		"btrfs": true,
-		"tmpfs": true,
-		"xfs":   true,
-		"zfs":   true,
+		"btrfs":   true,
+		"overlay": true,
+		"tmpfs":   true,
+		"xfs":     true,
+		"zfs":     true,
 	}
 
 	for _, mount := range mounts {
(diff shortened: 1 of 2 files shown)


jaerik:
email address updated

fs/fs.go
```diff
@@ -222,6 +223,10 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
 			}
 		}
 
+		if mount.Fstype == "overlay" {
+			mount.Source = fmt.Sprintf("%s_%d-%d", mount.Source, mount.Major, mount.Minor)
```
dashpole: why are we setting source here?

jaerik: When mount.Source is not unique to overlayfs mount in the mountinfo file, the mount.Source is "overlay" for all overlayfs mounts, it causes problems when it is used as key in the partitions map. Hence this is a way to make mount.Source unique.

dashpole: can you add a comment explaining that this makes makes mount.Source unique for overlay mounts?

PR #2238, opened 2019-05-12
Replace use of cgo for getting number of online CPUs

lorenz:
This is part of a series of PRs removing the dependence on cgo for cadvisor. Together with https://github.com/mindprince/gonvml/pull/11 and a forthcoming PR for the cpuload collector all of cadvisor and kubelet are buildable without cgo.

PR #2239, opened 2019-05-13
Remove use of cgo for cgroup stats and refactoring of constants

lorenz:
This is part of a patch series to remove the dependence on cgo in cadvisor, see https://github.com/google/cadvisor/pull/2238.

This removes the cgo-enabled defs.go file for netlink communication and replaces it with constants from the updated `golang.org/x/sys/unix` package. It also clears up confusion regarding the use of `__TASKSTATS_CMD_MAX+1` (which happens to be the way the kernel defines its constants) instead of `CGROUPSTATS_CMD_GET`.


lorenz:
The other PR is now here, tests now pass.

dashpole:
This is great.  Thanks for the contribution.

PR #2252, opened 2019-06-10
Allow storage drivers to read the spec details as well.

mcdan:
Per quick discussion in slack[0] here is the patch to pass the spec information long to the storage driver. I am not really certain if this is passing the cached copy as requested but it follows the existing patterns for other information going into the struct.

[0] - https://kubernetes.slack.com/archives/C0BP8PW9G/p1560178234038700

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -631,9 +631,18 @@ func (c *containerData) updateStats() error {
 		}
 		return err
 	}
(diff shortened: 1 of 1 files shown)


dashpole:
Sorry this sat for so long.  It would probably be best to include the changes to storage drivers here as well, so I can tell why we are adding the spec to the container info.

Issue #1522, opened 2016-11-01
Unable to get network stats from pid


gl328518397:
my OS is redhat7,
docker is 1.8
my cmd is 
"
 docker run --volume=/:/rootfs:ro  \
--volume=/var/run:/var/run:rw \
--volume=/sys:/sys:ro \
--volume=/var/lib/docker/:/var/lib/docker:ro \
--volume=/cgroup:/cgroup:ro  \
--publish=8096:8080 \
--detach=true \
--link influxsrv:influxsrv \
--privileged=true \
 google/cadvisor:latest  \
  -v=2 \
 -storage_driver=influxdb \
 -storage_driver_db=cadvisor \
 -storage_driver_host=127.0.0.1:8086 
"

i say logs -->>
Unable to get network stats from pid 34215: couldn't read network stats: failure opening /rootfs/proc/34215/net/dev: open /rootfs/proc/34215/net/dev: no such file or directory
I1101 06:01:03.346357       1 helpers.go:101] Unable to get network stats from pid 32943: couldn't read network stats: failure opening /rootfs/proc/32943/net/dev: open /rootfs/proc/32943/net/dev: no such file or directory
I1101 06:01:04.145622       1 helpers.go:101] Unable to get network stats from pid 34215: couldn't read network stats: failure opening /rootfs/proc/34215/net/dev: open /rootfs/proc/34215/net/dev: no such file or directory
I1101 06:01:05.720426       1 helpers.go:101] Unable to get network stats from pid 32943: couldn't read network stats: failure opening /rootfs/proc/32943/net/dev: open /rootfs/proc/32943/net/dev: no such file or directory
I1101 06:01:07.009749       1 helpers.go:101] Unable to get network stats from pid 32943: couldn't read network stats: failure opening /rootfs/proc/32943/net/dev: open /rootfs/proc/32943/net/dev: no such file or directory
I1101 06:01:07.710892       1 helpers.go:101] Unable to get network stats from pid 34215: couldn't read network stats: failure opening /rootfs/proc/34215/net/dev: open /rootfs/proc/34215/net/dev: no such file or directory
I1101 06:01:09.289906       1 helpers.go:101] Unable to get network stats from pid 34215: couldn't read network stats: failure opening /rootfs/proc/34215/net/dev: open /rootfs/proc/34215/net/dev: no such file or directory
[...]

PR #2255, opened 2019-06-15
Monitor docker restart container events and update pid for the container process.

Dennisparchkov:
Network stats correctly work after containers are restarted: issue #1522
I've been running this for a couple of weeks now and haven't seen any issues affecting performance. 
Pid of containers is updated.

Note:
- This solution resets the cumulative values in network stats so the documentation will need updating to show this. 
- I can't seem to synchronise `GetStats()` to when the containers are restarted. Therefore, as soon as a container is restarted (with `docker restart ...`), the error `Unable to get network stats from pid 34215` might appear once before the pid is updated. This as a knock-on effect on the front-end. 
- The front-end will receive a single empty network stat (from the issue mention above) causing a error on rendering the graph. This will fix itself when the time stamp of that specific empty stat is outside the time range of the graph. This can be solved by not returning empty stats on the API but i have't looked into it.

### DIFF
--- a/container/docker/event.go
+++ b/container/docker/event.go
@@ -1,0 +1,109 @@
+package docker
+
+import (
+	"context"
+	"sync"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/events"
+	"github.com/docker/docker/api/types/filters"
+	dclient "github.com/docker/docker/client"
+	containerlibcontainer "github.com/google/cadvisor/container/libcontainer"
+	"k8s.io/klog"
+)
+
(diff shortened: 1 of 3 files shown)


Dennisparchkov:
I signed it!

dashpole:
If you are still working on this, feel free to rebase, and fix the e2e failure.  It looks like it was long enough ago that the results aren't available anymore.  In general, this change seems useful, so i'm happy to review once tests are passing.

Dennisparchkov:
I'll see what I can do

PR #1891, opened 2018-02-19
add PIDs stats to exported prometheus container metrics

adrien-f:
Greetings !

We were in need of tracking global PIDs usage across our containers via Prometheus so we set out forking and adding this little feature. We're running this new feature in production (easy debian repackaging !) and we're really satisfied so far. 

cadvisor-container-stats

As this is my first PR for this project, let me know if you need more informations or tests.

Thanks :wave: 

Signed-off-by: Adrien Fillon <adrien.fillon@cdiscount.com>

PR #2257, opened 2019-06-20
add pids stats

KielChan:
We will collect pids metrics in kubelet by enable subcontainers to detect Burstable, Guaranteed and BestEffort pod pids stats.
And also it will be used to detect threads leaked if used jvm or some other virtual machine.
 The output maybe like this:
```
			"processes": {
				"process_count": 1,
				"fd_count": 8,
                                 "threads_current": 66,
                                 "threads_max": 0
			}
```

in `kubepods` and if subcontainers enabled.

`threads_max` will be detected in pod metrics `Processes` Spec `limit` field if pids_limit feature enabled in kubelet.

Ref: https://github.com/google/cadvisor/pull/1891


derekwaynecarr:
this is great to see.

fyi @sjenning @mrunalp @RobertKrawitz

metrics/prometheus.go
```diff
@@ -415,6 +415,32 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.PidsMetrics) {
```
dashpole: We already have container_processes.  How does this differ from that metric?

info/v1/container.go
```diff
@@ -41,6 +41,10 @@ type MemorySpec struct {
 	SwapLimit uint64 `json:"swap_limit,omitempty"`
 }
 
+type PidsSpec struct {
+	Limit uint64 `json:"limit,omitempty"`
```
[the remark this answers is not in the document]
KielChan: just ref 
```
type MemorySpec struct {
	// The amount of memory requested. Default is unlimited (-1).
	// Units: bytes.
	Limit uint64 `json:"limit,omitempty"`

	// The amount of guaranteed memory.  Default is 0.
	// Units: bytes.
	Reservation uint64 `json:"reservation,omitempty"`

	// The amount of swap space requested. Default is unlimited (-1).
	// Units: bytes.
	SwapLimit uint64 `json:"swap_limit,omitempty"`
}
```

its value equals to pids.max, and it is a static variable if used pids limit feature in kubelet

metrics/prometheus.go
KielChan: `ProcessCount`  just collect processes num, and it is not equals to threads num in container.

info/v1/container.go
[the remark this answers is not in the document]
KielChan: just ref 
```
type MemorySpec struct {
	// The amount of memory requested. Default is unlimited (-1).
	// Units: bytes.
	Limit uint64 `json:"limit,omitempty"`

	// The amount of guaranteed memory.  Default is 0.
	// Units: bytes.
	Reservation uint64 `json:"reservation,omitempty"`

	// The amount of swap space requested. Default is unlimited (-1).
	// Units: bytes.
	SwapLimit uint64 `json:"swap_limit,omitempty"`
}
```

its value equals to pids.max, and it is a static variable if used pids limit feature in kubelet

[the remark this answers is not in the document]
KielChan: just ref 
```
type MemorySpec struct {
	// The amount of memory requested. Default is unlimited (-1).
	// Units: bytes.
	Limit uint64 `json:"limit,omitempty"`

	// The amount of guaranteed memory.  Default is 0.
	// Units: bytes.
	Reservation uint64 `json:"reservation,omitempty"`

	// The amount of swap space requested. Default is unlimited (-1).
	// Units: bytes.
	SwapLimit uint64 `json:"swap_limit,omitempty"`
}
```

its value equals to pids.max, and it is a static variable if used pids limit feature in kubelet

metrics/prometheus.go
KielChan: `ProcessCount`  just collect processes num, and it is not equals to threads num in container.

metrics/prometheus.go
```diff
@@ -415,6 +415,32 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:      "container_pids_max",
+				help:      "The number of container max pids, infinity if value is zero",
```
sjenning: reword:
The maximum number of processes allowed in the container, infinity if value is zero

metrics/prometheus.go
```diff
@@ -415,6 +415,32 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{
+						value:     float64(s.Pids.Max),
+						timestamp: s.Timestamp,
+					}}
+				},
+			},
+			{
+				name:      "container_pids_current",
+				help:      "The number of container current pids",
```
sjenning: reword:
The current number of processes in the container

container/common/helpers.go
```diff
@@ -143,7 +152,7 @@ func readString(dirpath string, file string) string {
 
 func readUInt64(dirpath string, file string) uint64 {
 	out := readString(dirpath, file)
-	if out == "" {
+	if out == "" || out == "max" {
```
sjenning: why is this needed?

sjenning: ah nm, `max` is the value in `/sys/fs/cgroup/pids/.../pids.max` when the max is unbounded :+1:

metrics/prometheus.go
sjenning: I think I would expect these to match.  `ProcessCount` counts the number of pids listed in the `cgroups.proc` file in the `cpu` hierarchy where the new stat just reads the `pids.current` in the `pids` hierarchy.

info/v1/container.go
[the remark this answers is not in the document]
sjenning: This was also confusing to be at first, but I can see the rationale for the `Max` vs `Limit` as kubelet refers to this as `Limit` for all other resources. :+1:

sjenning:
LGTM, other than some rewording nits in the in Prom metric descriptions

I do yield that we already have a metric for the number of processes in a container, but I think reading straight out of the pids cgroup, which is used for pid limit enforcement, is the better way.

metrics/prometheus.go
KielChan: pids.current equals to processes + threads num，its value is like pstree-l entries num

container/common/helpers.go
KielChan: That's right.

metrics/prometheus.go
KielChan: I will fix it

KielChan: I have fixed it

metrics/prometheus.go
KielChan: I have fixed it too

container/factory.go
```diff
@@ -53,6 +53,7 @@ const (
 	NetworkUsageMetrics     MetricKind = "network"
 	NetworkTcpUsageMetrics  MetricKind = "tcp"
 	NetworkUdpUsageMetrics  MetricKind = "udp"
+	PidsMetrics             MetricKind = "pids"
```
dashpole: Lets stick with ProcessMetrics

info/v1/container.go
```diff
@@ -565,13 +572,19 @@ type ProcessStats struct {
...
+}
+
 type ContainerStats struct {
 	// The time of this stat point.
 	Timestamp time.Time    `json:"timestamp"`
 	Cpu       CpuStats     `json:"cpu,omitempty"`
 	DiskIo    DiskIoStats  `json:"diskio,omitempty"`
 	Memory    MemoryStats  `json:"memory,omitempty"`
 	Network   NetworkStats `json:"network,omitempty"`
+	Pids      PidsStats    `json:"pids,omitempty"`
```
dashpole: Lets stick with ProcessStats

info/v1/container.go
```diff
@@ -41,6 +41,10 @@ type MemorySpec struct {
 	SwapLimit uint64 `json:"swap_limit,omitempty"`
 }
 
+type PidsSpec struct {
```
dashpole: Can we call this ProcessSpec for consistency?

info/v1/container.go
```diff
@@ -565,13 +572,19 @@ type ProcessStats struct {
 	FdCount uint64 `json:"fd_count"`
 }
 
+type PidsStats struct {
```
dashpole: Lets add these to ProcessStats

metrics/prometheus.go
dashpole: This is different from threads, right?  Or if this is a threads metric, lets name it that way

info/v1/container.go
KielChan: ok, I will move it

info/v1/container.go
KielChan: fine, I will fix it today

metrics/prometheus.go
KielChan: @dashpole it's threads metrics, I will rename it by prefix like `Threads(Max|Current)`

metrics/prometheus.go
```diff
@@ -1045,7 +1045,30 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 					return metricValues{{value: float64(s.Processes.FdCount), timestamp: s.Timestamp}}
 				},
 			},
+			{
+				name:      "container_pids_max",
```
dashpole: `container_threads_max`

metrics/prometheus.go
```diff
@@ -1045,7 +1045,30 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				valueType: prometheus.GaugeValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{
+						value:     float64(s.Processes.ThreadsMax),
+						timestamp: s.Timestamp,
+					}}
+				},
+			},
+			{
+				name:      "container_pids_current",
```
dashpole: `container_threads`.  To be consistent with other metrics, omit the `_current` suffix.

metrics/prometheus.go
KielChan: fine, I will fix it.

container/raw/handler.go
```diff
@@ -77,6 +77,9 @@ func newRawContainerHandler(name string, cgroupSubsystems *libcontainer.CgroupSu
 	pid := 0
 	if isRootCgroup(name) {
 		pid = 1
+
+		// delete pids from cgroup paths because /sys/fs/cgroup/pids/pids.current not exist
+		delete(cgroupPaths, "pids")
```
dashpole: This means we won't have current or max threads for the root cgroup, right?

That would be useful to have if we want to have the kubelet perform eviction when we are running low.  I don't think it needs to be a requirement for this PR, but we should consider adding this as a future improvement.  We would want to get the limit from `/proc/sys/kernel/threads-max`, and (probably?) would have to scan `/proc` for total thread count.

PR #2260, opened 2019-06-28
Add socketCount for container process stat

qingwave:
**What**
Add metrics `container_threads` and `container_sockets` for container。
As follows:
```
	Processes: info.ProcessStats{
		ProcessCount: 1,
		ThreadCount:  10,
		FdCount:      5,
		SocketCount:  3,
	},
```

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -144,7 +144,7 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
 }
 
 func processStatsFromProcs(rootFs string, cgroupPath string) (info.ProcessStats, error) {
-	var fdCount uint64
+	var fdCount, socketCount uint64
 	filePath := path.Join(cgroupPath, "cgroup.procs")
 	out, err := ioutil.ReadFile(filePath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


qingwave:
@googlebot signed

qingwave:
I signed it!

qingwave:
@googlebot

dashpole:
ping here when the cla is green

qingwave:
I signed it!

qingwave:
I signed it!

dashpole:
FYI there is another PR in progress to get thread stats from the `pid` cgroup.

dashpole:
Can you change the PR title and description to indicate that this only adds sockets?

Actually, what is the use-case for reporting the number of sockets?  It doesn't look like linux places any limits on the number of sockets, except the limit on FDs.

qingwave:
> Can you change the PR title and description to indicate that this only adds sockets?
> 
> Actually, what is the use-case for reporting the number of sockets? It doesn't look like linux places any limits on the number of sockets, except the limit on FDs.
Practically, I get the socket number by fd's link, and cannot find a dirct way to get it from a file or somewhere.

dashpole:
right, I'm just curious what your use-case is for monitoring the number of sockets each container is using.  Outside of running out of FDs, what problems can occur when too many sockets are open?

qingwave:
There are some nginx services in our k8s cluster, sometimes the nginx server containers receive too many requests within minutes, which the host will drop packets. Thus, we need the socket number to find the trouble-maker container on time and set the suitable connections for the server.

dashpole:
Which limit causes the host to drop packets?  A common pattern in cAdvisor is to monitor both the usage, and the limit so operators (or even eventually the kubelet or other actors) can see when usage is approaching the limit and take action to prevent disruption.

qingwave:
Lost packets maybe another thing, I think the container socket number is a kind of resource that should be exported, the user can  adjust config by the socket usage.

qingwave:
Lost packets maybe another thing, I think the container socket number is a kind of resource that should be exported, the user can  adjust config by the socket usage.

dashpole:
Ok, sounds reasonable.  Thanks!

Issue #2264, opened 2019-07-03
Cadvisor using too much memory


woosley:
see below screenshot, from time to time, the cadvisor process on some k8s nodes would consume a lot of memory and eventually lead to system OOM. 
![image](https://user-images.githubusercontent.com/61531/60567748-91059800-9d9d-11e9-884e-71bc0358a1d1.png)

cadvisor version: v0.30.2

kabakaev:
@dashpole @woosley, it's nice that you've set 2GB memory limit so that cadvisor would be killed and restarted regularly. But it is clearly a workaround. Do you have an idea why it needs so much RAM?

dashpole:
@woosley can you say more about your setup?  What container runtime are you using, and which exporters/storage backends are you using?  How many containers are running on your node?

If you restart it with `--profiling=true`, you should be able to access the pprof endpoint at /debug/pprof

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2265, opened 2019-07-03
set memory limitation for cadvisor 

woosley:
see issue #2264 
this patch adds default memory limitation for Cadvisor in K8s so k8s will restart it.


woosley:
I signed it!

PR #2269, opened 2019-07-18
Use k8s.io/utils/inotify and resolved godep conflicts

Pothulapati:
Part of https://github.com/kubernetes/kubernetes/issues/80092

As `github.com/sigma/go-inotify` is archieved, the package has been forked into `k8s.io/utils/inotify` which is being planned to use in all k8s repos.

More about this https://github.com/kubernetes/kubernetes/issues/75526

@dims 

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>


Pothulapati:
I signed it!

Issue #2259, opened 2019-06-25
Config knob to disable collecting root cgroup stats


sashankreddya:
We have seen a performance issue while reading memory.stat file for “rootCgroup”. The time taken to read "/sys/fs/cgroup/memory/memory.stat" is of the order of "3~8s". This also correlates to one of the earlier discussion thread as in [here](https://github.com/google/cadvisor/issues/1774#issuecomment-406314361) seen by others of the openSource community. Once we disable collecting root cgroup stats with some thing like below, the CPU spike reduced from close to 100% to 2%.

`
func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
        if disable_rootCgroup_stats && isRootCgroup(self.name) {
            return nil, nil
        }
`

It would be nice if we can have a config knob to not collect root cgroup stats if set.

smaftoul:
Fixed in #2283 , can be closed !

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2283, opened 2019-08-07
Provide a config knob to disable collecting root Cgroup stats #2259

sashankreddya:
#2259

### DIFF
--- a/container/raw/handler.go
+++ b/container/raw/handler.go
@@ -227,6 +227,9 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 }
 
 func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
+	if *disableRootCgroupStats && isRootCgroup(self.name) {
+		return nil, nil
+	}
 	stats, err := self.libcontainerHandler.GetStats()
 	if err != nil {
 		return stats, err
--- a/container/raw/factory.go
+++ b/container/raw/factory.go
@@ -30,6 +30,7 @@ import (
 )
 
 var dockerOnly = flag.Bool("docker_only", false, "Only report docker containers in addition to root stats")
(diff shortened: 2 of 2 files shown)


sashankreddya:
@dashpole : Looks like some test eco system issue in pulling changes (pull-cadvisor-e2e).

dashpole:
```
container/raw/factory.go container/raw/factory.go
```

sashankreddya:
> ```
> The following files are not properly formatted:
> container/raw/factory.go container/raw/factory.go
> ```

Fixed it.

container/raw/handler.go
```diff
@@ -227,6 +227,9 @@ func (self *rawContainerHandler) getFsStats(stats *info.ContainerStats) error {
 }
 
 func (self *rawContainerHandler) GetStats() (*info.ContainerStats, error) {
+	if *disableRootCgroupStats && isRootCgroup(self.name) {
```
dashpole: can we do this in CanHandleAndAccept in factory.go similar to the `docker_only` flag?

sashankreddya: Sure will test it and update you.

sashankreddya: @dashpole :

We cant do it in "CanHandleandAccept" currently as [this step](https://github.com/google/cadvisor/blob/master/manager/manager.go#L273) throws error that it failed to find the container ("/") as the [container creation step for "/"](https://github.com/google/cadvisor/blob/master/manager/manager.go#L268) skips creating the container with message 

> 

I0810 00:05:34.490103  103510 factory.go:170] Factory "raw" can handle container "/", but ignoring.
I0810 00:05:34.490112  103510 manager.go:908] ignoring container "/"

>

dashpole: Ok.  Thanks for looking into it.

trallnag:
@dashpole documentation on `disable_root_cgroup_stats` option is still missing in https://github.com/google/cadvisor/blob/master/docs/runtime_options.md

PR #2284, opened 2019-08-08
container/crio: retry getting pid if 0

sjenning:
When a container is created, cadvisor creates a handler, triggered by inotify.  In the case of a crio managed container, the crio handler immediately queries crio for the pid of the just created container.  Due to performance concerns in crio, the pid might not yet be known and thus returns 0.  If the pid is 0, then network stats are not collected for the lifetime of the container or the lifetime of cadvisor, whichever is shorter.

https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L80-L83

This PR changes the crio handler to requery crio for the pid if the pid was unknown during handler initialization or the last `GetStat()` call.  Once the pid becomes known, the libcontainer handler with the non-zero pid is cached and no further crio queries are needed.

Fixes https://github.com/openshift/origin/issues/23492

@mrunalp @dashpole

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -33,6 +33,9 @@ import (
 )
 
 type crioContainerHandler struct {
+	client crioClient
+	name   string
+
 	machineInfoFactory info.MachineInfoFactory
 
 	// Absolute path to the cgroup hierarchies of this container.
@@ -68,6 +71,9 @@ type crioContainerHandler struct {
 	reference info.ContainerReference
 
 	libcontainerHandler *containerlibcontainer.Handler
+	cgroupManager       *cgroupfs.Manager
+	rootFs              string
+	pidKnown            bool
 }
 
 var _ container.ContainerHandler = &crioContainerHandler{}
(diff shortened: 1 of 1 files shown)


mrunalp:
:+1:

sjenning:
fyi @Reamer

Reamer:
Hi @sjenning,
I will try this fix today in my test environment.

Reamer:
I think you have a small typo.
```patch
diff --git a/container/crio/handler.go b/container/crio/handler.go
index 7d0c5a83..d9eab06e 100644
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -296,7 +296,7 @@ func (self *crioContainerHandler) getLibcontainerHandler() *containerlibcontaine
 	}
 
 	self.pidKnown = true
-	self.libcontainerHandler = containerlibcontainer.NewHandler(self.cgroupManager, self.rootFs, cInfo.Pid, self.ignoreMetrics)
+	self.libcontainerHandler = containerlibcontainer.NewHandler(self.cgroupManager, self.rootFs, cInfo.Pid, self.includedMetrics)
 
 	return self.libcontainerHandler
 }
```

sjenning:
ah yes, I made this patch against origin 3.11 vendor tree and that changed in master

PR #2289, opened 2019-08-16
Support build this project by docker

purebluesong:
Building in docker will do not make some effects on the local go environment


purebluesong:
@googlebot I signed it!

PR #2292, opened 2019-08-19
support multi tmpfs in cadvisor

zouyee:
support multi tmpfs in cadvisor
fix https://github.com/kubernetes/kubernetes/issues/81575

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -176,9 +176,11 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
 		if !strings.HasPrefix(mount.Fstype, "ext") && !supportedFsType[mount.Fstype] {
 			continue
 		}
-		// Avoid bind mounts.
+		// Avoid bind mounts, exclude tmpfs.
 		if _, ok := partitions[mount.Source]; ok {
-			continue
+			if mount.Fstype != "tmpfs" {
+				continue
+			}
 		}
 
 		hasPrefix := false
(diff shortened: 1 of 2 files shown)


zouyee:
related PR: https://github.com/google/cadvisor/pull/1667
/cc @dashpole

dashpole:
I don't fully understand the issue.  Would you mind explaining it in detail?

zouyee:
Have multiple tmpfs mounts which have different major:minor pairs, e.g.:

```
root@hostname:/# cat /proc/self/mountinfo | grep tmpfs

404 394 0:18 / /dev/shm rw,nosuid,nodev shared:22 - tmpfs tmpfs rw,seclabel
1715 278 0:36 /var/lib/kubernetes/ssl/kubelet /var/lib/kubernetes/ssl ro,relatime shared:1 - tmpfs tmpfs rw,seclabel,mode=755
...
1775 278 0:36 /var/lib/kubelet /var/lib/kubelet rw,relatime shared:1 - tmpfs tmpfs rw,seclabel,mode=755
1820 278 0:36 /etc/kubernetes /etc/kubernetes rw,relatime shared:1 - tmpfs tmpfs rw,seclabel,mode=755
1845 278 0:36 /etc/ssl/certs /etc/ssl/certs ro,relatime shared:1 - tmpfs tmpfs rw,seclabel,mode=755
1877 278 0:36 /etc/hosts /etc/hosts rw,relatime shared:1 - tmpfs tmpfs rw,seclabel,mode=755
```
[_, ok := partitions[mount.Source]](https://github.com/google/cadvisor/blob/master/fs/fs.go#L181)
If mount.Source is `tmpfs`, other mountpoints which source also is tmpfs, will be escape.

dashpole:
Right, I see why it exits.  I'm trying to figure out why this works for other filesystems, but not for tmpfs. Is mount.Source normally unique?

zouyee:
mount.Source is unique.

dashpole:
You will have to rebase.  Ping the PR after you do that so I can merge it.

fs/fs.go
```diff
@@ -176,8 +176,8 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
 		if !strings.HasPrefix(mount.Fstype, "ext") && !supportedFsType[mount.Fstype] {
 			continue
 		}
-		// Avoid bind mounts.
-		if _, ok := partitions[mount.Source]; ok {
+		// Avoid bind mounts, exclude tmpfs.
+		if _, ok := partitions[mount.Source]; ok && mount.Fstype != "tmpfs" {
```
dims: can we please move `mount.Fstype != "tmpfs"` into a separate `if` condition?

zouyee: > can we please move `mount.Fstype != "tmpfs"` into a separate `if` condition?

Done

zouyee:
@dashpole PTAL

PR #2291, opened 2019-08-16
Provide PSR information for containers.

mJace:
Allow cadvisor to provide PSR information when doing `ps` to containers now.
PSR is a `ps` output option, it shows which cpu that current process runs on.

The related issue:
https://github.com/google/cadvisor/issues/2290

PR #2296, opened 2019-08-22
Fix printed output for prow script

dashpole:
We the makes it print 'make assets FORCE=true', rather than run it (which just exits 1 since it fails).

Related: #2291

/assign @mJace


mJace:
LGTM

dashpole:
Merging w/o approval based on https://github.com/google/cadvisor/pull/2296#issuecomment-524142220

PR #2304, opened 2019-09-03
Add hugepage info to v1 node structure

ohsewon:
Purpose of this PR: Add the new field which describes the number of pre-allocate hugepages per NUMA node to v1 node structure.

The node structure is used to describe the resources of the NUMA node, inside the MachineInfo structure.
The new field will describe the number of pre-allocate hugepages per NUMA node.
The purpose of the new filed is to put additional information for future usage.
The future usage means guaranteeing alignment of node resources such as CPU, GPU, NIC, and hugepages by Kubelet component like Topology Manager.

Signed-off-by: sewon.oh <sewon.oh@samsung.com>


ohsewon:
@googlebot I signed it!

bg-chun:
This PR is related to the below issue and KEP.
Issue: https://github.com/kubernetes/kubernetes/issues/80716
KEP: https://github.com/kubernetes/enhancements/pull/1199

And also mentioned in the below slack thread.
https://kubernetes.slack.com/archives/C0BP8PW9G/p1567484868027900

dashpole:
Let me know once the KEP is approved, and i'll take a look at this.

bg-chun:
@ohsewon 
plz check it out.
```
W0904 02:58:03.092] # github.com/google/cadvisor/machine [github.com/google/cadvisor/machine.test]
W0904 02:58:03.093] machine/topology_test.go:63:7: node.HugePagesInfo undefined (type v1.Node has no field or method HugePagesInfo)
W0904 02:58:03.094] machine/topology_test.go:63:35: topology[i].HugePagesInfo undefined (type v1.Node has no field or method HugePagesInfo)
I0904 02:58:04.368] Makefile:38: recipe for target 'vet' failed
W0904 02:58:04.468] make: *** [vet] Error 2
```
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/google_cadvisor/2304/pull-cadvisor-e2e/1169081552542371840/build-log.txt

machine/machine.go
```diff
@@ -191,6 +194,45 @@ func getNodeIdFromCpuBus(cpuBusPath string, threadId int) (int, error) {
...
+			return nil, err
+		}
+
+		pageNum, err := strconv.ParseUint(string(bytes.TrimSpace(num)), 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		hugePagesInfo = append(hugePagesInfo, info.HugePagesInfo{
+			PageSize: pageSize / 1024, // Convert to kB.
```
odinuge: Isn't this size already in `kB`?

A test case would be suitable I guess.

machine/machine.go
```diff
@@ -191,6 +194,45 @@ func getNodeIdFromCpuBus(cpuBusPath string, threadId int) (int, error) {
...
+	hugePagesInfo := []info.HugePagesInfo{}
+	path := filepath.Join(nodePath, fmt.Sprintf("node%d/hugepages", nodeIndex))
+	files, err := ioutil.ReadDir(path)
+	// Ignore if per-node info is not available.
+	if err != nil {
+		klog.Errorf("failed to get hugepages information for node %d: %v", nodeIndex, err)
+		return nil, nil
+	}
+
+	for _, file := range files {
```
odinuge: Is there a reason we cannot reuse some of the code from https://github.com/google/cadvisor/blob/f5e7ddf27add8bf6b3ead954d5b09d895a8597d7/machine/info.go#L66? The concept is the same here as it is there.

machine/machine.go
```diff
@@ -191,6 +194,45 @@ func getNodeIdFromCpuBus(cpuBusPath string, threadId int) (int, error) {
 	return nodeId, nil
 }
 
+/* Look for per-node hugepages info using node id */
```
odinuge: Can you switch to normal `//` comments here to keep the consistency?

machine/machine.go
odinuge: In the normal hugepage variant they use the pageSize directly, and since it is parsed from `hugepages-<xyz>kB` that should be correct.

https://github.com/google/cadvisor/blob/f5e7ddf27add8bf6b3ead954d5b09d895a8597d7/machine/info.go#L90

info/v1/machine.go
```diff
@@ -38,9 +38,10 @@ type FsInfo struct {
 type Node struct {
 	Id int `json:"node_id"`
 	// Per-node memory
-	Memory uint64  `json:"memory"`
-	Cores  []Core  `json:"cores"`
-	Caches []Cache `json:"caches"`
+	Memory    uint64          `json:"memory"`
+	HugePages []HugePagesInfo `json:"huge_pages"`
```
derekwaynecarr: this should be `hugepages` so its consistent with serialization in MachineInfo

machine/machine.go
derekwaynecarr: Agreed.

machine/machine.go
derekwaynecarr: can we consolidate how we parse capacity in `GetHugePagesInfo` in info.go with how we would parse here?  I am fine with the regex approach, I just want us to have a common approach.

machine/machine.go
derekwaynecarr: it seems we should be able to reuse the common logic for how we parse /sys/mm/kernel/hugepages with what we parse here so we dont do it differently.  can we consolidate on a single approach and reuse what can be reused?

odinuge: Yes, I agree that we should reuse as much as possible. This is handled together in the Kernel, so imo. it makes perfect sense to just implement this once in cadvisor. The only thing that differ between them is the filepath, and maybe a bit different error handling, but the implementation should be similar.

machine/machine.go
bg-chun: It is intended for the consistency.
Author(of this PR) and I referred below functions in this file.
Those functions have similar usage(parsing info from the given path) and they have comment style like below.
```
/* Look for sysfs cpu path containing core_id */ 
/* Such as: sys/bus/cpu/devices/cpu0/topology/core_id */ 
```
Take look below functions.
https://github.com/google/cadvisor/blob/f5e7ddf27add8bf6b3ead954d5b09d895a8597d7/machine/machine.go#L135
https://github.com/google/cadvisor/blob/f5e7ddf27add8bf6b3ead954d5b09d895a8597d7/machine/machine.go#L159

info/v1/machine.go
bg-chun: agreed

machine/machine.go
bg-chun: @derekwaynecarr @odinuge 
`it seems we should be able to reuse the common logic`
`can we consolidate on a single approach`
`Yes, I agree that we should reuse as much as possible.`
=> I totally agree with it. :)

IMO, I think the below strategy is the best option for us.
1) Move `GetHugePagesInfo()` from `info.go` to `machine.go`.
 _(It means that we will use existing way to parse hugepage capacity, not a regex.)_
2) Then make it(`machine.GetHugePagesInfo()`) to take filepath and to parse capacity from a given file for `reusability`.
_(it will be `machine.GetHugePagesInfo(path string)`)_
3) Now, we can reuse it(`machine.GetHugePagesInfo(path string)`) in both of `info.go` and `machine.go`.
4) In `info.go`, `machine.GetHugePagesInfo(path string)` will be used to parse capacity at the machine level.
5) in 'machine.go', `machine.GetHugePagesInfo(path string)` will be used to parse capacity at the NUMA node level.

How about it?

info/v1/machine.go
ohsewon: fixed it.

machine/machine.go
ohsewon: I removed duplicated logic.

machine/machine.go
```diff
@@ -352,6 +393,15 @@ func addNode(nodes *[]info.Node, id int) (int, error) {
 			}
 			node.Memory = uint64(m)
 		}
+		// Look for per-node hugepages info using node id
+		// Such as: /sys/devices/system/node/node%d/hugepages
+		hugepagesDirectory := nodePath + "/node" + string(id) + "/hugepages/"
```
odinuge:
```suggestion
		hugepagesDirectory := fmt.Sprintf("%s/node%d/hugepages/", nodePath, id)
```

machine/machine.go
```diff
@@ -191,6 +193,45 @@ func getNodeIdFromCpuBus(cpuBusPath string, threadId int) (int, error) {
 	return nodeId, nil
 }
 
+// GetHugePagesInfo returns information about pre-allocated huge pages
+func GetHugePagesInfo(hugepagesDirectory string) ([]info.HugePagesInfo, error) {
```
odinuge: Maybe add a comment about what directories this function wants.

odinuge: nit: This makes it easy to write some simple tests. 

There are already some tests her: https://github.com/google/cadvisor/blob/master/machine/topology_test.go. The coverage of unit tests is low, but some tests never hurt I guess. :smile:

ohsewon: I added unit test and comment

derekwaynecarr:
Thank you for making the update and adding tests!

/lgtm
/approve

derekwaynecarr:
@dashpole are you able to help merge this?

dashpole:
Done.

PR #2309, opened 2019-09-06
[RFC] cgroup: initial support for cgroups v2

giuseppe:
add some initial support for cgroups v2.  Not all the stats supported on cgroups v1 are supported, e.g. it is not possible to read percpu usage.


container/common/helpers.go
```diff
@@ -110,9 +118,25 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
...
-			spec.Memory.Limit = readUInt64(memoryRoot, "memory.limit_in_bytes")
-			spec.Memory.SwapLimit = readUInt64(memoryRoot, "memory.memsw.limit_in_bytes")
-			spec.Memory.Reservation = readUInt64(memoryRoot, "memory.soft_limit_in_bytes")
+			if cgroups.IsCgroup2UnifiedMode() {
+				if utils.FileExists(path.Join(memoryRoot, "memory.max")) {
+					spec.Memory.Reservation = readUInt64(memoryRoot, "memory.high")
+					spec.Memory.Limit = readUInt64(memoryRoot, "memory.max")
+					spec.Memory.SwapLimit = readUInt64(memoryRoot, "memory.swap.max")
+				} else {
+					limit, err := machine.GetMachineMemoryCapacity()
```
dashpole: Is the cgroup always unlimited when memory.max doesn't exist?  Could it be affected by the parent cgroup?

If at all possible, we should keep the behavior between the two modes the same.

container/common/helpers.go
```diff
@@ -155,6 +183,9 @@ func readUInt64(dirpath string, file string) uint64 {
 	if out == "" || out == "max" {
 		return 0
 	}
+	if out == "max" {
```
dashpole: This is impossible to hit, since we just checked `out == "" || out == "max"` above.

manager/manager.go
```diff
@@ -918,11 +918,11 @@ func (m *manager) createContainerLocked(containerName string, watchSource watche
...
 		return err
 	}
-	devicesCgroupPath, err := handler.GetCgroupPath("devices")
+	memoryCgroupPath, err := handler.GetCgroupPath("memory")
 	if err != nil {
-		klog.Warningf("Error getting devices cgroup path: %v", err)
+		klog.Warningf("Error getting memory cgroup path: %v", err)
 	} else {
-		cont.nvidiaCollector, err = m.nvidiaManager.GetCollector(devicesCgroupPath)
+		cont.nvidiaCollector, err = m.nvidiaManager.GetCollector(memoryCgroupPath)
```
dashpole: I don't think this is what we want to do... The nvidia manager needs the devices cgroup to discover devices.

giuseppe: should we skip this part for cgroups v2 since there is no devices controller?

dashpole: I think the way it was written already handles the case where the devices cgroup doesn't exist.

container/common/helpers.go
giuseppe: good point, yes we should definitely look into the parent directory and see if any limit is set there.  I've changed it in the new version I've pushed

container/common/helpers.go
giuseppe: ouch, sure.  Dropped it

manager/manager.go
giuseppe: I've guarded it with a `if !cgroups.IsCgroup2UnifiedMode() {` otherwise it prints a warning on cgroups v2

container/common/helpers.go
dashpole: Sorry, that was more of a question than an assertion about the file not existing...  I wasn't sure under what conditions memory.max wouldn't exist.  I took a look at the v2 documentation for it, and the file should exist for non-root cgroups: https://www.kernel.org/doc/Documentation/cgroup-v2.txt.  We just need to handle the case where it is the root cgroup, and use the machine limit as you did before.  I don't think we need to search the higherarchy, since that maintains the same behavior as cgroups v1.

I looked at the cgroup v1 implementation, and it looks like the default value is a very high integer.  For me, it was 9223372036854771712 bytes instead of the machine limit.

giuseppe: with cgroups v2 it is possible though to enable the memory controller only for the parent cgroup, where `memory.max` exists and not enable it for the child directory.  In that case the limit still exists.

```
mkdir -p /sys/fs/cgroup/foo/bar
# ls /sys/fs/cgroup/foo/memory.max /sys/fs/cgroup/foo/bar/memory.max
ls: cannot access '/sys/fs/cgroup/foo/bar/memory.max': No such file or directory
/sys/fs/cgroup/foo/memory.max
```

In this case processes added to  `/sys/fs/cgroup/foo/bar` are still limited in memory by   '/sys/fs/cgroup/foo/memory.max`.

dashpole: oh, ok.  Makes sense.

Issue #2311, opened 2019-09-11
Collect/report file descriptor limits for containers


sahilrihan:
Currently cadvisor reports the number of open fds per container. It would be useful to also report the fd limits configured for a container. The limit can be combined with the actual usage to generate a percentage-based alert to warn service owners.

sahilrihan:
@sashankreddya , @dashpole : since you were involved with #1994 which added counts for open fds, what are your thoughts?

sahilrihan:
nofile can be set for containers just like it's set for normal system users. For docker it can be set at container launch time using the --ulimit flag.

sahilrihan:
I can put up a diff if it's useful

dashpole:
Feel free to just put up a PR and reference this issue.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2315, opened 2019-09-17
Add support for emitting file descriptor limits for containers.

sahilrihan:
This change is intended to address issue #2311.

It adds a ContainerSpec field for Ulimits and populates it for docker containers from the HostConfig. I only emitted the fd soft limit since that's the one we need right now. In theory we could emit all ulimits that are populated (and generate metric names from the resource names), but it seemed better to avoid emitting unnecessary metrics.

The fd soft limit is emitted as container_spec_file_descriptor_limit_soft, which is longer than I would like, but seems to be consistent with the naming convention.

An alternative approach that might be more "container agnostic" involves looking up all the processes in the cgroup, and emitting the ulimits for the "parent/init/session" process for the cgroup. However, it might be more "fragile".

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -43,6 +44,8 @@ type Handler struct {
 	includedMetrics container.MetricSet
 	pidMetricsCache map[int]*info.CpuSchedstat
 }
+
+var whitelistedUlimits = [...]string{"max_open_files"}
 
 func NewHandler(cgroupManager cgroups.Manager, rootFs string, pid int, includedMetrics container.MetricSet) *Handler {
 	return &Handler{
@@ -125,7 +128,7 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
 		if !ok {
 			klog.V(4).Infof("Could not find cgroups CPU for container %d", h.pid)
 		} else {
-			stats.Processes, err = processStatsFromProcs(h.rootFs, path)
+			stats.Processes, err = processStatsFromProcs(h.rootFs, path, h.pid)
 			if err != nil {
 				klog.V(4).Infof("Unable to get Process Stats: %v", err)
 			}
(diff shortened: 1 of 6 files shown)


sahilrihan:
@googlebot I signed it!

container/docker/handler.go
```diff
@@ -218,6 +222,7 @@ func newDockerContainerHandler(
 		Aliases:   []string{strings.TrimPrefix(ctnr.Name, "/"), id},
 		Namespace: DockerNamespace,
 	}
+	handler.ulimits = ctnr.HostConfig.Ulimits
```
dashpole: What would we need to do to get Ulimits in a container-runtime agnostic way?

sahilrihan: As mentioned in the PR: An alternative approach that might be more "container agnostic" involves looking up all the processes in the cgroup, and emitting the ulimits for the "parent/init/session" process for the cgroup. However, it might be more "fragile".

Also, I'm not sure if container ulimit should be inferred for container-runtimes that don't explicitly provide the ability to set ulimits.

dashpole: Right.  Why would it be more fragile?  Are you suggesting a change in container runtime implementation would break the metric?

How does the process -> ulimit lookup work?

sahilrihan: The difference is that in this case we're reading the value from the container config. It's what the owner configured at container launch time.

For process->ulimit we'd have find the processes in the cgroups by looking at /sys/fs/cgroup/*/<container>/cgroup.procs, then selecting the process and reading the limits from /proc/<pid>/limits

To do it correctly we probably need to read all the cgroups and find the lowest number pid across all of them. The lowest pid is not necessarily the first pid since pids can wrap around, so ideally we would use some equivalent of pstree, or get the parent pid from the container runtime (I haven't looked into how to do this part).

dashpole: I did a little bit of digging, and I think it should be easy to implement this in a container-runtime agnostic way.

We already get the PID of the container for all container runtimes.  For docker, it is [ctnr.State.Pid](https://github.com/google/cadvisor/blob/master/container/docker/handler.go#L212).

I checked to make sure the PID has the limits specified with `--ulimit` applied to it, following the [docker run example in the docs](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit):

$ docker run --ulimit nofile=1024:1024 --rm debian sh -c "sleep 60; ulimit -n" &
$ docker ps | grep debian 
42048ce156e3 .....
$ docker inspect 42048ce156e3 | grep Pid
            "Pid": 2834013,
            "PidMode": "",
            "PidsLimit": 0,
$ cat /proc/2834013/limits | grep "open files"
Max open files            1024                 1024                 files

I did the same for a container without --ulimit specified, and got:
Max open files            1048576              1048576              files

That shows that the limits are being applied to the PID specified in container.State.Pid, which is what we have access to in cAdvisor already.

To implement this, we would want to add this logic to https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L122.

sahilrihan: Force pushed a container agnostic version per your suggestions.

metrics/prometheus.go
```diff
@@ -1079,6 +1079,23 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
 				},
 			},
+			{
+				name:        "container_root_process_ulimits_soft",
+				help:        "Soft ulimit values for the container root process.",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"ulimit_name"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					values := make(metricValues, 0, len(s.Processes.Ulimits))
+					for _, ulimit := range s.Processes.Ulimits {
```
sahilrihan: @dashpole : this will generate 16 metrics for each container. if that's too much I can scope it down to just file descriptors for now, and people can emit other ulimits as and when they're needed.

container/docker/handler.go
sahilrihan: @dashpole : ping. updated to be container agnostic per your suggestion

container/libcontainer/handler.go
```diff
@@ -143,7 +144,47 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
...
+			soft := strings.TrimSpace(fields[1])
+			soft_num, err := strconv.Atoi(soft)
+			if err != nil {
+				soft_num = -1
+			}
+
+			hard := strings.TrimSpace(fields[2])
+			hard_num, err := strconv.Atoi(hard)
+			if err != nil {
+				hard_num = -1
```
dashpole: I'm assuming -1 means unlimited?  How hard would it be to use the node-level limit if no container limit exists?

metrics/prometheus.go
dashpole: Yeah, that is a lot.  I would prefer keeping the prometheus metrics the same as the json metrics, though.  If we do filtering, lets do it at collection time.

Usually, it starts out as a constant whitelist defined in the file.  We may need to add it as a flag eventually, but we can start with it hard-coded.

metrics/prometheus.go
```diff
@@ -1079,6 +1079,23 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 					}
 				},
 			},
+			{
+				name:        "container_root_process_ulimits_soft",
```
dashpole: you can remove "root_process" from the name.  That is implied with other metrics (e.g. threads_max).  You can leave it in the help text if you want.

metrics/prometheus.go
```diff
@@ -1079,6 +1079,23 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				help:        "Soft ulimit values for the container root process.",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"ulimit_name"},
```
dashpole: omit "_name" from the label name.

container/libcontainer/handler.go
sahilrihan: @dashpole : yes, -1 means unlimited in general, with the exception of nice and priority. I think it might be better to drop the metric in that case, or only emit the soft/hard limit depending on which one has a valid value.

W.r.t. your second question: Do you mean host-level limit when you say node-level? The limits are inherited from dockerd (which in turn gets it from system settings or configurations), if they are not specified at the container level. https://docs.docker.com/engine/reference/commandline/dockerd/#default-ulimit-settings
So if you're concerned about what values to default to, that's already taken care of by Linux/dockerd.

dashpole: Yeah, I mean host-level.

So are you saying that the ulimit for the container already takes host-level limits into account?  I.E. if it is -1, then that means it is unlimited on the host?

sahilrihan: Yes. If it's not set at the container level, it inherits the value from dockerd, which in turn gets its ulimits from systemd, etc. In general, linux processes inherit ulimit values from their parent during a fork: http://danielzhou82.github.io/blog/2014/09/30/on-the-confusing-ulimit/  The process can set soft limit up to hard limit, but cannot change the hard limit, unless it is a privileged process.

dashpole: Thanks for explaining.

container/docker/handler.go
dashpole: Thanks.

sahilrihan:
@dashpole : updated with review feedback. Also added a unit test for parsing errors, and included -1 explanation to help message.

metrics/prometheus.go
dashpole: Can we add a whitelist to container/libcontainer/handler.go to only store ulimit for FD limits?
Or it can include others if you think more are commonly used.

sahilrihan: Sorry about that, I didn't fully follow what you meant the first time around. Added a whitelist array to handler.go

PR #2319, opened 2019-10-01
Update the reference to Source attribute via safer Get API for mesos containers.

sashankreddya:
GetSource API for mesosContainers ensures to return a safer empty string on non-existence unlike the previous reference which would crash.


sashankreddya:
@dashpole

PR #2320, opened 2019-10-02
Drop support for rkt - which is now archived

dims:
- https://github.com/cncf/toc/pull/262
- https://www.cncf.io/blog/2019/08/16/cncf-archives-the-rkt-project/

Change-Id: I4bc38990251fe12be521a242ff9b9c81e2d03501


dims:
@dashpole WDYT?

dashpole:
This makes sense to me.  I assume kubernetes has already dropped support (if it still has any in-tree support)?

dims:
@dashpole way long ago (a few months :)

https://github.com/kubernetes/kubernetes/pulls?utf8=%E2%9C%93&q=is%3Apr+rkt+in%3Atitle

PR #2325, opened 2019-10-14
runtime option for housekeeping jitter

Astrologer:
Jitter itself is a good thing, but it can cause some non obvious side effects in case of prometheus.

For example I want to collect metrics every 30 seconds. I have a bunch of containers so I'm using cadvisor to collect metrics from them with housekeeping_interval = 30 seconds and I have prometheus exporter to collect data from cadvisors and some other subsysterms (and it collects every 30 seconds too).

Such setup couses a chain of issues. First off all, real interval isn't a 30 seconds, it's 30-60 seconds and this part isn't documented. As a result cadvisor has less measurments than prometheus exporter. Looks... not terrible but it can cause signiffican fluctuations in exported cpu usage metric for example.

Cadvisor exports cpu seconds as a counter, like 10, 20, 30 and so on. But if prometheus exporter is getting two equal values in a row (cadvisor is a bit slower in our case, since interval is 30-60 seconds) it thinks that cpu usage is 0, because nothing were changed for the last 30 seconds.

So I'm forsed to set housekeeping interval to 15 seconds to make sure that metrics are collecte at least once. I personally would prefer to be able to shrink jitter factor if needed.

### DIFF
--- a/docs/runtime_options.md
+++ b/docs/runtime_options.md
@@ -79,6 +79,7 @@ Per-container housekeeping is run once on each container cAdvisor tracks. This t
 --global_housekeeping_interval=1m0s: Interval between global housekeepings
 --housekeeping_interval=1s: Interval between container housekeepings
 --max_housekeeping_interval=1m0s: Largest interval to allow between container housekeepings (default 1m0s)
+--housekeeping_jitter_factor=1.0: Housekeeping interval jitter factor. Full interval equals interval + factor * random(interval) (default 1)
 ```
 
 ## HTTP
--- a/manager/container.go
+++ b/manager/container.go
@@ -46,6 +46,7 @@ import (
 // Housekeeping interval.
 var enableLoadReader = flag.Bool("enable_load_reader", false, "Whether to enable cpu load reader")
 var HousekeepingInterval = flag.Duration("housekeeping_interval", 1*time.Second, "Interval between container housekeepings")
+var housekeepingJitterFactor = flag.Float64("housekeeping_jitter_factor", 1.0, "Housekeeping interval jitter factor. Full interval equals interval + factor * random(interval)")
 
 // cgroup type chosen to fetch the cgroup path of a process.
 // Memory has been chosen, as it is one of the default cgroups that is enabled for most containers.
@@ -94,10 +95,10 @@ type containerData struct {
 }
 
 // jitter returns a time.Duration between duration and duration + maxFactor * duration,
-// to allow clients to avoid converging on periodic behavior.  If maxFactor is 0.0, a
+// to allow clients to avoid converging on periodic behavior.  If maxFactor is less than 0.0, a
 // suggested default value will be chosen.
 func jitter(duration time.Duration, maxFactor float64) time.Duration {
-	if maxFactor <= 0.0 {
+	if maxFactor < 0.0 {
 		maxFactor = 1.0
 	}
 	wait := duration + time.Duration(rand.Float64()*maxFactor*float64(duration))
@@ -439,7 +440,7 @@ func (self *containerData) nextHousekeepingInterval() time.Duration {
 		}
 	}
 
-	return jitter(self.housekeepingInterval, 1.0)
+	return jitter(self.housekeepingInterval, *housekeepingJitterFactor)
 }
 
 // TODO(vmarmol): Implement stats collecting as a custom collector.


Astrologer:
@googlebot I signed it!

dashpole:
I agree with the point about documentation.  Feel free to open a PR to update it to something that would have made it clearer for you.  Feel free to update the help text for the flag as well if that would have been useful.

> But if prometheus exporter is getting two equal values in a row (cadvisor is a bit slower in our case, since interval is 30-60 seconds) it thinks that cpu usage is 0, because nothing were changed for the last 30 seconds

> it can cause signiffican fluctuations in exported cpu usage metric for example

We attach timestamps to metrics for this reason.  When metrics are collected out-of-band (rather than  at query time), this can be an issue.  But with timestamps, prometheus should treat getting the same two metrics with the same timestamp as essentially not collecting a metric, and handle it correctly.  It should not be causing fluctuations in CPU usage with timestamps attached to metrics.

It looks like you stumbled on the recommended solution for this problem, which is setting the prometheus scrape interval to be at least 2x the max collection interval.  If we can solve your issues without adding additional configuration to cAdvisor, that would be preferable.

Astrologer:
>  If we can solve your issues without adding additional configuration to cAdvisor, that would be preferable.

Ok let's think about issues around the jitter. First point is that when I want to run something once per second I suppose it to be run about 3600 times per hour, at least ruphly assimptoticaly. It means that idealy, jitter offset shold be simmetric (+- some value). Second point is the size of the jitter offset. 100% looks like overkill and it will produse very irregular measurements. From my point of view 10% should be more than enough to create some drifting effect and produce even distribution of load. With such default configuration there is no to much sence to add addition configuration.

Will it be ok from your point of view? I didn't considered such case from the begginning because it's a behavioral change and it may cause an issue for someone.

dashpole:
> First point is that when I want to run something once per second I suppose it to be run about 3600 times per hour, at least ruphly assimptoticaly. It means that idealy, jitter offset shold be simmetric (+- some value).

I'm not sure we can change the behavior of housekeeping at this point, as that wouldn't be backwards-compatible for users that have already set the value.  I do think we should explain this better in our documentation.

> Second point is the size of the jitter offset. From my point of view 10% should be more than enough to create some drifting effect and produce even distribution of load. With such default configuration there is no to much sence to add addition configuration.

Looking at #1053 and #1057, it looks like they made the decision on how much jitter to add based on some performance tests they were running, which I don't have insight into today.  I'm not against decreasing the amount of jitter in theory, but i'd rather not touch that knob unless you are also planning to do some fairly involved benchmarking.

dashpole:
BTW, did you figure out if you are seeing issues with fluctuations caused by this as I mentioned above?  I want to make sure that if we do have bugs there that those get fixed...

Astrologer:
>  i'd rather not touch that knob unless you are also planning to do some fairly involved benchmarking.

We are trapped then. I have quite big cluster but it's rather a bunch of small clusters, so I have no idea how to reproduce the load from the issue above. I haven't found any reasons behind specific value in the conversation. Anyway, I can't say that update is critical for me since I always can use my own updated version. I just wanted to share some concerns  about quite conterintuitive behavior.

>BTW, did you figure out if you are seeing issues with fluctuations caused by this as I mentioned above? I want to make sure that if we do have bugs there that those get fixed...

Yap that's for sure. But, as you said, in some sense it's an issue of the prometheus exporter or rather prometheus architectural approach. Prometheus doesn't suppose to use timestamps or at least doesn't like that:

> Accordingly, you should not set timestamps on the metrics you expose, let Prometheus take care of that. If you think you need timestamps, then you probably need the Pushgateway instead.

dashpole:
Prometheus does support explicit timestamps.  See here: https://prometheus.io/docs/practices/instrumentation/#offline-processing.  We use the [NewMetricWithTimestamp](https://github.com/prometheus/client_golang/blob/master/prometheus/metric.go#L172) function from the official go client library to instantiate our metrics.  What do you mean by it being an issue with the prometheus exporter?

Astrologer:
>What do you mean by it being an issue with the prometheus exporter?

I mean that if cAdvisor has timestamps then we have an issue somewhere downstream. In the prometheus itself I can see very even row of numbers, intervals are very precise and in case of missed measurement I can see the same value twice in the row. So timestamp was lost somewhere during scraping, saving or retriving... and all these things were made on the prometheus side.

Astrologer:
ok, it looks like a root cause is the version of the cAdvisor. We have 0.31 (Sep 8 2018) and timestamps support was added in the 0.9 version of the prometheus/client_golang (Oct 15 2018)

Issue #2327, opened 2019-10-18
`store_container_labels` option conflicts with `docker_env_metadata_whitelist`


Colstuwjx:
Hi,

I'm using cadvisor, in order to ignore some labels which we really don't care, I have taken `store_container_labels` and `whitelisted_container_labels` to make cadvisor ONLY exposes whitelisted container labels, it works fine.

But another requirement comes in, and things broken. we want to expose some docker environment as cadvisor prom labels, and we found `docker_env_metadata_whitelist` option, but we strangely find that it didn't work as expect. 

After deep into the code, I found that the [DefaultContainerLabels](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L1144) function implements `docker_env_metadata_whitelist` feature, and [BaseContainerLabels](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L1167)  loss it and implements another `store_container_labels` feature.

Maybe we need to also implements `docker_env_metadata_whitelist` feature inside `BaseContainerLabels` function?

Thanks.

dashpole:
Yeah, it looks like we just skip env vars all-together.  We should probably add those.  Feel free to open a PR for that.

alyssahardy:
I'm interested in this fix as well.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2330, opened 2019-10-23
Fixed missing white listed env labels(#2327)

Colstuwjx:


dashpole:
The only problem is that this is a change in behavior for non-docker runtimes, since they will now get env vars added to their metrics.  Can we also introduce a `env_metadata_whitelist` flag that isn't docker-specific and deprecate the docker-specific one?

alyssahardy:
Are there any plans to fix this bug anytime soon?

Colstuwjx:
> Are there any plans to fix this bug anytime soon?
 
@alyssahardy I have no time to work on this PR anymore, feel free to continue this job :)

bobbypage:
I'm going to close this for now since there isn't any update, please re-open if appropriate.

PR #2331, opened 2019-10-25
Update CRI-O storage driver support

saschagrunert:
Some storage drivers seems to be missing which are now supported by
CRI-O. Beside that, we removed the non-existing overlay2 driver.


dashpole:
@mrunalp can you take a quick look at this?

dashpole:
cc @runcom

giuseppe:
LGTM

PR #2338, opened 2019-11-18
Add support for go module

dqminh:
This adds initial support for go modules and remove Godeps.


dashpole:
/ok-to-test

Thanks a bunch for taking this on!

It looks like this bumps a number of dependency versions.  It would be helpful for me to be able to see which ones were updated, and why the update is being made.  I'm not against updating dependencies, but it would be best to push any that aren't strictly required for this migration to a different PR in-case any of those cause issues so we don't have to revert the whole thing...

dashpole:
```
go: downloading github.com/jteeuwen/go-bindata v3.0.7+incompatible
go: extracting github.com/jteeuwen/go-bindata v3.0.7+incompatible
go: finding github.com/jteeuwen/go-bindata v3.0.7+incompatible
flag provided but not defined: -nometadata
   -debug
     	Do not embed the assets, but provide the embedding API. Contents will still be loaded from disk.
   -ignore value
     	Regex pattern to ignore
   -nocompress
     	Assets will *not* be GZIP compressed when this flag is specified.
   -nomemcopy
     	Use a .rodata hack to get rid of unnecessary memcopies. Refer to the documentation to see what implications this carries.
   -o string
     	Optional name of the output file to be generated. (default "./bindata.go")
   -pkg string
     	Package name to use in the generated code. (default "main")
   -prefix string
     	Optional path prefix to strip off asset names.
   -tags string
     	Optional set of build tags to include.
   -version

     	Displays version information.
 Usage: go-bindata [options] <input directories>
```

It looks like the version of go-bindata might have changed?

The e2e tests are currently failing: https://k8s-testgrid.appspot.com/sig-node-cadvisor#cadvisor-e2e
I opened https://github.com/google/cadvisor/pull/2339 to fix the current error, but you might be running into other ones that aren't caused by this PR...

dqminh:
@dashpole looks like the original go-bindata repository was foobar, i have switched the one we used here to https://github.com/kevinburke/go-bindata which seems to be in better shape.

The only other version bump we did here was related to some runc dependencies that go module seems to get wrong, i spotcheck a few other dependencies and they seem to stay the same.

dashpole:
Ok, thanks for the explanation.  I tried and failed to get a version diff, so I just reviewed all of the diffs in `vendor/`.  Hopefully we don't have to revert this, since that would be painful...

Thanks a bunch for your work here, it should make the project more maintainable.

PR #2340, opened 2019-11-18
Close files

CallanTaylor:


container/libcontainer/handler.go
```diff
@@ -414,6 +414,8 @@ func udpStatsFromProc(rootFs string, pid int, file string) (info.UdpStat, error)
 		return udpStats, fmt.Errorf("couldn't read udp stats: %v", err)
 	}
 
+	r.Close()
```
dashpole: can we move this to just after we handle the error from os.Open, and `defer` it?  That way we ensure it is called if scanUdpStats(r) returns an error as well.

container/libcontainer/handler_test.go
```diff
@@ -87,6 +87,8 @@ func TestScanUDPStats(t *testing.T) {
 	if stats != udpstats {
 		t.Errorf("Expected %#v, got %#v", udpstats, stats)
 	}
+
+	r.Close()
```
dashpole: same as above.  Please `defer r.Close()` after we handle the error from os.Open()

container/libcontainer/handler.go
```diff
@@ -408,12 +408,12 @@ func udpStatsFromProc(rootFs string, pid int, file string) (info.UdpStat, error)
 	if err != nil {
 		return udpStats, fmt.Errorf("failure opening %s: %v", udpStatsFile, err)
 	}
-
-	udpStats, err = scanUdpStats(r)
```
dashpole: I don't think we want to remove this call.

PR #2350, opened 2019-12-04
Upgrading containerd and related dependencies

elweb9858:
In order to vendor in this PR (https://github.com/microsoft/hcsshim/pull/730) to kubernetes, the containerd/containerd dependency must also be upgraded to this version:
github.com/containerd/containerd v1.3.0-beta.2.0.20190828155532-0293cbd26c69

This repo currently relies on this version of containerd:
github.com/containerd/containerd v1.0.2

container/containerd/client.go imports github.com/containerd/containerd/dialer, which has since been moved to github.com/containerd/containerd/pkg/dialer. This is why cadvisor's containerd version must be bumped in order for kubernetes to bump the version as well. 

This PR bumps the containerd version to v1.3.0-beta.2.0.20190828155532-0293cbd26c69, which is what the hcsshim bump requires, but another option is to bump to the latest version, v1.3.2.


dashpole:
cc @Random-Liu 
Is there a minor/patch release we can use instead of bumping to a random commit?

dashpole:
I1204 23:02:42.041] container/containerd/client.go container/containerd/client.go

elweb9858:
@dashpole I don't think container/containerd/client.go is improperly formatted, all I did was change "github.com/containerd/containerd/dialer" to "github.com/containerd/containerd/**pkg**/dialer". Do you think this could be failing because it's not compatible with the kubernetes containerd version?

dashpole:
I don't think so... The error message definitely indicates it failed because of go formatting.  I'm not sure why your change caused it, but trying to format the file is probably your best bet.

elweb9858:
> cc @Random-Liu
> Is there a minor/patch release we can use instead of bumping to a random commit?

The earliest release with the [change ](https://github.com/containerd/containerd/commit/927517de3696b147923dedda385b0f9e705afda5#diff-7a0a01ce6a27b082c63c57b797ffe654) I need is [containerd 1.2.0-beta.0](https://github.com/containerd/containerd/releases/tag/v1.2.0-beta.0), which unfortunately is a major version bump. I'm happy to use a tagged release instead of the commit, though.

elweb9858:
@googlebot I signed it!

elweb9858:
@dashpole  closing this PR & will open up a similar one on my personal account

Issue #1633, opened 2017-04-08
Add storage_driver_password_file option for security


basvandijk:
For security reasons it would be better not to pass secrets via the command line because other processes can read the command line. So I'm requesting a feature that besides the option:

```
-storage_driver_password=some-secret
```

we get the option:

```
-storage_driver_password_file=/run/keys/my-secret
```

dashpole:
cc @timstclair security-related

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

Issue #1570, opened 2017-01-11
Authentication on Elasticsearch storage driver 


digiwhite1980:
Hi,

I would like to use the cAdvisor container to monitor docker statistics within Elasticsearch. We have shield installed on our Elasticsearch cluster. This means we need to add basic authentication when connection to an Elastic node. The help files do not show the option to add basic authentication when using Elasticsearch as the storage driver.

Anybody any suggestions?

Thnx

grokbot:
I am not 100% sure but you might be able to use http://username:password@elasticsearch:9200 as the host.

digiwhite1980:
Hi,

I've tried the solution above but it did not work out. Further I took a loot at the current pull requests. There is a pull request regarding logstash / JSON storage driver. This  pull request seems like a good solution for the current problem I'm facing. Could we get an update on this pull request as it's been out there for over a half year.

grokbot:
I am considering making a temporary docker image until the JSON storage adapter gets merged in. I am in a similar boat as you it seems.

wroujoulah:
+1

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2356, opened 2019-12-11
Db password fromfile

adrianp-sti:
A simple patch to allow the reading of a db password from a file as opposed to having to type the password on the command line when starting cadvisor.  This can be useful with e.g. docker secrets where you can specify "-storage_driver_password_file=/run/secrets/cadvisor_influxdb_password".

InfluxDB looks to be the only storage engine with authentication so it's only valid for that at the moment but would be trivial to extend to other storage engines if they gain authentication functionality e.g. #1570 

This should address #1633 

Tested with Docker 19.03.5 against InfluxDB 1.7.9

### DIFF
--- a/storage/common_flags.go
+++ b/storage/common_flags.go
@@ -21,6 +21,7 @@ import (
 
 var ArgDbUsername = flag.String("storage_driver_user", "root", "database username")
 var ArgDbPassword = flag.String("storage_driver_password", "root", "database password")
+var ArgDbPasswordFile = flag.String("storage_driver_password_file", "", "database password from file")
 var ArgDbHost = flag.String("storage_driver_host", "localhost:8086", "database host:port")
 var ArgDbName = flag.String("storage_driver_db", "cadvisor", "database name")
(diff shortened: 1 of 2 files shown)


adrianp-sti:
@googlebot I signed it!

storage/common_flags.go
```diff
@@ -21,6 +21,7 @@ import (
 
 var ArgDbUsername = flag.String("storage_driver_user", "root", "database username")
 var ArgDbPassword = flag.String("storage_driver_password", "root", "database password")
+var ArgDbPasswordFile = flag.String("storage_driver_password_file", "", "database password from file")
```
dashpole: can you add that it takes precedence over --storage_driver_password?

adrianp-sti: note added.

dashpole:
I1218 21:43:09.064] >> checking go formatting
I1218 21:43:09.575] The following files are not properly formatted:
I1218 21:43:09.575] storage/influxdb/influxdb.go storage/influxdb/influxdb.go

dashpole:
Feel free to reopen if you are still working on this.

Issue #1932, opened 2018-04-18
cAdvisor occasionally gets into a state where it has no container metadata


bboreham:
From https://github.com/kubernetes/kubernetes/issues/39812#issuecomment-381096812

The high-level symptom is that `curl /metrics` returns blank for image, namespace, etc.

example line from broken node:
```
container_cpu_system_seconds_total{container_name="",id="/kubepods/burstable/podf8b4aff6914beba1622e8dac2c24ddf2/ca26d89b90d306290a36891aa612cb06ff60f54b89ac7016d982e8a2cfa1036d/\"\"",image="",name="",namespace="",pod_name=""} 126.33
```
example line from working node:
```
container_cpu_system_seconds_total{container_name="fluentd-loggly",id="/kubepods/burstable/pode29b14ff-380e-11e8-b4ff-0a97ed59c75e/86fbd797b1ad353b890e548bfc742871be4cbebad5713a32b45257c72ea1656a",image="quay.io/weaveworks/fluentd-loggly@sha256:1b4f74d1cb976175c114802279c2f3f7ac3011740ba141503ef55def1e9b95c3",name="k8s_fluentd-loggly_fluentd-loggly-fl8cp_monitoring_e29b14ff-380e-11e8-b4ff-0a97ed59c75e_2",namespace="monitoring",pod_name="fluentd-loggly-fl8cp"} 95.1
```

We have seen this on two clusters in recent weeks, running k8s v1.9.3.

It appears that kubelet's view of the universe has diverged significantly from Docker's hence it does not have the metadata to tag container metrics.  Lots of these in kubelet logs:

```
Apr 13 10:32:38 ip-172-20-2-209 kubelet[885]: E0413 10:32:38.395138     885 manager.go:1103] Failed to create existing container: /kubepods/besteffort/pod52908ce2-3c8a-11e8-9d5c-0a41257e78e8/f6276ace92468f578cddd61be4babd0ea3e03c3ad79735137a54ea4e72fdee08: failed to identify the read-write layer ID for container "f6276ace92468f578cddd61be4babd0ea3e03c3ad79735137a54ea4e72fdee08". - open /var/lib/docker/image/overlay2/layerdb/mounts/f6276ace92468f578cddd61be4babd0ea3e03c3ad79735137a54ea4e72fdee08/mount-id: no such file or directory
[...]

PR #2359, opened 2019-12-18
Retry to get docker root and log when error

chenkaiyue:
Cadvisor in kubelet can't connect to docker daemon when docker and kubelet start together and docker daemon hasn't fully initialized, cadvisor can't get the docker root dir from docker daemon, and will get this value from kubelet's --docker-root flag,  its default value is /var/lib/docker, but when the user's docker dir is not a standard dir,  this will cause the lack of monitoring. 

such as the case metioned in  https://github.com/google/cadvisor/issues/1932#issuecomment-383044255, it is a case of node reboot. We also find when start a node, this problem will occasionally happen again, and will have lots of errors such as `Oct 27 23:35:32 tgamesdk-prod-caster-sh-app-14 kubelet[19472]: E1027 23:35:32.931339   19472 manager.go:1130] Failed to create existing container: /kubepods/burstable/pod52a1ee02-f091-11e9-a887-ca99c560e26a/9a20cfa298899496b45ff29fe1562e81986d4ce8dab9267d8803ed01c177aef1: failed to identify the read-write layer ID for container "9a20cfa298899496b45ff29fe1562e81986d4ce8dab9267d8803ed01c177aef1". - open /var/lib/docker/image/aufs/layerdb/mounts/9a20cfa298899496b45ff29fe1562e81986d4ce8dab9267d8803ed01c177aef1/mount-id: no such file or directory`, it  is same with https://github.com/google/cadvisor/issues/1932.

Docker and kubelet always start together when a node start, we can't ensure that docker daemon has fully initialized when kubelet cadvisor connect docker daemon for getting docker root, so we add some retries here and log when error.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/blang/semver"
 	dockertypes "github.com/docker/docker/api/types"
@@ -49,6 +50,12 @@ var ArgDockerCA = flag.String("docker-tls-ca", "ca.pem", "path to trusted CA")
 // The namespace under which Docker aliases are unique.
 const DockerNamespace = "docker"
 
+// The retry times for getting docker root dir
+const rootDirRetries = 5
+
+//The retry period for getting docker root dir, Millisecond
+const rootDirRetryPeriod time.Duration = 1000 * time.Millisecond
+
 // Regexp that identifies docker cgroups, containers started with
 // --cgroup-parent have another prefix than 'docker'
 var dockerCgroupRegexp = regexp.MustCompile(`([a-z0-9]{64})`)
@@ -72,10 +79,16 @@ var (
 
 func RootDir() string {
 	dockerRootDirOnce.Do(func() {
-		status, err := Status()
-		if err == nil && status.RootDir != "" {
-			dockerRootDir = status.RootDir
-		} else {
+		for i := 0; i < rootDirRetries; i++ {
+			status, err := Status()
+			if err == nil && status.RootDir != "" {
+				dockerRootDir = status.RootDir
+				break
+			} else {
+				time.Sleep(rootDirRetryPeriod)
+			}
+		}
+		if dockerRootDir == "" {
 			dockerRootDir = *dockerRootDirFlag
 		}
 	})


chenkaiyue:
@googlebot I signed it!

container/docker/factory.go
```diff
@@ -72,11 +79,21 @@ var (
 
 func RootDir() string {
 	dockerRootDirOnce.Do(func() {
-		status, err := Status()
-		if err == nil && status.RootDir != "" {
-			dockerRootDir = status.RootDir
-		} else {
-			dockerRootDir = *dockerRootDirFlag
+		var remaining, initTime, succeed = GetRootDirMaxTime, GetRootDirInitTime, false
```
dashpole: This is pretty hard to read... Can we just do:

```
for i  := 0; i < rootDirRetries; i++ {
    status, err := Status()
    if err == nil && status.RootDir != "" {
        dockerRootDir = status.RootDir
        break
    } else {
        time.Sleep(rootDirRetryPeriod)
    }
}
if dockerRootDir == "" {
    dockerRootDir = *dockerRootDirFlag
}

Comment (chenkaiyue) [reply container/docker/factory.go]: Thanks for your review!
My solution is just sleep for different milliseconds each time if get root failed, and each time is twice as much as last time. But it is really hard to read than this...
Shall we wait for some more reviewers to comment, and i will correct this if most of us agree with this easier way.
Thanks again for your review!

**Comment by chenkaiyue** [comment]
@dashpole gives a more easier way, while mine is binary exponential backoff. My original intention is to relieve the call pressure of docker daemon. which way should we choose, thanks for advices~

File: container/docker/factory.go (hunk 1 above)
Comment (chenkaiyue) [reply container/docker/factory.go]: I think for a while, it is really not necessary to make this retry complicated, i will correct it, thanks for your advice~

Comment (chenkaiyue) [reply container/docker/factory.go]: I have update the code, thanks your advice ~

Comment (dashpole) [reply container/docker/factory.go]: yeah, sadly I'm the only reviewer active for this repository.

**Comment by dashpole** [comment]
/ok-to-test

you will have to remove the extra imported files

**Comment by dashpole** [comment]
I1219 17:29:34.896] >> checking go formatting
I1219 17:29:35.537] The following files are not properly formatted:
I1219 17:29:35.537] container/docker/factory.go container/docker/factory.go

**Comment by chenkaiyue** [comment]
> /ok-to-test
> 
> you will have to remove the extra imported files

@dashpole Thx! The test has passed~

File: container/docker/factory.go [hunk 2]
```diff
@@ -49,6 +50,12 @@ var ArgDockerCA = flag.String("docker-tls-ca", "ca.pem", "path to trusted CA")
 // The namespace under which Docker aliases are unique.
 const DockerNamespace = "docker"
 
+// The retry times for getting docker root dir
+const rootDirRetries = 5
+
+//The retry period for getting docker root dir, Millisecond
+const rootDirRetryPeriod = 1000
```
dashpole: Please make the constant be a time.Duration, rather than an integer.  That way we don't have to worry about units.  I.E. `rootDirRetryPeriod = time.Second`

chenkaiyue: Ok~I have corrected it.

PR #2376, opened 2020-01-13
Ignore CPU clock for mips64

liupeng0518:
Add mips64 to clock speed ignore list.


liupeng0518:
@googlebot I signed it!

PR #2382, opened 2020-01-20
[WIP] Switch from docker/docker to k8s.io/utils for mount

dims:
We are using `docker/docker` for parsing mount information. we already have support for this in k8s.io/utils. Let us please switch to that.


dims:
cc @liggitt

PR #2383, opened 2020-01-20
Switch from docker/docker to k8s.io/utils for mount

dims:
We are using docker/docker for parsing mount information. we already have support for this in k8s.io/utils. Let us please switch to that.


dims:
the major/minor stuff is bothering me, fixing it in k8s.io/utils ( https://github.com/kubernetes/utils/pull/133 )

dashpole:
ping this again once the k/utils update is done, and i'll merge it.

dims:
@dashpole this is ready now.

PR #2391, opened 2020-02-10
Skip TestTopology on non-amd64

zhsj:
It fails on s390x and aarch64, since GetTopology
function is platform specified.


zhsj:
Find another failed test on s390x, so I add another commit to skip TestTopologyWithSimpleCpuinfo

PR #2401, opened 2020-02-24
Check extraErr before accessing extraUsage.Bytes

tedyu:
In realFsHandler#update, extraUsage.Bytes is accessed without checking extraErr.

This PR adds check before referencing extraUsage.Bytes

### DIFF
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -90,7 +90,7 @@ func (fh *realFsHandler) update() error {
 	fh.Lock()
 	defer fh.Unlock()
 	fh.lastUpdate = time.Now()
-	if fh.rootfs != "" && rootErr == nil {
+	if fh.rootfs != "" && rootErr == nil && extraErr == nil {
 		fh.usage.InodeUsage = rootUsage.Inodes
 		fh.usage.TotalUsageBytes = rootUsage.Bytes + extraUsage.Bytes
 	}
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -90,12 +90,13 @@ func (fh *realFsHandler) update() error {
 	fh.Lock()
 	defer fh.Unlock()
 	fh.lastUpdate = time.Now()
-	if fh.rootfs != "" && rootErr == nil && extraErr == nil {
+	if fh.rootfs != "" && rootErr == nil {
 		fh.usage.InodeUsage = rootUsage.Inodes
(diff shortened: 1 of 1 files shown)


tedyu:
@dashpole
Tests passed.

container/common/fsHandler.go
```diff
@@ -90,7 +90,7 @@ func (fh *realFsHandler) update() error {
 	fh.Lock()
 	defer fh.Unlock()
 	fh.lastUpdate = time.Now()
-	if fh.rootfs != "" && rootErr == nil {
+	if fh.rootfs != "" && rootErr == nil && extraErr == nil {
```
dashpole: I believe it is valid for fh.extraDir to be unset, which would cause GetDirUsage above to return an uninitialized UsageInfo struct.  Thankfully, the zero-value of the Bytes field is 0, so the current setup is technically correct.

I agree that it would be more readable to check the error, but we don't want to not set TotalUsageBytes if we get one.

Can we instead do a separate check on extraErr == nil, and then do `fh.usage.TotalUsageBytes += extraUsage.Bytes`?

tedyu: @dashpole 
Please take another look.

tedyu:
@dashpole 
Please take another look.

container/common/fsHandler.go
dashpole: Just remove the check for extraErr == nil here, and we are all good to go.

tedyu: Addressed, @dashpole

container/common/fsHandler.go
```diff
@@ -92,10 +92,11 @@ func (fh *realFsHandler) update() error {
 	fh.lastUpdate = time.Now()
 	if fh.rootfs != "" && rootErr == nil {
 		fh.usage.InodeUsage = rootUsage.Inodes
-		fh.usage.TotalUsageBytes = rootUsage.Bytes + extraUsage.Bytes
+		fh.usage.TotalUsageBytes = rootUsage.Bytes
 	}
 	if fh.extraDir != "" && extraErr == nil {
 		fh.usage.BaseUsageBytes = rootUsage.Bytes
```
dashpole: It looks like this line (fh.usage.BaseUsageBytes = rootUsage.Bytes) should be moved up to the previous if statement.  Not sure how we messed that up :).  Feel free to do that here, or i'll merge this the way it is.

tedyu: @dashpole 
Addressed.

tedyu:
@dashpole 
All tests passed.

PR #2402, opened 2020-02-24
Adjust period after each fh.update() call

tedyu:
In realFsHandler#trackUsage, the error return from the first call to fh.update() is ignored.

This PR extracts the period adjustment logic into a func and calls this func in place of original fh.update()

### DIFF
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -106,31 +106,30 @@ func (fh *realFsHandler) update() error {
 }
 
 func (fh *realFsHandler) trackUsage() {
-	fh.update()
 	longOp := time.Second
 	for {
+		start := time.Now()
+		if err := fh.update(); err != nil {
+			klog.Errorf("failed to collect filesystem stats - %v", err)
+			fh.period = fh.period * 2
+			if fh.period > maxBackoffFactor*fh.minPeriod {
+				fh.period = maxBackoffFactor * fh.minPeriod
+			}
+		} else {
+			fh.period = fh.minPeriod
+		}
+		duration := time.Since(start)
+		if duration > longOp {
+			// adapt longOp time so that message doesn't continue to print
+			// if the long duration is persistent either because of slow
+			// disk or lots of containers.
+			longOp = longOp + time.Second
+			klog.V(2).Infof("fs: disk usage and inodes count on following dirs took %v: %v; will not log again for this container unless duration exceeds %v", duration, []string{fh.rootfs, fh.extraDir}, longOp)
+		}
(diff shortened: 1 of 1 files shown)


tedyu:
@dashpole 
Tests passed.

container/common/fsHandler.go
```diff
@@ -106,23 +106,26 @@ func (fh *realFsHandler) update() error {
...
-			if err := fh.update(); err != nil {
-				klog.Errorf("failed to collect filesystem stats - %v", err)
-				fh.period = fh.period * 2
-				if fh.period > maxBackoffFactor*fh.minPeriod {
-					fh.period = maxBackoffFactor * fh.minPeriod
-				}
-			} else {
-				fh.period = fh.minPeriod
-			}
+			periodAdjuster()
```
dashpole: Can we instead just move this logic to the top of the loop?  I.E.
```golang
for {
    ... all of the logic ...
    select {
    case <-fh.stopChan:
        return
    case <-time.After(fh.period):
    }
}
```

Also, lets include the logging for long operations in there as well.

tedyu: @dashpole 
Done.

container/common/fsHandler.go
```diff
@@ -106,23 +106,22 @@ func (fh *realFsHandler) update() error {
...
-			if err := fh.update(); err != nil {
-				klog.Errorf("failed to collect filesystem stats - %v", err)
-				fh.period = fh.period * 2
-				if fh.period > maxBackoffFactor*fh.minPeriod {
-					fh.period = maxBackoffFactor * fh.minPeriod
-				}
-			} else {
-				fh.period = fh.minPeriod
-			}
 			duration := time.Since(start)
```
dashpole: please move the start, and duration logging here up as well

container/common/fsHandler.go
```diff
@@ -106,31 +106,30 @@ func (fh *realFsHandler) update() error {
...
+		if err := fh.update(); err != nil {
+			klog.Errorf("failed to collect filesystem stats - %v", err)
+			fh.period = fh.period * 2
+			if fh.period > maxBackoffFactor*fh.minPeriod {
+				fh.period = maxBackoffFactor * fh.minPeriod
+			}
+		} else {
+			fh.period = fh.minPeriod
+		}
+		start := time.Now()
```
dashpole: start needs to be defined before the update.  The current change makes duration ~= 0 always.

tedyu: @dashpole 
Corrected.

PR #2403, opened 2020-02-24
Verify that fsType is found for device

tedyu:
In crioContainerHandler#getFsStats, we come out of the loop over mi.Filesystems when device is found.
However, in case the device is not found, we should return info.FsStats with limit of 0.

This PR adds check against the device not found case.

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -272,6 +272,9 @@ func (self *crioContainerHandler) getFsStats(stats *info.ContainerStats) error {
 		}
 	}
 
+	if fsType == "" {
+		return fmt.Errorf("unable to determine fs type for device: %v", device)
+	}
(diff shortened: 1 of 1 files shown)


tedyu:
@dashpole 
Can you take a look ?

container/crio/handler.go
```diff
@@ -262,16 +262,21 @@ func (self *crioContainerHandler) getFsStats(stats *info.ContainerStats) error {
...
 	for _, fs := range mi.Filesystems {
 		if fs.Device == device {
 			limit = fs.Capacity
 			fsType = fs.Type
+			deviceFound = true
 			break
 		}
 	}
 
+	if !deviceFound {
```
dashpole: I would prefer checking if `fsType == ""`, rather than using a bool in the loop

tedyu: @dashpole 
Please take another look.

PR #2404, opened 2020-02-25
Chk scanner err

tedyu:
In scanAdvancedTcpStats, after scanning is done, we call json.Marshal followed by returning scanner.Err() .
If scanning has had error, it should be returned prior to calling json.Marshal .

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -477,6 +477,9 @@ func scanAdvancedTcpStats(advancedStats *info.TcpAdvancedStat, advancedTcpStatsF
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


tedyu:
ping @dashpole

container/libcontainer/handler.go
```diff
@@ -477,6 +477,9 @@ func scanAdvancedTcpStats(advancedStats *info.TcpAdvancedStat, advancedTcpStatsF
 			}
 		}
 	}
+	if scanner.Err() != nil {
```
dashpole: I believe we do this to return best-effort results if the error occurs mid-way through scanning.

PR #2410, opened 2020-03-02
support export envs with prefix

KielChan:
This is a feature that cadvisor can export envs with kinds of prefix pattern.

 It can be set by flag `--docker-env-metadata-whitelist` in `kubelet`, and with this feature, users can access pods' envs exposed to collect more metrics.

For example, if user want to expose some metrics collected by `prometheus` sdk, or metrics about runtime like java `jmx` and so on, they can set `--docker-env-metadata-whitelist` and acquire metrics detail by kubelet API with `Subcontainers` option enabled.

Eventually, I wanna to set envs exposed to be `Upper` cased(Lower cased now).

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -51,7 +51,6 @@ const (
 )
 
 type dockerContainerHandler struct {
-
 	// machineInfoFactory provides info.MachineInfo
 	machineInfoFactory info.MachineInfoFactory
 
@@ -253,11 +252,16 @@ func newDockerContainerHandler(
 
 	// split env vars to get metadata map.
 	for _, exposedEnv := range metadataEnvs {
+		if exposedEnv == "" {
+			// no envs provided, len(metadataEnvs) == 1, metadataEnvs[0] == ""
+			continue
+		}
+
 		for _, envVar := range ctnr.Config.Env {
 			if envVar != "" {
 				splits := strings.SplitN(envVar, "=", 2)
-				if len(splits) == 2 && splits[0] == exposedEnv {
-					handler.envs[strings.ToLower(exposedEnv)] = splits[1]
+				if len(splits) == 2 && strings.HasPrefix(splits[0], exposedEnv) {
+					handler.envs[strings.ToLower(splits[0])] = splits[1]
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


KielChan:
Environments can also be used for TSDB tags. If exposed with a pattern, we can do more with it.

container/docker/handler.go
```diff
@@ -253,11 +252,16 @@ func newDockerContainerHandler(
 
 	// split env vars to get metadata map.
 	for _, exposedEnv := range metadataEnvs {
+		if exposedEnv == "" {
```
dashpole: I don't understand this part.  Is this fixing a bug?

dashpole:
I don't seem to see the `docker-env-metadata-whitelist` in the kubelet.  If it isn't already exposed, we probably shouldn't add it.

KielChan:
exec cmd like `kubelet --help`(master branch),
```
      --docker-env-metadata-whitelist string                                                                      a comma-separated list of environment variable keys that needs to be collected for docker containers (DEPRECATED: This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.)
```

Fine, it is `DEPRECATED` in kubelet flags.

But, if exposed metrics with cadivsor daemonset, we can also add this feature to exposed some tags matched some kind of prefix pattern. And dump this tags to TSDB like `victoria metrics` or `influxdb`.

container/docker/handler.go
KielChan:
```golang
func main() {
	a := ""
	b := "test"
	if strings.Contains(b, a) {
		fmt.Println("yes")
	}
}
```

if like before, it's fine for totally match. But if matched with `strings.Contains`, `""` may be wrong

KielChan:
And we can do more with these environments exposed  like service discovery.

KielChan:
ref: cmd/kubelet/app/options/globalflags_linux.go

dashpole:
Ok, this seems reasonable.  Can you update the flag help text to describe the new behavior?

KielChan:
> Ok, this seems reasonable. Can you update the flag help text to describe the new behavior?

Fine

KielChan:
CLA checks hang maybe caused by introducing in `Sign Off` in `git commit -s`, I will remove it soon and create another PR to fix it

KielChan:
moved to: https://github.com/google/cadvisor/pull/2430

Issue #2388, opened 2020-02-05
Hardware PMU - performance counters using Linux perf and /resctrl interface


ss7pro:
Hey,

In Intel we're working on a solution that is providing online workload characterization using hardware PMU using Linux perf interface. In addition we're also using RDT through /resctrl interface to extract memory bandwidth and cache usage. Full list of metrics that we collect is documented here:

https://github.com/intel/workload-collocation-agent/blob/master/docs/metrics.rst

We would like to migrate all telemetry collection and query to cAdvisor. Do you think it make sense to add those kind of metrics to cAdvisor and if yes do you anticipate any major issues in that approach?

dashpole:
cAdvisor exposes metrics associated with cgroups.  From looking at you metric list, it looks like most metrics are associated with tasks, although looking through some of the other docs, it looks like "task" can at least sometimes mean cgroup.  Are all "task" metrics associated with cgroups?

iwankgb:
@dashpole some of the task metrics come from resctrl filesystem (they are related to Intel RDT - cache and memory bandwidth monitoring). Task concept is similar to pod in Kubernetes: it always is a cgroup, it might include other cgroups and it might rely on other kernel interfaces to collect metrics too.

dashpole:
Given a particular cgroup, is it possible to get the associated task?

Are the kernel interfaces it relies on documented anywhere?

iwankgb:
Resctrl is [documented in kernel](https://www.kernel.org/doc/html/latest/x86/resctrl_ui.html). 

Regarding identifying a task by cgroup - I'm not sure if I understand your question; are you asking about implementation of a task in Workload Collocation Agent?

iwankgb:
BTW, as far as I can tell RDT and resctrl are already being handled using runc's `IntelRdtManager`, so we will be able to utilize [existing functionality](https://github.com/google/cadvisor/blob/master/vendor/github.com/opencontainers/runc/libcontainer/container_linux.go#L183
), I hope.

dashpole:
Oh, great.  I see it is already integrated into libcontainer.  In that case, it will be easy to integrate.

iwankgb:
cAdvisor is using libcontainer to collect all the metrics from cgroups - am I right? If this is the case then we will proceed with some PRs in libcontainer and then we will get back to you once we are ready to integrate them to cAdvisor.

dashpole:
Yes.  cAdvisor uses libcontainer to collect metrics from cgroups.

[...]

PR #2415, opened 2020-03-03
Nvm and dimm information

iwankgb:
Goal of this PR is to expose various types of memory and their size at `/api/v2.0/machine` endpoint. Non-volatile RAM capacity and performance differs from standard DDR and we believe that it's worth exposing it to cAdvisors users. No breaking changes are introduced.

Our goal is to use this information to improve workload scheduling in Kubernetes clusters to optimize  non-volatile memory usage and get rid of metrics collecting code from [intel/workload-collocation-agent](https://github.com/intel/workload-collocation-agent), and replace it with cAdvisor.

Summary of changes:
* Extending `MachineInfo` struct.
* Traversing memory controller part of [EDAC sysfs subsystem](https://www.kernel.org/doc/html/v4.13/driver-api/edac.html) to gather information about type and capacity of installed DIMMs.
* Fake filesystem and tests for new functionality.

`/api/v2.0/machine` endpoint will be extended with following JSON snippet:
```json
"memory_dimm_count":20,
"memory_by_type":{
	"Non-volatile-RAM":{
		"Capacity":2168421613568,
		"DimmCount":8
	},
	"Unbuffered-DDR4":{
		"Capacity":412316860416,
		"DimmCount":12}
	},
}
```
EDAC is available since kernel 3.6.

This PR is related to #2388.


info/v1/machine.go
```diff
@@ -167,6 +167,13 @@ type MachineInfo struct {
 	// The amount of memory (in bytes) in this machine
 	MemoryCapacity uint64 `json:"memory_capacity"`
 
+	// Number of memory DIMMs in this machine.
+	// 0 if number is uknown (e.g. in virtualized envirtonment).
+	MemoryDimmCount uint `json:"memory_dimm_count"`
```
dashpole: Can this be inferred from the DimmCount in MemoryByType below?

machine/machine.go
```diff
@@ -101,6 +105,67 @@ func GetMachineMemoryCapacity() (uint64, error) {
...
+	names, err := ioutil.ReadDir(edacPath)
+	// On some architectures (such as ARM) memory controller device may not exist.
+	// If this is the case then we ignore error and return empty slice.
+	_, ok := err.(*os.PathError)
+	if err != nil && ok {
+		return memory, dimmCount, nil
+	} else if err != nil {
+		return memory, dimmCount, err
+	}
+	isController := regexp.MustCompile("mc[0-9]+")
```
dashpole: nit: you can declare the compiled regexp outside of the function so we don't recompile each time?

machine/machine.go
```diff
@@ -101,6 +105,67 @@ func GetMachineMemoryCapacity() (uint64, error) {
...
+	isDimm := regexp.MustCompile("dimm[0-9]+")
+	for _, controllerDir := range names {
+		controller := controllerDir.Name()
+		if isController.MatchString(controller) {
```
dashpole: nit: this is getting pretty deeply nested, which makes it hard to read.  One way we can reduce that is by changing
```golang
if isController.MatchString(controller) {
    ...
}
```
to 
```golang
if !isController.MatchString(controller) {
    continue
}
...
```
This applies to both regexp.

info/v1/machine.go
iwankgb: Sure, it can. If you want I can remove it but in my opinion cost of calculating it and sending it too a client is pretty low while we still expose general information about all the DIMMs. It's your call :)

iwankgb:
@dashpole - updated.

info/v1/machine.go
dashpole: I would prefer to keep the API as minimal as possible.

iwankgb:
@dashpole - updated, no overall DIMM count :)

PR #2416, opened 2020-03-04
fix GetDirUsage error ignored

gongguan:
Don't ignore error of filepath.Walk


gongguan:
@googlebot I signed it!

PR #2418, opened 2020-03-05
Handle the new 5.0 linux format for oom messages

### LINKED ISSUE #2417: OOMWatcher does not detect victim container correctly on some OSs
Linux 5.0
Ubuntu 18.04

The [OOM Parser code](https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser.go#L29) looks for `Task in (.*) killed as a result of limit of (.*)`, which doesn't seem to match what this kernel is putting out...

I am seeing:

```
[180042.118480] oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=ef807430361e6e82b45db92e2e9b6fbec98f419b12c591e655c1a725565e73a8,mems_allowed=0,oom_memcg=/kubepods/burstable/podfbdfe8e3-1c87-4ff2-907
c-b2ec8e25d012,task_memcg=/kubepods/burstable/podfbdfe8e3-1c87-4ff2-907c-b2ec8e25d012/ef807430361e6e82b45db92e2e9b6fbec98f419b12c591e655c1a725565e73a8,task=manager,pid=966,uid=0
```
instead.

This causes all OOMs to be reported as system OOMs, since we default the ContainerName and VictimContainerName to "/".

dashpole:
Fixes #2417, which caused all OOMs on >5.0 kernels to be considered system ooms.

This adds parsing for the new format.  When this format is found, we can "finish" immediately, since the message has all of the information we want in a single place.  If the format is not found, we fall back to looking for the old message.  It also grabs the Constraint that is hit, which may be useful for users that have topology constraints.

### DIFF
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -26,7 +26,9 @@ import (
 )
 
 var (
-	containerRegexp = regexp.MustCompile(`Task in (.*) killed as a result of limit of (.*)`)
+	legacyContainerRegexp = regexp.MustCompile(`Task in (.*) killed as a result of limit of (.*)`)
+	// Starting in 5.0 linux kernels, the OOM message changed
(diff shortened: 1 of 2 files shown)


dims:
@dashpole i've reviewed this. LGTM

dashpole:
Merging based on lgtm above.

PR #2419, opened 2020-03-06
Measuring perf events - chapter I

iwankgb:
Purpose of this PR is to introduce support for measuring perf events to cAdvisor. It will eventually consist of the following:

- [x] Configuration primitives allowing to describe what events are to be measured (user-friendly, using `libpfm` to translate event names to `perf_event_attr`
- [x] Perf events collector (ungrouped core events support)
- [x] Perf events manager responsible to container lifecycle control (ungrouped core events support)

This is part of effort discussed in #2388


iwankgb:
@dashpole, I would appreciate if you looked at this very PR. At the moment it consists of documentation and proposed configuration API. Before proceeding with further changes I want to know what you think about it and make sure that we can agree on a user-friendly implementation.

Originally, we were planning to use `runc` to handle perf subsystem but to do so we would have to update all the projects that interact with `runc` including projects that interact indirectly. For instance, if we updated `runc` configuration and OCI runtime spec then we would have to update `containerd` too and Kubernetes' CRI so that it is possible to pass perf configuration from `PodSpec` through CRI to containerd that would pass it to `runc` using OCI runtime spec. 

We believe that configuring perf subsystem in cAdvisor still makes sense as this subsystem is not responsible for any resource allocation but purely for monitoring. We also assume that same events will be measured for all the containers.

dashpole:
Whoops, missed your comment earlier... I had been ignoring it because of the WIP tag.  I'll take a look today

dashpole:
> Originally, we were planning to use runc to handle perf subsystem but to do so we would have to update all the projects that interact with runc including projects that interact indirectly. For instance, if we updated runc configuration and OCI runtime spec then we would have to update containerd too and Kubernetes' CRI so that it is possible to pass perf configuration from PodSpec through CRI to containerd that would pass it to runc using OCI runtime spec.

Yeah, I doubt it would be accepted in the pod spec.  Do you know if perf events could be enabled using OCI hooks?  Those seem to be the standard place to do container configuration.

> We believe that configuring perf subsystem in cAdvisor still makes sense as this subsystem is not responsible for any resource allocation but purely for monitoring. We also assume that same events will be measured for all the containers.

I definitely agree that cAdvisor collecting these events would be useful.  I have seen some pretty neat sidecar-container perf tools, so I know these metrics can be useful.  I was initially surprised that we would want to measure the same events across all containers, as I assumed it would hurt application performance.  But from some reading, it doesn't seem to be _that_ bad...

We haven't done a great job in the past with having this sort of configuration in cAdvisor.  If there were a way to just have cAdvisor collect the metrics, and have something else do the configuration, that would fit more cleanly within our scope.  But absent that, I think this is a reasonable feature to add.

iwankgb:
Sorry for the delayed reply, I had to take few days off.
@dashpole, I don't think it is feasible (while it's perfectly possible) to handle perf events configuration in an OCI hook. In order to do so one would have to call `perf_event_open` syscall in the hook and then do one of the following:
* pass returned file descriptor to cAdvisor - doable but imperfect (only containers launched after launching cAdvisor could be monitored or we would have to create a long-running hook process)
* you OCI hook to launch a sidecar container responsible for perf events handling - this is scenario that we want to deprecate by adding perf support to cAdvisor.
I can't think about any other ways of handling configuration step outside of cAdvisor.

What do you think about providing two configuration schemas:
* *advanced* - as already described in the PR
* *simple* - limiting number of events to number of available counters, using only named events, no grouping support?

docs/runtime_options.md
```diff
@@ -134,7 +134,85 @@ cAdvisor stores the latest historical data in memory. How long of a history it s
...
+        ]
+      }
+    ]
+  ],
+  "interval": "5s"
+}
+```
+
+In the example above number:
+* `INST_RETIRED.ANY_P` (number of instructions retired on 2nd Generation Intel® Xeon® Scalable Processor) identified by
```
dashpole: are events specific to processors?  Are there any common ones that nearly everyone would be able to use?

perf/config.go
```diff
@@ -0,0 +1,93 @@
...
+
+type Events struct {
+	// List of perf events to be measured. Events in this list
+	// will not be grouped. See group_fd argument documentation
+	// at man perf_event_open.
+	NonGrouped []Event `json:"non_grouped"`
+
+	// List of groups of events to be measured. See group_fd
+	// argument documentation at man perf_event_open.
+	Grouped [][]Event `json:"grouped"`
```
dashpole: Can you help me understand of the value of having groups:

1. Each event in the group still gets its own metric stream, right?
2. Grouping event A and B just impacts scheduling to _somehow_ make metrics better.  From the man page: 

`An event group is scheduled onto the CPU as a unit: it will be put onto the CPU only if all of the events in the group can be put onto the CPU.  This means that the values of the member events can be meaningfully compared—added, divided (to get ratios), and so on—with each other, since they have counted events for the same set of executed instructions." `

I don't quite understand how Events can be scheduled by the kernel.  I thought these "Events" were things we were looking for the occurrence of, not things we were running (i.e. scheduling)

perf/config.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Interval between each measurement.
+	Interval Duration `json:"interval"`
```
dashpole: Unless we have a good reason to, I would prefer using the housekeeping interval with jitter like we do for other metrics.  Otherwise, starting many containers at once (a common pattern) leads to large spikes in usage from cAdvisor collecting all container metrics at once.

docs/runtime_options.md
```diff
@@ -134,7 +134,85 @@ cAdvisor stores the latest historical data in memory. How long of a history it s
...
+`config[0xc0]` will be measured as non-grouped event
+* `UNC_M_CAS_COUNT.WR` (all DRAM Write CAS commands issued or `cas_count_write`) identified by `config[0xC04]` and 
+`UNC_M_CAS_COUNT.RD` (all DRAM Read CAS commands issued or `cas_count_read`) identified by `config[0xc304]` will be 
+measured as grouped events. In their case `type` attribute indicates memory controller.
+
+##### Translating event name to config values
+
+In order to determine what values must be used to configure particular event you should follow a procedure:
+1. Identify the event in `perf list` output.
```
dashpole: Is this something cAdvisor could do for users?   Should our API be higher-level, or is that not possible?

dashpole:
As long as we have good documentation and examples, I prefer having a single configuration option.

I thought about it after suggesting it, and the file-descriptor API makes it pretty much impossible to implement cleanly.

In general, this is probably the right direction to go.

container/libcontainer/helpers.go
```diff
@@ -104,14 +104,15 @@ func getCgroupSubsystemsHelper(allCgroups []cgroups.Mount, disableCgroups map[st
...
-	"devices": {},
+	"cpu":        {},
+	"cpuacct":    {},
+	"memory":     {},
+	"pids":       {},
+	"cpuset":     {},
+	"blkio":      {},
+	"io":         {},
+	"devices":    {},
+	"perf_event": {},
```
dashpole: Does this exist and work similarly in cgroups v2?

docs/runtime_options.md
iwankgb: Yes, events are different even between various generation of Xeons. If you take AMD's Naples and Rome CPUs you will see similar differences not mentioning other architectures such as ARM.

perf/config.go
iwankgb: Event grouping is extremely useful when you want to calculate a derivative metric - it is guaranteed that events that you collect were calculated in exactly same period of time.
Kernel is responsible for "scheduling" - it instructs hardware performance counters (that are just pieces of silicon) to react to occurrence of specific perf events.

perf/config.go
iwankgb: :+1:

container/libcontainer/helpers.go
iwankgb: It does exists and I believe it is similar but I have not investigated it yet.

docs/runtime_options.md
iwankgb: I will look into it. There are some libraries (in and out of kernel) that can help.

docs/runtime_options.md
iwankgb: I missed one of your questions: yes, some of the events (e.g. instructions retired, cycles) are supported across all or majority of Linux architectures.

dashpole: Great!  Once we have this in, it would be useful to have an example configmap in deploy/kubernetes as a follow-up, so it is easy for users to get started with the feature.

perf/config.go
dashpole: I'll try and rephrase to see if I get it now:  The kernel is essentially making the process of _updating_ the group of performance counters atomic so you can't ever measure a time where one counter is updated, but others are not?

Does it also change the metric you get back?  I.E. is there a reason why we don't just have one large group?

iwankgb: Yes, I think I that atomicity is very good name for it :) 
When you want to assure atomicity then you can't measure more events that you have hardware perf counters available (8 per physical core/4 per HT on current generation Xeons) so using single big group is simply impossible. If you are antes to measure more events that you have counters available then your events (or group of them) will be multiplexed and results will have to be scaled.

dashpole: Really appreciate the explanation.  It all makes sense now.

If I were to create a group of 100 perf counters, how would cAdvisor fail?  Is there a way we can fail on startup if it won't work?

iwankgb: In such case syscall (perf_event_open) will fail. I suppose that EINVAL will be returned so the error itself will not be very descriptive. 
It is possible to determine properties of hardware counters from cpuinfo, though and such information could be used to validate configuration before calling perf_event_open.

dashpole: Or even just calling perf_event_open with the config on the root cgroup during startup would be good.  That way we can crash cAdvisor if it isn't valid

iwankgb:
A quick note after chat with @dashpole on Slack:
* It's fine to use `libpfm` to make perf event configuration more user friendly
* It must be possible to build cAdvisor without `cgo` and/or `libpfm`
* There must be clear message (perhaps a failure to start cAdvisor) if it's built without `cgo` and/or `libpfm`support and user tries to launch perf event monitoring.

iwankgb:
@dashpole - I would appreciate if you took a look at the code. At the moment it works for `libcontainer` derivatives and metrics are not exposed to any data storage. I wanted to implement non-grouped events collecting for one of the supported environments and proceed with further changes later. Let me know if it's fine with you.
One part that this PR is missing are integration tests. I am just looking at how they can be implemented and will proceed with them before doing anything else (e.g. extending Prometheus exporter output).

perf/collector_libpfm.go
```diff
@@ -0,0 +1,257 @@
...
+	if !ok {
+		c.cpuFiles[name] = map[int]readerCloser{}
+	}
+	c.cpuFiles[name][cpu] = perfFile
+}
+
+func (c *collector) setupNonGrouped() error {
+	if !isLibpfmInitialized {
+		klog.Warning("libpfm4 is not initialized, cannot proceed with setting perf events up")
+		return errors.New("libpfm4 is not initialized, cannot proceed with setting perf events up")
```
dashpole, line 161: any reason to use errors.New here?  We generally just use fmt.Errorf()

perf/manager_libpfm.go
```diff
@@ -0,0 +1,57 @@
...
+
+type manager struct {
+	events   Events
+	numCores int
+	stats.NoopSetupDestroy
+}
+
+func NewManager(configFile string, numCores int) (stats.Manager, error) {
+	if configFile == "" {
+		return nil, nil
```
dashpole, line 38: Ah, I see.  I prefer adding a noop manager when the config file isn't specified.

perf/config.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"k8s.io/klog"
+	"strconv"
+)
+
+type Events struct {
+	// List of perf events' names to be measured. Events in this list
+	// will not be grouped. See group_fd argument in documentation
+	// at man perf_event_open
+	NonGrouped []string `json:"non_grouped"`
```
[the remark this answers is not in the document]
iwankgb: Sure, we can make such an assumption.

info/v1/container.go
```diff
@@ -816,6 +816,26 @@ type AcceleratorStats struct {
...
+	Value uint64
+
+	// Name is human readable name of an event.
+	Name string
+
+	// Time when event was collected.
+	Time time.Time
+
+	// CPU that perf event was measured on.
+	Cpu int
```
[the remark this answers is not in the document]
iwankgb: Unfortunately when we measure perf events per cgroup we must create separate counter for each CPU. I will look into `--disable-metrics` though and make sure that no events are measured for disabled CPUs.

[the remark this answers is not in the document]
iwankgb: So: if `--disable_metrics` argument includes `percpu` then we should not measure any oncore perf events. In my opinion it should not affect uncore events<sup>[1]</sup> as they are not CPU related. Do you agree?
<sup>[1]</sup> https://en.wikipedia.org/wiki/Uncore, https://software.intel.com/en-us/blogs/2014/07/11/documentation-for-uncore-performance-monitoring-units, https://github.com/andikleen/pmu-tools/blob/master/ucevent/README.md

info/v1/container.go
```diff
@@ -816,6 +816,26 @@ type AcceleratorStats struct {
 	DutyCycle uint64 `json:"duty_cycle"`
 }
 
+// PerfStat represents value of a single monitored perf event.
+type PerfStat struct {
+	// Indicates scaling ratio for an event: time_enabled/time_running.
+	// value 1.0 indicates that no multiplexing occurred.
+	// See: https://lwn.net/Articles/324756/
+	ScalingRatio float64
```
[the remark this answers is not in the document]
iwankgb: I added a bit of explanation and documented, that `Value` is scaled.

perf/collector_libpfm.go
```diff
@@ -0,0 +1,257 @@
...
+
+	"golang.org/x/sys/unix"
+	"k8s.io/klog"
+)
+
+type collector struct {
+	cgroupPath   string
+	events       Events
+	cpuFiles     map[string]map[int]readerCloser
+	cgroup       *os.File
```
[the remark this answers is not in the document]
iwankgb: I am surprised, but:
* such approach works like a charm
* quick analysis of kernel code shows that we should be safe closing this file immediately.

info/v1/container.go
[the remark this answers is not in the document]
iwankgb: After a bit longer consideration I am not that sure that applying `--disable_metrics` to perf events makes sense. Nothing will be accounted for unless configuration file is provided and relying on second flag can make it more complicated.

iwankgb:
@dashpole, I think I did my best to merge you feedback into this PR :) The last remainig question (unless I'm missing something) is [`--disable_metric` question](https://github.com/google/cadvisor/pull/2419#discussion_r404417370). Next round of review will be appreciated.

info/v1/container.go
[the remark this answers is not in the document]
dashpole: Ok, agreed.  I would say if it is "easy" to degrade behavior, then we should.  Otherwise, the current implementation is fine.

perf/config.go
```diff
@@ -0,0 +1,72 @@
...
+
+type Events struct {
+	// List of perf events' names to be measured. Any value found in
+	// output of perf list can be used.
+	Events [][]Event `json:"events"`
+
+	// List of custom perf events' to be measured. It is impossible to
+	// specify some events using their names and in such case you have
+	// to provide lower level configuration.
+	CustomEvents [][]CustomEvent `json:"custom_events"`
```
dashpole, line 33: I was asking a slightly different question.  Can "simple" and "custom" events be grouped together?  My thinking was that we could have part of the config for defining all of the events, and another for specifying their groupings.  E.g.

```json
{
  "events": [
    ["instructions", "instructions_retired"]
  ],
  "custom_events":   [
    {
      "type": 4,
      "config": [
        "0x5300c0"
      ],
      "name": "instructions_retired"
    }
  ]
}
```

iwankgb: Oh, ok. I didn't get this. I bet it is doable :)

dashpole:
I went through and resolved all the comments that have been dealt with.  The open ones either still need to be resolved, or should be resolved in a follow-up

perf/collector_libpfm.go
```diff
@@ -0,0 +1,257 @@
...
+				continue
+			}
+			perfData := &ReadFormat{}
+			reader := bytes.NewReader(buf)
+			err = binary.Read(reader, binary.LittleEndian, perfData)
+			now := time.Now()
+			if err != nil {
+				klog.Warningf("Unable to decode from binary format read from perf_event file (event: %q, CPU: %d) for %q", name, cpu, c.cgroupPath)
+			}
+			klog.Infof("Read metric for event %q for cpu %d from cgroup %q: %d", name, cpu, c.cgroupPath, perfData.Value)
```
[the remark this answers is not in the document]
iwankgb: is `V(4)` fine or you want to have it removed completely?

iwankgb:
@dashpole done.

perf/collector_libpfm.go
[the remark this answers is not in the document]
dashpole: V(4) is fine

perf/collector_libpfm.go
```diff
@@ -0,0 +1,250 @@
...
+func NewCollector(cgroupPath string, events Events, numCores int) stats.Collector {
+	return &collector{cgroupPath: cgroupPath, events: events, cpuFiles: map[string]map[int]readerCloser{}, numCores: numCores}
+}
+
+func (c *collector) UpdateStats(stats *info.ContainerStats) error {
+	c.cpuFilesLock.Lock()
+	defer c.cpuFilesLock.Unlock()
+
+	stats.PerfStats = []info.PerfStat{}
+	klog.V(4).Infof("Attempting to update perf_event stats from cgroup %q", c.cgroupPath)
```
dashpole, line 74: This should probably be V(5)

perf/collector_libpfm.go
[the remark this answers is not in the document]
dashpole: Actually, lets make it V(5), since it is per-container, per housekeeping.

perf/collector_libpfm.go
```diff
@@ -0,0 +1,250 @@
...
+	for _, group := range c.events.Events {
+		if len(group) == 1 {
+			customEvent, ok := c.events.eventToCustomEvent[group[0]]
+			if ok {
+				c.setupRawNonGrouped(customEvent, cgroupFd)
+			} else {
+				c.setupNonGrouped(string(group[0]), cgroupFd)
+			}
+		} else {
+			klog.Info("Grouped events are not supported yet")
```
dashpole, line 125: Can we validate this during startup?  Is this something you are planning to implement before we submit the change?

perf/collector_libpfm.go
```diff
@@ -0,0 +1,250 @@
...
+	_, ok := c.cpuFiles[name]
+	if !ok {
+		c.cpuFiles[name] = map[int]readerCloser{}
+	}
+	c.cpuFiles[name][cpu] = perfFile
+}
+
+func (c *collector) setupNonGrouped(name string, cgroup int) error {
+	if !isLibpfmInitialized {
+		klog.Warning("libpfm4 is not initialized, cannot proceed with setting perf events up")
```
dashpole, line 164: We should probably handle the error in the caller and log there.  That way we can provide additional context about the call.

Same goes for other places where we log and return the same error.

perf/collector_libpfm.go
```diff
@@ -0,0 +1,250 @@
...
+	for cpu = 0; cpu < c.numCores; cpu++ {
+		pid, groupFd, flags := cgroup, -1, unix.PERF_FLAG_FD_CLOEXEC|unix.PERF_FLAG_PID_CGROUP
+		fd, err := unix.PerfEventOpen(config, pid, cpu, groupFd, flags)
+		if err != nil {
+			klog.Errorf("Setting up perf event %#v failed: %q", config, err)
+			return
+		}
+		perfFile := os.NewFile(uintptr(fd), name)
+		if perfFile == nil {
+			klog.Warningf("Unable to create os.File from file descriptor %#v", fd)
```
dashpole, line 147: Is there a reason why this is a warning rather than an error?  Is this expected in some cases?

Should we be returning these errors and logging them at the caller?

perf/collector_libpfm.go
```diff
@@ -0,0 +1,250 @@
...
+	config.Size = uint32(unsafe.Sizeof(unix.PerfEventAttr{}))
+}
+
+func (c *collector) Destroy() {
+	c.cpuFilesLock.Lock()
+	defer c.cpuFilesLock.Unlock()
+
+	for name, files := range c.cpuFiles {
+		for cpu, file := range files {
+			klog.V(4).Infof("Closing perf_event file descriptor for cgroup %q, event %q and CPU %d", c.cgroupPath, name, cpu)
```
dashpole, line 226: This should also be V(5).

perf/config.go
```diff
@@ -0,0 +1,95 @@
...
+	// List of perf events' names to be measured. Any value found in
+	// output of perf list can be used.
+	Events [][]Event `json:"events"`
+
+	// List of custom perf events' to be measured. It is impossible to
+	// specify some events using their names and in such case you have
+	// to provide lower level configuration.
+	CustomEvents []CustomEvent `json:"custom_events"`
+
+	eventToCustomEvent map[Event]*CustomEvent
```
dashpole, line 37: I'd prefer storing this on the collector struct, rather than in the config API.

perf/collector_libpfm.go
iwankgb: If we consider this an error then we should probably `Destroy()` the collector as it will be in half-baked state. What do you think?

dashpole: Yeah.  I would prefer propagating errors from Setup() up to the manager, and having the manager initiate the teardown.  We can then log to notify that metrics for that container will not be available, like the GPU metrics does.

iwankgb: I'm just looking at GPU implementation as I really want to keep them consistent. I will probably file another PR with some changes to interfaces defined in `stats` and their implementation in `accelerators`.

dashpole: ok, sgtm

perf/collector_libpfm.go
iwankgb: I was originally thinking about ignoring such setting but I now I think that returning noop manager early might be much better idea.

iwankgb:
^^ this is weird.

iwankgb:
@dashpole I think I have applied all requested changes.

manager/manager.go
```diff
@@ -928,7 +946,17 @@ func (m *manager) createContainerLocked(containerName string, watchSource watche
 		} else {
 			cont.nvidiaCollector, err = m.nvidiaManager.GetCollector(devicesCgroupPath)
 			if err != nil {
-				klog.V(4).Infof("GPU metrics may be unavailable/incomplete for container %q: %v", cont.info.Name, err)
+				klog.Infof("GPU metrics may be unavailable/incomplete for container %s: %s", cont.info.Name, err)
```
dashpole, line 949: Keep this at V(4)?

perf/collector_libpfm.go
```diff
@@ -0,0 +1,261 @@
...
+				continue
+			}
+			perfData := &ReadFormat{}
+			reader := bytes.NewReader(buf)
+			err = binary.Read(reader, binary.LittleEndian, perfData)
+			if err != nil {
+				klog.Warningf("Unable to decode from binary format read from perf_event file (event: %q, CPU: %d) for %q", name, cpu, c.cgroupPath)
+				continue
+			}
+			klog.V(4).Infof("Read metric for event %q for cpu %d from cgroup %q: %d", name, cpu, c.cgroupPath, perfData.Value)
```
dashpole, line 91: V(5) here.

docs/runtime_options.md
```diff
@@ -134,7 +134,67 @@ cAdvisor stores the latest historical data in memory. How long of a history it s
...
+* [perf Examples](http://www.brendangregg.com/perf.html) on Brendan Gregg's blog
+* [Kernel Perf Wiki](https://perf.wiki.kernel.org/index.php/Main_Page)
+* `man perf_event_open`
+* [perf subsystem](https://github.com/torvalds/linux/tree/v5.6/kernel/events) in Linux kernel
+
+See example configuration below:
+```json
+{
+  "events": [
+    ["instructions"]
```
dashpole, line 173: add "instructions_retired" as a second group here.

dashpole:
After this is rebased, I think we can merge and iterate from there.

iwankgb:
@dashpole rebased.

dashpole:
The remaining follow-ups i'm aware of are:

1. handle groups
2. example kubernetes configmap
3. cgroups v2

Issue #2406, opened 2020-02-27
no hugepage metrics, any reason we only add MemoryStats not HugetlbStats from cgroup stats?


sulixu:
looks like cgroup has the hugepage metrics support: https://github.com/google/cadvisor/blob/master/vendor/github.com/opencontainers/runc/libcontainer/cgroups/fs/hugetlb.go#L46 

any reason cadvisor's handler does not add hugepage stats?

https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L706

dashpole:
I believe you are the first to show interest in these metrics.  If you would like to contribute these changes, let me know, and I can give you a few pointers to get started.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2420, opened 2020-03-06
Add hugetlb stats 

sulixu:
Adding /sys/fs/cgroup/hugetlb statistics as discussed in #2406


sulixu:
@googlebot I signed it!

info/v1/container.go
```diff
@@ -833,12 +844,12 @@ type ProcessStats struct {
...
-	Cpu       CpuStats     `json:"cpu,omitempty"`
-	DiskIo    DiskIoStats  `json:"diskio,omitempty"`
-	Memory    MemoryStats  `json:"memory,omitempty"`
-	Network   NetworkStats `json:"network,omitempty"`
-
+	Timestamp time.Time               `json:"timestamp"`
+	Cpu       CpuStats                `json:"cpu,omitempty"`
+	DiskIo    DiskIoStats             `json:"diskio,omitempty"`
+	Memory    MemoryStats             `json:"memory,omitempty"`
+	Hugetlb   map[string]HugetlbStats `json:"hugetlb"`
```
dashpole: I believe we should tag this with omitempty.

sulixu:
> Can you update cadvisor.go as well? This should probably not be enabled by default, and should at least be possible to disable.

sulixu:
> Can you update cadvisor.go as well? This should probably not be enabled by default, and should at least be possible to disable.

I've make hugetlb disabled by default, to enable it need: 
--disable_metrics=tcp,udp,advtcp,sched,process

tested with: 
1. --disable_metrics=tcp,udp,advtcp,sched,process,hugetlb with no container_hugetlb_* stats
2.--disable_metrics=tcp,udp,advtcp,sched,process,  container_hugetlb_* stats showed up

dashpole:
```
I0317 00:51:05.780] >> checking go formatting
I0317 00:51:06.572] The following files are not properly formatted:
I0317 00:51:06.572] container/factory.go
```

PR #2421, opened 2020-03-10
NVM power usage

iwankgb:
This PR introduces information about NVM power usage to `/api/v2.0/machine`. Average power budget affects DIMM latency and memory bandwidth available and therefore can be used to optimize scheduling decisions in a cluster.
The PR uses external C library to obtain power information, but:
* The only alternative was to use an `ipmctl` subprocess while checking if the CLI tool is installed.
* I firmly believe that using a API when available is much better idea than the above.
* Building cAdvisor with `libipmctl` support requires using a build tag (see build docs) so users not using persistent memory will not be affected.

This is first PR using `libipmctl` but we would like to use the library for more cases when we retrieve persistent memory related information.

The PR is related to #2388.


dashpole:
Some users want to build cAdvisor without cgo.  See https://github.com/mindprince/gonvml/pull/13/files as an example of build rules for making that possible

info/v1/machine.go
```diff
@@ -176,6 +176,9 @@ type MachineInfo struct {
 	// Memory capacity and number of DIMMs by memory type
 	MemoryByType map[string]*MemoryInfo `json:"memory_by_type"`
 
+	// Average power budget for NVM devices configured in BIOS.
+	NVMAvgPowerBudget uint `json:"nvm_avg_power_budget"`
```
dashpole: Is this the only NVM-related thing we plan to have here?  If not, it would be nice to have a NVMInfo struct to group them.

iwankgb: There will be more. I'm going to update it then :)

iwankgb:
@dashpole both raised issues have been addressed.

info/v1/machine.go
```diff
@@ -176,6 +176,8 @@ type MachineInfo struct {
 	// Memory capacity and number of DIMMs by memory type
 	MemoryByType map[string]*MemoryInfo `json:"memory_by_type"`
 
+	NVMInfo NVMInfo `json:"nvm_info"`
```
katarzyna-z: Please add a comment here

katarzyna-z: Now reponse on /api/v2.0/machine request has following form:
`{
  ...
  "nvm_info": {
    "avg_power_budget": 0
  },
  "hugepages": [
    {
      "page_size": 1048576,
      "num_pages": 0
    },
    {
      "page_size": 2048,
      "num_pages": 2
    }
  ],`
  ...
}
I would like to see:
```
 "nvm": {
    "avg_power_budget": 0
  },
```
to have similar form as it is for the rest of keys on API (e.g. hugepages).

iwankgb: Done!

PR #2428, opened 2020-03-15
Drop dependency on operatingsytem package in docker/docker

dims:
Avoid dependency on docker/docker


dims:
cc @mattjmcnaughton

mattjmcnaughton:
Exciting!

dashpole:
looks like this is more than just a straight copy from docker/docker, and handles some additional cases, such as darwin, etc.  Can you update the description with the improvements you made?

dims:
@dashpole it's a re-implementation as i did not want to do a straight copy, it covers the same platforms as the docker/docker one. will write some notes below:

- `getOperatingSystem` is a private method, so we are not exposing this to the world
- operatingsystem_unix.go is for all the unix os(es)
- darwin and freebsd do not support files for looking up the release, so run `uname` (same command as used in the docker package)
- on the other unixes, we look for 2 files one after another (similar to the stuff in docker), but we do not use shellwords to parse, we just use a regexp to find the key with `PRETTY_NAME` instead of reading line by line
- for windows, we use the code from winstats, but make sure we mint the same version string as the docker/docker one used to.

PR #2430, opened 2020-03-17
support to expose environments with prefix specified

KielChan:
This is a feature that cadvisor can export envs with kinds of prefix pattern. It can be set by flag --docker-env-metadata-whitelist in kubelet. If exec cmd like `kubelet --help`(master branch),
```
      --docker-env-metadata-whitelist string                                                                      a comma-separated list of environment variable keys that needs to be collected for docker containers (DEPRECATED: This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.)
```

ref: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/globalflags_linux.go#L57

If exposed metrics with cadvisor daemonset, we can also add this feature to exposed tags matched some kind of prefix pattern. And dump these tags to TSDB like `victoria metrics` or `influxdb`.

And we can do more with these environments exposed like service discovery.
For example, if user want to expose some metrics collected by prometheus sdk, or metrics about runtime like java jmx and so on, they can set --docker-env-metadata-whitelist and acquire metrics detail by kubelet API with Subcontainers option enabled.

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -60,7 +60,7 @@ const rootDirRetryPeriod time.Duration = 1000 * time.Millisecond
 // --cgroup-parent have another prefix than 'docker'
 var dockerCgroupRegexp = regexp.MustCompile(`([a-z0-9]{64})`)
 
-var dockerEnvWhitelist = flag.String("docker_env_metadata_whitelist", "", "a comma-separated list of environment variable keys that needs to be collected for docker containers")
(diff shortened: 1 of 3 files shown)


KielChan:
I have added some test cases to prove that new method has no side effects. And `dockerEnvWhiteList` can work as before if flag content does not change, and new feature can do more.

PR #2431, opened 2020-03-17
Extending machine API by NVM memory mode capacity

katarzyna-z:

NVM memory mode capacity is visible on API in following form:
```
 "nvm": {
    "memory_mode_capcity": 429496729600,
    "avg_power_budget": 15
  }, 
``` 

related to #2388

### DIFF
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -223,7 +223,10 @@ type MemoryInfo struct {
 
 type NVMInfo struct {
 	// The total NVM capacity in bytes for memory mode.
-	MemoryModeCapcity uint `json:"memory_mode_capcity"`
+	MemoryModeCapcity uint64 `json:"memory_mode_capcity"`
+
+	//The total NVM capacity in bytes for app direct mode.
+	AppDirectModeCapcity uint64 `json:"app direct_mode_capcity"`
 
 	// Average power budget for NVM devices configured in BIOS.
 	AvgPowerBudget uint `json:"avg_power_budget"`
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -222,6 +222,9 @@ type MemoryInfo struct {
 }
 
 type NVMInfo struct {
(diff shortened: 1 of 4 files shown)


info/v1/machine.go
```diff
@@ -222,6 +222,9 @@ type MemoryInfo struct {
 }
 
 type NVMInfo struct {
+	// The total NVM capacity in bytes for memory mode.
+	MemoryModeCapcity uint `json:"memory_mode_capcity"`
```
dashpole: probably want a uint64 here?

dashpole: I may have missed this in the power budget below as well... do you know if there is a range of possible values for that?

katarzyna-z:
I decided to extend machine API also by NVM app direct mode capacity in this PR. It is a small change but this value could be useful for someone in the future. Current API form:
"""
"nvm": {
    "memory_mode_capcity": 429496729600,
    "app direct_mode_capcity": 1735166787584,
    "avg_power_budget": 15
  },
"""

info/v1/machine.go
katarzyna-z: Of course uint64 is a better choice for memory capacity. I corrected this.
power budget is in watts and I think that this value does not exceed uint range but I added information about unit of power budget in the comment.

info/v1/machine.go
```diff
@@ -222,7 +222,13 @@ type MemoryInfo struct {
 }
 
 type NVMInfo struct {
-	// Average power budget for NVM devices configured in BIOS.
+	// The total NVM capacity in bytes for memory mode.
+	MemoryModeCapcity uint64 `json:"memory_mode_capcity"`
+
+	//The total NVM capacity in bytes for app direct mode.
+	AppDirectModeCapcity uint64 `json:"app direct_mode_capcity"`
```
Creatone: Fix typo, ```Capcity``` -> ```Capacity```

PR #2432, opened 2020-03-17
Moving Nvidia interfaces

iwankgb:
Moving Nvidia interfaces to stats package so that they can be used outside of the accelerators package. This is related to #2419 and will be used there eventually.

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -129,7 +129,7 @@ func (nm *nvidiaManager) Destroy() {
 // GetCollector returns a collector that can fetch nvidia gpu metrics for nvidia devices
 // present in the devices.list file in the given devicesCgroupPath.
 func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector, error) {
-	nc := &NvidiaCollector{}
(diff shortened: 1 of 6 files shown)


accelerators/nvidia.go
```diff
@@ -16,6 +16,7 @@ package accelerators
 import (
 	"bufio"
 	"fmt"
+	"github.com/google/cadvisor/stats"
```
dashpole: nit: group cAdvisor imports together below

manager/container.go
```diff
@@ -17,6 +17,7 @@ package manager
 import (
 	"flag"
 	"fmt"
+	"github.com/google/cadvisor/stats"
```
dashpole: nit: group cAdvisor imports together below

manager/manager.go
```diff
@@ -18,6 +18,7 @@ package manager
 import (
 	"flag"
 	"fmt"
+	"github.com/google/cadvisor/stats"
```
dashpole: nit: group cAdvisor imports together below

iwankgb:
@dashpole fixed!

PR #2437, opened 2020-03-20
isolate binary dependencies from library dependencies

liggitt:
This PR:

1. Moves the main package files to a `cmd/` subdirectory (and adds a doc.go to the top-level cadvisor package)
2. Moves all storage integrations and container runtime integrations used only by the cadvisor binary to `/cmd/internal/...`
3. Moves all api/healthz/http/pages packages to `cmd/internal/...`
4. Adds a `github.com/google/cadvisor/cmd` module definition used only for building the main cadvisor package (not used as a library), and updates make/build/test scripts to build and test packages in the submodule
5. Adds a task to tidy the top-level and cmd go.mod files in the right order


liggitt:
@dims, my initial testing with this looked good... can you rebuild the https://github.com/kubernetes/kubernetes/pull/86975 branch replacing cadvisor with the tip of this branch:

https://github.com/liggitt/cadvisor/commits/isolate

currently, that's this commit:
```
require (
...
github.com/google/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c
)

replace (
...
github.com/google/cadvisor => github.com/liggitt/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c
)
```

dims:
@liggitt DONE! please see updated https://github.com/kubernetes/kubernetes/pull/86975

liggitt:
The go.mod / go.sum changes in that look way better… we stopped picking up the bad new deps I was concerned about and actually dropped some old bad ones like mesos. I'll review the other changes more closely in a bit and we can get CI green, but this is a really positive signal

dims:
@liggitt @dashpole i don't think we need the systemd container plugin either after reviewing the original PR:
https://github.com/google/cadvisor/pull/1235

The main thing about that PR was the the `systemd` handler was set *before* the `raw` handler so it skipped the raw handler. Currently the way we have the imports we add `systemd` at the very end, so it basically is a no-op.

liggitt:
is import order deterministic?

liggitt:
I also looked at other unused packages we can move to cmd/internal, and it looks like the api, healthz, http, and pages packages can all be moved. That would let k8s avoid the abbot/go-http-auth dep.

dims:
@liggitt we store it in a map, so it's not deterministic, but we store it in a map, and we iterate over it, so the original intent of being able to run systemd before raw is gone. And since this has not posed a problem so far, i'd say we don't need the systemd plugin any more

```
var plugins = make(map[string]Plugin)
````

**Comment by liggitt** [comment]
I'm not sure that means we don't need systemd in kube, it might mean there's a bug we need to fix. Will dig on Monday.

**Comment by dims** [comment]
@liggitt ack :) please start with https://github.com/google/cadvisor/pull/1235 and https://github.com/google/cadvisor/issues/1211

**Comment by liggitt** [comment]
at that point, runtime registrations got appended: https://github.com/google/cadvisor/blob/d01934a3e409b2fde078a62ee317c3b8b3250d25/container/factory.go#L74-L81

**Comment by dims** [comment]
@liggitt ah i missed that nuance. thanks!

**Comment by liggitt** [comment]
I updated with the api/http/healthz/pages move if you want to update the k/k branch to verify we're good, then this should be ready to go

**Comment by dims** [comment]
LGTM @liggitt 

updating the k/k patch is painful, let us get this in and then i'll rev it up with the SHA. once we confirm there, then we can tag a new cadvisor release.

**Comment by liggitt** [comment]
I did a quick update on top of your k/k branch, with this diff:

```diff
diff --git a/go.mod b/go.mod
index ac8797000da..a9e4b3f8d4f 100644
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/groupcache v0.0.0-20160516000752-02826c3e7903
 	github.com/golang/mock v1.3.1
-	github.com/google/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c
+	github.com/google/cadvisor v0.36.1-0.20200323135411-a801a825ac75
 	github.com/google/go-cmp v0.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.1
@@ -174,7 +174,6 @@ replace (
 	github.com/PuerkitoBio/purell => github.com/PuerkitoBio/purell v1.1.1
 	github.com/PuerkitoBio/urlesc => github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578
 	github.com/StackExchange/wmi => github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6
-	github.com/abbot/go-http-auth => github.com/abbot/go-http-auth v0.0.0-20140618235127-c0ef4539dfab
 	github.com/agnivade/levenshtein => github.com/agnivade/levenshtein v1.0.1
 	github.com/ajstarks/svgo => github.com/ajstarks/svgo v0.0.0-20180226025133-644b8db467af
 	github.com/alecthomas/template => github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc
@@ -307,7 +306,7 @@ replace (
 	github.com/golangplus/fmt => github.com/golangplus/fmt v0.0.0-20150411045040-2a5d6d7d2995
 	github.com/golangplus/testing => github.com/golangplus/testing v0.0.0-20180327235837-af21d9c3145e
 	github.com/google/btree => github.com/google/btree v1.0.0
-	github.com/google/cadvisor => github.com/liggitt/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c
+	github.com/google/cadvisor => github.com/liggitt/cadvisor v0.36.1-0.20200323135411-a801a825ac75
 	github.com/google/go-cmp => github.com/google/go-cmp v0.3.1
 	github.com/google/go-github => github.com/google/go-github v17.0.0+incompatible
 	github.com/google/go-querystring => github.com/google/go-querystring v1.0.0
diff --git a/go.sum b/go.sum
index 00ad33efc38..283f5d8035b 100644
--- a/go.sum
+++ b/go.sum
@@ -43,7 +43,6 @@ github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbt
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV/sSk/8dngufqelfh6jnri85riMAaF/M=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
-github.com/abbot/go-http-auth v0.0.0-20140618235127-c0ef4539dfab/go.mod h1:Cz6ARTIzApMJDzh5bRMSUou6UMSp0IEXg9km/ci7TJM=
 github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
 github.com/ajstarks/svgo v0.0.0-20180226025133-644b8db467af/go.mod h1:K08gAheRH3/J6wwsYMMT4xOr94bZjxIelGM0+d/wbFw=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
@@ -330,8 +329,8 @@ github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348/go.mod h1:B69LEHPfb2qLo0BaaOLcbitczOKLWTsrBG9LczfCD4k=
 github.com/libopenstorage/openstorage v1.0.0 h1:GLPam7/0mpdP8ZZtKjbfcXJBTIA/T1O6CBErVEFEyIM=
 github.com/libopenstorage/openstorage v1.0.0/go.mod h1:Sp1sIObHjat1BeXhfMqLZ14wnOzEhNx2YQedreMcUyc=
-github.com/liggitt/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c h1:RW9jt+VNi7gUMvx0JPl4B7c3aLfsMARMILZqh8bCrEo=
-github.com/liggitt/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c/go.mod h1:xKg/kEkH4bY+JBbsbaWbrlMkR7BaTxYZkq6x46r17b8=
+github.com/liggitt/cadvisor v0.36.1-0.20200323135411-a801a825ac75 h1:FLQls1t0hRzwg17Ari3Nh7NdABbcZHhPI6g/i9pGrkk=
+github.com/liggitt/cadvisor v0.36.1-0.20200323135411-a801a825ac75/go.mod h1:OTAGfVYYYA6MI5yPZ1/81PGoh6gfXpcQIyb3mXX7C3M=
 github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de h1:9TO3cAIGXtEhnIaL+V+BEER86oLrvS+kWobKpbJuye0=
 github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de/go.mod h1:zAbeS9B/r2mtpb6U+EI2rYA5OAXxsYw6wTamcNW+zcE=
 github.com/lithammer/dedent v1.1.0 h1:VNzHMVCBNG1j0fh3OrsFRkVUwStdDArbgBWoPAffktY=
diff --git a/vendor/modules.txt b/vendor/modules.txt
index acc572b85a5..cc61e33762a 100644
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -350,7 +350,7 @@ github.com/golang/protobuf/ptypes/timestamp
 github.com/golang/protobuf/ptypes/wrappers
 # github.com/google/btree v1.0.0 => github.com/google/btree v1.0.0
 github.com/google/btree
-# github.com/google/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c => github.com/liggitt/cadvisor v0.36.1-0.20200321044906-efa9742d7f7c
+# github.com/google/cadvisor v0.36.1-0.20200323135411-a801a825ac75 => github.com/liggitt/cadvisor v0.36.1-0.20200323135411-a801a825ac75
 github.com/google/cadvisor/accelerators
 github.com/google/cadvisor/cache/memory
 github.com/google/cadvisor/client/v2
```

nothing under vendor was affected, and the go.mod/go.sum looked great.

dashpole:
All set to merge?

liggitt:
pushing a couple doc cleanups, one second

liggitt:
Updated these two files:
* https://github.com/google/cadvisor/pull/2437/files#diff-3343d98ae65b027b3f09302fa43ab879
  * Installed all the container runtimes included in the library version for testing, with a comment about what should be done with them as they are shifted into cmd/internal gradually
* https://github.com/google/cadvisor/pull/2437/files#diff-5caea3b0ee8ee5c54a41bf195b261c51R5-R12
  * Added comments about the cmd -> cadvisor module dependency

liggitt:
@dashpole, all set from my side

liggitt:
fwiw, these were no uses found for the packages moved to cmd/internal:
* https://grep.app/search?q=google/cadvisor/storage/&filter[lang][0]=Go
* https://grep.app/search?q=google/cadvisor/container/mesos&filter[lang][0]=Go
* https://grep.app/search?q=google/cadvisor/api&filter[lang][0]=Go
* https://grep.app/search?q=google/cadvisor/healthz&filter[lang][0]=Go
* https://grep.app/search?q=google/cadvisor/http&filter[lang][0]=Go
* https://grep.app/search?q=google/cadvisor/pages&filter[lang][0]=Go

as a follow up, it might be good to add a verify check that the library module doesn't start depending on the cmd module accidentally

PR #2438, opened 2020-03-20
Drop vendor dir in favor of module builds

liggitt:
Drops the vendor dir (unused by default when building with go modules, no script verifying it was correct/up-to-date)

Split out of https://github.com/google/cadvisor/pull/2437 for separate review/merge

cc @dashpole


dims:
LGTM

PR #2439, opened 2020-03-20
Build go-bindata in a tmpdir

liggitt:
Builds go-bindata in a tmp dir to avoid mutating the go.mod/go.sum file of this module

Split out of #2437 for separate review/merge

cc @dashpole


dashpole:
good catch

dims:
LGTM

PR #2455, opened 2020-03-30
hugepage metrics on 0.34

sulixu:


dashpole:
We don't generally cherrypick features back to previous releases.  Is there a particular reason this needs to be cherrypicked?

PR #2456, opened 2020-04-01
Issue fix: get wrong machine info on Arm64 guest

lubinszARM:
The latest kubernetes deployment on Arm64 VM-s(official ubuntu18.04) always fails.
Because k8s always get num_cores=0 from cAdvisor on Arm64 VM-s.

The reason is that, there is no cache info on Arm64 VM-s.
So the process of getMachineInfo was blocked by the step of getCacheInfo
on Arm64 guest.

And the good news is, we can get cache info on Arm64 hosts.

When this patch was merged, I will deliver a patch to update the version
of cAdvisor in kubernetes as soon as possible.

Signed-off-by: bblu <bin.lu@arm.com>

### DIFF
--- a/utils/sysinfo/sysinfo.go
+++ b/utils/sysinfo/sysinfo.go
@@ -220,9 +220,12 @@ func GetNodesInfo(sysFs sysfs.SysFs) ([]info.Node, int, error) {
 		}
 
 		allLogicalCoresCount += len(cpuDirs)
+
+		// On some Linux platforms(such as Arm64 guest kernel), cache info may not exist.
+		// So, we should ignore error here.
 		err = addCacheInfo(sysFs, &node)
 		if err != nil {
-			return nil, 0, err
+			klog.Warningf("Found node without cache information, nodeDir: %s", nodeDir)
 		}
 
 		node.Memory, err = getNodeMemInfo(sysFs, nodeDir)
@@ -269,9 +272,11 @@ func getCPUTopology(sysFs sysfs.SysFs) ([]info.Node, int, error) {
 		}
 		node.Cores = cores
 
+		// On some Linux platforms(such as Arm64 guest kernel), cache info may not exist.
+		// So, we should ignore error here.
 		err = addCacheInfo(sysFs, &node)
 		if err != nil {
-			return nil, 0, err
+			klog.Warningf("Found cpu without cache information, cpuPath: %s", cpus)
 		}
 		nodes = append(nodes, node)
 	}
--- a/utils/sysinfo/sysinfo_test.go
+++ b/utils/sysinfo/sysinfo_test.go
@@ -273,6 +273,104 @@ func TestGetNodesWithoutMemoryInfo(t *testing.T) {
 	assert.Equal(t, 0, cores)
 }
 
+func TestGetNodesInfoWithoutCacheInfo(t *testing.T) {
+	fakeSys := &fakesysfs.FakeSysFs{}
+
+	nodesPaths := []string{
+		"/fakeSysfs/devices/system/node/node0",
+		"/fakeSysfs/devices/system/node/node1",
+	}
+	fakeSys.SetNodesPaths(nodesPaths, nil)
+
+	cpusPaths := map[string][]string{
+		"/fakeSysfs/devices/system/node/node0": {
+			"/fakeSysfs/devices/system/node/node0/cpu0",
+			"/fakeSysfs/devices/system/node/node0/cpu1",
+		},
+		"/fakeSysfs/devices/system/node/node1": {
+			"/fakeSysfs/devices/system/node/node0/cpu2",
+			"/fakeSysfs/devices/system/node/node0/cpu3",
+		},
+	}
+	fakeSys.SetCPUsPaths(cpusPaths, nil)
+
+	coreThread := map[string]string{
+		"/fakeSysfs/devices/system/node/node0/cpu0": "0",
+		"/fakeSysfs/devices/system/node/node0/cpu1": "0",
+		"/fakeSysfs/devices/system/node/node0/cpu2": "1",
+		"/fakeSysfs/devices/system/node/node0/cpu3": "1",
+	}
+	fakeSys.SetCoreThreads(coreThread, nil)
+
+	memTotal := "MemTotal:       32817192 kB"
+	fakeSys.SetMemory(memTotal, nil)
+
+	hugePages := []os.FileInfo{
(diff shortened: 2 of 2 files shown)


lubinszARM:
Hi @dashpole
Can you help to review it?
Thanks.

dims:
@lubinszARM i'd rather see the full patch ... not sure why we need to merge this change quickly

lubinszARM:
> i'd rather see the full patch ... not sure why we need to merge this change quickly
Hi, 

Currently, our conformance tests were blocked due to this issue.
https://k8s-testgrid.appspot.com/conformance-arm#Periodic%20Arm64%20conformance%20test

Because our conformance tests were deployed in Arm64 VM-s, and the guest kernel (official ubuntu18.04） doesn't show the cache information.

iwankgb:
Why have you decided to ignore errors on all the platforms rather then on ARM only?

utils/sysinfo/sysinfo.go
```diff
@@ -212,10 +212,10 @@ func GetNodesInfo(sysFs sysfs.SysFs) ([]info.Node, int, error) {
...
 
 		allLogicalCoresCount += logicalCoreCount
-		err = addCacheInfo(sysFs, &node)
-		if err != nil {
-			return nil, 0, err
-		}
+
+		// On some Linux platforms(such as Arm64 guest kernel), cache info may not exist.
+		// So, we should ignore error here.
+		_ = addCacheInfo(sysFs, &node)
```
dashpole, line 218: Can we `klog.Warningf` here at least?  It will log frequently on Arm, but this is just meant to fix the conformance test temporarily.

lubinszARM:
> Why have you decided to ignore errors on all the platforms rather then on ARM only?

Not all ARM platforms, just only some ARM VM-s, such as official ubuntu-18.04 virtual machine.
cAdvisor can work fine on Arm host.

utils/sysinfo/sysinfo.go
lubinszARM: Done.

lubinszARM:
Hi @dashpole 
Any idea?

dashpole:
I believe this was fixed by https://github.com/google/cadvisor/pull/2471

lubinszARM:
> I believe this was fixed by #2471

Hi,
I have compiled the code with your patch.
Please see the log:
root@ubuntu-03:/go/src/github.com/google/cadvisor# git log | grep kk-fix-cpus-without-nodes
    Merge pull request #2471 from katarzyna-z/kk-fix-cpus-without-nodes

It seems this issue is not gone. Please see the machine info as reference:
{"num_cores":0,"num_physical_cores":2,"num_sockets":1,"cpu_frequency_khz":0,"memory_capacity":8329531392,"memory_by_type":{},"nvm":{"memory_mode_capacity":0,"app direct_mode_capacity":0,"avg_power_budget":0},"hugepages":[{"page_size":1048576,"num_pages":0},{"page_size":2048,"num_pages":0},{"page_size":32768,"num_pages":0},{"page_size":64,"num_pages":0}],"machine_id":"af15b5a662c84009a24a3a245edf00a0","system_uuid":"62f5f880-9aa0-4f6a-b468-7287cc9fd704","boot_id":"174f24ab-23a6-4445-8b88-b1cd812c27bd","filesystems":[{"device":"/run/user/1000","capacity":832950272,"type":"vfs","inodes":1016788,"has_inodes":true},{"device":"tmpfs","capacity":832950272,"type":"vfs","inodes":1016788,"has_inodes":true},{"device":"/run","capacity":832954368,"type":"vfs","inodes":1016788,"has_inodes":true},{"device":"/dev/mapper/ubuntu--01--vg-root","capacity":45747568640,"type":"vfs","inodes":2853424,"has_inodes":true},{"device":"/dev/shm","capacity":4164763648,"type":"vfs","inodes":1016788,"has_inodes":true},{"device":"/run/lock","capacity":5242880,"type":"vfs","inodes":1016788,"has_inodes":true},{"device":"/sys/fs/cgroup","capacity":4164763648,"type":"vfs","inodes":1016788,"has_inodes":true}],"disk_map":{"253:0":{"name":"dm-0","major":253,"minor":0,"size":46749712384,"scheduler":"none"},"253:1":{"name":"dm-1","major":253,"minor":1,"size":1027604480,"scheduler":"none"},"8:0":{"name":"sda","major":8,"minor":0,"size":48318382080,"scheduler":"mq-deadline"}},"network_devices":[{"name":"enp1s0","mac_address":"52:54:00:e1:49:c6","speed":-1,"mtu":1500}],"topology":null,"cloud_provider":"Unknown","instance_type":"Unknown","instance_id":"None"}

lubinszARM:
I also added a test case of 'TestGetNodesInfoWithoutCacheInfo' for this issue.

dashpole:
cc @katarzyna-z 
I'm going to go ahead with this mitigation for now to unblock the kubernetes conformance tests.  This includes a unit test to verify we don't break them again.

lubinszARM:
Hi @dims 
Can we update the version of cAdvisor in k8s side?
Thanks.

dims:
@lubinszARM go for it

lubinszARM:
Hi @dashpole 
Can you add a new tag for cadvisor, such as v0.36.1?
So that, I can use this new tag in kubernetes.

Thanks.

dims:
@lubinszARM there's a few more things that need to get in. we typically do this just before code freeze of kubernetes to do a final sync.

utils/sysinfo/sysinfo.go
```diff
@@ -220,9 +220,12 @@ func GetNodesInfo(sysFs sysfs.SysFs) ([]info.Node, int, error) {
 		}
 
 		allLogicalCoresCount += len(cpuDirs)
+
+		// On some Linux platforms(such as Arm64 guest kernel), cache info may not exist.
+		// So, we should ignore error here.
 		err = addCacheInfo(sysFs, &node)
 		if err != nil {
-			return nil, 0, err
+			klog.Warningf("Found node without cache information, nodeDir: %s", nodeDir)
```
alexellis, line 228: Will this appear in the syslog / journal by default? If so, that may be a concern and I would encourage using a lower log level, or a flag to suppress it.

I came here via an issue on the k3s repo after having a log fill up with this message. Turns out, this code gets hit a fair bit.

lubinsz: @alexellis 
Unfortunately, this is still in the syslog.
I will resubmit a PR to solve your troubles

PR #2460, opened 2020-04-02
CRI-O client should check Unix Transport error

tedyu:
Signed-off-by: Ted Yu <yuzhihong@gmail.com>

Currently the error return from configureUnixTransport is ignored for CRI-O client construction.

This PR checks the error return.

PR #2461, opened 2020-04-02
Construct crio client once

tedyu:
Signed-off-by: Ted Yu <yuzhihong@gmail.com>

This PR stores crio client in a global variable.
Client() would return this client instance.

### DIFF
--- a/container/crio/client.go
+++ b/container/crio/client.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"sync"
 	"syscall"
 	"time"
 )
@@ -26,6 +27,12 @@ import (
 const (
 	CrioSocket            = "/var/run/crio/crio.sock"
 	maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
+)
+
+var (
+	theClient      crioClient
+	clientErr      error
+	crioClientOnce sync.Once
 )
 
 // Info represents CRI-O information as sent by the CRI-O server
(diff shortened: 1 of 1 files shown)


container/crio/client.go
```diff
@@ -28,6 +29,11 @@ const (
 	maxUnixSocketPathSize = len(syscall.RawSockaddrUnix{}.Path)
 )
 
+var (
+	theClient      crioClient
+	clientErr      error
```
dashpole, line 34: clientErr is always nil.  No need to store it.

tedyu: @dashpole 
Can you take a look at #2460 as well ?
I intend to rebase this after #2460 goes in.

tedyu: I have rebased the PR.

container/crio/client.go
```diff
@@ -70,16 +77,21 @@ func configureUnixTransport(tr *http.Transport, proto, addr string) error {
...
-	}
-	c := &http.Client{
-		Transport: tr,
-	}
-	return &crioClientImpl{
-		client: c,
-	}, nil
+	crioClientOnce.Do(func() {
+		tr := new(http.Transport)
+		if err := configureUnixTransport(tr, "unix", CrioSocket); err != nil {
```
dashpole, line 82: I think this can just be:
```golang
if clientErr = configureUnixTransport(...; err != nil {
    return
}
```

tedyu: @dashpole 
Your comment has been addressed.

dashpole:
cc @runcom @dims

dims:
LGTM logic seems to be same used in docker/client.go

PR #2467, opened 2020-04-03
Populate IP address for cri-o release with dual stack support

tedyu:
Signed-off-by: Ted Yu <yuzhihong@gmail.com>

via commit 2f2e88c93931f678201ef65ffd80161b1072f5f8 :
  Add dual stack IPv6 support

cri-o may return multiple IPs:
```
-       IP              string            `json:"ip_address"`
+       IPs             []string          `json:"ip_addresses"`
```
This PR introduces an internal struct, containerInfoDualStack, to crio client.
If IP is not obtained by parsing with single IP struct, we try with parsing using containerInfoDualStack struct.

### DIFF
--- a/container/crio/client.go
+++ b/container/crio/client.go
@@ -45,6 +45,7 @@ type ContainerInfo struct {
 	LogPath     string            `json:"log_path"`
 	Root        string            `json:"root"`
 	IP          string            `json:"ip_address"`
+	IPs         []string          `json:"ip_addresses"`
 }
 
 type crioClient interface {
@@ -117,6 +118,7 @@ func (c *crioClientImpl) ContainerInfo(id string) (*ContainerInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+	cInfo := ContainerInfo{}
 	resp, err := c.client.Do(req)
 	if err != nil {
 		return nil, err
@@ -129,9 +131,14 @@ func (c *crioClientImpl) ContainerInfo(id string) (*ContainerInfo, error) {
 		return nil, fmt.Errorf("Error finding container %s: Status %d returned error %s", id, resp.StatusCode, resp.Body)
 	}
 
-	cInfo := ContainerInfo{}
 	if err := json.NewDecoder(resp.Body).Decode(&cInfo); err != nil {
 		return nil, err
+	}
+	if len(cInfo.IP) > 0 {
+		return &cInfo, nil
+	}
+	if len(cInfo.IPs) > 0 {
+		cInfo.IP = cInfo.IPs[0]
 	}
 	return &cInfo, nil
 }


dims:
@tedyu it would be better to add both the old field and new field into the same data structure (use pointer to string/string array , omitempty etc) instead of calling the http request twice ... no?

tedyu:
I am working on a new version where there is at most one extra http call.
I thought about using two IP fields but didn't choose that approach because the presence of two related fields is confusing.
In the current formation, we know that the one field would carry the actual value.

container/crio/client.go
```diff
@@ -47,13 +47,28 @@ type ContainerInfo struct {
 	IP          string            `json:"ip_address"`
 }
 
+// containerInfoDualStack corresponds to cri-o release which supports dual stack
+type containerInfoDualStack struct {
```
dashpole, line 51: This wrapper seems strange.  I'd prefer just adding to the ContainerInfo struct.  I believe it should still be able to unmarshal from older versions...

container/crio/client.go
```diff
@@ -117,21 +133,49 @@ func (c *crioClientImpl) ContainerInfo(id string) (*ContainerInfo, error) {
...
-
-	cInfo := ContainerInfo{}
-	if err := json.NewDecoder(resp.Body).Decode(&cInfo); err != nil {
+	cInfoDualStack := containerInfoDualStack{}
+	if err := json.NewDecoder(resp.Body).Decode(&cInfoDualStack); err != nil {
 		return nil, err
 	}
+	if len(cInfoDualStack.IPs) > 0 {
+		cInfo.IP = cInfoDualStack.IPs[0]
+		c.dualStack = dualStackSupported
```
dashpole, line 178: Do we use this information anywhere?  What is the purpose of storing and adding it?

container/crio/client.go
```diff
@@ -117,21 +133,49 @@ func (c *crioClientImpl) ContainerInfo(id string) (*ContainerInfo, error) {
...
 	}
-
-	cInfo := ContainerInfo{}
-	if err := json.NewDecoder(resp.Body).Decode(&cInfo); err != nil {
+	cInfoDualStack := containerInfoDualStack{}
+	if err := json.NewDecoder(resp.Body).Decode(&cInfoDualStack); err != nil {
 		return nil, err
 	}
+	if len(cInfoDualStack.IPs) > 0 {
+		cInfo.IP = cInfoDualStack.IPs[0]
```
dashpole, line 177: How does this add ipv6 support?  It looks like we still just take a single IP from the response, which is presumably the IPV4 address

tedyu: The handler just needs one address:
```
	handler.ipAddress = cInfo.IP
```

container/crio/client.go
tedyu: Dropped

container/crio/client.go
tedyu: @dashpole 
Comment has been addressed.

container/crio/client.go
dashpole: Ah, got it.  Thanks

dims:
much better

LGTM

PR #2469, opened 2020-04-06
libcontainer, cgroupv2: adapt to new API

giuseppe:
adapt the handlers to use the new libcontainer API for cgroup v2.


giuseppe:
I've not re-vendored github.com/opencontainers/runc as part of this PR, although the issue for cgroup v2 is fixed: https://github.com/opencontainers/runc/pull/2297

@dashpole should I re-vendor runc here or do we wait for the next release?

dims:
@giuseppe let's request runc folks to do a rc11 please! looks like a bunch of things have piled up already - https://github.com/opencontainers/runc/compare/v1.0.0-rc10...HEAD and the last release was in Jan.

giuseppe:
@AkihiroSuda could we have a new release for runc?

AkihiroSuda:
cc @cyphar

AkihiroSuda:
Maybe we should merge cgroup2 integration tests before rc11.

https://github.com/opencontainers/runc/pull/2295

container/libcontainer/handler.go
```diff
@@ -60,9 +60,23 @@ func NewHandler(cgroupManager cgroups.Manager, rootFs string, pid int, includedM
...
-	if err != nil {
-		return nil, err
+	var cgroupStats *cgroups.Stats
+	readCgroupStats := true
+	if cgroups.IsCgroup2UnifiedMode() {
+		p, err := h.cgroupManager.GetUnifiedPath()
+		if err != nil {
+			return nil, err
+		}
+		if p == fs2.UnifiedMountpoint {
```
dashpole, line 70: what is this checking, and why do we not want to read cgroup stats anymore?

manager/manager.go
```diff
@@ -148,11 +148,15 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig
 	}
 
 	// Detect the container we are running on.
-	selfContainer, err := cgroups.GetOwnCgroupPath("cpu")
-	if err != nil {
-		return nil, err
+	selfContainer := "/"
+	var err error
+	if !cgroups.IsCgroup2UnifiedMode() {
```
dashpole, line 153: why doesn't this work in cgroup2 unified mode?

container/libcontainer/handler.go
```diff
@@ -728,8 +742,13 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 		ret.Memory.HierarchicalData.Pgmajfault = v
 	}
 
+	inactiveFileKeyName := "total_inactive_file"
+	if cgroups.IsCgroup2UnifiedMode() {
+		inactiveFileKeyName = "inactive_file"
```
dashpole, line 747: is this actually the same as total_inactive_file?  I thought total included all children cgroup's inactive_file?

giuseppe: in cgroupv2 there is only `inactive_file` that has the same semantic as `total_inactive_file` in v1.  

I've verified it locally with:

```
#  mkdir -p /sys/fs/cgroup/a/b/c/d
...(move process to the new cgroup, enable the memory controller and open/read some random files)...
/ # grep inactive_file /sys/fs/cgroup/a/b/c/d/memory.stat
inactive_file 405504
/ # grep inactive_file /sys/fs/cgroup/a/b/c/memory.stat
inactive_file 405504
/ # grep inactive_file /sys/fs/cgroup/a/b/memory.stat
inactive_file 405504
/ # grep inactive_file /sys/fs/cgroup/a/memory.stat
inactive_file 405504
```

manager/manager.go
giuseppe: GetOwnCgroup just fails on cgroup v2 at the moment:

https://github.com/opencontainers/runc/blob/dd8d48ede80d6538cab42673699a4f6704e58912/libcontainer/cgroups/utils.go#L312-L314

container/libcontainer/handler.go
giuseppe: it is checking if it is the root cgroup, if it can help I can change the variable name to make it more explicit.
With cgroup v2 there are no stats in the root cgroup, foe example:

```
$ ls /sys/fs/cgroup/memory.stat /sys/fs/cgroup/init.scope/memory.stat
ls: cannot access '/sys/fs/cgroup/memory.stat': No such file or directory
/sys/fs/cgroup/init.scope/memory.stat
```

dashpole: a comment would also be helpful

manager/manager.go
dashpole: Can we add a warning if running in unified mode, and open an issue to track removing this once getowncgroup is working?

container/libcontainer/handler.go
giuseppe: thanks for the review.  Added a comment

manager/manager.go
giuseppe: added a warning and reported the issue to opencontainers/runc

giuseppe:
@dims @dashpole it seems a runc release is going to take more time.

Would it be possible to re-vendor for now using the current master and then we can update the dependency again once the release is out?

A new cAdvisor+libcontainer is necessary for the Kubelet to work on cgroup v2

dims:
@giuseppe looks like they are trying to fix something for us!! https://github.com/opencontainers/runc/pull/2364#issuecomment-621546746 So no, i would not recommend updating now until they are ready.

giuseppe:
> @giuseppe looks like they are trying to fix something for us!! [opencontainers/runc#2364 (comment)](https://github.com/opencontainers/runc/pull/2364#issuecomment-621546746) So no, i would not recommend updating now until they are ready.

the devices cgroup is not used by the Kubelet, so the fix won't affect Kubernetes

mrunalp:
It is going to take some to establish a release cadence in runc that lets us iterate quick enough on cgroups v2. I propose that keep updating runc and then pin to a release which gives us a way to keep making progress on cgroups v2.

giuseppe:
@dashpole @dims given the reasoning above and that we will probably need to iterate more times until we get cgroup v2 right, are you fine with the current PR so we can unblock the Kubelet?

dims:
@giuseppe sounds reasonable. please update this PR to latest then?

giuseppe:
> @giuseppe sounds reasonable. please update this PR to latest then?

yes sure.  I've squashed the runc update commit into the `libcontainer, cgroupv2: adapt to new API` commit, otherwise the commit will fail to build during a git bisect.

PR #2473, opened 2020-04-06
Check error return from common.GetSpec

tedyu:
crioContainerHandler#GetSpec() doesn't check the error return from common.GetSpec.

This PR adds the check.

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -225,6 +225,9 @@ func (self *crioContainerHandler) needNet() bool {
 func (self *crioContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	hasFilesystem := self.includedMetrics.Has(container.DiskUsageMetrics)
 	spec, err := common.GetSpec(self.cgroupPaths, self.machineInfoFactory, self.needNet(), hasFilesystem)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


container/crio/handler.go
```diff
@@ -328,7 +331,6 @@ func (self *crioContainerHandler) GetStats() (*info.ContainerStats, error) {
 }
 
 func (self *crioContainerHandler) ListContainers(listType container.ListType) ([]info.ContainerReference, error) {
-	// No-op for Docker driver.
```
dashpole, line 331: Any reason why we are removing the comment here?

container/crio/handler.go
```diff
@@ -225,6 +225,9 @@ func (self *crioContainerHandler) needNet() bool {
 func (self *crioContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	hasFilesystem := self.includedMetrics.Has(container.DiskUsageMetrics)
 	spec, err := common.GetSpec(self.cgroupPaths, self.machineInfoFactory, self.needNet(), hasFilesystem)
+	if err != nil {
+		return info.ContainerSpec{}, err
```
dashpole, line 229: I think we still want to set the Lables, Envs, and Image here... I don't think this is needed.

PR #2477, opened 2020-04-08
Returning more error information from stats.Manager and stats.Collector

iwankgb:
@dashpole I would appreciate if you took a look - I need these changes to proceed with #2419.

Not changed:
* when `nil` and real `error` are returned.

Changed:
* `stats.NoopCollector` might be returned sometimes.
* `stats.Manager.Setup()` may return an error
* `stats.Collector.Setup()` may return an error

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -99,9 +99,6 @@ var initializeNVML = func(nm *nvidiaManager) error {
 	if err != nil {
 		return fmt.Errorf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
 	}
-	if numDevices == 0 {
-		return nil
-	}
 	klog.V(1).Infof("NVML initialized. Number of nvidia devices: %v", numDevices)
 	nm.nvidiaDevices = make(map[int]gonvml.Device, numDevices)
 	for i := 0; i < int(numDevices); i++ {
@@ -140,8 +137,7 @@ func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector
 		err := initializeNVML(nm)
 		if err != nil {
 			nm.Unlock()
-			klog.Warningf("Unable to initialize NVML: %s", err)
-			return &stats.NoopCollector{}, nil
+			return &stats.NoopCollector{}, err
 		}
 	}
 	nm.Unlock()
@@ -150,7 +146,7 @@ func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector
 	}
 	nvidiaMinorNumbers, err := parseDevicesCgroup(devicesCgroupPath)
 	if err != nil {
-		return nil, err
+		return &stats.NoopCollector{}, err
 	}
 
 	for _, minor := range nvidiaMinorNumbers {
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -100,7 +100,7 @@ var initializeNVML = func(nm *nvidiaManager) error {
 		return fmt.Errorf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
 	}
 	if numDevices == 0 {
-		return fmt.Errorf("no nvidia devices found")
+		return nil
 	}
 	klog.V(1).Infof("NVML initialized. Number of nvidia devices: %v", numDevices)
 	nm.nvidiaDevices = make(map[int]gonvml.Device, numDevices)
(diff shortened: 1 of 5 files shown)


accelerators/nvidia.go
```diff
@@ -89,34 +88,34 @@ func detectDevices(vendorId string) bool {
...
 	}
 	nm.nvmlInitialized = true
 	numDevices, err := gonvml.DeviceCount()
 	if err != nil {
-		klog.Warningf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
-		return
+		return fmt.Errorf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
+	}
+	if numDevices == 0 {
+		return fmt.Errorf("no nvidia devices found")
```
dashpole, line 103: Is this actually an error scenario?  Isn't it valid to have no devices?

accelerators/nvidia.go
```diff
@@ -131,28 +130,30 @@ func (nm *nvidiaManager) Destroy() {
 func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector, error) {
 	nc := &nvidiaCollector{}
 
-	if !nm.devicesPresent {
-		return nc, nil
+	if !nm.devicesPresent || len(nm.nvidiaDevices) == 0 {
```
dashpole, line 133: I think len(nm.nvidiaDevices) will be zero if !nm.nvmlInitialized is false?  I think that is why this case was handled below in the previous version...

accelerators/nvidia.go
```diff
@@ -132,27 +131,32 @@ func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector
...
 	if !nm.nvmlInitialized {
-		initializeNVML(nm)
-	}
-	if !nm.nvmlInitialized || len(nm.nvidiaDevices) == 0 {
-		nm.Unlock()
-		return nc, nil
+		err := initializeNVML(nm)
+		if err != nil {
+			nm.Unlock()
+			klog.Warningf("Unable to initialize NVML: %s", err)
```
dashpole, line 143: Since this is called on each container creation, we probably don't want to Warningf here.  E.g. if nvml isn't installed on the system, this is expected to log.   I think if we just return the error, it will be logged at V(4), which is probably what we want

accelerators/nvidia.go
```diff
@@ -132,27 +131,32 @@ func (nm *nvidiaManager) GetCollector(devicesCgroupPath string) (stats.Collector
...
+		}
 	}
 	nm.Unlock()
+	if len(nm.nvidiaDevices) == 0 {
+		return &stats.NoopCollector{}, nil
+	}
 	nvidiaMinorNumbers, err := parseDevicesCgroup(devicesCgroupPath)
 	if err != nil {
-		return nc, err
+		return nil, err
```
dashpole, line 153: should this be a noopcollector?

accelerators/nvidia.go
```diff
@@ -89,34 +88,34 @@ func detectDevices(vendorId string) bool {
...
+		return fmt.Errorf("Could not initialize NVML: %v", err)
 	}
 	nm.nvmlInitialized = true
 	numDevices, err := gonvml.DeviceCount()
 	if err != nil {
-		klog.Warningf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
-		return
+		return fmt.Errorf("GPU metrics would not be available. Failed to get the number of nvidia devices: %v", err)
+	}
+	if numDevices == 0 {
```
dashpole, line 102: I think you can remove this if statement entirely?  It won'y enter into the for loop below, and will just return nil anyways.

manager/manager.go
```diff
@@ -264,7 +264,12 @@ func (self *manager) Start() error {
 	}
 
 	// Setup collection of nvidia GPU metrics if any of them are attached to the machine.
-	self.nvidiaManager.Setup()
+	err = self.nvidiaManager.Setup()
+	if err != nil {
+		// We intentionally do nothing here. If Setup() fails then GetCollector()
+		// will return stats.NoopCollector
+		klog.Warningf("Setting up nvidia manager failed: %s", err)
```
dashpole, line 271: Since this is expected in some cases, can we just add something like: "NVidia GPU metrics will not be available"?  Every warning we add generates "Look at the warning I found; it caused ______!" bugs from users...

iwankgb:
@dashpole fixed.

manager/manager.go
dashpole: We can leave this for another PR

PR #2479, opened 2020-04-09
manager#Exists should take read lock

tedyu:
manager#Exists doesn't perform write operation.
Taking read lock should suffice.


tedyu:
@dashpole 
Can you review this ?

PR #2480, opened 2020-04-09
Add missing fields to ContainerSpec#Eq

tedyu:
There were some recently added fields of ContainerSpec missing in Eq().

This PR adds them.


tedyu:
@dashpole 
Tests passed.

PR #2489, opened 2020-04-10
Integration test for Docker and perf stats

iwankgb:
Integration tests for Docker container and perf stats.


.github/workflows/test.yml
```diff
@@ -41,3 +41,9 @@ jobs:
       - name: Run integration tests
         run: |
           cd $GITHUB_WORKSPACE/go/src/github.com/google/cadvisor && make docker-test-integration
+      - name: Upload cAdvisor log file
```
iwankgb, line 44: It's really useful to have logs available when build fails.

build/integration.sh
```diff
@@ -26,7 +26,7 @@ printf "" # Refresh sudo credentials if necessary.
 function start {
   set +e  # We want to handle errors if cAdvisor crashes.
   echo ">> starting cAdvisor locally"
-  GORACE="halt_on_error=1" ./cadvisor --docker_env_metadata_whitelist=TEST_VAR --v=4 --logtostderr &> "$log_file"
+  GORACE="halt_on_error=1" ./cadvisor --docker_env_metadata_whitelist=TEST_VAR --v=6 --logtostderr $CADVISOR_ARGS &>> "$log_file"
```
iwankgb, line 29: I increased verbosity and replaced `>` with `>>` to make sure that no message will be missed and that logs from both runs will be saved.

build/integration.sh
```diff
@@ -56,5 +56,5 @@ while [ "$(curl -Gs http://localhost:8080/healthz)" != "ok" ]; do
 done
 
 echo ">> running integration tests against local cAdvisor"
-./api.test --vmodule=*=2
-./healthz.test --vmodule=*=2
+./api.test --vmodule=*=2 -test.v
```
iwankgb, line 59: `-test.v` makes it much easier to see what tests were executed or failed.

iwankgb:
@dashpole, you will find a proposal for extending integration tests in this PR. What changes is that integration tests are being run for "normal" build of cAdvisor (i.e. with no tags) and for a custom one (with `libpfm` support). I was trying to make each of the test runs independent but it seems to be impossible to use list of objects in build matrix configuration so I decided to settle with shell script based approach.
I have a question about `pull-cadvisor-e2e` tests: as far as I understand integrations tests are executed using `test-integration` make target; is this target expected to run on any distribution? To be able to execute perf-related tests I need to install an additional package and it's a distro-specific operation...

build/integration-in-docker.sh
```diff
@@ -19,35 +19,57 @@ set -e
...
+    ${EXTRA_DOCKER_OPTS} \
+    --privileged \
+    --cap-add="sys_admin" \
+    --entrypoint="" \
+    gcr.io/k8s-testimages/bootstrap \
+    bash -c "apt update && apt install $PACKAGES && \
+CADVISOR_ARGS="$CADVISOR_ARGS" /usr/local/bin/runner.sh build/integration.sh"
+}
+
+run_tests "-race" "sudo"
```
dashpole, line 73: Is there a reason why we want to do perf tests separately?  Are they expected to fail in some scenarios?

integration/tests/api/perf_test.go
```diff
@@ -0,0 +1,57 @@
...
+	containerID := fm.Docker().RunPause()
+	waitForContainerInfo(fm, containerID)
+
+	info, err := fm.Cadvisor().Client().DockerContainer(containerID, &v1.ContainerInfoRequest{
+		NumStats: 1,
+	})
+
+	assert.Nil(t, err)
+	assert.Len(t, info.Stats, 1)
+	assert.Greater(t, len(info.Stats[0].PerfStats), 0, "Length of info.Stats[0].PerfStats is not greater tahn zero")
```
dashpole, line 42: s/tahn/than

integration/tests/api/perf_test.go
```diff
@@ -0,0 +1,57 @@
...
+	for k, stat := range info.Stats[0].PerfStats {
+		//Everything beyond name is non-deterministic unfortunately.
+		assert.Contains(t, []string{"context-switches", "cpu-migrations-custom"}, stat.Name, "Wrong metric name for key %d: %#v", k, stat)
```
dashpole, line 45: Are we guaranteed to have more than 0 of either of these events?  Even that check can sometimes catch problems...

dashpole:
cc @dims 
in case you want to take a look

build/integration-in-docker.sh
dashpole: Just read the comment above.  I think this is a reasonable solution, although I think we should consider having a separate target.  That way users who just want to run the "distro agnostic" tests can do so without having errors mixed in.

dims:
+1 to "perf tests separately" please!

looks reasonable otherwise :)

integration/tests/api/perf_test.go
iwankgb: In `waitForContainerInfo()` we wait for `v1.ContainerInfo` to be available. Perf stats are part of this structure and I would consider it a bug if they were not available.

iwankgb:
@dashpole @dims - is such approach acceptable? Changes are:
* there are shell files that should be sourced before running non-standard integration tests suite
* same `make` target is used in both cases
* paths to the shell files are part of build matrix
* documentation is provided.

Eventually it would make sense to execute unit tests in Docker too, I think. There are some unit tests against libpfm4 part of code that can be executed without installing the library. What is you opinion?

dims:
@iwankgb works for me!

PR #2490, opened 2020-04-10
Make a copy of MachineInfo in GetMachineInfo()

tedyu:
Fixes #2487

The data race was due to manager.machineInfo being read after manager#GetMachineInfo() returns.

This PR makes a copy of MachineInfo and returns the pointer to the copy.

PR #2491, opened 2020-04-11
Make a copy of the maps of MachineInfo

tedyu:
For manager#GetMachineInfo(), we should duplicate the maps inside MachineInfo when copying.

This completes #2490

### DIFF
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -211,6 +211,44 @@ type MachineInfo struct {
 	InstanceID InstanceID `json:"instance_id"`
 }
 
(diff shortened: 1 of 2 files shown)


dims:
do we want to use mohae/deepcopy?

tedyu:
mohae/deepcopy would be a new dependency.

We can wait for a few more instances where deepcopy is needed before it is adopted.

In the current formation, we don't incur cost of reflection.

tedyu:
@dashpole 
Can you take a look ?

Issue #2482, opened 2020-04-09
Perf: exposing metrics to all cAdvisor storage.StorageDriver implementations


iwankgb:
It must be possible to store metrics in using any of available StorageDrivers.

katarzyna-z:
Partially addressed in https://github.com/google/cadvisor/pull/2612, https://github.com/google/cadvisor/pull/2638, and https://github.com/google/cadvisor/pull/2593

katarzyna-z:
Perf metrics are available in ElasticSearch, Kafka and Redis without any changes in code.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2494, opened 2020-04-15
Exposing perf metrics in Prometheus

iwankgb:
This is part of #2482


metrics/prometheus.go
```diff
@@ -1716,6 +1717,16 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 				}
 			}
 		}
+		if c.includedMetrics.Has(container.PerfMetrics) {
+			for _, metric := range stats.PerfStats {
+				labelNames := append(labels, "cpu")
```
dashpole, line 1722: we should still add rawLabels here.

metrics/prometheus.go
```diff
@@ -1716,6 +1717,16 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
...
+				labelValues := append(values, strconv.Itoa(metric.Cpu))
```
dashpole, line 1723: Would anyone ever want to aggregate over all events?  It may be better to give the metric a static name, e.g. perf_events_total, and add the event name as a label.  In general, putting non-constant things in metric names is an anti-pattern in prometheus.

iwankgb: I use non-constant metric name only to pass scaling information. Here CPU ID is added as a label because I believe that passing raw values is better idea than aggregating them in a way that would suite all the use cases.

katarzyna-z:
Please add information about new Prometheus metrics to documentation, see: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

metrics/prometheus.go
iwankgb: I think that this is where `labels` and `values` [are coming from](https://github.com/google/cadvisor/blob/ba6fa06c73215e151a95a735a24e0e74d8da272a/metrics/prometheus.go#L1644).

metrics/prometheus.go
iwankgb: For instance low value of scaling ration indicates that event was being measured for very short time and value might not be accurate. I had an internal discussion about aggregating perf stats this week and we were not able to find a solution that could be universal enough.

metrics/prometheus.go
dashpole: Oh, can we put this up in the table above instead of down here?

metrics/prometheus.go
dashpole: I'm not quite sure what problems it could cause... I'm proposing

`perf_events_total{cpu=1, event="instruction"}`

That can be broken down in the same way that `container_perf_instructions{cpu=1}` can.  But I can also then write more generic queries, like `rate(perf_events_total[1m]) by (event)` to graph the rate of all perf events.  That way, someone could create a cAdvisor dashboard that is decoupled from their perf config, making it possible to re-use it.

What were the concerns you had about having a constant metric name?

metrics/prometheus.go
iwankgb: `cpu` is perf specific - I would prefer to keep it local, to be honest.

dashpole: The table above allows specifying extra labels that are only applicable to the perf metric.

metrics/prometheus.go
iwankgb: Got it. Looks perfectly reasonable to me now :)

metrics/prometheus.go
iwankgb: Sorry, but I can't get what table you are referring to. Could you point me at line of code, please?

dashpole: https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L1474

iwankgb:
@dashpole @katarzyna-z I think it should be much better now.

PR #2495, opened 2020-04-16
Add working set size metric

katarzyna-z:
see: https://github.com/brendangregg/wss#wsspl-referenced-page-flag
- Introduce map of collectors in containerData
- Introduce map of managers in Manager

The working set size indicates how much memory  a container needs to keep working. It's intrusive metric because collection of metric can influence kernel page reclaim policy and add latency (this is mentioned in [documentation](https://raw.githubusercontent.com/katarzyna-z/cadvisor/31a447c6fdcfb2df01a0e82968fb7d7f55ec4b45/docs/storage/prometheus.md)). Wss metric is disabled by default.


katarzyna-z:
I'm working on fix for data race indicated in tests:
```
==================
WARNING: DATA RACE
Read at 0x00c0004d2db8 by goroutine 50:
  encoding/json.sliceEncoder.encode()
      /usr/local/go/src/reflect/value.go:1071 +0x137
  encoding/json.sliceEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:760 +0x7b
  encoding/json.structEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:664 +0x40d
  encoding/json.structEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:635 +0xa0
  encoding/json.ptrEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:810 +0xfc
  encoding/json.ptrEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:805 +0x7b
  encoding/json.arrayEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:791 +0xe3
  encoding/json.arrayEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:784 +0x7b
  encoding/json.sliceEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:765 +0xda
  encoding/json.sliceEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:760 +0x7b
  encoding/json.structEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:664 +0x40d
  encoding/json.structEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:635 +0xa0
  encoding/json.mapEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:706 +0x36e
  encoding/json.mapEncoder.encode-fm()
      /usr/local/go/src/encoding/json/encode.go:682 +0x7b
  encoding/json.(*encodeState).reflectValue()
      /usr/local/go/src/encoding/json/encode.go:337 +0x93
  encoding/json.(*encodeState).marshal()
      /usr/local/go/src/encoding/json/encode.go:309 +0xcc
  encoding/json.Marshal()
      /usr/local/go/src/encoding/json/encode.go:161 +0x73
  github.com/google/cadvisor/cmd/internal/api.writeResult()
      /go/src/github.com/google/cadvisor/cmd/internal/api/handler.go:126 +0x5d
  github.com/google/cadvisor/cmd/internal/api.(*version1_2).HandleRequest()
      /go/src/github.com/google/cadvisor/cmd/internal/api/versions.go:236 +0x570
  github.com/google/cadvisor/cmd/internal/api.(*version1_3).HandleRequest()
      /go/src/github.com/google/cadvisor/cmd/internal/api/versions.go:272 +0x104
  github.com/google/cadvisor/cmd/internal/api.handleRequest()
      /go/src/github.com/google/cadvisor/cmd/internal/api/handler.go:121 +0x766
  github.com/google/cadvisor/cmd/internal/api.RegisterHandlers.func1()
      /go/src/github.com/google/cadvisor/cmd/internal/api/handler.go:51 +0x75
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2007 +0x51
  net/http.(*ServeMux).ServeHTTP()
      /usr/local/go/src/net/http/server.go:2387 +0x288
  net/http.(*ServeMux).ServeHTTP()
      /usr/local/go/src/net/http/server.go:2387 +0x288
  net/http.serverHandler.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2802 +0xce
  net/http.(*conn).serve()
      /usr/local/go/src/net/http/server.go:1890 +0x837

Previous write at 0x00c0004d2db8 by goroutine 61:
  github.com/google/cadvisor/perf.(*collector).UpdateStats()
      /go/src/github.com/google/cadvisor/perf/collector_libpfm.go:102 +0x95a
  github.com/google/cadvisor/manager.(*containerData).updateStats()
      /go/src/github.com/google/cadvisor/manager/container.go:658 +0x514
  github.com/google/cadvisor/manager.(*containerData).housekeepingTick()
      /go/src/github.com/google/cadvisor/manager/container.go:535 +0x23a
  github.com/google/cadvisor/manager.(*containerData).housekeeping()
      /go/src/github.com/google/cadvisor/manager/container.go:483 +0x2ee

Goroutine 50 (running) created at:
  net/http.(*Server).Serve()
      /usr/local/go/src/net/http/server.go:2928 +0x5b5
  net/http.(*Server).ListenAndServe()
      /usr/local/go/src/net/http/server.go:2825 +0x102
  main.main()
      /usr/local/go/src/net/http/server.go:3081 +0x10e9

Goroutine 61 (running) created at:
  github.com/google/cadvisor/manager.(*containerData).Start()
      /go/src/github.com/google/cadvisor/manager/container.go:108 +0x4c
  github.com/google/cadvisor/manager.(*manager).createContainerLocked()
      /go/src/github.com/google/cadvisor/manager/manager.go:1030 +0xa4e
  github.com/google/cadvisor/manager.(*manager).createContainer()
      /go/src/github.com/google/cadvisor/manager/manager.go:926 +0xbd
  github.com/google/cadvisor/manager.(*manager).watchForNewContainers.func1()
      /go/src/github.com/google/cadvisor/manager/manager.go:1182 +0x41a
==================
```

info/v1/container.go
```diff
@@ -899,6 +899,9 @@ type ContainerStats struct {
 
 	// Statistics originating from perf events
 	PerfStats []PerfStat `json:"perf_stats,omitempty"`
+
+	// Working set size
+	Wss uint64 `json:"wss,omitempty"`
```
dashpole, line 904: Should we nest this under MemoryStats?

cmd/cadvisor.go
```diff
@@ -132,7 +134,7 @@ func (ml *metricSetValue) Set(value string) error {
 }
 
 func init() {
-	flag.Var(&ignoreMetrics, "disable_metrics", "comma-separated list of `metrics` to be disabled. Options are 'disk', 'diskIO', 'network', 'tcp', 'udp', 'percpu', 'sched', 'process', 'hugetlb'.")
+	flag.Var(&ignoreMetrics, "disable_metrics", "comma-separated list of `metrics` to be disabled. Options are 'disk', 'diskIO', 'network', 'tcp', 'udp', 'percpu', 'sched', 'process', 'hugetlb', 'wss'.")
```
dashpole, line 137: We already have a concept which we refer to as the working set... (memory_usage_bytes - total_inactive_file).  Is there a different name we could give this?

docs/storage/prometheus.md
```diff
@@ -78,6 +78,7 @@ Metric name | Type | Description | Unit (where applicable)
 `container_spec_memory_reservation_limit_bytes` | Gauge | Memory reservation limit for the container | bytes
 `container_start_time_seconds` | Gauge | Start time of the container since unix epoch | seconds
 `container_tasks_state` | Gauge | Number of tasks in given state (`sleeping`, `running`, `stopped`, `uninterruptible`, or `ioawaiting`) |
+`container_wss` | Gauge |  Container referenced bytes during last measurements cycle based on Referenced field in /proc/smaps file, with /proc/PIDs/clear_refs set to 1 after defined number of cycles configured through `wss_reset_interval` cAdvisor parameter.</br>Warning: this is intrusive collection, because can influence kernel page reclaim policy and add latency. Refer to https://github.com/brendangregg/wss#wsspl-referenced-page-flag for more details. | bytes
```
dashpole, line 81: Spitballing here, but can't we still rely on cgroup memory accounting to be correct?  Would it be enough just to use proc/PIDs/clear_refs to force reclaim?

The way to check would be to see if our memory metrics (e.g. usage_bytes, working_set or rss) match what we get from the new container_wss metric when it is turned on.

manager/manager.go
```diff
@@ -44,18 +44,26 @@ import (
...
 var logCadvisorUsage = flag.Bool("log_cadvisor_usage", false, "Whether to log the usage of the cAdvisor container")
 var eventStorageAgeLimit = flag.String("event_storage_age_limit", "default=24h", "Max length of time for which to store events (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is a duration. Default is applied to all non-specified event types")
 var eventStorageEventLimit = flag.String("event_storage_event_limit", "default=100000", "Max number of events to store (per type). Value is a comma separated list of key values, where the keys are event types (e.g.: creation, oom) or \"default\" and the value is an integer. Default is applied to all non-specified event types")
 var applicationMetricsCountLimit = flag.Int("application_metrics_count_limit", 100, "Max number of application metrics to store (per container)")
+var wssResetInterval = flag.Uint64("wss_reset_interval", 5, "Reset interval for working set size metric, number of measurement cycles after which referenced bytes are cleared")
```
dashpole, line 66: Does "cycle" == "housekeeping interval" here?  Would it be more intuitive to measure in seconds?  Or even better, could we just reset it every housekeeping interval, and not have a knob at all?

manager/manager.go
```diff
@@ -953,11 +970,21 @@ func (m *manager) createContainerLocked(containerName string, watchSource watche
...
 			if err != nil {
 				klog.Infof("perf_event metrics will not be available for container %s: %s", cont.info.Name, err)
 			}
 		}
+
+		cpuCgroupPath, err := handler.GetCgroupPath("cpu")
+		if err != nil {
+			klog.Warningf("Error getting cpu cgroup path: %q", err)
+		} else {
+			cont.collectors[wssMetricsProviderKey], err = m.metricsManagers[wssMetricsProviderKey].GetCollector(cpuCgroupPath)
```
dashpole, line 983: instead of using the manager model, can we instead put this in the container collector code?  I don't think there is any initialization or anything like that that needs to be done.  This seems more along the lines of process scheduler metrics: https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L72

info/v1/container.go
katarzyna-z: All memory stats come from memory cgroup so I think that wss does not fit to them.

cmd/cadvisor.go
katarzyna-z: The metric could be named 'container_referenced_bytes' then on `disabled_metrics` there will be `referenced` option.

manager/manager.go
katarzyna-z: It could be good to have a mechanism to set a different interval for setting clear_refs than housekeeping interval. Sometimes it is better to measure wss in longer intervals. I've found following information:

> What you intend to use WSS for will direct how you measure it. Consider these three scenarios:
A) You are sizing main memory for an application, with the intent to keep it from paging (swapping). WSS will be measured in bytes over a long interval, say, one minute.
B) You are optimizing CPU caches. WSS will be measured in unique cachelines accessed over a short interval, say, one second or less. The cacheline size is architecture dependent, usually 64 bytes.
C) You are optimizing the TLB caches (translation lookaside buffer: a memory management unit cache for virtual to physical translation). WSS will be measured in unique pages accessed over a short interval, say, one second or less. The pagesize is architecture and OS configuration dependent, 4 Kbytes is commonly used.

It comes from http://www.brendangregg.com/wss.html#Estimation

If I measure time for setting clear_refs in seconds I need to introduce a new timer. Do you think that is a good idea?

manager/manager.go
katarzyna-z: I moved getting of wss into libcontainer/handler.go. Do you think that is an good idea to move the rest of the code from wss package into libcontainer?

docs/storage/prometheus.md
ppalucki: Interval to reset "referenced flags" should be optional and orthogonal to "container_referenced_bytes" collection, the "referenced_bytes" itself may be useful for application analysis (referenced bytes are mapped to process pages with accessed/young bits) and are also cleared by kswapd during periodic scan or low memory conditions, but...

You're right about that "memcg concept" and big thanks for pointing that (but the formula should also include inactive_anon_page). memcg approach is based on pages belonging to different LRU list (active/inactive) rather than "accessed bit" raised by HW . My gut felling is that, "referenced approach" is more accurate - but cannot dispute more without empirical results - let us perform more detailed analysis comparing both approaches before proceeding.

dashpole: Looking at the man page for clear_refs, I can't actually tell if it is actually inducing reclaim or primarily updating accounting references.  My suggestion is probably only reasonable if it is doing the former...

manager/manager.go
dashpole: Actually, does it even make sense to have this differ from the housekeeping interval?  If my HK interval is 10 seconds, and wss reset is 6, won't I get higher and higher measurements each housekeeping until it is reset?  It seems like we would want to collect at the same interval that we reset at.

I guess this begs the question of whether this is meant to be used just for benchmarking, or whether this is something we expect to be enabled for production monitoring.  The warnings everywhere make this seem like mostly a benchmarking tool, in which case we could just tell users to use the housekeeping interval to tune the wss reset interval.  But if it is important to support different collection and rss reset intervals, then we would have to add this config.

Thinking about it more, I do agree that basing this off the HK interval is probably better than a separate timer.  We would need to reimplement jitter to prevent spikes, and it is good to be able to collect immediately before resetting, and NOT collect right after a reset.

cmd/cadvisor.go
dashpole: That works for me, thanks!

manager/manager.go
dashpole: Yes, that would be good.

manager/manager.go
katarzyna-z: It is useful to observe how reference bytes growth in time and observe if we get only refernced bytes or it is a working set size (memory needed to keep working),  see: 
image

Summing up, I think that it is reasonable to introduce a new parameter which depends on house keeping interval.

docs/storage/prometheus.md
katarzyna-z: @ppalucki I know that you've made some experiments, please share you observations ;)

katarzyna-z:
@dashpole It's now ready for review.

docs/storage/prometheus.md
dashpole: That works for me.  Thanks for checking

manager/manager.go
dashpole: sgtm

cmd/cadvisor.go
```diff
@@ -87,6 +87,7 @@ var (
 		container.ProcessSchedulerMetrics:        struct{}{},
 		container.ProcessMetrics:                 struct{}{},
 		container.HugetlbUsageMetrics:            struct{}{},
+		container.ReferencedMetric:               struct{}{},
```
dashpole, line 90: nit: even though we only have one metric, lets still use the plural: `ReferencedMetrics`.

dashpole, line 90: WDYT about `ReferencedMemoryMetrics` instead?  Just Referenced seems a little generic...

dashpole: and similarly `referenced_memory` for the flag string

container/libcontainer/handler.go
```diff
@@ -81,6 +92,19 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
...
 
+	if h.includedMetrics.Has(container.ReferencedMetric) {
+		h.cycles++
+		pids, err := h.cgroupManager.GetPids()
+		if err != nil {
+			klog.V(4).Infof("Could not get PIDs for container %d: %v", h.pid, err)
+		} else {
+			stats.Referenced, err = referencedBytesStat(pids, h.cycles, *referencedResetInterval)
+			if err != nil {
+				klog.V(4).Infof("Unable to get working set size: %v", err)
```
dashpole, line 103: s/working set size/referenced memory

container/libcontainer/handler.go
```diff
@@ -318,6 +342,92 @@ func schedulerStatsFromProcs(rootFs string, pids []int, pidMetricsCache map[int]
...
+	}
+
+	err = clearReferencedBytes(pids, cycles, resetInterval)
+	if err != nil {
+		return uint64(0), err
+	}
+	return referencedKBytes * 1024, nil
+}
+
+func getReferencedBytes(pids []int) (uint64, error) {
```
dashpole, line 360: getReferencedKBytes?

container/libcontainer/handler.go
```diff
@@ -318,6 +342,92 @@ func schedulerStatsFromProcs(rootFs string, pids []int, pidMetricsCache map[int]
...
+	referencedKBytes := uint64(0)
+	readSmapsContent := false
+	foundMatch := false
+	for _, pid := range pids {
+		smapsFilePath := fmt.Sprintf(smapsFilePathPattern, pid)
+		smapsContent, err := ioutil.ReadFile(smapsFilePath)
+		if err != nil {
+			klog.V(3).Infof("Cannot read %s file, err: %s", smapsFilePath, err)
```
dashpole, line 368: For logging that is expected in some scenarios, lets lower to V(5)

container/libcontainer/handler.go
```diff
@@ -318,6 +342,92 @@ func schedulerStatsFromProcs(rootFs string, pids []int, pidMetricsCache map[int]
...
+		if !readSmapsContent {
+			klog.Warningf("Cannot read smaps files for any PID from %s", "CONTAINER")
+		} else if !foundMatch {
+			klog.Warningf("Not found any information about referenced bytes in smaps files for any PID from %s", "CONTAINER")
+		}
+	}
+	return referencedKBytes, nil
+}
+
+func clearReferencedBytes(pids []int, cycles uint64, resetInterval uint64) error {
```
dashpole, line 405: Just to make sure I understand, if clearReferencedBytes errors, we will wait another `resetInterval` cycles before attempting to reset again, right?  If so, is that better than always trying to clear after an error?

katarzyna-z: You're right, if clearReferencedBytes return error, we will wait another `resetInterval` but error occurs only if there is a serious issue in system (problem with writing into existing file or problem with closing previously opened file). I don't think that is a good idea to force clearing in case of errors.
User can easily observe if referenced bytes were cleared or not, seeing value of referenced bytes.

Since your last review I've introduced an option to switch off clearing of referenced bytes (setting `referenced_reset_interval` to 0). It's documented [here](https://github.com/katarzyna-z/cadvisor/blob/7ab5e27909c5b72363641ed6e683be42488e0365/container/libcontainer/handler.go#L45). During experiments it's more user friendly to set 0 than very long reset interval if it's desired to observe referenced bytes in longer period.

PR #2499, opened 2020-04-17
Go version must be passed to docker

iwankgb:
Instead of installing requested Go version in a VM we should use proper version of container image.


iwankgb:
@dims you may want to verify it.

PR #2503, opened 2020-04-20
golanci-lint based cleanup

iwankgb:
Is something like this useful? Majority of fixes is about not checked errors (in and out of tests). After applying more complex ones it should be possible to use `golangci-lint` and I think it could benefit the project.

### DIFF
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -107,11 +107,8 @@ func getProcessNamePid(line string, currentOomInstance *OomInstance) (bool, erro
 
 // uses regex to see if line is the start of a kernel oom log
 func checkIfStartOfOomMessages(line string) bool {
-	potential_oom_start := firstLineRegexp.MatchString(line)
-	if potential_oom_start {
-		return true
-	}
-	return false
+	potentialOomStart := firstLineRegexp.MatchString(line)
+	return potentialOomStart
 }
 
 // StreamOoms writes to a provided a stream of OomInstance objects representing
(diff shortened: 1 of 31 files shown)


utils/oomparser/oomparser.go
```diff
@@ -107,11 +107,8 @@ func getProcessNamePid(line string, currentOomInstance *OomInstance) (bool, erro
 
 // uses regex to see if line is the start of a kernel oom log
 func checkIfStartOfOomMessages(line string) bool {
-	potential_oom_start := firstLineRegexp.MatchString(line)
-	if potential_oom_start {
-		return true
-	}
-	return false
+	potentialOomStart := firstLineRegexp.MatchString(line)
```
dashpole, line 110: just `return firstLineRegexp.MatchString(line)`?

iwankgb:
`golint` is one of supported linters. Besides this it is capable of detecting unchecked errors, deadcode, unused properties/functions/variables/struts and many more (it looks like about 30 linters are supported).
I was thinking about using default set of linters for starters and idea of adding `golint` to this set sounds really good. I don't think it makes sense to look for more at the moment as some tuning will be required and the community should choose what is important and what is not.

dashpole:
That sounds reasonable.

iwankgb:
At the moment some check from default suite fail - fixing them seem to be non-trivial and I wanted to keep this PR as uncomplicated as possible.

dashpole:
Ok, lets iterate then.

PR #2512, opened 2020-04-23
 Using unexported return type from exported function shall be forbidden

iwankgb:
I think we should be rather safe. I haven't check it (I don't really know how) but I don't think that any project would rely on unexported struct properties.

### DIFF
--- a/events/handler_test.go
+++ b/events/handler_test.go
@@ -48,7 +48,8 @@ func initializeScenario(t *testing.T) (*events, *Request, *info.Event, *info.Eve
 	fakeEvent := makeEvent(createOldTime(t), "/")
 	fakeEvent2 := makeEvent(time.Now(), "/")
 
-	return NewEventManager(DefaultStoragePolicy()), NewRequest(), fakeEvent, fakeEvent2
+	manager := NewEventManager(DefaultStoragePolicy())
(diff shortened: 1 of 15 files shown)


events/handler_test.go
```diff
@@ -48,7 +48,8 @@ func initializeScenario(t *testing.T) (*events, *Request, *info.Event, *info.Eve
 	fakeEvent := makeEvent(createOldTime(t), "/")
 	fakeEvent2 := makeEvent(time.Now(), "/")
 
-	return NewEventManager(DefaultStoragePolicy()), NewRequest(), fakeEvent, fakeEvent2
+	manager := NewEventManager(DefaultStoragePolicy())
+	return manager.(*events), NewRequest(), fakeEvent, fakeEvent2
```
dashpole, line 52: can we change the return type to EventManager instead of casting, or would that break the test?

iwankgb: I would have broken it as one of the tests relied on internal structures of `events` instead of using interface methods. Fixed :)

dashpole:
I0423 21:02:33.744] >> checking go formatting
I0423 21:02:35.174] The following files are not properly formatted:
I0423 21:02:35.174] events/handler_test.go

PR #2513, opened 2020-04-23
Getting rid of depracated grpc function calls

iwankgb:


iwankgb:
@dashpole I tested it locally against containerd instance and all the stats seemed to be available as expected.

container/containerd/client.go
```diff
@@ -62,29 +64,37 @@ func Client(address, namespace string) (ContainerdClient, error) {
...
 			grpc.WithUnaryInterceptor(unary),
 			grpc.WithStreamInterceptor(stream),
 		)
 
-		conn, err := grpc.Dial(dialer.DialAddress(address), gopts...)
+		ctx, cancel := context.WithTimeout(context.Background(), connectionTimeout)
+		conn, err := grpc.DialContext(ctx, dialer.DialAddress(address), gopts...)
 		if err != nil {
 			retErr = err
+			cancel()
```
dashpole, line 89: should we just `defer cancel()` after the `context.WithTimeout(...` call?

iwankgb: Sounds legit.

container/containerd/client.go
```diff
@@ -48,6 +49,7 @@ var ctrdClient ContainerdClient = nil
 
 const (
 	maxBackoffDelay   = 3 * time.Second
+	baseBackoffDelay  = 100 * time.Millisecond
```
dashpole, line 52: Where did this number come from?

dashpole:
Just a few nits...

container/containerd/client.go
iwankgb: I had to set it to some meaningful value and I thought that 100ms might make sense.

iwankgb:
@googlebot I consent.

iwankgb:
@dashpole my comment at googlebot seems to be ignored.

dashpole:
It looks like it thinks the first three commits came from a different account

iwankgb:
I guess that it might be caused be me using another email address. All commits use same address now :)

iwankgb:
@dashpole the last remaining comment is yours about value of `baseBackoffDelay`. I decided to set to 100ms but I'm open to change it to any other value. My gut feeling is that this one isn't that bad. Take a look at [the docs](https://github.com/grpc/grpc-go/blob/master/backoff/backoff.go#L30), please.

dashpole:
sorry, it just fell off my radar

PR #2516, opened 2020-04-25
add zipkin

MarcelHillmann:
As operator I would like to see how the Prometheus netwerk is working.

To improve this target, I added the opentracing-go framework, and implemented a Zipkin Observer.

Thx
Marcel

Signed-off-by: Marcel <marcel@mahillmann.de>


cmd/cadvisor.go
```diff
@@ -192,6 +204,10 @@ func main() {
 	// Register Prometheus collector to gather information about containers, Go runtime, processes, and machine
 	cadvisorhttp.RegisterPrometheusHandler(mux, resourceManager, *prometheusEndpoint, containerLabelFunc, includedMetrics)
 
+	muxHandle := nethttp.Middleware(opentracing.GlobalTracer(), mux, nethttp.OperationNameFunc(func(r *http.Request) string {
```
dashpole, line 207: What would the operation name be if we didn't rename it here?

cmd/cadvisor.go
```diff
@@ -145,6 +153,10 @@ func main() {
 	defer klog.Flush()
 	flag.Parse()
 
+	if os.Getenv("ZIPKIN") != "" {
```
dashpole, line 156: Please use flags instead of env vars for configuration.  This applies here and elsewhere.

cmd/cadvisor.go
dashpole: nit: we should be able to assign this back to mux so we don't have to modify variable names below.

MarcelHillmann: 1. the default operation name is `"HTTP " + http.Method` ex: `HTTP GET`. That is a bit useless.
2. the mux variable is alllocated as `ServerMux` and a reassignement is not possible. Because the `nethttp.Middleware` returns a `http.Handle`

cmd/cadvisor.go
MarcelHillmann: I will refactore

MarcelHillmann:
> I don't think a direct integration with zipkin is what we want here. Our experience with metric storage has been that adding direct integrations with metric backends (see https://github.com/google/cadvisor/tree/master/cmd/internal/storage) is that they are hard to maintain. I would strongly prefer using the opentelemetry (OTLP) format, and running an OT Collector to route traces to your backend of choice.

Sorry, I'm a bit lost in translation.
The part with OTLP sounds interessting and I checked the HP, I could change my implementation.
I have no clue what you meen with the metric backend.

Sorry Marcel

dashpole:
> Sorry, I'm a bit lost in translation.
The part with OTLP sounds interessting and I checked the HP, I could change my implementation.
I have no clue what you meen with the metric backend.

The point is that cAdvisor maintains ~10 different integrations with metrics backends.  For tracing, I would like to support only a single exporter.  Using OTLP would allow users to route to any trace backend, and allows cAdvisor to only support a single trace exporter.

MarcelHillmann:
Thanks, I will create a new PR with otlp.

Thx Marcel

PR #2518, opened 2020-04-27
Remove go-cmp from from cache comparison

katarzyna-z:

As indicated [here](https://godoc.org/github.com/google/go-cmp/cmp) go-cmp should not be used in production environment because  it may panic if it cannot compare the values.


dashpole:
Can we remove it from go modules as well?

katarzyna-z:
Now, go-cmp is an indirect dependency.

katarzyna-z:
I see that integration tests failed because docker return version as "Unknown" : `Expect "Unknown" to match "\d+\.\d+\.\d+"`. The same issue is for #2520 .

PR #2519, opened 2020-04-27
Introduce new machine metrics:

katarzyna-z:
- machine_cpu_cache_capacity_bytes
- machine_node_hugepages_count
- machine_node_memory_capacity_bytes
- machine_nvm_avg_power_budget_watts
- machine_thread_siblings_count

Introduce timestamp for machine related information (it is possible to reconfigure huge pages without restart so it seems to be useful to have timestamp for this information)

### DIFF
--- a/metrics/prometheus_machine.go
+++ b/metrics/prometheus_machine.go
@@ -15,6 +15,8 @@
 package metrics
 
 import (
+	"strconv"
+
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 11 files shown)


metrics/prometheus_machine.go
```diff
@@ -90,6 +99,42 @@ func NewPrometheusMachineCollector(i infoProvider) *PrometheusMachineCollector {
...
+				extraLabels: []string{prometheusNodeLabelName, prometheusCoreLabelName, prometheusTypeLabelName, prometheusLevelLabelName},
+				getValues: func(machineInfo *info.MachineInfo) metricValues {
+					return getCaches(machineInfo)
+				},
+			},
+			{
+				name:        "machine_thread_siblings_count",
+				help:        "Number of CPU thread siblings.",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{prometheusNodeLabelName, prometheusCoreLabelName, prometheusThreadLabelName},
```
dashpole, line 115: what is the cardinality of each of these?  This may end up being a very expensive metric...

dashpole:
If a timestamp isn't specified, the scrape time is used by prometheus as the timestamp.  We probably want to add timestamps wherever possible in all spec metrics.  IIRC, we update machine info every 5 minutes, so it would be helpful to add timestamps since collection and scrape times can differ substantially.

metrics/prometheus_machine.go
katarzyna-z: number of `machine_thread_siblings_count` metrics will be equal to number of CPU threads on the machine, on my machine 72...
Adding `machine_thread_siblings_count` metric I thought about exposing the most information from topology and at this moment I don't have better idea for it so I think we can get rid of this metric for now.

katarzyna-z:
I think that timestamp can be added to all machine metrics to avoid difference between collection and scrape time.

metrics/prometheus_machine.go
katarzyna-z: I see that similar case will be for `machine_cpu_cache_capacity_bytes`, on my machine I'll have 54 metrics in each collection. I would like to group `machine_cpu_cache_capacity_bytes`, `machine_node_hugepages_count`, `machine_node_memory_capacity_bytes` and `machine_thread_siblings_count` metrics as `cpu_topology` and make them disable on default. I think about adding `cpu_topology` as an option to `--diable-metrics` parameter. What do you think?

dashpole: Thats a great idea.

PR #2492, opened 2020-04-13
Make a copy of containerInfo in containerData#GetInfo

tedyu:
This addresses existing comment by making a copy:
```
        // Make a copy of the info for the user.
```

PR #2520, opened 2020-04-27
Simplify making copy

tedyu:
This PR simplifies making copy, introduced via #2491 and #2492


tedyu:
@dashpole 
Can you take a look ?

PR #2525, opened 2020-04-29
Add perf uncore events support.

Creatone:
It provides support for uncore perf events in which PMUs are not associated with a core but a whole socket.

### DIFF
--- a/docs/runtime_options.md
+++ b/docs/runtime_options.md
@@ -158,6 +158,106 @@ automatically.
 * `grouping` - in scenario when accounted for events are used to calculate derivative metrics, it is reasonable to
 measure them in transactional manner: all the events in a group must be accounted for in the same period of time. Keep
 in mind that it is impossible to group more events that there are counters available.
+* `uncore events` - events which can be counted by PMUs outside core.
+* `PMU` - Performance Monitoring Unit
+
+#### Getting config values
+Using perf tools:
+* Identify the event in `perf list` output. 
+* Execute command: `perf stat -I 5000 -vvv -e EVENT_NAME`
+* Find `perf_event_attr` section on `perf stat` output, copy config and type field to configuration file.
+
+```
+------------------------------------------------------------
+perf_event_attr:
+  type                             18
+  size                             112
+  config                           0x304
+  sample_type                      IDENTIFIER
+  read_format                      TOTAL_TIME_ENABLED|TOTAL_TIME_RUNNING
+  disabled                         1
+  inherit                          1
+  exclude_guest                    1
+------------------------------------------------------------
+```
+* Configuration file should look like: 
+```json
+{
+  "core": {
+    "events": [
+      ["EVENT_NAME"]
+    ],
+    "custom_events": [
+      {
+        "type": 18,
+        "config": [
+          "0x304"
+        ],
+        "name": "EVENT_NAME"
+      }
+    ]
+  },
+  "uncore": {
+    "events": [
+      ["EVENT_NAME"]
+    ],
+    "custom_events": [
+      {
+        "type": 18,
+        "config": [
+          "0x304"
+        ],
+        "name": "EVENT_NAME"
+      }
+    ]
+  }
+}
+```
+
+Config values can be also obtain from: 
+* [Intel® 64 and IA32 Architectures Performance Monitoring Events](https://software.intel.com/content/www/us/en/develop/download/intel-64-and-ia32-architectures-performance-monitoring-events.html)
+
+
+##### Uncore Events configuration
+Uncore Event name should be in form `PMU_PREFIX/event_name` where **PMU_PREFIX** mean
+that statistics would be counted on all PMUs with that prefix in name.
+
+Let's explain this by example: 
+
+```json
+{
+  "uncore": {
+    "events": [
+      ["uncore_imc/cas_count_read"],
+      ["uncore_imc_0/cas_count_write"],
+      ["cas_count_all"]
+    ],
+    "custom_events": [ 
+      {
+        "config": [
+          "0x304"
+        ],
+        "name": "uncore_imc_0/cas_count_write"
+      },
+      {
+        "type": 19,
+        "config": [
+          "0x304"
+        ],
+        "name": "cas_count_all"
+      }
+    ]
+  }
+}
+```
+
+- `uncore_imc/cas_count_read` - because of `uncore_imc` type and no entry in custom events,
+    it would be counted by **all** Integrated Memory Controller PMUs with config provided from libpfm package.
+    (using this function: https://man7.org/linux/man-pages/man3/pfm_get_os_event_encoding.3.html)
+
+- `uncore_imc_0/cas_count_write` - because of `uncore_imc_0` type and entry in custom events it would be counted by `uncore_imc_0` PMU with provided config.
+
+- `uncore_imc_1/cas_count_all` - because of entry in custom events with type field, event would be counted by PMU with **19** type and provided config.
 
 ### Further reading
 
@@ -165,16 +265,17 @@ in mind that it is impossible to group more events that there are counters avail
 * [Kernel Perf Wiki](https://perf.wiki.kernel.org/index.php/Main_Page)
 * `man perf_event_open`
 * [perf subsystem](https://github.com/torvalds/linux/tree/v5.6/kernel/events) in Linux kernel
+* [Uncore Performance Monitoring Reference Manuals](https://software.intel.com/content/www/us/en/develop/articles/intel-sdm.html#uncore)
 
 See example configuration below:
 ```json
 {
-  "events": [
-    ["instructions"],
-    ["instructions_retired"]
-  ],
-  "custom_events": [
-    [
+  "core": {
+    "events": [
+      ["instructions"],
+      ["instructions_retired"]
+    ],
+    "custom_events": [
       {
         "type": 4,
         "config": [
@@ -183,7 +284,20 @@ See example configuration below:
         "name": "instructions_retired"
       }
     ]
-  ]
+  },
+  "uncore": {
+    "events": [
+      ["uncore_imc/cas_count_read"]
+    ],
+    "custom_events": [
+      {
+        "config": [
+          "0xc04"
+        ],
+        "name": "uncore_imc/cas_count_read"
+      }
+    ]
+  }
 }
 ```
 
@@ -194,6 +308,9 @@ interface that majority of users will rely on.
 * `instructions_retired` will be measured as non-grouped event and is specified using an advanced API that allows
 to specify any perf event available (some of them are not named and can't be specified with plain string). Event name 
 should be a human readable string that will become a metric name.
+* `cas_count_read` will be measured as uncore non-grouped event on all Integrated Memory Controllers Performance Monitoring Units because of unset `type` field and
+`uncore_imc` prefix.
+
 
 ## Storage driver specific instructions:
 
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -874,6 +874,32 @@ type ResctrlStats struct {
 	Cache           []CacheStats           `json:"cache,omitempty"`
 }
 
+// PerfUncoreStat represents value of a single monitored perf uncore event.
+type PerfUncoreStat struct {
+	// Indicates scaling ratio for an event: time_running/time_enabled
+	// (amount of time that event was being measured divided by
+	// amount of time that event was enabled for).
+	// value 1.0 indicates that no multiplexing occurred. Value close
+	// to 0 indicates that event was measured for short time and event's
+	// value might be inaccurate.
+	// See: https://lwn.net/Articles/324756/
+	ScalingRatio float64 `json:"scaling_ratio"`
+
+	// Value represents value of perf event retrieved from OS. It is
+	// normalized against ScalingRatio and takes multiplexing into
+	// consideration.
+	Value uint64 `json:"value"`
+
+	// Name is human readable name of an event.
+	Name string `json:"name"`
+
+	// Socket that perf event was measured on.
+	Socket int `json:"socket"`
+
+	// PMU is Performance Monitoring Unit which collected these stats.
+	PMU string `json:"pmu"`
+}
+
 type UlimitSpec struct {
 	Name      string `json:"name"`
 	SoftLimit int64  `json:"soft_limit"`
@@ -925,6 +951,10 @@ type ContainerStats struct {
 
 	// Statistics originating from perf events
 	PerfStats []PerfStat `json:"perf_stats,omitempty"`
+
+	// Statistics originating from perf uncore events.
+	// Applies only for root container.
+	PerfUncoreStats []PerfUncoreStat `json:"perf_uncore_stats,omitempty"`
 
 	// Referenced memory
 	ReferencedMemory uint64 `json:"referenced_memory,omitempty"`
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -71,6 +71,19 @@ func (n *Node) FindCore(id int) (bool, int) {
 	return false, -1
 }
 
+// FindCoreByThread returns bool if found Core with same thread as provided and it's index in Node Core array.
+// If it's not found, returns false and -1.
+func (n *Node) FindCoreByThread(thread int) (bool, int) {
+	for i, n := range n.Cores {
+		for _, t := range n.Threads {
+			if t == thread {
+				return true, i
+			}
+		}
+	}
+	return false, -1
+}
+
 func (n *Node) AddThread(thread int, core int) {
 	var coreIdx int
 	if core == -1 {
--- a/info/v2/conversion.go
+++ b/info/v2/conversion.go
@@ -155,6 +155,9 @@ func ContainerStatsFromV1(containerName string, spec *v1.ContainerSpec, stats []
 		if len(val.PerfStats) > 0 {
 			stat.PerfStats = val.PerfStats
 		}
+		if len(val.PerfUncoreStats) > 0 {
+			stat.PerfUncoreStats = val.PerfUncoreStats
+		}
 		if len(val.Resctrl.MemoryBandwidth) > 0 || len(val.Resctrl.Cache) > 0 {
 			stat.Resctrl = val.Resctrl
 		}
@@ -212,6 +215,9 @@ func DeprecatedStatsFromV1(cont *v1.ContainerInfo) []DeprecatedContainerStats {
 		}
 		if len(val.PerfStats) > 0 {
 			stat.PerfStats = val.PerfStats
+		}
+		if len(val.PerfUncoreStats) > 0 {
+			stat.PerfUncoreStats = val.PerfUncoreStats
 		}
 		if len(val.Resctrl.MemoryBandwidth) > 0 || len(val.Resctrl.Cache) > 0 {
 			stat.Resctrl = val.Resctrl
--- a/machine/info.go
+++ b/machine/info.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/nvm"
@@ -30,8 +32,6 @@ import (
 	"github.com/google/cadvisor/utils/sysinfo"
 
 	"k8s.io/klog/v2"
-
-	"golang.org/x/sys/unix"
 )
 
 const hugepagesDirectory = "/sys/kernel/mm/hugepages/"
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -212,7 +212,7 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig
 	newManager.machineInfo = *machineInfo
 	klog.V(1).Infof("Machine: %+v", newManager.machineInfo)
 
-	newManager.perfManager, err = perf.NewManager(perfEventsFile, machineInfo.NumCores)
+	newManager.perfManager, err = perf.NewManager(perfEventsFile, machineInfo.NumCores, machineInfo.Topology)
 	if err != nil {
 		return nil, err
 	}
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1545,11 +1545,11 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
-	if c.includedMetrics.Has(container.PerfMetrics) {
+	if includedMetrics.Has(container.PerfMetrics) {
 		c.containerMetrics = append(c.containerMetrics, []containerMetric{
 			{
-				name:        "container_perf_metric",
-				help:        "Perf event metric",
+				name:        "container_perf_events_total",
+				help:        "Perf event metric.",
 				valueType:   prometheus.CounterValue,
 				extraLabels: []string{"cpu", "event"},
 				getValues: func(s *info.ContainerStats) metricValues {
@@ -1565,8 +1565,8 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 				},
 			},
 			{
-				name:        "container_perf_metric_scaling_ratio",
-				help:        "Perf event metric scaling ratio",
+				name:        "container_perf_events_scaling_ratio",
+				help:        "Perf event metric scaling ratio.",
 				valueType:   prometheus.GaugeValue,
 				extraLabels: []string{"cpu", "event"},
 				getValues: func(s *info.ContainerStats) metricValues {
@@ -1575,6 +1575,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 						values = append(values, metricValue{
 							value:     metric.ScalingRatio,
 							labels:    []string{strconv.Itoa(metric.Cpu), metric.Name},
+							timestamp: s.Timestamp,
+						})
+					}
+					return values
+				},
+			},
+			{
+				name:        "container_perf_uncore_events_total",
+				help:        "Perf uncore event metric.",
+				valueType:   prometheus.CounterValue,
+				extraLabels: []string{"socket", "event", "pmu"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					values := make(metricValues, 0, len(s.PerfUncoreStats))
+					for _, metric := range s.PerfUncoreStats {
+						values = append(values, metricValue{
+							value:     float64(metric.Value),
+							labels:    []string{strconv.Itoa(metric.Socket), metric.Name, metric.PMU},
+							timestamp: s.Timestamp,
+						})
+					}
+					return values
+				},
+			},
+			{
+				name:        "container_perf_uncore_events_scaling_ratio",
+				help:        "Perf uncore event metric scaling ratio.",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"socket", "event", "pmu"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					values := make(metricValues, 0, len(s.PerfUncoreStats))
+					for _, metric := range s.PerfUncoreStats {
+						values = append(values, metricValue{
+							value:     metric.ScalingRatio,
+							labels:    []string{strconv.Itoa(metric.Socket), metric.Name, metric.PMU},
 							timestamp: s.Timestamp,
 						})
 					}
--- a/metrics/prometheus_fake.go
+++ b/metrics/prometheus_fake.go
@@ -648,6 +648,22 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 							Cpu:          1,
 						},
 					},
+					PerfUncoreStats: []info.PerfUncoreStat{
+						{
+							ScalingRatio: 1.0,
+							Value:        1231231512.0,
+							Name:         "cas_count_read",
+							Socket:       0,
+							PMU:          "uncore_imc_0",
+						},
+						{
+							ScalingRatio: 1.0,
+							Value:        1111231331.0,
+							Name:         "cas_count_read",
+							Socket:       1,
+							PMU:          "uncore_imc_0",
+						},
+					},
 					ReferencedMemory: 1234,
 					Resctrl: info.ResctrlStats{
 						MemoryBandwidth: []info.MemoryBandwidthStats{
--- a/metrics/testdata/prometheus_metrics
+++ b/metrics/testdata/prometheus_metrics
@@ -327,18 +327,26 @@ container_network_udp_usage_total{container_env_foo_env="prod",container_label_f
 container_network_udp_usage_total{container_env_foo_env="prod",container_label_foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",udp_state="listen",zone_name="hello"} 0 1395066363000
 container_network_udp_usage_total{container_env_foo_env="prod",container_label_foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",udp_state="rxqueued",zone_name="hello"} 0 1395066363000
 container_network_udp_usage_total{container_env_foo_env="prod",container_label_foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",udp_state="txqueued",zone_name="hello"} 0 1395066363000
-# HELP container_perf_metric Perf event metric
-# TYPE container_perf_metric counter
-container_perf_metric{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 123 1395066363000
-container_perf_metric{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 321 1395066363000
-container_perf_metric{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 456 1395066363000
-container_perf_metric{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 789 1395066363000
-# HELP container_perf_metric_scaling_ratio Perf event metric scaling ratio
-# TYPE container_perf_metric_scaling_ratio gauge
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.66666666666 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.5 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.33333333333 1395066363000
+# HELP container_perf_events_total Perf event metric.
+# TYPE container_perf_events_total counter
+container_perf_events_total{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 123 1395066363000
+container_perf_events_total{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 321 1395066363000
+container_perf_events_total{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 456 1395066363000
+container_perf_events_total{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 789 1395066363000
+# HELP container_perf_events_scaling_ratio Perf event metric scaling ratio.
+# TYPE container_perf_events_scaling_ratio gauge
+container_perf_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1 1395066363000
+container_perf_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.66666666666 1395066363000
+container_perf_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.5 1395066363000
+container_perf_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.33333333333 1395066363000
+# HELP container_perf_uncore_events_total Perf uncore event metric.
+# TYPE container_perf_uncore_events_total counter
+container_perf_uncore_events_total{container_env_foo_env="prod",container_label_foo_label="bar",event="cas_count_read",id="testcontainer",image="test",name="testcontaineralias",pmu="uncore_imc_0",socket="0",zone_name="hello"} 1.231231512e+09 1395066363000
+container_perf_uncore_events_total{container_env_foo_env="prod",container_label_foo_label="bar",event="cas_count_read",id="testcontainer",image="test",name="testcontaineralias",pmu="uncore_imc_0",socket="1",zone_name="hello"} 1.111231331e+09 1395066363000
+# HELP container_perf_uncore_events_scaling_ratio Perf uncore event metric scaling ratio.
+# TYPE container_perf_uncore_events_scaling_ratio gauge
+container_perf_uncore_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",event="cas_count_read",id="testcontainer",image="test",name="testcontaineralias",pmu="uncore_imc_0",socket="0",zone_name="hello"} 1 1395066363000
+container_perf_uncore_events_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",event="cas_count_read",id="testcontainer",image="test",name="testcontaineralias",pmu="uncore_imc_0",socket="1",zone_name="hello"} 1 1395066363000
 # HELP container_processes Number of processes running inside the container.
 # TYPE container_processes gauge
 container_processes{container_env_foo_env="prod",container_label_foo_label="bar",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1 1395066363000
--- a/perf/collector_libpfm.go
+++ b/perf/collector_libpfm.go
@@ -31,18 +31,21 @@ import (
 	"sync"
 	"unsafe"
 
-	info "github.com/google/cadvisor/info/v1"
 	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
+
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/stats"
 )
 
 type collector struct {
 	cgroupPath         string
-	events             Events
+	events             PerfEvents
 	cpuFiles           map[string]map[int]readerCloser
 	cpuFilesLock       sync.Mutex
 	numCores           int
 	eventToCustomEvent map[Event]*CustomEvent
+	uncore             stats.Collector
 }
 
 var (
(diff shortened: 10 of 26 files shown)


perf/uncore.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if len(cpus) == 0 {
+		return nil, fmt.Errorf("there is no PMU with event type: %#v", gotType)
+	}
+
+	return cpus, nil
+}
+
+func setupUncore() error {
+	err := filepath.Walk("/sys/devices", func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
Creatone: For me, these strings has better readability.

perf/collector_libpfm.go
```diff
@@ -154,25 +178,104 @@ func (c *collector) registerEvent(config *unix.PerfEventAttr, name string, cgrou
...
-	klog.V(5).Infof("Setting up non-grouped perf event %s", name)
+	setAttributes(config)
+	klog.V(5).Infof("perf_event_attr struct prepared: %#v", config)
+	return config
+}
+
+func (c *collector) setupRawNonGroupedUncoreMemoryController(event *CustomEvent, cgroup int) error {
+	klog.V(5).Infof("Setting up non-grouped raw perf uncore memory controller event %#v", event)
+
+	if event.Type == perfTypeHardware {
```
iwankgb, line 231: `event.Type` 0 indicates on-core events, doesn't it?

perf/collector_libpfm.go
```diff
@@ -154,25 +178,104 @@ func (c *collector) registerEvent(config *unix.PerfEventAttr, name string, cgrou
...
+		if err != nil {
+			return err
+		}
+		return c.registerUncoreMemoryControllerEvent(config, string(event.Name), cgroup, cpus)
+	}
+}
+
+func (c *collector) registerUncoreMemoryControllerEvent(config *unix.PerfEventAttr, name string, cgroup int, cpus []uint32) error {
+	for _, cpu := range cpus {
+		pid, groupFd, flags := cgroup, -1, unix.PERF_FLAG_FD_CLOEXEC|unix.PERF_FLAG_PID_CGROUP
```
iwankgb, line 259: What about renaming `cgroup` argument to `pid`?

perf/collector_libpfm.go
```diff
@@ -154,25 +178,104 @@ func (c *collector) registerEvent(config *unix.PerfEventAttr, name string, cgrou
...
 		}
 
-		c.addEventFile(name, cpu, perfFile)
+		c.addEventFile(name, config.Type, cpu, perfFile)
 	}
 	return nil
 }
 
-func (c *collector) addEventFile(name string, cpu int, perfFile *os.File) {
+func (c *collector) addEventFile(name string, pmuType uint32, cpu int, perfFile *os.File) {
```
iwankgb, line 186: `pmu` will be better argument name, I think.

perf/collector_libpfm.go
```diff
@@ -154,25 +178,104 @@ func (c *collector) registerEvent(config *unix.PerfEventAttr, name string, cgrou
...
+			return fmt.Errorf("unable to create os.File from file descriptor %#v", fd)
+		}
+
+		c.addEventFile(name, config.Type, int(cpu), perfFile)
+	}
+
+	return nil
+}
+
+func getPerfEventAttr(name string) (*unix.PerfEventAttr, error) {
```
iwankgb, line 275: You allocate memory in line 280, then return Go pointer allocated in line 284 but in line 281 you defer call to `free()`. `perfEventAttr` becomes a [dangling pointer](https://en.wikipedia.org/wiki/Dangling_pointer). `free()` must be called after `perfEventAttr` is no longer used.

perf/uncore.go
```diff
@@ -0,0 +1,99 @@
...
+func setupUncore() error {
+	cpumaskRegexp := regexp.MustCompile("[-,\n]")
+	err := filepath.Walk("/sys/devices", func(path string, info os.FileInfo, err error) error {
+		if info.IsDir() {
+			if strings.Contains(info.Name(), "uncore_imc") {
+				buf, err := ioutil.ReadFile(filepath.Join(path, "type"))
+				if err != nil {
+					return err
+				}
+				typeString := strings.TrimSpace(string(buf))
```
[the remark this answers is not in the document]
Creatone: Yes, https://golang.org/ref/spec#Types

dashpole:
Let me know when this is no longer a WIP and i'll start reviewing

Creatone:
@dashpole @katarzyna-z @iwankgb @piowag

machine/info.go
```diff
@@ -17,6 +17,7 @@ package machine
 import (
 	"bytes"
 	"flag"
+	"github.com/google/cadvisor/perf"
```
katarzyna-z, line 20: please add this import into separate group of imports

machine/machine.go
```diff
@@ -313,3 +313,22 @@ func getNumCores() int {
 
 	return numCPU
 }
+
+func GetPerfUncore() []info.PerfStat {
```
katarzyna-z, line 317: Why this function returns const values?

metrics/prometheus_machine.go
```diff
@@ -192,6 +192,29 @@ func NewPrometheusMachineCollector(i infoProvider, includedMetrics container.Met
...
+				name:        "machine_perf_uncore_metric",
+				help:        "Perf uncore event metric.",
+				valueType:   prometheus.CounterValue,
+				extraLabels: []string{"cpu", "event", "type"},
+				getValues: func(machineInfo *info.MachineInfo) metricValues {
+					values := make(metricValues, 0, len(machineInfo.PerfUncoreStats))
+					for _, metric := range machineInfo.PerfUncoreStats {
+						values = append(values, metricValue{
+							value:     float64(metric.Value),
+							labels:    []string{strconv.Itoa(metric.Cpu), metric.Name, strconv.FormatUint(uint64(metric.Type), 10)},
```
katarzyna-z, line 208: Is possible to determined cpu assigned for metric for uncore perf metrics?

Creatone:
#2488

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,360 @@
...
+// #cgo CFLAGS: -I/usr/include
+// #cgo LDFLAGS: -lpfm
+// #include <perfmon/pfmlib.h>
+// #include <stdlib.h>
+import "C"
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"golang.org/x/sys/unix"
```
katarzyna-z, line 29: Could you check if grouping of imports is correct here using e.g. goimports?

perf/uncore_libpfm_test.go
```diff
@@ -0,0 +1,111 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Uncore perf events logic tests.
+package perf
+
+import (
+	"github.com/stretchr/testify/assert"
```
katarzyna-z, line 21: please correct grouping of imports

manager/manager.go
```diff
@@ -194,7 +194,12 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig
 		rawContainerCgroupPathPrefixWhiteList: rawContainerCgroupPathPrefixWhiteList,
 	}
 
-	machineInfo, err := machine.Info(sysfs, fsInfo, inHostNamespace)
+	newManager.perfUncoreCollector, err = perf.NewUncoreCollector(perfEventsFile)
```
katarzyna-z, line 197: imo, it'll be good if cAdvisor will log that there is a problem with perf uncore configuration but other machine information will be exposed, could you check this behavior?

perf/collector_libpfm.go
```diff
@@ -214,6 +206,34 @@ func createPerfEventAttr(event CustomEvent) *unix.PerfEventAttr {
...
+
+	perfEventAttr := (*unix.PerfEventAttr)(perfEventAttrMemory)
+	fstr := C.CString("")
+	event.fstr = unsafe.Pointer(fstr)
+	event.attr = perfEventAttrMemory
+	event.size = C.ulong(unsafe.Sizeof(event))
+
+	cSafeName := C.CString(name)
+
+	pErr := C.pfm_get_os_event_encoding(cSafeName, C.PFM_PLM0|C.PFM_PLM3, C.PFM_OS_PERF_EVENT_EXT, unsafe.Pointer(&event))
```
dashpole, line 225: Why do we need to switch to EXT?

Creatone: We don't need to. I forgot to change it back.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+	pmuTypeFilename    = "type"
+	pmuCpumaskFilename = "cpumask"
+	systemDevicesPath  = "/sys/devices"
+)
+
+var imcPMUs uncorePMUs
+
+func getIMCPMUs() (uncorePMUs, error) {
+	if len(imcPMUs) == 0 {
+		var err error
```
iwankgb, line 61: You don't need to declare the variable, use `:=` in the line below.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+	configFile, err := os.Open(configPath)
+	if err != nil {
+		formatedError := fmt.Errorf("unable to read configuration file %q: %q", configPath, err)
+		klog.Warningf("Perf uncore metrics will not be available: %s", formatedError)
+		return &NoopUncoreCollector{}
+	}
+
+	config, err := parseConfig(configFile)
+	if err != nil {
+		formatedError := fmt.Errorf("unable to read configuration file %q: %q", configPath, err)
```
iwankgb, line 152:
```suggestion
		formatedError := fmt.Errorf("unable to parse configuration file %q: %q", configPath, err)
```

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+const (
+	imcDirectoryName   = "uncore_imc"
+	pmuTypeFilename    = "type"
+	pmuCpumaskFilename = "cpumask"
+	systemDevicesPath  = "/sys/devices"
+)
+
+var imcPMUs uncorePMUs
+
+func getIMCPMUs() (uncorePMUs, error) {
```
iwankgb, line 59: Consider renaming the function to: `getMemoryControllerPMUs()`. It will be much more understandable :)

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+	for _, pmu := range *pmus {
+		if pmu.typeOf == gotType {
+			return pmu.cpus, nil
+		}
+	}
+
+	return nil, fmt.Errorf("there is no pmu with event type: %#v", gotType)
+}
+
+func getUncoreIMCPMUs(devicesPath string) (uncorePMUs, error) {
```
iwankgb, line 83: Consider renaming to `getUncoreMemoryControllerPMUs()`.

perf/collector_libpfm.go
```diff
@@ -39,7 +39,7 @@ import (
 type collector struct {
 	cgroupPath         string
 	events             Events
-	cpuFiles           map[string]map[int]readerCloser
+	cpuFiles           map[string]map[uint32]map[int]readerCloser
```
iwankgb, line 42: Is this change necessary? You created another struct that handles uncore perf events.

perf/collector_libpfm.go
```diff
@@ -214,6 +206,34 @@ func createPerfEventAttr(event CustomEvent) *unix.PerfEventAttr {
 	return config
 }
 
+func getPerfEventAttr(name string) (*unix.PerfEventAttr, error) {
```
iwankgb, line 209: It seems to me that this is the only meaningful change in this file now. Everything else can be undone (see my comment to `collector.cpuFiles`).

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+)
+
+type pmu struct {
+	name   string
+	typeOf uint32
+	cpus   []uint32
+}
+
+const (
+	imcDirectoryName   = "uncore_imc"
```
iwankgb, line 51: s/imcDirectoryName/memoryControllerDir/

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+	pmuTypeFilename    = "type"
+	pmuCpumaskFilename = "cpumask"
+	systemDevicesPath  = "/sys/devices"
+)
+
+var imcPMUs uncorePMUs
+
+func getIMCPMUs() (uncorePMUs, error) {
+	if len(imcPMUs) == 0 {
```
iwankgb, line 60: You can [improve this check](https://tour.golang.org/moretypes/12). 0-length slice can be perfectly valid.

perf/uncore_libpfm.go
iwankgb: I would appreciate if this function was tested :)

perf/noop.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package perf
+
+import (
+	info "github.com/google/cadvisor/info/v1"
+	"github.com/google/cadvisor/stats"
+)
+
+type NoopUncoreCollector struct {
```
iwankgb, line 22: Consider `stats.NoopCollector` struct instead.
Especially if, as @dashpole suggested, stats are to be visible on root container.

perf/types.go
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package perf
+
+import v1 "github.com/google/cadvisor/info/v1"
+
+type UncoreCollector interface {
```
iwankgb, line 19: Consider using `stats.Collector` interface.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func parseUncoreEvents(events Events) (UncoreEvents, map[Event]*CustomEvent) {
```
iwankgb, line 203: Take a look at `perf.mapEventsToCustomEvents`. I think that this function can be simplified.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+	eventToCustomEvent := map[Event]*CustomEvent{}
+	for _, uncoreEvent := range events.UncoreEvents.MemoryController {
+		for _, customEvent := range events.CustomEvents {
+			if uncoreEvent[0] == customEvent.Name {
+				eventToCustomEvent[customEvent.Name] = &customEvent
+			}
+		}
+	}
+	return events.UncoreEvents, eventToCustomEvent
```
iwankgb, line 212: There is no reason to return `events.UncoreEvents` as `events` were passed as argument and must be available in callers scope.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+				if perfData.TimeEnabled != 0 {
+					scalingRatio = float64(perfData.TimeRunning) / float64(perfData.TimeEnabled)
+				}
+				stat := info.PerfStat{
+					Value:        uint64(float64(perfData.Value) / scalingRatio),
+					Name:         name,
+					ScalingRatio: scalingRatio,
+					Cpu:          cpu,
+					Type:         pmu,
+				}
```
iwankgb, line 266: Could you reuse some code from `collector.UpdateStats()`?

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+func (c *uncoreCollector) setupRawNonGroupedUncoreMemoryController(event *CustomEvent) error {
+	klog.V(5).Infof("Setting up non-grouped raw perf uncore memory controller event %#v", event)
+
+	pmus, err := getIMCPMUs()
+	if err != nil {
+		return err
+	}
+
+	if event.Type == 0 {
+		// PMU isn't set. Register event for all memory PMUs.
```
iwankgb, line 284: It must be clearly stated in the documentation.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,369 @@
...
+		for _, pmu := range pmus {
+			newEvent := CustomEvent{
+				Type:   pmu.typeOf,
+				Config: event.Config,
+				Name:   event.Name,
+			}
+			klog.V(5).Info(newEvent.Type)
```
iwankgb, line 291: This log message will not be very helpful.

info/v1/machine.go
```diff
@@ -214,6 +214,9 @@ type MachineInfo struct {
 
 	// ID of cloud instance (e.g. instance-1) given to it by the cloud provider.
 	InstanceID InstanceID `json:"instance_id"`
+
+	// Perf uncore stats.
+	PerfUncoreStats []PerfStat `json:"perf_uncore_stats"`
```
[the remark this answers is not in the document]
Creatone: Firstly, perf events aren't hierarchical like other cgroups. So root perf cgroup "/" doesn't have pids from others cgroups.

Secondly, we cannot gather uncore events per cgroup.
For ex.
```sudo perf stat -e uncore_imc_0/cas_count_read/ -vv -p 100```
Returns error code: -22

So if we want uncore events, we need to pass ```pid = -1``` calling ```perf_event_open```.

File: perf/uncore_libpfm.go (hunk 14 above)
Comment (Creatone) [reply perf/uncore_libpfm.go:61]: If I use ``:=`` it would create local ``imcPMUs``. I want to use the global one from the package.

Comment (iwankgb) [reply perf/uncore_libpfm.go:61]: Sounds reasonable.

File: perf/uncore_libpfm.go (hunk 16 above)
Comment (Creatone) [reply perf/uncore_libpfm.go:59]: I think that you mean ``getUncoreIMCPMUs()`` which is properly tested.

Comment (iwankgb) [reply perf/uncore_libpfm.go:59]: No, I meant this one. As far as I understand this function was created in order to make sure that no duplicated entries will be added to `imcPMUs`. In my opinion it's important to prove that it works as expected.

File: perf/uncore_libpfm.go (hunk 24 above)
Comment (Creatone) [reply perf/uncore_libpfm.go:203]: Yes, it can be simplified but there'll be also info about core events.

File: perf/uncore_libpfm.go (hunk 25 above)
Comment (Creatone) [reply perf/uncore_libpfm.go:212]: Yes, but in callers scope, it would be a elegant one line:
```c.events, c.eventToCustomEvent = parseUncoreEvents(events)```

File: perf/uncore_libpfm.go (hunk 24 above)
Comment (iwankgb) [reply perf/uncore_libpfm.go:203]: It makes me think that `perf.mapEventsToCustomEvents` should be modified not to include `UncoreEvents` then.

File: perf/uncore_libpfm.go (hunk 25 above)
Comment (iwankgb) [reply perf/uncore_libpfm.go:210]: It's up to you ultimately but take into consideration that such approach tricks user into thinking that this functions changes `Events.UncoreEvents` while it only reads the property. It is at least misleading and I believe that it introduces unnecessary chaos to the code.

File: perf/uncore_libpfm.go (hunk 16 above)
Comment (Creatone) [reply perf/uncore_libpfm.go]: Done

File: perf/config.go [hunk 30]
```diff
@@ -29,18 +29,25 @@ type Events struct {
 	// output of perf list can be used.
 	Events [][]Event `json:"events"`
 
+	// List of uncore perf events to be measured.
+	UncoreEvents UncoreEvents `json:"uncore_events"`
```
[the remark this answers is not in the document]
Creatone: These events are counted differently. That's why they are separated in configuration file.

perf/collector_libpfm.go
Creatone: I thought that it would be a good idea to have a `type` label in all perf metrics. It's useful information, for example, to check if my metric has the same type as provided in the configuration file.

perf/collector_libpfm.go
Creatone: See comment above.

perf/noop.go
Creatone: See my response to @dashpole comment.

perf/types.go
Creatone: `UncoreCollector` interface is more comfortable for me. I would like to get only `[]PerfStat`, nor create new `ContainerStats` and update it.

perf/uncore_libpfm.go
Creatone: It's looking good now?

docs/runtime_options.md
```diff
@@ -158,13 +158,25 @@ automatically.
...
 in mind that it is impossible to group more events that there are counters available.
+* `PMU` - Performance Monitoring Unit
+* `uncore events` - events which can be counted by PMUs outside core.
+
+#### Getting config values
+Using perf tools:
+* Identify the event in `perf list` output. 
+* Execute command: `perf stat -I 5000 -vvv -e [EVENT_NAME]`
+* Find `perf_event_attr` section on `perf stat` output
+
```
katarzyna-z, line 169: Could you provide here an example of perf stat output and config prepared from  this output?

metrics/prometheus_machine.go
```diff
@@ -192,9 +192,58 @@ func NewPrometheusMachineCollector(i infoProvider, includedMetrics container.Met
...
+					return values
+				},
+			},
+		}...)
+	}
+
 	return c
 }
 
+func getNUMANode(cpu int, nodes []info.Node) string {
```
katarzyna-z, line 238: To have an order in this file I suggest moving this function below, under all functions with PrometheusMachineCollector receiver, near getMemoryByType function.

Creatone: Good idea, I also provided unit tests.

info/v1/machine.go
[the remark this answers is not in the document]
Creatone: Oh, I understand now... I'll try to change it.

Thanks @ss7pro for explaining it.

metrics/prometheus.go
```diff
@@ -1568,20 +1568,62 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+		machineInfo, err := c.infoProvider.GetMachineInfo()
+		if err != nil {
+			klog.Errorf("Cannot obtain information about machine topology!")
+		} else {
+			c.containerMetrics = append(c.containerMetrics, []containerMetric{
+				{
+					name:        "container_perf_uncore_metric",
+					help:        "Perf uncore event metric.",
+					valueType:   prometheus.CounterValue,
+					extraLabels: []string{"node", "event", "type"},
```
ppalucki, line 1594: regarding the "node" label, perf uncores cannot be collected by "logical" NUMA nodes, it would be more appropriate to name it "socket" label

perf/uncore_libpfm.go
ppalucki: according kernel [perf_type_id](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/perf_event.h#L29) "0" value means hardware counter 

it is a little confusing why "0" here has other meaning, could you explain this?

info/v1/container.go
```diff
@@ -846,6 +846,9 @@ type PerfStat struct {
 
 	// CPU that perf event was measured on.
 	Cpu int `json:"cpu"`
+
+	// Type is type of an event.
+	Type uint32 `json:"type"`
```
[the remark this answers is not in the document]
Creatone: This indicates from whom or what data is collected.
See ``type`` section: http://man7.org/linux/man-pages/man2/perf_event_open.2.html

For example, ``type=4`` describes that statistics come from "raw" implementation-specific event.

Creatone:
@dashpole I changed this PR to cover all perf uncore events.

perf/types.go
iwankgb: I don't think that your justification is still relevant.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,347 @@
...
+
+const (
+	uncorePMUPrefix    = "uncore"
+	pmuTypeFilename    = "type"
+	pmuCpumaskFilename = "cpumask"
+	systemDevicesPath  = "/sys/devices"
+	rootPerfEventPath  = "/sys/fs/cgroup/perf_event"
+)
+
+var readedUncorePMUs uncorePMUs
```
iwankgb, line 56: shouldn't it be `readUncorePMUs`? What does this variable represent? Can global state be avoided?

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,347 @@
...
+				if err != nil {
+					return err
+				}
+				pmus[info.Name()] = *pmu
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return pmus, err
```
iwankgb, line 122: You should probably `return nil, err` - otherwise you might return half-baked `pmus`.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,347 @@
...
+		// Uncore metric doesn't exists for cgroups, only for entire platform.
+		return &NoopUncoreCollector{}
+	}
+
+	collector := &uncoreCollector{}
+
+	err := collector.setup(events)
+	if err != nil {
+		//formatedError := fmt.Errorf("unable to setup uncore perf event collector: %v", err)
+		//klog.V(5).Info("Perf uncore metrics will not be available: %s", formatedError)
```
iwankgb, line 147: We should probably log some information.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,347 @@
...
+	uncorePMUPrefix    = "uncore"
+	pmuTypeFilename    = "type"
+	pmuCpumaskFilename = "cpumask"
+	systemDevicesPath  = "/sys/devices"
+	rootPerfEventPath  = "/sys/fs/cgroup/perf_event"
+)
+
+var readedUncorePMUs uncorePMUs
+
+type uncorePMUsSet []pmu
```
iwankgb, line 58: Do we need both types: `uncorePMUsSet` and `uncorePMUs`? I think we can use just one instead of two of them.

info/v1/container.go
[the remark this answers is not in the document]
dashpole: Can we use string constants instead?

metrics/prometheus.go
```diff
@@ -1581,6 +1582,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 					return values
 				},
 			},
+			{
+				name:        "container_perf_uncore_metric",
```
dashpole, line 1586: @iwankgb I don't know how I missed this on the perf metrics above, but prometheus metrics should not end in "metric".  Please change the other perf metrics as well.  They should probably be "container_perf_events_total" and "container_perf_uncore_events_total"

metrics/prometheus.go
```diff
@@ -1581,6 +1582,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				help:        "Perf uncore event metric.",
+				valueType:   prometheus.CounterValue,
+				extraLabels: []string{"socket", "event", "type"},
```
dashpole, line 1589: type should probably be either configuration, or a metric label, but not both.

metrics/prometheus.go
```diff
@@ -1581,6 +1582,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				getValues: func(s *info.ContainerStats) metricValues {
+					values := make(metricValues, 0, len(s.PerfUncoreStats))
+					for _, metric := range s.PerfUncoreStats {
+						values = append(values, metricValue{
+							value:     float64(metric.Value),
+							labels:    []string{machine.GetSocketFromCPU(metric.Cpu), metric.Name, strconv.FormatUint(uint64(metric.Type), 10)},
```
dashpole, line 1595: It seems like it would be more consistent with other metrics to use the CPU instead of the socket.  Is there a particular reason why we are using socket here?

metrics/prometheus.go
```diff
@@ -1581,6 +1582,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+							timestamp: s.Timestamp,
+						})
+					}
+					return values
+				},
+			},
+			{
+				name:        "container_perf_uncore_metric_scaling_ratio",
```
dashpole, line 1603: same comment as above for removing "metric" from the metric name.

perf/config.go
```diff
@@ -29,18 +29,23 @@ type Events struct {
...
+
 	// List of custom perf events' to be measured. It is impossible to
 	// specify some events using their names and in such case you have
 	// to provide lower level configuration.
 	CustomEvents []CustomEvent `json:"custom_events"`
 }
 
+type UncoreEvents map[string][][]Event
```
dashpole, line 41: what is the string in this case?  please add a comment to this type

metrics/prometheus.go
Creatone: Uncore PMUs (Performance Monitoring Units) are inside CPU but outside of cores. This means that they count events per the whole CPU.
To count events we need to call ``perf_event_open`` with CPU argument, so we use CPUs from ``cpumask`` ( for ex. ``/sys/devices/uncore_imc_0/cpumask`` ). These CPUs correspond to Sockets, and there is only one CPU per Socket.

Creatone:
@dashpole

info/v2/conversion.go
```diff
@@ -155,6 +155,9 @@ func ContainerStatsFromV1(containerName string, spec *v1.ContainerSpec, stats []
 		if len(val.PerfStats) > 0 {
 			stat.PerfStats = val.PerfStats
 		}
+		if len(val.PerfUncoreStats) > 0 {
```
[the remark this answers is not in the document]
Creatone: Done.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,390 @@
...
+	// Maping from event name, pmu type, cpu.
+	c.cpuFiles = make(map[string]map[string]map[int]readerCloser)
+	c.events = events.UncoreEvents
+	c.eventToCustomEvent = parseUncoreEvents(events)
+	c.cpuFilesLock.Lock()
+	defer c.cpuFilesLock.Unlock()
+
+	for _, group := range c.events {
+		if len(group) > 1 {
+			klog.Warning("grouping uncore perf events is not supported!")
```
dashpole, line 177: is the plan to support this eventually, or is grouping uncore events never something we can support?

perf/testing/perf.json
```diff
@@ -3,13 +3,18 @@
...
+  ],
   "custom_events": [
     {
-      "type": 4,
+      "type": 18,
       "config": [
-        "0x5300c0"
+        "0x5300"
       ],
-      "name": "instructions_retired"
```
dashpole, line 12: don't we still need the instructions_retired custom event, or is that unused?

perf/config.go
```diff
@@ -29,6 +29,9 @@ type Events struct {
 	// output of perf list can be used.
 	Events [][]Event `json:"events"`
 
+	// List of uncore perf events to be measured.
+	UncoreEvents [][]Event `json:"uncore_events"`
```
dashpole, line 33: Since these events are completely distinct from "regular" perf events as far as defining CustomEvent and grouping, should they be in a separate struct?  I am fine with them being in the same configuration file, but it may make plumbing configuration easier, and lead to less confusion when writing the rules about what is possible.

perf/uncore_libpfm.go
Creatone: Yes, after the merge I want to work on PR with grouping events.

perf/config.go
Creatone: Can we stay with the same configuration?

Uncore is part of Perf so, in my opinion, it should be one configuration file.

dashpole: I'm fine with them being in the same file, but I think they should be nested separately to be clear that custom uncore events can only be listed in the UncoreEvent section.

info/v2/conversion.go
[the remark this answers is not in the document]
katarzyna-z: Have you checked that `PerfUncoreStats` are available on http://127.0.0.1:8080/api/v2.0/stats? As far as I remember you should add them also to `DeprecatedStatsFromV1` function.

machine/machine.go
```diff
@@ -313,3 +313,57 @@ func getNumCores() int {
...
+		propertyPath := filepath.Join(sysCPUPath, sysFsCPUTopology, sysFsCPUPhysicalPackageID)
+		propertyVal, err := ioutil.ReadFile(propertyPath)
+		if err != nil {
+			return nil, err
+		}
+
+		cpuDir := path.Base(sysCPUPath)
+
+		// Get CPU number from directory name ( ex. cpu10 ).
+		cpu, err := strconv.ParseInt(cpuDir[3:], 10, 0)
```
katarzyna-z, line 357: Is it possible that `cpuDir` has length equal to 3 or less? Consider length checking before accessing the third element of `cpuDir`.

machine/machine.go
```diff
@@ -313,3 +313,57 @@ func getNumCores() int {
...
+		if err != nil {
+			return nil, err
+		}
+
+		readData[int(cpu)], err = strconv.Atoi(strings.TrimSpace(string(propertyVal)))
```
katarzyna-z, line 362: `strconv.ParseInt` returns integer value you don't need to use here `int(cpu)`.

machine/machine.go
```diff
@@ -313,3 +313,57 @@ func getNumCores() int {
 
 	return numCPU
 }
+
+var cpuToSocket map[int]int
+
+// GetSocketFromCPU returns socket which is bounded to passed CPU.
+// If there is no information it would return -1.
+func GetSocketFromCPU(cpu int) int {
```
katarzyna-z, line 321: This function retives infromation from sysfs so it fits better to [sysinfo](https://github.com/google/cadvisor/tree/master/utils/sysinfo) or [sys](https://github.com/google/cadvisor/tree/master/utils/sysfs), please move it if it is possible.

machine/machine.go
```diff
@@ -313,3 +313,57 @@ func getNumCores() int {
...
+	}
+
+	if _, ok := cpuToSocket[cpu]; !ok {
+		return -1
+	}
+
+	return cpuToSocket[cpu]
+}
+
+func readCPUToSocket(basePath string) (map[int]int, error) {
```
katarzyna-z, line 338: This function also retrives infromation from sysfs so it fits better to [sysinfo](https://github.com/google/cadvisor/tree/master/utils/sysinfo) or [sys](https://github.com/google/cadvisor/tree/master/utils/sysfs), please move it if it is possible.

machine/machine.go
```diff
@@ -313,3 +313,57 @@ func getNumCores() int {
...
+
+var cpuToSocket map[int]int
+
+// GetSocketFromCPU returns socket which is bounded to passed CPU.
+// If there is no information it would return -1.
+func GetSocketFromCPU(cpu int) int {
+	// Check if information is cached.
+	if cpuToSocket == nil {
+		var err error
+		cpuToSocket, err = readCPUToSocket(cpuBusPath)
```
katarzyna-z, line 325: Please write a unit tests to cover the case when `physical_package_id` file is not available.

perf/collector_libpfm_test.go
```diff
@@ -20,10 +20,12 @@ package perf
 import (
 	"bytes"
 	"encoding/binary"
-	"github.com/stretchr/testify/assert"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
katarzyna-z, line 25: Why is this import in separated group? It seems that it can be in group with `info "github.com/google/cadvisor/info/v1"` and `"github.com/google/cadvisor/stats"`.

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,389 @@
...
+// #cgo CFLAGS: -I/usr/include
+// #cgo LDFLAGS: -lpfm
+// #include <perfmon/pfmlib.h>
+// #include <stdlib.h>
+import "C"
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"github.com/google/cadvisor/machine"
```
katarzyna-z, line 29: move "github.com/google/cadvisor/machine" and "github.com/google/cadvisor/stats" to different group of imports

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,389 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func parseEventName(eventName string) (string, string) {
+	splittedEvent := strings.SplitN(eventName, "/", 2)
```
katarzyna-z, line 205: Please document that first `/` separates pmu prefix and event name.

perf/uncore_libpfm_test.go
```diff
@@ -0,0 +1,194 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Uncore perf events logic tests.
+package perf
+
+import (
+	"golang.org/x/sys/unix"
```
katarzyna-z, line 21: Move this import to different group.

docs/runtime_options.md
```diff
@@ -158,13 +158,96 @@ automatically.
...
+      "config": [
+        "0x304"
+      ],
+      "name": "uncore_imc_1/cas_count_all"
+    }
+  ]
+}
+```
+
+- ```uncore_imc/cas_count_read``` - because of prefix and no entry in custom events,
```
katarzyna-z, line 235: I know what you mean by prefix but I think that user can have a problem with understanding.

docs/runtime_options.md
```diff
@@ -158,13 +158,96 @@ automatically.
...
+```
+
+Config values can be also obtain from: 
+* [Intel® 64 and IA32 Architectures Performance Monitoring Events](https://software.intel.com/content/www/us/en/develop/download/intel-64-and-ia32-architectures-performance-monitoring-events.html)
+
+
+#### Uncore Events configuration
+Configuration of perf uncore events has special behaviour.
+
+Let's explain this by example:
```
katarzyna-z, line 208: It valuable that it's example here but I think that documentation should have also general description of configuration method.

docs/runtime_options.md
```diff
@@ -158,13 +158,96 @@ automatically.
...
+    }
+  ]
+}
+```
+
+Config values can be also obtain from: 
+* [Intel® 64 and IA32 Architectures Performance Monitoring Events](https://software.intel.com/content/www/us/en/develop/download/intel-64-and-ia32-architectures-performance-monitoring-events.html)
+
+
+#### Uncore Events configuration
```
katarzyna-z, line 205: Methods described in "Getting config values" section apply also for uncore perf events so having "Uncore Events configuration" section could make confusion.  Example mentioned below could be a subsection of "Getting config values"

docs/runtime_options.md
```diff
@@ -194,6 +284,9 @@ interface that majority of users will rely on.
 * `instructions_retired` will be measured as non-grouped event and is specified using an advanced API that allows
 to specify any perf event available (some of them are not named and can't be specified with plain string). Event name 
 should be a human readable string that will become a metric name.
+* `cas_count_read` will be measured as uncore non-grouped event on all Integrated Memory Controllers Performance Monitoring Units because of unset `type` field and
+``uncore_imc`` prefix.
```
katarzyna-z, line 288: Wrong formatting of ``uncore_imc`` ? It seems that single "`" will be enough here.

info/v2/conversion.go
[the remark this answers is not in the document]
Creatone: Done.

perf/config.go
Creatone: Is it looking good now?

dashpole: yes, thanks

machine/machine.go
Creatone: ``strconv.ParseInt`` returns int64, we need int32.

perf/collector_libpfm_test.go
Creatone: It's from a different vendor?

machine/machine.go
Creatone: Functionality moved to ``utils/sysinfo/GetSocketFromCPU``.

machine/machine.go
Creatone: Functionality moved to ``utils/sysinfo/GetSocketFromCPU``.

machine/machine.go
Creatone: Functionality moved to ``utils/sysinfo/GetSocketFromCPU``.

metrics/prometheus.go
```diff
@@ -1565,8 +1565,8 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 				},
 			},
 			{
-				name:        "container_perf_metric_scaling_ratio",
-				help:        "Perf event metric scaling ratio",
+				name:        "container_perf_events_total_scaling_ratio",
```
dashpole, line 1568: this shouldn't include `total`.  `total` indicates that the metric is cumulative, which I don't believe this is.  `total` should only be placed at the end of the metric when appropriate.

metrics/prometheus.go
```diff
@@ -1581,6 +1581,40 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+							value:     float64(metric.Value),
+							labels:    []string{strconv.Itoa(metric.Socket), metric.Name, metric.PMU},
+							timestamp: s.Timestamp,
+						})
+					}
+					return values
+				},
+			},
+			{
+				name:        "container_perf_uncore_events_total_scaling_ratio",
```
dashpole, line 1602: remove `total` as well here.

metrics/testdata/prometheus_metrics
```diff
@@ -327,18 +327,18 @@ container_network_udp_usage_total{container_env_foo_env="prod",container_label_f
...
-# HELP container_perf_metric_scaling_ratio Perf event metric scaling ratio
-# TYPE container_perf_metric_scaling_ratio gauge
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 1 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="0",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.66666666666 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.5 1395066363000
-container_perf_metric_scaling_ratio{container_env_foo_env="prod",container_label_foo_label="bar",cpu="1",event="instructions_retired",id="testcontainer",image="test",name="testcontaineralias",zone_name="hello"} 0.33333333333 1395066363000
+# HELP container_perf_events_total Perf event metric.
```
dashpole, line 330: can you also add testing for the uncore events?  I think it requires modifying metrics/prometheus_fake.go

docs/runtime_options.md
```diff
@@ -158,23 +158,124 @@ automatically.
...
+  }
+}
+```
+
+Config values can be also obtain from: 
+* [Intel® 64 and IA32 Architectures Performance Monitoring Events](https://software.intel.com/content/www/us/en/develop/download/intel-64-and-ia32-architectures-performance-monitoring-events.html)
+
+
+##### Uncore Events configuration
+Uncore Event name should be in form `PMU_PREFIX/event_name` where **PMU_PREFIX** mean
```
dashpole, line 222: apologies if this has already been discussed, but why do we need to be able to provide a prefix for the PMU?

docs/runtime_options.md
```diff
@@ -158,23 +158,124 @@ automatically.
...
+      ["uncore_imc/cas_count_read"],
+      ["uncore_imc_0/cas_count_write"],
+      ["uncore_imc_1/cas_count_all"]
+    ],
+    "custom_events": [ 
+      {
+        "config": [
+          "0x304"
+        ],
+        "name": "uncore_imc_0/cas_count_write"
```
dashpole, line 240: Is the PMU prefix actually required to define a custom event?  Should we only have the event name (not the PMU prefix) here?

docs/runtime_options.md
```diff
@@ -158,23 +158,124 @@ automatically.
...
+          "0x304"
+        ],
+        "name": "uncore_imc_1/cas_count_all"
+      }
+    ]
+  }
+}
+```
+
+- `uncore_imc/cas_count_read` - because of `uncore_imc` prefix and no entry in custom events,
```
iwankgb, line 254: As far as I understand `prefix` is basically event `type`. I think single name should be used across code and documentation to make it consistent and easier to grasp.

metrics/testdata/prometheus_metrics
Creatone: Done

docs/runtime_options.md
Creatone: In this example, there is no type provided so we need PMU name in prefix

info/v1/machine.go
```diff
@@ -71,6 +71,17 @@ func (n *Node) FindCore(id int) (bool, int) {
...
 
+func (n *Node) FindCoreByThread(thread int) (bool, int) {
+	for i, n := range n.Cores {
+		for _, t := range n.Threads {
+			if t == thread {
+				return true, i
+			}
+		}
+	}
+	return false, -1
```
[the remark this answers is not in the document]
Creatone: Done.

docs/runtime_options.md
Creatone: Could you check it now, please? :)

docs/runtime_options.md
dashpole: Is there always an integer `type`, or do we sometimes need the string prefix?  In what cases can't we use the integer type?

Creatone: Let me explain this. 
If you don't pass `type` in custom events, PMU will be taken from `PMU_PREFIX`.
`PMU_PREFIX` is the name of PMU.

Creatone: We don't need both ways. Because setting type field in custom events was before uncore events, I added that possibility.
I decided also to add `perf stat` tool way of handling PMU ( **PMU_PREFIX/event_name** ).

perf/manager_libpfm_test.go
```diff
@@ -18,42 +18,45 @@
 package perf
 
 import (
+	"testing"
+
+	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/stats"
+
```
katarzyna-z, line 25: unnecessary empty line

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,392 @@
...
+// #cgo CFLAGS: -I/usr/include
+// #cgo LDFLAGS: -lpfm
+// #include <perfmon/pfmlib.h>
+// #include <stdlib.h>
+import "C"
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"github.com/google/cadvisor/utils/sysinfo"
```
katarzyna-z, line 29: This import should be in different group of imports

docs/runtime_options.md
```diff
@@ -158,23 +158,124 @@ automatically.
...
+        "name": "cas_count_all"
+      }
+    ]
+  }
+}
+```
+
+- `uncore_imc/cas_count_read` - because of `uncore_imc` type and no entry in custom events,
+    it would be counted by **all** Integrated Memory Controller PMUs with config provided from libpfm package.
+    (using this function: https://www.man7.org/linux/man-pages/man3/pfm_get_perf_event_encoding.3.html)
```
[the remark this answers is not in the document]
Creatone: We use `pfm_get_os_event_encoding()` . It's just wrong link :P

katarzyna-z:
I think that it's worth to squash commits.

Creatone:
> I think that it's worth to squash commits.

Done :)

perf/uncore_libpfm.go
```diff
@@ -0,0 +1,392 @@
...
+	return eventName, pmuPrefix
+}
+
+func obtainPMUs(want string, gotPMUs uncorePMUs) []pmu {
+	var pmus []pmu
+	if want == "" {
+		return pmus
+	}
+	for _, pmu := range gotPMUs {
+		if strings.Contains(pmu.name, want) {
```
dashpole, line 223: should we also use HasPrefix here?

docs/runtime_options.md
dashpole: I've had some time to dig into this.  Thanks to @iwankgb for explaining some things to me.  I'm not a fan of having "/" -delineated strings in our config, and generally prefer explicit fields for each input, but also understand that this is used elsewhere.  I'll let you decide what to do.

metrics/prometheus_fake.go
```diff
@@ -648,6 +648,22 @@ func (p testSubcontainersInfoProvider) SubcontainersInfo(string, *info.Container
 							Cpu:          1,
 						},
 					},
+					PerfUncoreStats: []info.PerfUncoreStat{
+						{
+							ScalingRatio: 1.0,
+							Value:        1231231512.0,
+							Name:         "uncore_imc_0/cas_count_read",
```
dashpole, line 655: While i'm alright with having the event name be PMU/Name for the configuration, i'd like us to separate them out for reporting of metrics. IIUC, based on prometheus metrics below, this name includes the PMU prefix.

Creatone: It's separated. Sorry, I missed change this test case.

In a real scenario prometheus metric will look like:
``container_perf_uncore_events_total{event="cas_count_read",id="/",pmu="uncore_imc_0",socket="0"} 2.106316711e+10 1592929424570``

iwankgb:
Congratulations, @Creatone :)

Creatone:
@dashpole I needed to change one integration test due to use `strings.HasPrefix`. Now everything should pass :)
@iwankgb Thanks!

PR #2530, opened 2020-05-01
add open telemetry

MarcelHillmann:
Hi,

the second try for #2516 without Zipkin ;)

Thanks
Marcel


cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+
+var (
+	traceSampler      = flag.String("trace_sampler", "always", "choose a sampler 'never', 'always' or 'probability'")
+	traceSamplerParam = flag.Float64("trace_sampler_param", 0, "used for the 'probability' sampler")
+	traceServiceName  = flag.String("trace_name", "cadvisor", "use localEndpoint service name")
+	traceTags         tracingTags
+)
+
+func init() {
+	flag.Var(&traceTags, "trace_tags", "additional tags for tracing (eg. foo=bar)")
```
dashpole, line 45: What is the use case for adding tags via flags?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+	"k8s.io/klog/v2"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+var (
+	traceSampler      = flag.String("trace_sampler", "always", "choose a sampler 'never', 'always' or 'probability'")
+	traceSamplerParam = flag.Float64("trace_sampler_param", 0, "used for the 'probability' sampler")
+	traceServiceName  = flag.String("trace_name", "cadvisor", "use localEndpoint service name")
```
dashpole, line 40: I think hard-coding this as cadvisor is fine to start with.  We don't need a flag.

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+	"go.opentelemetry.io/otel/sdk/trace"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/klog/v2"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+var (
+	traceSampler      = flag.String("trace_sampler", "always", "choose a sampler 'never', 'always' or 'probability'")
```
dashpole, line 38: can we combine this and trace_sampler_param into `trace_probability`, where 0 => never and 100 => always?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+
+func init() {
+	flag.Var(&traceTags, "trace_tags", "additional tags for tracing (eg. foo=bar)")
+}
+
+func InitTrace(traceEndpoint, collectorCert string) *otlp.Exporter {
+	exporterOptions := []otlp.ExporterOption{otlp.WithAddress(traceEndpoint)}
+
+	if collectorCert != "" {
+		if creds, err := credentials.NewClientTLSFromFile(collectorCert, ""); err != nil {
```
dashpole, line 52: nit: This will be easier to read if we don't if/else when the `if` exits:
change
```golang
if foo, err := bar(); err != nil {
    klog.Fatal(err)
} else {
   // use foo
}
```
to 
```golang
foo, err := bar()
if err != nil {
    klog.Fatal(err)
}
// use foo
```

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (w *wrapperHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	if w.callback != nil {
+		w.callback(writer, request)
+	} else {
+		panic(fmt.Errorf("not yet implemented"))
+	}
+}
+
+func modifyHeader(request *http.Request) bool {
```
dashpole, line 129: what is the purpose of this function?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,192 @@
...
+			trace.WithResourceAttributes(getTraceTags(traceTags)...),
+		}
+
+		if traceProvider, err := trace.NewProvider(providerOpts...); err != nil {
+			klog.Fatal(err)
+		} else {
+			global.SetTraceProvider(traceProvider)
+
+			apiB3, traceCtx, corrCtx := apiTrace.B3{SingleHeader: false}, apiTrace.DefaultHTTPPropagator(), correlation.DefaultHTTPPropagator()
+			options := []propagation.Option{propagation.WithInjectors(apiB3, traceCtx, corrCtx),
```
dashpole, line 77: i'm surprised we need to specify the propagation format...  I don't see this done in any examples.  Are you sure this is required?

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: I have at moment 14 cAdvisor installation, and would like to differ the ServiceName in Zipkin and Jaeger.

That the reason for `trace_name`.

Thx Marcel

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: Yes off course, but I don't like it if the WYSIWYG pattern is brocken.
But I coud refractor that.

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: I changed it, I missed the os.Exit on log.Fatal

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: If the traceId has only 16 digets, you have to pad it on teh left side with "0".

That is the propose from the opentelemetry-go Gitter.

https://github.com/open-telemetry/opentelemetry-go/issues/686

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: The default implementation dosn't use the B3 headers.

https://github.com/open-telemetry/opentelemetry-go/blob/master/api/global/internal/state.go#L124

I'm not happy with that implementation, I have to find a independend solution or more flags 
:(

MarcelHillmann:
@googlebot I fixed it.

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: If you need more flags for the search to differ the Traces.

Like k8s.* oder cloud.* https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions

MarcelHillmann:
@googlebot I fixed it.

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: I added a own HTTPExtracor, which checks all available Extractors.

It is not possible to register all HTTPExtractor, because it override a prepared Context, if the next failed.

cmd/internal/opentelemetry/wrapperMux.go
MarcelHillmann: I removed it and added trace_probability as Float64

cmd/internal/opentelemetry/wrapperMux.go
Aneurysm9: This should be handled in the otel-go b3 propagator.  We don't want every consumer to have to modify trace headers like this.  I hope that we'll be able to come to a quick agreement on how we want to handle 64-bit B3 trace ID values across the OpenTelemetry ecosystem and get a quick fix implemented in the Go propagator.

cmd/internal/opentelemetry/wrapperMux.go
dashpole: is it possible to add the host name or something else as a tag to differentiate?  Passing a different flag to each instance is probably not ideal.

cmd/internal/opentelemetry/wrapperMux.go
dashpole: Dumb question, but is there a difference between the W3C and B3 format?

cmd/internal/opentelemetry/wrapperMux.go
dashpole: I would prefer discovering these from the host wherever possible, rather than setting them via flags.  If we do go the route of flags, i'd prefer specific flags for each attribute, rather than a k/v list.  But i'm hoping we don't have to do that.

cmd/internal/opentelemetry/wrapperMux.go
Aneurysm9: Closing the loop, this was fixed in https://github.com/open-telemetry/opentelemetry-go/pull/698.

cmd/internal/opentelemetry/wrapperMux.go
Aneurysm9: Yes, the [W3C Trace Context Recommendation](https://www.w3.org/TR/trace-context/#traceparent-header) specifies a `traceparent` header that includes the trace ID, parent span ID, and trace flags.  The [B3 format](https://github.com/openzipkin/b3-propagation#http-encodings) specifies both multiple-header and single-header encodings of the same information with provision for additional information.  The single-header B3 encoding is nearly identical to the W3C `traceparent` encoding, but for the `version` field and `flags` vs. `sampled`.

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opentelemetry
+
+import (
+	"context"
```
iwankgb, line 18: Could you run `goimports` on this file, please?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+	} else {
+		exporterOptions = append(exporterOptions, otlp.WithInsecure())
+	}
+
+	exporter, err := otlp.NewExporter(exporterOptions...)
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	sampler := getTraceSamplerProbability() //*traceSampler, *traceSamplerParam)
```
iwankgb, line 62: Remove the comment, please.

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+			return newCtx
+		}
+	}
+	return orgCtx
+}
+
+func getTraceSamplerProbability() trace.Sampler {
+	probability := *traceProbability
+	if probability < 0 && probability > 1 {
+		klog.Errorf("invalid probability value %f, probability is allowed between 0.0 and 1.0", probability)
```
iwankgb, line 96: Shouldn't you `return trace.NeverSample()` here? Otherwise we are going to hit `return trace.ProbabilitySampler(probability)` with invalid probability.

cmd/cadvisor.go
```diff
@@ -170,7 +173,12 @@ func main() {
 		klog.Fatalf("Failed to create a manager: %s", err)
 	}
 
-	mux := http.NewServeMux()
+	// setup open telemetry
+	if *traceEndpoint != "" {
+		defer opentelemetry.InitTrace(*traceEndpoint, *collectorCert).Stop()
```
iwankgb, line 178: I think I understand why `Stop()` call is deferred but why is `InitTrace()` call deferred?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+		trace.WithConfig(trace.Config{DefaultSampler: sampler}),
+	}
+
+	traceProvider, err := trace.NewProvider(providerOpts...)
+	if err != nil {
+		klog.Fatal(err)
+	}
+	global.SetTraceProvider(traceProvider)
+
+	apiB3Single, apiB3, traceCtx, corrCtx := apiTrace.B3{SingleHeader: false}, apiTrace.B3{SingleHeader: false}, apiTrace.DefaultHTTPPropagator(), correlation.DefaultHTTPPropagator()
```
iwankgb, line 74: How is `apiB3Single` different from `apiB3`? They seem to be identically configured instance of the same struct.

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (w *wrapperHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	if w.callback != nil {
+		w.callback(writer, request)
+	} else {
+		panic(fmt.Errorf("not yet implemented"))
+	}
+}
+
+func getTraceTags(tags []string) []kv.KeyValue {
```
iwankgb, line 148: This function seems to be unused.

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+		} else {
+			keyValue = append(keyValue, kv.String(tKey, val))
+		}
+	}
+
+	klog.Infof("add tracings tags: %s", ToString(keyValue))
+	return keyValue
+}
+
+func ToString(keyValues []kv.KeyValue) string {
```
iwankgb, line 190: What is this function needed for? Are you trying to generate some Golang code?

dashpole:
cc @SergeyKanzhelev

cmd/internal/opentelemetry/tracingTags.go
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opentelemetry
+
+import (
+	"strings"
+)
+
+type tracingTags []string
```
SergeyKanzhelev, line 21: I cannot find where it's used

cmd/go.mod
```diff
@@ -23,18 +23,19 @@ require (
 	github.com/golang/snappy v0.0.0-20150730031844-723cc1e459b8 // indirect
 	github.com/influxdb/influxdb v0.9.6-0.20151125225445-9eab56311373
 	github.com/klauspost/crc32 v0.0.0-20151223135126-a3b15ae34567 // indirect
-	github.com/kr/pretty v0.1.0 // indirect
```
SergeyKanzhelev, line 26: curious why adding opentelemetry removed this indirect dependency

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+	"google.golang.org/grpc/credentials"
+	"k8s.io/klog/v2"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+const traceServiceName = "cadvisor"
+
+var traceProbability = flag.Float64("trace_probability", 1, "used for sampler (0=never, 1=always and probability is 0.1 to 0.99)")
```
SergeyKanzhelev, line 41: I see `trace_` prefix is used here and for the OTLP collector address. If variables would be standardized in OpenTelemetry, would you be open to reuse names from there instead of inventing new names? Or these flags has it's own naming convention?

cmd/internal/opentelemetry/wrapperMux.go
```diff
@@ -0,0 +1,203 @@
...
+		w.callback(writer, request)
+	} else {
+		panic(fmt.Errorf("not yet implemented"))
+	}
+}
+
+func getTraceTags(tags []string) []kv.KeyValue {
+	keyValue := []kv.KeyValue{
+		kv.String("service.name", traceServiceName),
+		kv.String("service.version", fmt.Sprintf("semver: %s git:%s", version.Info["version"], version.Info["revision"])),
```
SergeyKanzhelev, line 151: I think the idea is either use one or another, not two together. Do you need both?

cmd/cadvisor.go
```diff
@@ -77,6 +78,8 @@ var rawCgroupPrefixWhiteList = flag.String("raw_cgroup_prefix_whitelist", "", "A
 
 var perfEvents = flag.String("perf_events_config", "", "Path to a JSON file containing configuration of perf events to measure. Empty value disabled perf events measuring.")
 
+var traceEndpoint = flag.String("trace_endpoint", "", "Url host:port to the OTLP collector")
```
SergeyKanzhelev, line 81: why are you reading this variable here? Would it be better to hide this logic inside the `InitTrace`?

bobbypage:
I'm going to close this for now since there hasn't been any updates in a while, please re-open if appropriate.

PR #2531, opened 2020-05-02
Using upstream golangci-lint

iwankgb:
Replacing `golangci-lint` from my fork with upstream version.


iwankgb:
@dashpole we have to download release because `go get -u -v github.com/golangci/golangci-lint/cmd/golangci-lint` was producing errors that I was not able to reproduce on any available environment.

PR #2532, opened 2020-05-02
Some linting errors

iwankgb:
The only remaining errors (besides fixed in #2513) are those from `utils/cpuload/netlink/netlink.go` - do you think it's worth fixing them? `netlink.addAttribute()` is used in two places and in both cases buffers are based on slice of bytes; I can't imagine when it would be impossible to write to them.


dashpole:
entirely up to you whether or not to fix them.  It generally seems like a good thing to clean up the linter errors, but no one has touched the load reader code in years...

PR #2535, opened 2020-05-06
machine, topology_test: not read system file

giuseppe:
GetClockSpeed() uses the information from the file
/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq when it exists.

Make sure it is not used for tests.


giuseppe:
@sayaoailun PTAL

sayaoailun:
> @sayaoailun PTAL

@giuseppe Yes, you are right! I didn't notice this problem.

PR #2538, opened 2020-05-08
Fix description of cpu_cores metrics in docs

ymyzk:
The description of machine_cpu_cores and machine_cpu_physical_cores was swapped.


dashpole:
This seems correct, since the "physical" one has it in the name.  @katarzyna-z am I missing something?

ymyzk:
I checked [this code](https://github.com/google/cadvisor/blob/438d294d3bcc7b7c50393e7b72f257dfc5545cbe/metrics/prometheus_machine.go#L79-L94) as well. fyi :slightly_smiling_face:

katarzyna-z:
Yes, it's correct. We just overlooked this ;) LGTM
@dashpole please merge this pull request

PR #2541, opened 2020-05-13
Run integration tests on s390x

rajaskakodkar:
Added a job in Github Actions to run integration tests on s390x arch under qemu. 
Updated crc32 version to support s390x arch.
Updated `kubernetes/pause` image to `k8s.gcr.io/pause:3.2` multi arch image which supports s390x arch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,3 +72,42 @@ jobs:
       with:
         name: cadvisor.log
         path: ${{ github.workspace }}/go/src/github.com/google/cadvisor/cadvisor.log
+  test-integration-s390x:
+    runs-on: ubuntu-18.04
+    name: Build on s390x
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+          path: go/src/github.com/google/cadvisor
+      - uses: uraimo/run-on-arch-action@v1.0.9
+        with:
+          architecture: s390x
+          distribution: ubuntu18.04
+          run: |
+            set -euo pipefail
+            apt-get update
+            apt-get install -y --no-install-recommends \
+            apt-transport-https \
+            ca-certificates \
+            socat \
+            curl \
+            gnupg2 \
+            software-properties-common \
+            lsb-release \
+            build-essential \
+            git \
+            sudo
+            curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+            add-apt-repository \
+               "deb [arch=s390x] https://download.docker.com/linux/ubuntu \
+               $(lsb_release -cs) stable"
+            apt-get update
+            apt-cache madison docker-ce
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -104,7 +104,6 @@ jobs:
                "deb [arch=s390x] https://download.docker.com/linux/ubuntu \
                $(lsb_release -cs) stable"
             apt-get update
-            apt-cache madison docker-ce
```
dims: This is helpful to spot newer versions when they get added to the docker debian repository. please do not remove it.

rajaskakodkar: Added a commit for the same.

Makefile
```diff
@@ -24,6 +24,11 @@ else
   Dockerfile_tag := '.''$(arch)'
 endif
 
+ifeq ($(arch), s390x)
+	race_flag := ''
```
iwankgb, line 28: Why have you decided not to look for race conditions on s390x?

.github/workflows/test.yml
```diff
@@ -72,3 +72,49 @@ jobs:
       with:
         name: cadvisor.log
         path: ${{ github.workspace }}/go/src/github.com/google/cadvisor/cadvisor.log
+  test-integration-s390x:
```
iwankgb, line 75: It would be great if you could upload logs on error - they can help a lot when trying to understand why tests are failing, especially on architecture that some of us might not be able to get their hands on.

rajaskakodkar: Test cases are failing in Github Actions as docker is not coming up in s390x container under qemu. They pass in s390x container on s390x VM. After this change 
```
printf '{"iptables": false,"ip-masq": false,"bridge": "none" }' > /etc/docker/daemon.json
```
I could get docker up in s390x container under qemu but it fails to pull images, throws this error
```
Attempting next endpoint for pull after error: failed to register layer: Error processing tar file(exit status 1): 
```
cc @dims

Makefile
rajaskakodkar: race conditions are not supported on s390x.

.github/workflows/test.yml
iwankgb: According to the [documentation](https://docs.docker.com/engine/reference/commandline/dockerd/):

> `--iptables=false` prevents the Docker daemon from adding iptables rules. If multiple daemons manage iptables rules, they may overwrite rules set by another daemon. Be aware that disabling this option requires you to manually add iptables rules to expose container ports. If you prevent Docker from adding iptables rules, Docker will also not add IP masquerading rules, even if you set `--ip-masq` to `true`. Without IP masquerading rules, Docker containers will not be able to connect to external hosts or the internet when using network other than default bridge.

iwankgb: `uraimo/run-on-arch-action` will execute your commands (`run` on line 91) docker will  not be able to pull any images that are needed to run integration tests.

PR #2543, opened 2020-05-13
Reformat README

kshithijiyer:


dashpole:
Huh...

W0513 16:54:01.244] ++ git diff --name-only pages
W0513 16:54:01.294] fatal: ambiguous argument 'pages': unknown revision or path not in the working tree.

I don't think this is because of this PR

PR #2551, opened 2020-05-21
Add socket id into Core

AlexeyPerevalov:
Now cadvisor in MachineInfo.Topology keeps NUMA nodes.
There is no socket information, except number of sockets.
Since socket - NUMA nodes relationship could be many-to-many,
it's more reasonable to keep socket_id per Core, not per NUMA node.

This socket id could be used in kubernetes cpu manager to provide
per socket allocations.

Signed-off-by: Alexey Perevalov <alexey.perevalov@huawei.com>

### DIFF
--- a/info/v1/machine.go
+++ b/info/v1/machine.go
@@ -47,9 +47,10 @@ type Node struct {
 }
 
 type Core struct {
-	Id      int     `json:"core_id"`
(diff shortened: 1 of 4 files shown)


dashpole:
just needs CLA

AlexeyPerevalov:
@googlebot I signed it!

PR #2555, opened 2020-05-22
Add instruction for running cAdvisor with perf support on k8s

### LINKED ISSUE #2484: Perf: improve documentation by adding k8s example
An example of running cAdvisor with perf support on Kubernetes must be provided (including: ConfigMap, Pod etc).
Part of #2388.

katarzyna-z:

Fixes #2484


deploy/Dockerfile_perf
```diff
@@ -0,0 +1,27 @@
+FROM alpine:3.10
+
+RUN apk --no-cache add libc6-compat=1.1.22-r3 device-mapper=2.02.184-r0 findutils=4.6.0-r1 zfs=0.8.2-r0 build-base=0.5-r1 linux-headers=4.19.36-r0 && \
+    apk --no-cache add thin-provisioning-tools=0.7.1-r3 --repository http://dl-3.alpinelinux.org/alpine/edge/main/ && \ 
+    echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
+    rm -rf /var/cache/apk/*
+
+
+RUN wget https://sourceforge.net/projects/perfmon2/files/libpfm4/libpfm-4.10.1.tar.gz && \
```
dashpole, line 9: Is there a reason why this shouldn't just be the "normal" build of cAdvisor?  Would some users not want this installed, or does it come with any downsides?

iwankgb: In case of non-containerized cAdvisor perf requires users to install one package (available on Debian and Centos). It would be great to have it enabled by default and it could be really beneficial for users.
I'm not sure how keen the users would be (especially taking that cAdvisor is a Kubernetes dependency...)

dashpole: But the Dockerfile is only used for containerized users

iwankgb: Then it sound like a very good idea to me.

iwankgb:
LGTM :)

deploy/Dockerfile_perf
katarzyna-z: > Is there a reason why this shouldn't just be the "normal" build of cAdvisor? Would some users not want this installed, or does it come with any downsides?

I don't see downsides of having libpfm in normal build of cAdvisor. If cAdvisor binary is built without libpfm then introduced changes don't have any impact. 

If cAdvisor binary is built with  libpfm according to [this instruction](https://github.com/google/cadvisor/blob/master/docs/development/build.md#perf-support) then introduced changes enable of having perf metric exposed on API and on metric endpoint.

Changes to normal Dockerfile are visible in [this commit](https://github.com/google/cadvisor/pull/2555/commits/cd46e074899a6e131b35f52b1983c6d01b878e27).

katarzyna-z:
I added proposition of changes to have enabled perf metrics in release docker container by default, see [commit](https://github.com/google/cadvisor/pull/2555/commits/81d4a57d3946305dae08b34640c16e09f92d3276). 

I see that cAdvisor binary is available [here](https://github.com/google/cadvisor/releases/tag/v0.36.0) and this is why I added [the second build of cAdvisor](https://github.com/google/cadvisor/pull/2555/files#diff-fff440587f66b0e9a460aa8cf04ce6d2R56) (without libpfm4). If cAdvisor is build with libpfm4 there must be libpfm4 installed in runtime environment and for users it will an additional requirement.

build/release.sh
```diff
@@ -32,10 +32,19 @@ if ! git_user="$(git config --get user.email)"; then
   exit 1
 fi
 
-# Build the release.
 export BUILD_USER="$git_user"
 export BUILD_DATE=$( date +%Y%m%d ) # Release date is only to day-granularity
 export VERBOSE=true
+
+# Get and build libpfm4 for cAdvisor with perf support
+wget https://sourceforge.net/projects/perfmon2/files/libpfm4/libpfm-4.10.1.tar.gz
```
dashpole, line 40: I don't think we need to install libpfm4 as part of the release script.  The documentation in development/build.md should be sufficient.  I mostly just wanted to have libpfm used as to go flag for the docker build, which you added below.

deploy/kubernetes/README.md
```diff
@@ -24,6 +24,36 @@ To apply the daemonset to your cluster with example patches applied:
 kustomize build deploy/kubernetes/overlays/examples | kubectl apply -f -
 ```
 
+### cAdvisor with perf support on Kubernetes
+
+Example of modifications needed to deploy cAdvisor with perf support is provided in [overlays/examples_perf](overlays/examples_perf) directory (modification to daemonset and configmap with perf events configuration).
+
+To test example deployment from the cadvisor/ directory build binary with perf support
+```
+GO_FLAGS="-tags=libpfm,netgo" make build
```
dashpole, line 33: We don't need users to rebuild cadvisor if the release version has perf support, right?

deploy/kubernetes/overlays/examples_perf/cadvisor-perf.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: DaemonSet
+metadata:
+  name: cadvisor
+  namespace: cadvisor
+spec:
+  template:
+    spec:
+      containers:
+      - name: cadvisor
+        image: "<cadvisor_image>"
```
dashpole, line 12: reminder to remove this line, assuming users don't need to rebuild.

deploy/kubernetes/overlays/examples_perf/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+bases:
+- ../../base
+resources:
+- configmap.yaml
```
dashpole, line 4: i'd prefer just keeping this all in the examples folder.  Users can comment out the kustomizations they don't want there.

deploy/kubernetes/README.md
katarzyna-z: Yes, you're right;)

deploy/kubernetes/overlays/examples_perf/kustomization.yaml
katarzyna-z: I tried to add all files to examples folder but then kustomize returns following error:
```
$ kustomize build deploy/kubernetes/overlays/examples | kubectl apply -f -
Error: conflict between map[string]interface {}{"apiVersion":"apps/v1", "kind":"DaemonSet", "metadata":map[string]interface {}{"name":"cadvisor", "namespace":"cadvisor"}, "spec":map[string]interface {}{"template":map[string]interface {}{"spec":map[string]interface {}{"containers":[]interface {}{map[string]interface {}{"args":[]interface {}{"--housekeeping_interval=10s", "--max_housekeeping_interval=15s", "--event_storage_event_limit=default=0", "--event_storage_age_limit=default=0", "--disable_metrics=percpu,hugetlb,sched,tcp,udp,advtcp", "--docker_only", "--store_container_labels=false", "--whitelisted_container_labels=io.kubernetes.container.name, io.kubernetes.pod.name,io.kubernetes.pod.namespace"}, "env":[]interface {}{map[string]interface {}{"name":"LD_LIBRARY_PATH", "value":"/bin/nvidia/lib64/"}}, "name":"cadvisor", "securityContext":map[string]interface {}{"privileged":true}, "volumeMounts":[]interface {}{map[string]interface {}{"mountPath":"/dev", "name":"dev"}, map[string]interface {}{"mountPath":"/bin/nvidia/lib64/", "name":"libnvidia"}}}}, "volumes":[]interface {}{map[string]interface {}{"hostPath":map[string]interface {}{"path":"/dev"}, "name":"dev"}, map[string]interface {}{"hostPath":map[string]interface {}{"path":"/home/kubernetes/bin/nvidia/lib64/"}, "name":"libnvidia"}}}}}} and map[string]interface {}{"apiVersion":"apps/v1", "kind":"DaemonSet", "metadata":map[string]interface {}{"name":"cadvisor", "namespace":"cadvisor"}, "spec":map[string]interface {}{"template":map[string]interface {}{"spec":map[string]interface {}{"containers":[]interface {}{map[string]interface {}{"args":[]interface {}{"--perf_events_config=/etc/config/perf-non-hardware.json"}, "name":"cadvisor", "securityContext":map[string]interface {}{"privileged":true}, "volumeMounts":[]interface {}{map[string]interface {}{"mountPath":"/etc/config/", "name":"perf-volume"}}}}, "volumes":[]interface {}{map[string]interface {}{"configMap":map[string]interface {}{"name":"perf-config"}, "name":"perf-volume"}}}}}}
error: no objects passed to apply
```

It seems that kustomize has a problem with two patches which modify the same fields in different way, e.g. args

dashpole: ah, ok.  In that case, we can have a separate folder

build/release.sh
```diff
@@ -44,6 +46,10 @@ docker_tag="google/cadvisor:$VERSION"
 gcr_tag="gcr.io/cadvisor/cadvisor:$VERSION"
 docker build -t $docker_tag -t $gcr_tag -f deploy/Dockerfile .
 
+# Build the release binary without libpfm4 to not require libpfm4 in runtime environment
```
dashpole, line 49: Is this only required if perf is enabled, or will cadvisor not run at all if libpfm4 isn't enabled?  This would also mean perf support is only available in the containerized release version of cAdvisor .

deploy/Dockerfile
```diff
@@ -1,11 +1,19 @@
 FROM alpine:3.10
 MAINTAINER dengnan@google.com vmarmol@google.com vishnuk@google.com jimmidyson@gmail.com stclair@google.com
 
-RUN apk --no-cache add libc6-compat device-mapper findutils zfs && \
-    apk --no-cache add thin-provisioning-tools --repository http://dl-3.alpinelinux.org/alpine/edge/main/ && \
+RUN apk --no-cache add libc6-compat=1.1.22-r3 device-mapper==2.02.184-r0 findutils=4.6.0-r1 zfs=0.8.2-r0  build-base=0.5-r1 linux-headers=4.19.36-r0 && \
```
dashpole, line 4: why are we pinning dependencies to specific versions?  Is this required by the installation of libpfm4?

build/release.sh
katarzyna-z: If binary is built with libpfm4 ( with GO_FLAGS="-tags=libpfm,netgo") then cAdvisor does not run without libpfm4 in runtime environment and this mean that perf support will be available only in the containerized release version of cAdvisor .

deploy/Dockerfile
katarzyna-z: libpfm4 does not require any specific version of packages. I added versions to control installed versions. Should I go back to previous state here?

build/release.sh
katarzyna-z: May release contain two binaries? The first without perf support as it is now and the second with perf support (and later with ipmctl)?

dashpole: I'm fine with what you have for this PR.  I don't want to set the precedence that we will release additional binaries for optional features.

deploy/Dockerfile
dashpole: yeah, go back to the previous state.  Otherwise, i'll forget, and we will have packages with security problems...

PR #2556, opened 2020-05-23
oomparser_test: fix data race

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>
Capture the iterator variable pair in the goroutine to avoid data race.


iwankgb:
Looks unbelievably simple to me 👍.

PR #2563, opened 2020-05-29
Add Resctrl metrics.

Creatone:
This PR provides support for Resource Control statistics.


info/v1/container.go
```diff
@@ -848,6 +848,25 @@ type PerfStat struct {
 	Cpu int `json:"cpu"`
 }
 
+type MemoryBandwidthMonitoringStats struct {
```
dashpole, line 851: Maybe MemoryBandwidthStats?  Or is this metrics about the monitoring of memory bandwidth?

info/v1/container.go
```diff
@@ -848,6 +848,25 @@ type PerfStat struct {
...
+	// The 'mbm_total_bytes'.
```
dashpole, line 852: This looks like a reference to something.  Can you include a link to where this comes from?

info/v1/container.go
```diff
@@ -848,6 +848,25 @@ type PerfStat struct {
...
+	TotalBytes uint64 `json:"mbm_total_bytes,omitempty"`
+
+	// The 'mbm_local_bytes'.
+	LocalBytes uint64 `json:"mbm_local_bytes,omitempty"`
+}
+
+type CacheMonitoringTechnologyStats struct {
```
dashpole, line 859: Maybe just CacheStats?

metrics/prometheus.go
```diff
@@ -1595,6 +1595,82 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.ResctrlMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:        "container_mem_bandwidth_bytes",
```
dashpole, line 1601: use container_memory instead of container_mem to be consistent with other memory metrics.

metrics/prometheus.go
```diff
@@ -1595,6 +1595,82 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+						}
+					}
+					return metrics
+				},
+			},
+			{
+				name:        "container_mem_bandwidth_local_bytes",
+				help:        "Total local memory bandwidth usage statistics for container counted with RDT Memory Bandwidth Monitoring (MBM).",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"numa_node"},
```
dashpole, line 1622: I think we have been using node_id in other metrics.  See prometheus_machine.go.  Maybe we can share that constant?

metrics/prometheus.go
```diff
@@ -1595,6 +1595,82 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+							value:     float64(stats.LocalBytes),
+							timestamp: s.Timestamp,
+							labels:    []string{strconv.Itoa(numaNode)},
+						}
+					}
+					return metrics
+				},
+			},
+			{
+				name:        "container_mem_bandwidth_remote_bytes",
```
dashpole, line 1637: this metric is unnessessary, as it can be computed from the previous two metrics.

Alternatively, you can have a single metric, container_memory_bandwidth_bytes_total that is broken down by "local" vs "remote"

metrics/prometheus.go
dashpole: also, cumulative metrics should have the suffix `_total`

metrics/prometheus.go
```diff
@@ -1595,6 +1595,82 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+							timestamp: s.Timestamp,
+							labels:    []string{strconv.Itoa(numaNode)},
+						}
+					}
+					return metrics
+				},
+			},
+			{
+				name:        "container_llc_occupancy_bytes",
+				help:        "Last level cache usage statistics for container counted with RDT Memory Bandwidth Monitoring (MBM).",
```
dashpole, line 1656: This is not cumulative, right?

info/v1/container.go
Creatone: It's related to MBM ( Memory Bandwidth Monitoring ). See: https://01.org/cache-monitoring-technology

info/v1/container.go
Creatone: See: https://01.org/cache-monitoring-technology

metrics/prometheus.go
Creatone: This is not cumulative. It should be a gauge metric.

metrics/prometheus.go
Creatone: I agree with you, it can be done with Prometheus rules. Even RDT API exposes only Total and Local bytes.

info/v1/container.go
Creatone: Done.

info/v1/container.go
```diff
@@ -848,6 +848,26 @@ type PerfStat struct {
 	Cpu int `json:"cpu"`
 }
 
+// See: https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt
+type MemoryBandwidthMonitoringStats struct {
```
katarzyna-z, line 852: Exported type should have a descriptive comment starting with `MemoryBandwidthMonitoringStats` , see e.g. `PerfStat` type description

info/v1/container.go
katarzyna-z: Exported struct should have a descriptive comment starting with CacheMonitoringTechnologyStats, please add link to description

info/v1/container.go
```diff
@@ -848,6 +848,26 @@ type PerfStat struct {
...
+	// The 'mbm_local_bytes'.
+	LocalBytes uint64 `json:"mbm_local_bytes,omitempty"`
+}
+
+type CacheMonitoringTechnologyStats struct {
+	// The 'llc_occupancy'.
+	LLCOccupancy uint64 `json:"llc_occupancy,omitempty"`
+}
+
+type ResctrlStats struct {
```
katarzyna-z, line 865: Could you add description also this exported struct?

info/v1/container.go
```diff
@@ -902,6 +922,9 @@ type ContainerStats struct {
 
 	// Referenced memory
 	ReferencedMemory uint64 `json:"referenced_memory,omitempty"`
+
+	// Resource Control (resctrl) statistics.
```
katarzyna-z, line 926: The rest of comments in `ProcessStats` struct do not have dot at the end.

info/v2/container.go
```diff
@@ -170,6 +170,8 @@ type ContainerStats struct {
 	PerfStats []v1.PerfStat `json:"perf_stats,omitempty"`
 	// Referenced memory
 	ReferencedMemory uint64 `json:"referenced_memory,omitempty"`
+	// Resource Control (resctrl) statistics.
```
katarzyna-z, line 173: The rest of comments in `ContainerStats` struct do not have dot at the end, except `Timestamp` field

info/v2/conversion.go
```diff
@@ -155,6 +155,9 @@ func ContainerStatsFromV1(containerName string, spec *v1.ContainerSpec, stats []
 		if len(val.PerfStats) > 0 {
 			stat.PerfStats = val.PerfStats
 		}
+		if len(val.Resctrl.MemoryBandwidthMonitoring) > 0 || len(val.Resctrl.CacheMonitoringTechnology) > 0 {
```
katarzyna-z, line 158: As I remember, you need add values from `Resctrl` also to `DeprecatedStatsFromV1` function to have them on API v2.0.
Have you checked that newly added values are available on http://127.0.0.1:8080/api/v2.0/stats?

info/v1/container.go
Creatone: Is it wrong?

info/v2/conversion.go
Creatone: You are correct :)

info/v1/container.go
dashpole: Based on that, we should call this MemoryBandwithStats.  I realize this ruins the appreciation, but in this context, "monitoring" is unnecessary

metrics/prometheus.go
```diff
@@ -1595,6 +1595,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+							value:     float64(stats.LocalBytes),
+							timestamp: s.Timestamp,
+							labels:    []string{strconv.Itoa(numaNode)},
+						}
+					}
+					return metrics
+				},
+			},
+			{
+				name:        "container_llc_occupancy_bytes_total",
```
dashpole, line 1637: If this isn't cumulative, please remove the _total suffix.

metrics/prometheus.go
```diff
@@ -1595,6 +1595,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.ResctrlMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:        "container_memory_bandwidth_bytes_total",
+				help:        "Total memory bandwidth usage statistics for container counted with RDT Memory Bandwidth Monitoring (MBM).",
+				valueType:   prometheus.GaugeValue,
```
dashpole, line 1603: If this is cumulative, please use a counter type.  If it is not cumulative, please remove the _total suffix.

katarzyna-z:
please add new metrics to this table https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md#prometheus-container-metrics

PR #2565, opened 2020-05-30
Not collecting disk metric for root container when not requested to

### LINKED ISSUE #2544: Repeated logs: Not collecting filesystem statistics because file "/proc/diskstats" was not found
Running cadvisor on centos 7 and it shows in logs that for every 1 or 2 seconds it raises a message "Not collecting filesystem statistics because file "/proc/diskstats" was not found"
I am running centos on top of virtuozzo container which does't provide /proc/diskstats. In such case how to run cAdvisor.

iwankgb:
Fixes #2544


dashpole:
I think this makes sense, since disabling disk metrics removes the prometheus metrics from the endpoint anyways.

PR #2572, opened 2020-06-03
updating go-conenctions to v0.4.0

ykakarap:
Updating the "go-connections" version to v0.4.0 from v0.3.0

cc @dims


ykakarap:
@googlebot I signed it!

dashpole:
Is there a reason for the update?  Can you update the description?

dims:
@dashpole looking at https://github.com/docker/go-connections/compare/v0.3.0...v0.4.0 there are a couple of PRs in there for TLS and cipher updates, So requested @ykakarap to file an update we can then turn around and use in kubernetes/kubernetes.

PR #2573, opened 2020-06-03
Update opencontainers/runc

shyaamsn:
Update opencontainers/runc

This is required in order to update opencontainers/runc and google/cadvisor in https://github.com/kubernetes/kubernetes

https://github.com/kubernetes/kubernetes/pull/91366

cc @dims


shyaamsn:
@googlebot I signed it!

dashpole:
Are there any updates in particular you are looking to bring in with this change?

shyaamsn:
@dashpole This is a dependency for kubernetes/kubernetes#91366. Need to update runc here first before updating cadvisor and runc in k/k.

PR #2576, opened 2020-06-06
Using multistage Dockerfile and upgrading Alpine to 3.12

iwankgb:
I was trying to build Docker image on aarch64 and I failed miserably because of some incompatibilities between `glibc` and `musl`. Using multistage Dockerfile helps and, as a side-effect, shrinks the image to 150 megabytes.


iwankgb:
I have just managed to confirm that Docker image build in a new way is capable of collecting perf stats.

PR #2577, opened 2020-06-09
Change resources for cadvisor container in perf example

wacuuu:
Hi,
I tried to run cadvisor with perf example. A word about my setup: I have a machine in a cluster(I limited daemonset only to single node in 3 nodes cluster), which is running around 40 containers. The response times of curl for /metrics on default settings from example:
- running cadvisor with perf without container: did not measure, although it was pretty fast
- running cadvisor without perf with container: around 40 seconds
- running cadvisor with perf with container: after almost 20 minutes curl reports that i was unable to get all the data

I played a little bit with limits and it seems that this is CPU related. So i suggest this change in example and mention it in readme for examples.

### DIFF
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -33,6 +33,8 @@ To generate and apply the daemonset with patches for cAdvisor with perf support:
 kustomize build deploy/kubernetes/overlays/examples_perf | kubectl apply -f -
 ```
 
+Be aware, that perf gathering generates more data, and the CPU may be a bottleneck. If the stats take too long to get, try increasing CPU limits in daemonset configuration.
+
 ## Kustomization
 
 On your own fork of cAdvisor, create your own overlay directoy with your patches.  Copy patches from the example folder if you intend to use them, but don't modify the originals.  Commit your changes in your local branch, and use git to manage them the same way you would any other piece of code.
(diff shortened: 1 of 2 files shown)


wacuuu:
The problem with the CLA is that i don't know who is the issue nor do i know any of those people :/

iwankgb:
You can try to learn more about cAdvisor resource consumption using [flags `--profiling` and `--log_cadvisor_usage`](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md#debugging-and-logging).

dashpole:
It looks like the commit here was authored by a "Jakub Walecki", which is different from the account used to open this PR.  If this was your work, just make sure the email on the commit matches the email on your github.  Otherwise, I can't accept the commit.

dashpole:
> running cadvisor without perf with container: around 40 seconds

This seems excessive.  I haven't seen it this high before.  If that is the case, I don't think this is perf-specific, but more likely a general performance issue.  The suggestions above should be useful.

wacuuu:
Ok, I've learned my lesson with pprof and charts(a bit) on what takes how much time. Played a little bit with the limits. The conclusion is that most of the time goes for http handling and prometheus metric processing. 

Regarding the second thing, that to get Guarateed QoS group, this was not my intention, as the example does not have it. I did not have an idea how to differentiate the limits, so I have added those.

The question is: as this is not perf specyfic, is it worth to add this to example config for perfs and is it worth to mention it in the documentation regarding perf? Or, more bluntly, does this PR even make sense? :)

iwankgb:
I can't say I'm surprised. I can recall other issue related to Prometheus exposition and its performance that was raised recently. 
The best approach would probably be to improve Prometheus exposition format performance (might be hard) or use another  storage driver such as Kafka.

dashpole:
I don't think this PR is necessary.

wacuuu:
Oki, thanks for help anyway, this thought me something :)

PR #2586, opened 2020-06-13
Fix incorrect detection of images path

hakman:
Images path is incorrectly detected as the last one in the list of possible locations.
As all tests used the last value as the expected value, this issue was not detected earlier.

The issue has a pretty high risk as one can end up with:
* frequent unnecessary images GC
* no images GC at all
* nodes being tainted `node.kubernetes.io/disk-pressure: NoSchedule`

Ref: https://github.com/kubernetes/kubernetes/issues/91982


wu105:
@hakman Will cadvisor handle images/containers on arbitrary partitions?  E.g., we have variations like the following:

/dev/sda4  /var/lib/docker   Storage Driver btrfs
/dev/sdc1  /var/lib/docker   Storage Driver btrfs
/dev/sda1 /var/lib/containerd Storage Driver overlayfs

PR #2598, opened 2020-06-25
Add "-disable_metrics" column to prometheus metrics table.

Creatone:
This PR adds information about the `-disable_metrics` parameter to the Prometheus metrics table.


cmd/cadvisor.go
```diff
@@ -106,6 +106,7 @@ var (
 		container.HugetlbUsageMetrics:            struct{}{},
 		container.ReferencedMemoryMetrics:        struct{}{},
 		container.CPUTopologyMetrics:             struct{}{},
+		container.ResctrlMetrics:                 struct{}{},
```
dashpole, line 109: please add this to the default ignoreMetrics above as well.

Creatone: Done.

PR #2603, opened 2020-06-26
Allow on-demand metrics collection for prometheus

### LINKED ISSUE #1989: Allow on-demand houskeeping for Prometheus scraping

dqminh:
Fix #1989 
This is a rework of #2000 

This allows user to set v2.RequestOptions for PrometheusCollector to be
used when retrieving container stats. The options will only be set once
per collector to ensure that we don't race for concurrent metrics
request. This should not be a problem for prometheus scrapes as the
scrapes are usually in the same format.

One can design the on-demand metrics collection as such to minimize CPU
usage in housekeeping routine:

- assuming prometheus scrape interval is 60s
- increase housekeeping interval to a high value such as 300s
- set max_age paramether in scrape to a low value ( such as 1s ), so
that if we scrape just as housekeeping finish, we don't do any extra
work. When we scrape, if current metrics are older than 1s, we will
collect the metrics again.

Signed-off-by: Daniel Dao <dqminh89@gmail.com>

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/google/cadvisor/container"
 	info "github.com/google/cadvisor/info/v1"
+	v2 "github.com/google/cadvisor/info/v2"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog/v2"
@@ -97,13 +98,14 @@ type PrometheusCollector struct {
 	containerMetrics    []containerMetric
 	containerLabelsFunc ContainerLabelsFunc
 	includedMetrics     container.MetricSet
+	opts                v2.RequestOptions
 }
 
 // NewPrometheusCollector returns a new PrometheusCollector. The passed
 // ContainerLabelsFunc specifies which base labels will be attached to all
 // exported metrics. If left to nil, the DefaultContainerLabels function
 // will be used instead.
-func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetrics container.MetricSet, now clock.Clock) *PrometheusCollector {
+func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetrics container.MetricSet, now clock.Clock, opts v2.RequestOptions) *PrometheusCollector {
 	if f == nil {
 		f = DefaultContainerLabels
 	}
@@ -129,6 +131,7 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		},
 		includedMetrics: includedMetrics,
+		opts:            opts,
 	}
 	if includedMetrics.Has(container.CpuUsageMetrics) {
 		c.containerMetrics = append(c.containerMetrics, []containerMetric{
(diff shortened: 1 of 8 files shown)


metrics/prometheus.go
```diff
@@ -97,6 +99,12 @@ type PrometheusCollector struct {
 	containerMetrics    []containerMetric
 	containerLabelsFunc ContainerLabelsFunc
 	includedMetrics     container.MetricSet
+
+	// custom options to pass to stats collection routine. The option will be populated
+	// once in the first scrape so the scrape request should be the same each time (as
+	// usual with prometheus scrapes should already be in most cases)
+	updateOnce sync.Once
```
dashpole, line 106: Is it not possible to do this per-request?  It seems like this would have strange behavior if there are multiple callers.

dqminh: `Gatherer` interface only has method `Collect(ch)` which didnt accept any argument, and can be called concurrently, so do this per-request is hard to do with a singleton collector as we have now, as we would have to lock the request.
However, i looked at it again, and dont think this is a good idea any more, mainly because this prevents user from changing maxAge by changing prometheus configuration.

Alternative would be the previous approach in #2000 which is to create a collector per request ( like how node_exporter currently does it https://github.com/prometheus/node_exporter/blob/master/node_exporter.go#L96-L137 ). Actually now that i think about it abit more, that does seem like a better approach that what we have now, as registrry and collector are quite lightweight, and the collection is done infrequently.

metrics/prometheus.go
```diff
@@ -97,13 +98,14 @@ type PrometheusCollector struct {
...
 	includedMetrics     container.MetricSet
+	opts                v2.RequestOptions
 }
 
 // NewPrometheusCollector returns a new PrometheusCollector. The passed
 // ContainerLabelsFunc specifies which base labels will be attached to all
 // exported metrics. If left to nil, the DefaultContainerLabels function
 // will be used instead.
-func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetrics container.MetricSet, now clock.Clock) *PrometheusCollector {
+func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetrics container.MetricSet, now clock.Clock, opts v2.RequestOptions) *PrometheusCollector {
```
dashpole, line 108: Can you add tests for possible values of this parameter?

dqminh: added a test `TestNewPrometheusCollectorWithRequestOptions` to test that we propagated the request option to `GetRequestedContainersInfo`

alexw1982:
> This allows user to set v2.RequestOptions for PrometheusCollector to be
> used when retrieving container stats. The options will only be set once
> per collector to ensure that we don't race for concurrent metrics
> request. This should not be a problem for prometheus scrapes as the
> scrapes are usually in the same format.
> 
> One can design the on-demand metrics collection as such to minimize CPU
> usage in housekeeping routine:
> 
>     * assuming prometheus scrape interval is 60s
> 
>     * increase housekeeping interval to a high value such as 300s
> 
>     * set max_age paramether in scrape to a low value ( such as 1s ), so
>       that if we scrape just as housekeeping finish, we don't do any extra
>       work. When we scrape, if current metrics are older than 1s, we will
>       collect the metrics again.

How would I implement this feature. I get setting housekeeping interval to something beyond the prometheus scrape interval but where/how do I set max_age and the v2.RequestOptions for PrometheusCollector? I was unable to find any documentation on this...

PR #2606, opened 2020-07-01
Add command arguments to GetProcessList

gaainf:
It will be useful to see `cmd` format option in GetProcessList method. 

What is done:
1. Added `cmd` option to getPsOutput format inside GetProcessList method
2. Renamed previous `cmd` option with actual `comm` option
3. Updated ProcessInfo structure

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -289,12 +289,12 @@ func (cd *containerData) GetProcessList(cadvisorContainer string, inHostNamespac
 	if !inHostNamespace {
 		rootfs = "/rootfs"
 	}
-	format := "user,pid,ppid,stime,pcpu,pmem,rss,vsz,stat,time,comm,psr,cgroup"
+	format := "user,pid,ppid,stime,pcpu,pmem,rss,vsz,stat,time,comm,psr,cgroup:10000,cmd"
 	out, err := cd.getPsOutput(inHostNamespace, format)
 	if err != nil {
 		return nil, err
 	}
-	expectedFields := 13
+	expectedFields := 14
 	processes := []v2.ProcessInfo{}
 	lines := strings.Split(string(out), "\n")
 	for _, line := range lines[1:] {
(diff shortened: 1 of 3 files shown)


gaainf:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
@googlebot I signed it!

dashpole:
/ok-to-test

can you make sure there are unit tests for this?

manager/container.go
```diff
@@ -335,7 +335,10 @@ func (cd *containerData) GetProcessList(cadvisorContainer string, inHostNamespac
 		vs *= 1024
 		psr, err := strconv.Atoi(fields[11])
 		if err != nil {
-			return nil, fmt.Errorf("invalid pid %q: %v", fields[1], err)
+			// to avoid defunct processes
+			//return nil, fmt.Errorf("invalid psr %q: %v", fields[11], err)
```
dashpole, line 339: I think we still want to fail on an error here, right?

gaainf: Thank you for the comment. The reason of the commit is "to increase fault tolerance of  cadvisor". Unfortunately, process list is returned as empty in such case, when any external process (even not in container) appeared in the host system as defunct. It is regular case. I think, cadvisor should take into account and handle it correctly. But, "psr = -1" was the easiest and maybe not the best solution. Do you have an alternative variant?

dashpole: What error message were you getting here?

Is this related to adding the additional field, or is this fixing a preexisting problem?

gaainf: This is fixing a preexisting problem. Error message `invalid psr %q: %v <defunct> strconv.Atoi: parsing "<defunct>": invalid syntax` appears on attempt to convert string "<defunc>" to integer via strconv.Atoi.

dashpole: can we check for "defunct" specifically instead of ignoring the error entirely?

bobbypage:
I'm going to close this for now since there hasn't been any updates in a while, please re-open if appropriate.

This PR also is currently conflicting with the changes done in https://github.com/google/cadvisor/pull/2751

PR #2607, opened 2020-07-01
Improve information about metrics

katarzyna-z:
- Corrected information about perf metrics
- Add information about build flags required for metrics


dashpole:
Looks like the divider is missing between container_perf_events_total and container_perf_metric_scaling_ratio, and the divider above libpfm for container_perf_events_total.

katarzyna-z:
I don't see that... I see 5 dividers in lines with `container_perf_events_total` and 5 dividers in line with `container_perf_metric_scaling_ratio`

dashpole:
I only see it when displaying the rich diff... not actually sure why it happens.  We can fix it in a follow-up if needed.

dashpole:
huh, I don't see the problem anymore in the actual page: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md

PR #2612, opened 2020-07-07
Add stats to stdout storage

piowag:
Add stats to stdout storage: memory, hugetlb, perf, resctrl, referenced memory

### DIFF
--- a/cmd/internal/storage/stdout/stdout.go
+++ b/cmd/internal/storage/stdout/stdout.go
@@ -17,6 +17,7 @@ package stdout
 import (
 	"bytes"
 	"fmt"
+	"strconv"
 	"time"
 
 	info "github.com/google/cadvisor/info/v1"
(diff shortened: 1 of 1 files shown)


cmd/internal/storage/stdout/stdout.go
```diff
@@ -32,27 +33,62 @@ type stdoutStorage struct {
 }
 
 const (
-	colTimestamp = "timestamp"
-	// CPU Uasge
-	colCpuCumulativeUsage = "cpu_cumulative_usage"
```
dashpole, line 37: I'm fine with the rename, but just in-case someone is using these metrics, lets keep the old name around as well, with a link to an issue to remove it in a few releases (v0.39.0)

Creatone:
cc: @Creatone

dashpole:
This looks good.  Do you think we have enough metrics here where we need to start respecting --ignore-metrics?  I believe the current PR will just have zero values for disabled metrics, which is acceptable, but adds to the load for any backends that process the data.

dashpole (approved):
lgtm

PR #2621, opened 2020-07-17
Memory numa stats

katarzyna-z:

This pull request introduces information from memory.numa_stat as Prometheus metrics.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -422,7 +422,8 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 				getValues: func(s *info.ContainerStats) metricValues {
 					return metricValues{{value: float64(s.Memory.WorkingSet), timestamp: s.Timestamp}}
 				},
-			}, {
+			},
+			{
 				name:        "container_memory_failures_total",
 				help:        "Cumulative count of memory allocation failures.",
 				valueType:   prometheus.CounterValue,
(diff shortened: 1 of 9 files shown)


metrics/prometheus.go
```diff
@@ -454,6 +455,38 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+						[]string{"total", "container"}, s.Timestamp)...)
+					values = append(values, getNumaStatsPerNode(s.Memory.ContainerData.NumaStats.File,
+						[]string{"file", "container"}, s.Timestamp)...)
+					values = append(values, getNumaStatsPerNode(s.Memory.ContainerData.NumaStats.Anon,
+						[]string{"anon", "container"}, s.Timestamp)...)
+					values = append(values, getNumaStatsPerNode(s.Memory.ContainerData.NumaStats.Unevictable,
+						[]string{"unevictable", "container"}, s.Timestamp)...)
+
+					values = append(values, getNumaStatsPerNode(s.Memory.HierarchicalData.NumaStats.Total,
+						[]string{"total", "hierarchy"}, s.Timestamp)...)
```
dashpole, line 478: I suspect hierarchy vs container may need to be separate metrics by the same logic above.

metrics/prometheus.go
```diff
@@ -454,6 +455,38 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.MemoryNumaMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:        "container_memory_numa_pages",
+				help:        "Memory usage per numa node",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"type", "scope", "node"},
```
dashpole, line 464: I think we only add two labels to metrics below?

metrics/prometheus.go
```diff
@@ -454,6 +455,38 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.MemoryNumaMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:        "container_memory_numa_pages",
```
dashpole, line 461: is this measured in #pages, or bytes?  The help text should specify the units, and the suffix of the metric should be _<units>

metrics/prometheus.go
```diff
@@ -454,6 +455,38 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				help:        "Memory usage per numa node",
+				valueType:   prometheus.GaugeValue,
+				extraLabels: []string{"type", "scope", "node"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					values := make(metricValues, 0)
+
+					values = append(values, getNumaStatsPerNode(s.Memory.ContainerData.NumaStats.Total,
+						[]string{"total", "container"}, s.Timestamp)...)
```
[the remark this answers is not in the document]
katarzyna-z: In [documentation](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) it is written that "total" count is sum of file + anon + unevictable so I'll remove metrics with "total" label.

metrics/prometheus.go
katarzyna-z: I followed the pattern which was used for "container_memory_failures_total metric", see [this](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L426). There is "scope" label with values "container" or "hierarchy".

metrics/prometheus.go
katarzyna-z: [Here](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) it is written that memory.numa_stat contains pages and I see that in [runc](https://github.com/opencontainers/runc/blob/d65df61dc5f87e5be4459782ba7a72619f145dca/libcontainer/cgroups/fs/memory.go#L286) values are only read from file. I'll improve help text.

katarzyna-z:
@dashpole Could you take a look if new names of metrics are better?

metrics/prometheus.go
dashpole: I think this comment is still relevant

dashpole: Ah, I see now.

PR #2631, opened 2020-07-24
Add entry for libpfm related tests to Makefile

Creatone:
It must be possible to run libpfm related tests with Makefile.


dashpole:
I0724 16:17:04.048] >> running tests
W0724 16:17:04.151] go test: race flag may be set only once
W0724 16:17:04.153] run "go help test" or "go help testflag" for more information

Creatone:
Okk @dashpole, I found better solution for this case.

PR #2638, opened 2020-08-07
Add stats to statsd storage

piowag:
Add stats to stdout storage: memory, hugetlb, perf, resctrl, referenced memory

### DIFF
--- a/cmd/internal/storage/statsd/statsd.go
+++ b/cmd/internal/storage/statsd/statsd.go
@@ -15,6 +15,8 @@
 package statsd
 
 import (
+	"strconv"
+
 	client "github.com/google/cadvisor/cmd/internal/storage/statsd/client"
 	info "github.com/google/cadvisor/info/v1"
 	"github.com/google/cadvisor/storage"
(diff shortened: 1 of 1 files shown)


dashpole:
Same comment copied from the other PR:

This looks good. Do you think we have enough metrics here where we need to start respecting --ignore-metrics? I believe the current PR will just have zero values for disabled metrics, which is acceptable, but adds to the load for any backends that process the data.

katarzyna-z:
> Same comment copied from the other PR:
> 
> This looks good. Do you think we have enough metrics here where we need to start respecting --ignore-metrics? I believe the current PR will just have zero values for disabled metrics, which is acceptable, but adds to the load for any backends that process the data.

I think that it will be good to start respecting `disable_metrics` parameter also for other storages than Prometheus. My only doubt is how to clearly inform users about these changes. If we start now, for some of storages `disable_metrics` will be respected,  for others will not be respected.
Maybe it will be good to add this information to description in help, for example:
```
-disable_metrics metrics
        comma-separated list of metrics to be disabled. Parameter is used by Prometheus and statsd storages. Options are 'accelerator', 'cpu_topology','disk', 'diskIO', 'network', 'tcp', 'udp', 'percpu', 'sched', 'process', 'hugetlb', 'referenced_memory', 'resctrl'. (default tcp,advtcp,sched,udp,process,hugetlb,referenced_memory,cpu_topology,resctrl)
```

dashpole:
+1 to updating the flag description.  If you'd prefer, we can only gate the _new_ metrics with the respective disable_metric.

dashpole (approved):
lgtm

PR #2640, opened 2020-08-13
Remove exclude guest flag from perf event attrs. 

Creatone:
From https://man7.org/linux/man-pages/man2/perf_event_open.2.html :
```
exclude_guest (since Linux 3.2)
              When conducting measurements that include processes running VM
              instances (i.e., have executed a KVM_RUN ioctl(2)), do not
              measure events happening inside guest instances.  This is only
              meaningful outside the guests; this setting does not change
              counts gathered inside of a guest.  Currently, this function‐
              ality is x86 only.
```

Depend on the kernel version, `perf_event_open` can throw `invalid_argument` for uncore events with this flag set.
For core events it's also unnecessary.

### DIFF
--- a/perf/collector_libpfm.go
+++ b/perf/collector_libpfm.go
@@ -345,7 +345,7 @@ func createPerfEventAttr(event CustomEvent) *unix.PerfEventAttr {
 func setAttributes(config *unix.PerfEventAttr, leader bool) {
 	config.Sample_type = unix.PERF_SAMPLE_IDENTIFIER
 	config.Read_format = unix.PERF_FORMAT_TOTAL_TIME_ENABLED | unix.PERF_FORMAT_TOTAL_TIME_RUNNING | unix.PERF_FORMAT_GROUP | unix.PERF_FORMAT_ID
-	config.Bits = unix.PerfBitInherit | unix.PerfBitExcludeGuest
+	config.Bits = unix.PerfBitInherit
 
 	// Group leader should have this flag set to disable counting until all group would be prepared.
 	if leader {
--- a/perf/collector_libpfm_test.go
+++ b/perf/collector_libpfm_test.go
@@ -172,13 +172,13 @@ func TestSetGroupAttributes(t *testing.T) {
 	setAttributes(attributes, true)
 	assert.Equal(t, uint64(65536), attributes.Sample_type)
 	assert.Equal(t, uint64(0xf), attributes.Read_format)
-	assert.Equal(t, uint64(0x100003), attributes.Bits)
+	assert.Equal(t, uint64(0x3), attributes.Bits)
 
 	attributes = createPerfEventAttr(event)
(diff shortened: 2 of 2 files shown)


dashpole:
/ok-to-test
cc @katarzyna-z @iwankgb

iwankgb:
Can we make it configurable? We did it for a reason in WCA, I guess.

Creatone:
Before posting this PR, I had a discussion with @ppalucki and he doesn't know why this parameter was added.
Someone else was the author of this part of code, @gryf can you explain why this parameter is needed?

iwankgb:
What is this flag responsible for?

Creatone:
To not measure events inside VM.

Why omit VM when we depend on measures from cgroups?

gryf:
I don't remember the exact reason, why we exclude VMs from measurement in WCA. Perhaps it has something to do with specific environment we planned to measure.

ppalucki:
IMHO there is no reason to omit VM measurements by default at all - I strongly recommend removing this flag because it will result in misleading results for virtual environments - we're going to remove that from WCA as well.

gryf:
The only reason I could think of is to actually prevent to count anything from hypervisor, perhaps to detect a process, which could grab the resources VM wants, thus performance of VM is affected. I'd opt for @iwankgb proposal to make it configurable.

iwankgb:
@ppalucki @Creatone configurable with default value to include hypervisor?

Creatone:
Sure, but where keep value? Another flag or in a perf config file?

katarzyna-z:
I think that we can extend config file.

katarzyna-z:
@iwankgb What do you think about proposed form of configuration?

Creatone:
We assumed it would be possible when event grouping would be available.

iwankgb (approved):

dashpole:
Sorry for the slow review.  Based on the discussion, i'm a little concerned we don't have real use-cases for the additional configuration.  It sounds like someone could use this _in theory_, but would anyone actually set this in practice?

If we are unsure if it is needed, we should err on the side of omitting it for now, and doing what will work for most.  We can always add it in later if there is a real need.
If we _do_ think the configuration is justified, i'd like to see a bit more documentation explaining when someone would want to set it.  It sounds like there may be a small set of use-cases, so we could potentially enumerate them.

Creatone:
I agree with you. 
If this parameter won't be configurable now, I'll delete it.

iwankgb:
@bobbypage your involvement here would be more than appreciated.

Creatone:
@bobbypage thanks to this PR, perf events would be accurate for VMs. Could you check this PR?

dashpole (approved):
lgtm

dashpole:
Sorry, didn't see the most recent commit removing the flag.  This looks good now.

PR #2644, opened 2020-08-19
Use perf attributes from unix lib.

Creatone:
Use perf attributes from unix lib.


katarzyna-z (commented):
lgtm

dashpole (approved):
lgtm

PR #2645, opened 2020-08-19
Some optimization for process stats collection

dqminh:
- Use `Readdirnames` instead of `ioutil.Readdir` to collect fds
- Add a flag to disable socket Count collection, as `Readlink` is a lot of work for each fd

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -263,6 +263,34 @@ func processRootProcUlimits(rootFs string, rootPid int) []info.UlimitSpec {
 	return processLimitsFile(string(out))
 }
 
+func countFdAndSocket(dirPath string) (fdCount, socketCount uint64, err error) {
+	d, err := os.Open(dirPath)
+	if err != nil {
+		return fdCount, socketCount, err
+	}
+	defer d.Close()
+
+	names, err := d.Readdirnames(-1)
+	if err != nil {
+		return fdCount, socketCount, err
+	}
+	fdCount = uint64(len(names))
+
+	for _, name := range names {
+		fdPath := path.Join(dirPath, name)
+		linkName, err := os.Readlink(fdPath)
+		if err != nil {
+			klog.V(4).Infof("error while reading %q link: %v", fdPath, err)
+			continue
+		}
+		if strings.HasPrefix(linkName, "socket") {
+			socketCount++
+		}
+	}
+
+	return fdCount, socketCount, nil
+}
+
 func processStatsFromProcs(rootFs string, cgroupPath string, rootPid int) (info.ProcessStats, error) {
 	var fdCount, socketCount uint64
 	filePath := path.Join(cgroupPath, "cgroup.procs")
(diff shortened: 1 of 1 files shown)


dqminh:
@dashpole have you had time to look into this one ?

container/libcontainer/handler.go
```diff
@@ -43,6 +43,8 @@ var (
 	whitelistedUlimits      = [...]string{"max_open_files"}
 	referencedResetInterval = flag.Uint64("referenced_reset_interval", 0,
 		"Reset interval for referenced bytes (container_referenced_bytes metric), number of measurement cycles after which referenced bytes are cleared, if set to 0 referenced bytes are never cleared (default: 0)")
+	disableSocketCountStats = flag.Bool("disable_socket_count_stats", false,
```
dashpole, line 46: I think this is currently covered under the "process" stats category in --disable_metrics.  Does that not work for your usecase?

dqminh: Sorry for the late reply !

No, because we want to have process stats, we just don't need the socket counts ( partly because it's too costly to collect )

dashpole:
Sorry for the slow review.  How expensive are socket stats?  If we wanted to do this, we would need to add a new category for it in --disable_metrics, rather than a new flag.  But we would only do that if socket stats are very expensive.

dqminh:
@dashpole in the commit, i put in some numbers from when i test it

With args:

```
/cadvisor \
  -disable_metrics=sched,hugetlb,referenced_memory,cpu_topology,tcp,udp,advtcp \
  -housekeeping_interval=1s -profiling=true
```

and using `profile?seconds=10`, we have
- before: `processStatsFromProcs` takes 25.93% with `os.Readlink` takes 16.67%.
- after: `processStatsFromProcs` takes 11.82%

We have been running this for a while on our machines, here's a graph

Screenshot from 2020-11-05 10-43-31

The first significant bump is when we update cadvisor to https://github.com/google/cadvisor/commits/6f30891d89cc5b9cf93634048eb9f77c748a1437 (without these patches) from 0.30,
the reduction is with these patches. As you see they are still consuming more than before, but i would accept that is from extra metrics in existing categories that we are collecting. Would be ideal to be able to further remove unused metrics, but it's an acceptable cost for us at the moment.

dims:
@bobbypage this looks legit to me :)

dqminh:
@dashpole ping again. Is there anything we can do here ?

dashpole:
This should be a new category of disable metrics, instead of a new flag.

container/libcontainer/handler.go
```diff
@@ -276,15 +278,17 @@ func countFdAndSocket(dirPath string) (fdCount, socketCount uint64, err error) {
...
-	for _, name := range names {
-		fdPath := path.Join(dirPath, name)
-		linkName, err := os.Readlink(fdPath)
-		if err != nil {
-			klog.V(4).Infof("error while reading %q link: %v", fdPath, err)
-			continue
-		}
-		if strings.HasPrefix(linkName, "socket") {
-			socketCount++
+	if !noSocketCount {
```
kolyshkin, line 281: Perhaps add something like this instead?
```go
if noSocketCount
     return fdCount, socketCount, nil
}
```

eero-t:
GitHub action comments between the code state that there are undefined names being used.

> How expensive are socket stats? If we wanted to do this, we would need to add a new category for it in --disable_metrics, rather than a new flag.

I concur that this change is too insignificant to warrant its own option.  It should be part of -disable_metrics.

I think socket (count) item being dependent on process stats item is confusing to the users of that option though, and it should be made independent of process stats.

(If it's not made independent, then this PR documentation updates are incomplete, all places in docs mentioning process or socket stats should be updated to note that there's a dependency between their CLI options.)

Also, when there's a disable option for some item, outputting of the disabled data should also be skipped (e.g. in Prometheus output).  Currently that too is missing from this PR.

eero-t:
Here's (only build tested) code for adding "sock_count" to metrics one can use with -disable_metrics/-enable_metrics: https://github.com/eero-t/cadvisor/commits/sock-count-option

I'm not sure how its dependency on process stats option should be documented.

container/libcontainer/handler.go
```diff
@@ -276,22 +276,24 @@ func countFdAndSocket(dirPath string) (fdCount, socketCount uint64, err error) {
...
-	for _, name := range names {
-		fdPath := path.Join(dirPath, name)
-		linkName, err := os.Readlink(fdPath)
-		if err != nil {
-			klog.V(4).Infof("error while reading %q link: %v", fdPath, err)
-			continue
-		}
-		if strings.HasPrefix(linkName, "socket") {
-			socketCount++
+	if includeSocketCount {
```
kolyshkin, line 279: For less indentation / more readability, this can be 
```go
if !includeSocketCount {
	return
}
...
```

dqminh:
Ping again for review since it has been some time and i kinda forgot about this, sorry !

pacoxu:
This may help for fixing https://github.com/kubernetes/kubernetes/issues/99183. Is this PR still in progress?

PR #2657, opened 2020-09-03
Add /proc/vmstat metrics

kwisniewski98:
Add /proc/vmstat/ metrics to REST API and Prometheus
Signed-off-by: Wisniewski, Krzysztof2 <Krzysztof2.Wisniewski@intel.com>


machine/machine.go
```diff
@@ -122,6 +124,44 @@ func GetClockSpeed(procInfo []byte) (uint64, error) {
...
+		value, err := strconv.Atoi(splittedLine[1])
+		if err != nil {
+			klog.Infof("cannot parse vmstat's %s; %v. Skipping", splittedLine[0], err)
+			continue
+		}
+		stats[splittedLine[0]] = value
+
+	}
+	if len(stats) == 0 {
+		return nil, fmt.Errorf("no vmstat metrics found")
```
katarzyna-z, line 159: I think that info level will be enough for this case. What do you think?

dashpole:
Sorry for the slow review.  I'm not sure cAdvisor is the best place for this, as we primarily collect container-level metrics, or node configuration metrics, which we don't update frequently.

I think the prometheus node_exporter is probably a better place to collect this.  It looks like it already has vmstat metrics as well: https://github.com/prometheus/node_exporter/blob/master/collector/vmstat_linux.go.  Does that meet your needs?

PR #2663, opened 2020-09-07
Add Nodes "numastat" and "vmstat" metrics.

Creatone:


cmd/cadvisor.go
```diff
@@ -141,7 +143,7 @@ func (ml *metricSetValue) Set(value string) error {
 }
 
 func init() {
-	flag.Var(&ignoreMetrics, "disable_metrics", "comma-separated list of `metrics` to be disabled. Options are 'accelerator', 'cpu_topology','disk', 'diskIO', 'memory_numa', 'network', 'tcp', 'udp', 'percpu', 'sched', 'process', 'hugetlb', 'referenced_memory', 'resctrl'.")
+	flag.Var(&ignoreMetrics, "disable_metrics", "comma-separated list of `metrics` to be disabled. Options are 'accelerator', 'cpu_topology','disk', 'diskIO', 'memory_numa', 'network', 'tcp', 'udp', 'percpu', 'sched', 'process', 'hugetlb', 'referenced_memory', 'resctrl', 'numa_vmstat'.")
```
katarzyna-z, line 146: In documentation there is information that `memory_numa` option is used to enable "machine_node_numastat_pages" metric but I don't see this option here

katarzyna-z:
In another [pull request](https://github.com/google/cadvisor/pull/2657/files#diff-554ae30fe65ce5e268ad0f0bfcd9a084R43) there is introduced additional parameter for metrics from /proc/vmstat. Do you think that this parameter should be also used for vmstat metrics per numa node?

docs/storage/prometheus.md
```diff
@@ -106,8 +106,10 @@ Metric name | Type | Description | Unit (where applicable) | -disable_metrics pa
 `machine_dimm_capacity_bytes` | Gauge | Total RAM DIMM capacity (all types memory modules) value labeled by dimm type,<br>information is retrieved from sysfs edac per-DIMM API (/sys/devices/system/edac/mc/) introduced in kernel 3.6 | bytes | | |
 `machine_dimm_count` | Gauge | Number of RAM DIMM (all types memory modules) value labeled by dimm type,<br>information is retrieved from sysfs edac per-DIMM API (/sys/devices/system/edac/mc/) introduced in kernel 3.6 | | |
 `machine_memory_bytes` | Gauge | Amount of memory installed on the machine | bytes | |
-`machine_node_hugepages_count` | Gauge |  Numer of hugepages assigned to NUMA node | | cpu_topology |
+`machine_node_hugepages_count` | Gauge |  Number of hugepages assigned to NUMA node | | cpu_topology |
 `machine_node_memory_capacity_bytes` | Gauge |  Amount of memory assigned to NUMA node | bytes | cpu_topology |
+`machine_node_numastat_pages` | Gauge | Number of memory hit/miss statistics per NUMA node. | | memory_numa |
```
katarzyna-z, line 111: I see that you've used existing options for disable_metrics parameter ("memory_numa") , consider adding a new option as number of metrics is significantly different  when you enable "container_memory_numa_pages" metric (enabled by "memory_numa" option), which is collected per cgroup.

utils/sysinfo/sysinfo.go
```diff
@@ -523,3 +537,31 @@ func GetSocketFromCPU(topology []info.Node, cpu int) int {
...
+
+	if len(statsLines) == 0 {
+		return nil, nil
+	}
+
+	stats := make(map[string]int)
+	for _, line := range statsLines {
+		if nodeStatRegExp.Match([]byte(line)) {
+			words := strings.Split(line, " ")
+			value, err := strconv.Atoi(words[1])
```
katarzyna-z, line 556: Please check length of  `words` before accessing an element with index 1.

utils/sysinfo/sysinfo.go
```diff
@@ -523,3 +537,31 @@ func GetSocketFromCPU(topology []info.Node, cpu int) int {
...
+			if err != nil {
+				return nil, err
+			}
+			stats[words[0]] = value
+		} else {
+			return nil, fmt.Errorf("cannot obtain NUMA node stats from %q", nodeStatPath)
+		}
+	}
+
```
katarzyna-z, line 565: I suggest adding log (info level) when length of stats is equal to zero.

utils/sysinfo/sysinfo.go
Creatone: This is unnecessary as nodeStatRegExp already does.

dashpole:
See https://github.com/google/cadvisor/pull/2657#issuecomment-690852001.  I'm not sure this belongs in cAdvisor

PR #2665, opened 2020-09-07
add clean operation when watchForNewContainers/Start failed

nicktming:
when watchForNewContainers and rawContainerWatcher  Start fails and return err, but it does not have some clean operations since some ContainerWatchers already execute the Start() functions. so this commit tries to add clean operations when this happens.


nicktming:
@googlebot I signed it!

dashpole:
/ok-to-test

would you mind adding a description of why we want to make the change?

nicktming:
> /ok-to-test
> 
> would you mind adding a description of why we want to make the change?

ok.

dashpole (approved):
lgtm

PR #2677, opened 2020-09-23
Fix cpu discovery on custom kernel builds

odinuge:
When linux is compiled with CONFIG_HOTPLUG_CPU=n the "online" and
"offline" file does not exist. In that case, we know that all cpus will
be online.


odinuge:
/hold

Turns out this code is duplicated. Will look into it

odinuge:
Closed in favor of #2744

PR #2687, opened 2020-10-02
Move mount library dependency from utils/mount to mount-utils

brahmaroutu:
This PR is updating the dependency for the mount library. We have recently moved utils/mount to mount-utils under the repo https://github.com/kubernetes/mount-utils. This library is now available for kubernetes/kubernetes components under staging/src/k8s.io/mount-utils. Go module being k8s.io/mount-utils
Going forward, mount library will not be maintained under the repo https://github.com/kubernetes/utils.


brahmaroutu:
Not sure who could review, but looking at the latest commits, @dashpole @iwankgb can you please help here? 
This PR is important for us to remove dependency on utils/mount. This is the only module in kubernetes/kubernetes which is dependent on utils/mount and blocking our progress.

iwankgb:
@brahmaroutu, I had no idea that I can use commands here.

iwankgb:
lgtm

iwankgb:
@brahmaroutu I do not have all the superpowers, though. I can't merge it :(

dashpole (approved):
lgtm

brahmaroutu:
Awesome, thank you @iwankgb and @dashpole for the review.

dashpole:
@brahmaroutu we can't actually depend on a staging library in cAdvisor since cAdvisor is vendored into the kubelet.  This will have to be reverted.

PR #2689, opened 2020-10-02
Increase the readability of perf event logs.

Creatone:

### DIFF
--- a/perf/uncore_libpfm.go
+++ b/perf/uncore_libpfm.go
@@ -151,8 +151,7 @@ func NewUncoreCollector(cgroupPath string, events PerfEvents, cpuToSocket map[in
 
 	err := collector.setup(events, systemDevicesPath)
 	if err != nil {
-		formatedError := fmt.Errorf("unable to setup uncore perf event collector: %v", err)
-		klog.V(5).Infof("Perf uncore metrics will not be available: %s", formatedError)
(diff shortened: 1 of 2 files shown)


Creatone:
@bobbypage can you check this one ?

bobbypage:
LGTM, thanks.

perf/uncore_libpfm.go
```diff
@@ -151,8 +151,7 @@ func NewUncoreCollector(cgroupPath string, events PerfEvents, cpuToSocket map[in
 
 	err := collector.setup(events, systemDevicesPath)
 	if err != nil {
-		formatedError := fmt.Errorf("unable to setup uncore perf event collector: %v", err)
-		klog.V(5).Infof("Perf uncore metrics will not be available: %s", formatedError)
+		klog.Errorf("Perf uncore metrics will not be available: unable to setup uncore perf event collector: %s", err)
```
bobbypage, line 154: shouldn't we use %v instead of %s here for the error?

bobbypage (changes requested):
small nit regarding the formatting of error, otherwise LGTM

perf/uncore_libpfm.go
Creatone: Yes

bobbypage (approved):

PR #2691, opened 2020-10-04
Broken build configuration when custom build tags are used

iwankgb:
The problem seems to be broader then I expected. Apparently not only static analysis was ignoring build tags but also integration tests. This PR seems to be pretty critical to build pipeline, I think.


iwankgb:
@dashpole I think it works now. Review will be appreciated :)

dashpole (approved):

Issue #2692, opened 2020-10-06
Perf: support uncore pmu from libpfm library


Creatone:
It must be possible to use PMU name from libpfm library in uncore configuration:

Example of event from libpfm:
`skx_unc_imc0::UNC_M_CLOCKTICKS`

More info on the "EVENT STRINGS" section here: https://man7.org/linux/man-pages/man3/libpfm.3.html

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2693, opened 2020-10-06
[WIP] Support uncore pmu from libpfm library

Creatone:
#2692


Creatone:
The Libpfm interface does not allow to obtain of PMU names.

PR #2706, opened 2020-10-30
Add a badge for the current e2e test result

dashpole:
This should just display the results from the most recently merged PR if I understand it correctly.


bobbypage (approved):

bobbypage:
nice, LGTM!

PR #2707, opened 2020-11-02
[Fix] Avoid random values in unix.PerfEventAttr{}

Creatone:
To avoid unix.PerfEventAttr with random fields from `readPerfEventAttr`, fill memory with 0 values.
This fixes getting an "invalid argument" when opening perf events.

@kwisniewski98 @iwankgb @katarzyna-z


Creatone:
@bobbypage can you look at this?

bobbypage (approved):

PR #2672, opened 2020-09-15
Cherry-pick #2490  Make a copy of MachineInfo in GetMachineInfo() to v0.35

odinuge:
Ref. discussion in k/k here: https://github.com/kubernetes/kubernetes/pull/94176

PR #2708, opened 2020-11-04
Fix pull-cadvisor-e2e test due outdated asset file

bobbypage:
Ran:

```
FORCE=true build/assets.sh
```

it looks like `AssetDebug` was added in https://github.com/google/cadvisor/pull/2599 but `cmd/internal/pages/static/assets.go` wasn't updated.

This should fix the test failure on https://github.com/google/cadvisor/pull/2672


dashpole:
I think you will have to change it in the v0.35 branch to unblock that PR.  Its failing because it is the first PR to that branch in 2020

bobbypage:
yup you're right, the fix there is on 0.35, this shouldn't be needed as we discussed.

PR #2709, opened 2020-11-04
break assets to make sure tests work

dashpole:
This should fail tests
@bobbypage


bobbypage:
hmm, looks like this failed the lint tests but `pull-cadvisor-e2e` is still green...

PR #2710, opened 2020-11-05
Fix image location in quickstart section

jeteon:
The published image location no longer works but I found one that does in one of the releases. Hopefully it saves someone some time in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ sudo docker run \
   --name=cadvisor \
   --privileged \
   --device=/dev/kmsg \
-  gcr.io/cadvisor/cadvisor:$VERSION
+  gcr.io/google_containers/cadvisor:$VERSION
 ```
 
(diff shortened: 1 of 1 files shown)


bobbypage:
We've actually moved new versions from `gcr.io/google_containers/` to `gcr.io/cadvisor/cadvisor/`

`gcr.io/cadvisor/cadvisor` should include the following versions:

```
$ gcloud container images list-tags gcr.io/cadvisor/cadvisor
DIGEST        TAGS     TIMESTAMP
6f1225883baf  v0.35.1  2020-11-05T21:48:34
aa9694377d3b  v0.37.0  2020-07-08T01:54:56
c549e9b4b419  v0.35.0  2020-06-20T00:06:43
5ee8e2734bb7  v0.36.0  2020-06-19T23:59:55
```

are you using an older version than those?

PR #2713, opened 2020-11-06
Bump golang to 1.15

kolyshkin:
Go 1.15 was released in August, so let's use it where possible.
    
Go 1.13 is not supported since Go 1.15 became available,
so let's not use it for CI (but keep testing on 1.14).


kolyshkin:
@iwankgb @dashpole @bobbypage PTAL

bobbypage:
LGTM, thanks for the update @kolyshkin.

Perhaps we should update to 1.15 everywhere, not just for CI tests? Here's the other references I found of 1.13/1.14:

* https://github.com/google/cadvisor/blob/master/Makefile#L89
* https://github.com/google/cadvisor/blob/master/build/integration-in-docker.sh#L73
* https://github.com/google/cadvisor/blob/master/build/unit-in-container.sh#L47
* https://github.com/google/cadvisor/blob/master/deploy/canary/Dockerfile#L1
* https://github.com/google/cadvisor/blob/master/cmd/go.mod#L3
* https://github.com/google/cadvisor/blob/master/go.mod#L3

@dashpole thoughts?

dashpole:
I agree; consistency is a good idea

kolyshkin:
> Perhaps we should update to 1.15 everywhere, not just for CI tests

Done. No touching go.mod or docs, for obvious reasons.

bobbypage:
Thanks! And just to understand what's the negative implications of updating `go.mod` files as well? For example, it looks like in kubernetes as part of upgrade from 1.13 -> 1.15, updated their go.mod files: https://github.com/kubernetes/kubernetes/commit/247fcdfda5311bfc446c61c7bda67803e3527fdd

kolyshkin:
> what's the negative implications of updating go.mod files as well?

To my best knowledge (not an expert here), `go VERSION` line in `go.mod` serves two purposes:
1. denotes the minimal version requirement for the package
2. might be used by some golang tools to alter their behavior to stay backward-compatible

As for 1, it seems the current code is fine to be used with go 1.14, and a new golang release is not changing that fact.

As for 2, I think I saw that something like `go vet` becomes more strict when the version in go.mod is bumped, but it does not look like a good enough reason to me to bump it.

iwankgb:
Looks good to me and there is no reason to keep using 1.13. My understanding of `go.mod` is identical to @kolyshkin's.

dims (approved):
LGTM

bobbypage:
@kolyshkin got it, thanks for the clarification regarding go version update in `go.mod` files.

bobbypage:
Thanks again for the update! LGTM.

bobbypage (approved):

PR #2723, opened 2020-11-10
do not initialize libipmctl package when getting an error from nvm_init()

### LINKED ISSUE #2721: Libipmctl package initializes even if there is an error.
https://github.com/google/cadvisor/blob/ca49536c2a019f735d24b99471cea54200ef22fa/nvm/machine_libipmctl.go#L35-L45

If there is an error with libipmctl initialization it's only printed and `isNVMLibInitialized` is always set.

Creatone:
Fixes #2721

@iwankgb 


iwankgb:
Looks more then reasonable to me.

bobbypage (approved):

PR #2726, opened 2020-11-11
Revert mount-utils back to utils/mount

dims:
@brahmaroutu we need to revert the change in https://github.com/google/cadvisor/pull/2687 because we are causing a recursion

- mount-utils is in staging 
- k/k uses cadvisor
- cadvisor uses the k8s.io/mount-utils

So this is not going to work. So switching cadvisor back to `k8s.io/utils/mount`


brahmaroutu:
@dims  Sorry if this caused recursion. We already moved utils/mount to staging. With cAdvisor using util/mount we will not be able to retire this package. Can you please explain why third party software cannot use staging modules?

dims:
third party software can use staging modules IFF the third party software is not vendored back into k/k :)

dims:
we have to live with k8s.io/utils/mount or copy that stuff over to cadvisor or find another library which is equivalent.

brahmaroutu:
Let us address this in 1.21 ASAP. Since 90% of the work to move utils/mount into staging is already done, I would like to make this work without maintaining multiple copies of the mount library and if possible, without retreating the changes made so far.

bobbypage (approved):

PR #2734, opened 2020-11-18
Cherrypick to v0.37 - update docker client method

bobbypage:
* Cherrypick  update docker client method #2714 to v0.37
* Bump docker to https://github.com/moby/moby/tree/v19.03.8 to match that of k8s 1.19 - https://github.com/kubernetes/kubernetes/blob/release-1.19/go.mod#L39 
   ```
   github.com/docker/docker v1.4.2-0.20200309214505-aa6a9891b09c
   ```

### DIFF
--- a/cmd/go.sum
+++ b/cmd/go.sum
@@ -59,8 +59,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible h1:dvc1KSkIYTVjZgHf/CTC2diTYC8PzhaA5sFISRfNVrE=
 github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible/go.mod h1:J2gT2udsDAN96Uj4KfcMRqY0/ypR+oyYUYmja8H+y+w=
-github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0 h1:w3NnFcKR5241cfmQU5ZZAsf0xcpId6mWOupTvJlUX2U=
-github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0/go.mod h1:eEKB0N0r5NX/I1kEveEz05bcu8tLC/8azJZsviup8Sk=
+github.com/docker/docker v1.4.2-0.20200309214505-aa6a9891b09c h1:zviRyz1SWO8+WVJbi9/jlJCkrsZ54r/lTRbgtcaQhLs=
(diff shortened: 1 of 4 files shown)


bobbypage:
@JornShen it looks like there's some issues cherrypicking https://github.com/google/cadvisor/pull/2714 to v0.37:

From CI:

```
[/home/runner/work/cadvisor/cadvisor/container/docker/client.go:58:12: WithAPIVersionNegotiation not declared by package client -: could not load export data: no export data for \"github.com/google/cadvisor/container/docker\"]"
```

I think the problem is `WithAPIVersionNegotiation` was added in (https://github.com/moby/moby/commit/b26aa97914dab997d6f8ebf1cb9c019f97e3c254) looks like but cAdvisor v0.37 is using an older Docker version with hash `be7ac8be2ae0` [1] that doesn't have that method: https://github.com/moby/moby/blob/be7ac8be2ae072032a4005e8f232be3fc57e4127/client/options.go

[1] From v0.37 `go.mod`:
```
github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0
```

bobbypage:
@JornShen is there some way we can fix https://github.com/google/cadvisor/pull/2714 with the current version of docker on the v0.37 branch?

JornShen:
> @JornShen is there some way we can fix #2714 with the current version of docker on the v0.37 branch? I think we should avoid bumping docker version on an existing cherrypick version.

It's related to docker client version.  Is there problem for us to update the docker version here? 
If no, I think we can only update cadvisor dependency of k/k to v0.38.X inrelease-1.19 .
Is cadvisor version  strong binding to k8s version? like  release-1.19 must use  v0.37?

JornShen:
@dims  is there some method for us to  use which is same effect as `WithAPIVersionNegotiation` in old docker version?

dims:
@JornShen when you use `WithAPIVersionNegotiation` the newer client can talk to older servers.

liggitt:
was the late-initializing fix verified? I'm not seeing where setup() is retried. xref https://github.com/google/cadvisor/pull/2732#issuecomment-729892968

JornShen:
> @JornShen when you use `WithAPIVersionNegotiation` the newer client can talk to older servers.

@dims  I meam, are there other method which is same effect as `WithAPIVersionNegotiation` in old docker client?
@bobbypage  dont want to bump the docker version.In v0.37 version, cadvisor use old docker client `v0.7.3-0.20190327010347-be7ac8be2ae0` which dont have method `WithAPIVersionNegotiation`.

bobbypage:
@JornShen I looked at this again, I think it should be ok to update docker version. 

My concern is that when we go to vendor the cherrypicked cadvisor back into k/k we want to avoid bumping the docker version since cherrypick should ideally be as small as possible. 

However, k8s 1.19 seems to be on newer docker already compared to cadvisor v0.37:

From k8s 1.19 go.mod:

```
github.com/docker/docker v1.4.2-0.20200309214505-aa6a9891b09c
```
https://github.com/kubernetes/kubernetes/blob/release-1.19/go.mod#L39

However cAdvisor v0.37 is using:
```
github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0
```
https://github.com/google/cadvisor/blob/release-v0.37/go.mod#L17

Since k8s 1.19 is already using newer docker version (that also includes `WithAPIVersionNegotiation`) I think it should be fine to bump docker version in cadvisor v0.37 to match that in k8s 1.19.

PR #2744, opened 2020-11-29
Handling arm64: topology and online information

### LINKED ISSUE #2743: Cannot read number of sockets correctly, number of sockets set to 0 (ARM64)
I know ARM is not officially supported (notice I'm not using the official image either), so no expectations here at all for an actual fix.

Running Ubuntu 20.04 LTS on a Raspberry Pi 4B 8 GB in case that helps.

```sh
$ uname -a
Linux 5.4.0-1022-raspi #25-Ubuntu SMP PREEMPT Thu Oct 15 13:31:49 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

```yml
$ cat docker-compose.yml

...
  cadvisor:
    image: zcube/cadvisor:v0.37.0
    ports:
    - target: 8080
      published: 8080
      mode: host
    volumes:
    - /:/rootfs:ro
    - /var/run:/var/run:rw
    - /sys:/sys:ro
    - /var/lib/docker/:/var/lib/docker:ro
...
```

```sh
docker logs cadvisor_1
W1128 17:43:05.535985       1 nvidia.go:61] NVIDIA GPU metrics will not be available: no NVIDIA devices found
W1128 17:43:05.555422       1 sysinfo.go:203] Nodes topology is not available, providing CPU topology
W1128 17:43:05.556342       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu0/online: open /sys/devices/system/cpu/cpu0/online: no such file or directory
W1128 17:43:05.556449       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu1/online: open /sys/devices/system/cpu/cpu1/online: no such file or directory
W1128 17:43:05.556522       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu2/online: open /sys/devices/system/cpu/cpu2/online: no such file or directory
W1128 17:43:05.556619       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu3/online: open /sys/devices/system/cpu/cpu3/online: no such file or directory
E1128 17:43:05.556835       1 info.go:114] Failed to get system UUID: open /etc/machine-id: no such file or directory
W1128 17:43:05.558397       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:43:05.558482       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:43:05.558534       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:43:05.558593       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:43:05.558614       1 machine.go:72] Cannot read number of physical cores correctly, number of cores set to 0
W1128 17:43:05.558894       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:43:05.558948       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:43:05.558992       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:43:05.559045       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:43:05.559064       1 machine.go:86] Cannot read number of sockets correctly, number of sockets set to 0
W1128 17:43:05.779355       1 manager.go:288] Could not configure a source for OOM detection, disabling OOM events: open /dev/kmsg: no such file or directory
W1128 17:48:06.166638       1 sysinfo.go:203] Nodes topology is not available, providing CPU topology
W1128 17:48:06.167769       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu0/online: open /sys/devices/system/cpu/cpu0/online: no such file or directory
W1128 17:48:06.167986       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu1/online: open /sys/devices/system/cpu/cpu1/online: no such file or directory
W1128 17:48:06.168103       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu2/online: open /sys/devices/system/cpu/cpu2/online: no such file or directory
W1128 17:48:06.168197       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu3/online: open /sys/devices/system/cpu/cpu3/online: no such file or directory
E1128 17:48:06.168379       1 info.go:114] Failed to get system UUID: open /etc/machine-id: no such file or directory
W1128 17:48:06.169090       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:48:06.169271       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:48:06.169346       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:48:06.169415       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:48:06.169438       1 machine.go:72] Cannot read number of physical cores correctly, number of cores set to 0
W1128 17:48:06.169837       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:48:06.169950       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:48:06.170007       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:48:06.170071       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:48:06.170090       1 machine.go:86] Cannot read number of sockets correctly, number of sockets set to 0
W1128 17:53:06.165974       1 sysinfo.go:203] Nodes topology is not available, providing CPU topology
W1128 17:53:06.166788       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu0/online: open /sys/devices/system/cpu/cpu0/online: no such file or directory
W1128 17:53:06.166881       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu1/online: open /sys/devices/system/cpu/cpu1/online: no such file or directory
W1128 17:53:06.166959       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu2/online: open /sys/devices/system/cpu/cpu2/online: no such file or directory
W1128 17:53:06.167033       1 sysfs.go:348] unable to read /sys/devices/system/cpu/cpu3/online: open /sys/devices/system/cpu/cpu3/online: no such file or directory
E1128 17:53:06.167165       1 info.go:114] Failed to get system UUID: open /etc/machine-id: no such file or directory
W1128 17:53:06.167613       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:53:06.167677       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:53:06.167724       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:53:06.167778       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:53:06.167799       1 machine.go:72] Cannot read number of physical cores correctly, number of cores set to 0
W1128 17:53:06.168038       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu0 online state, skipping
W1128 17:53:06.168090       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu1 online state, skipping
W1128 17:53:06.168133       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu2 online state, skipping
W1128 17:53:06.168185       1 machine.go:253] Cannot determine CPU /sys/bus/cpu/devices/cpu3 online state, skipping
E1128 17:53:06.168200       1 machine.go:86] Cannot read number of sockets correctly, number of sockets set to 0

```

### LINKED ISSUE #2783: cAdvisor ignores multiple sockets when calculating unique CPU properties
When trying to determine number of unique CPU properties (e.g. core_id) cAdvisor will silently ignore multi-NUMA or multi-socket systems and report cores as if only single NUMA node or socket existed

iwankgb:
Fixes #2743 (to some extent). Fixes #2783 too.

According to [kernel docs](https://www.kernel.org/doc/html/latest/core-api/cpu_hotplug.html#using-cpu-hotplug) CONFIG_HOTPLUG_CPU is the only way of disabling or enabling cpu hotplug. It should therefore be safe to assume, that if `/sys/devices/system/cpu/cpuXXX/online` file does not exist then hotplug is disabled.

The above might not be true on some systems, apparently. Raspberry Pi 4 (Cortex-A72) seems to be one such system. 

Besides this:
* `GetMachineMemoryByType()` and `getUniqueCPUPropertyCount()` have been moved to `utils/sysfs` where they belong (these functions read from sysfs).
* all the edited files have been reformatted automatically to make linter happy.


iwankgb:
@bobbypage it seems that this fix could make some folks in Kubernetes community merrier: https://github.com/kubernetes/kubernetes/issues/95039. Might be worth being reviewed.

utils/sysfs/sysfs_arm64.go
```diff
@@ -0,0 +1,19 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sysfs
+
+func isCPUOnline(_ string) bool {
+	return true
```
bobbypage, line 18: can we comment why arm64 == cpus are always online?

bobbypage:
Thanks, LGTM for changes here. 

Only question/concern: the changes here basically make the assumption that arm64 does need to check `cpu/cpuXXX/online` at all; cpus are always online for arm64. Is this a reasonable assumption? 

I don't know too much about cpu hotplug / arm64 support, but just wondering based on https://github.com/google/cadvisor/pull/2744#issue-529189022, is something weird with Raspberry Pi 4 specifically? Just thinking that maybe there's some other ARM64 machines that do support cpu hotplug and might need the online check? LMK what you think, and if this is actually a real concern.

bobbypage:
See also related in https://github.com/google/cadvisor/pull/2677 regarding if kernel complied with CONFIG_HOTPLUG_CPU=n

iwankgb:
I looked at Graviton2 EC2 instance yesterday and it behaves as expected: online file is there. Perhaps better way to fix it would be to assume that lack of online file means lack of hotplug capabilities, irrespective of architecture.

bobbypage:
Yeah, that's kinda what I was thinking, perhaps that's a better fix? WDYT?

iwankgb:
It will require pretty extensive testing on various architectures and configurations. We should at least test:
- Hotplug enabled on x86 and arm
- Hotplug disabled on x86 and arm
- Hyperthreading disabled on x86
- hyperthreading enabled on x86
- RPi that is weird an unusual

I will try to conduct arm testing on Graviton2 and perhaps some other arm that I will be able to get hold of. X86 should not be an issue. 

My original intention was to avoid changes to x86 code - this is why I decided to limit changes to arm only.

utils/sysfs/sysfs.go
```diff
@@ -335,25 +341,143 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+		klog.V(4).Infof("Cannot convert %s to int: %s", state, err)
+		return false
+	}
+	return realState == desiredState
+}
+
+// getOnlineStateID looks for line similar to `206: online` from /sys/devices/system/cpu/hotplug/states
+// and extracts number from the beginning.
+// See: https://www.kernel.org/doc/html/latest/core-api/cpu_hotplug.html#testing-of-hotplug-states
+func getOnlineStateID(cpuPath string) (int, error) {
```
iwankgb, line 429: I think that it would make sense to cache result of this function: it can't change when system is running, I believe. Do you think that using package-level variable to cache the output makes sense?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +341,143 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
-	cpuPath := fmt.Sprintf("%s/online", dir)
-	content, err := ioutil.ReadFile(cpuPath)
+func (fs *realSysFs) IsCPUOnline(cpuPath string) bool {
+	return isCPUOnline(cpuPath)
+}
+
+// isCPUOnline verifies if a CPU in online. Takes path to /sys/devices/system/cpu/cpu*/ directory as a parameter.
+func isCPUOnline(cpuPath string) bool {
+	if isx86() && isZeroCPU(cpuPath) {
+		// It *is* possible to enable CPU hotplug for CPU0 but it requires setting a boot flag.
```
iwankgb, line 351: I believe it is safe, but I can't promise it. Do you think we should check kernel boot params?

iwankgb:
@bobbypage I have to tested all configurations yet, but I'm confident that code works on RPi4 and Graviton2. Can you take a look and let me know if this approach is fine with you? I will proceed with further testing then.

utils/sysfs/sysfs.go
bobbypage: is there any reason we need special logic for cpu0?

bobbypage: or actually just looking at my 32 core ubuntu GCE VM:

```
root@ubuntu-big:~# find /sys/devices/system/cpu/ | grep "online"
/sys/devices/system/cpu/cpu21/online
/sys/devices/system/cpu/cpu11/online
/sys/devices/system/cpu/cpu9/online
/sys/devices/system/cpu/cpu7/online
/sys/devices/system/cpu/cpu28/online
/sys/devices/system/cpu/cpu18/online
/sys/devices/system/cpu/cpu5/online
/sys/devices/system/cpu/cpu26/online
/sys/devices/system/cpu/cpu16/online
/sys/devices/system/cpu/cpu3/online
/sys/devices/system/cpu/cpu24/online
/sys/devices/system/cpu/online
/sys/devices/system/cpu/cpu14/online
/sys/devices/system/cpu/cpu1/online
/sys/devices/system/cpu/cpu22/online
/sys/devices/system/cpu/cpu12/online
/sys/devices/system/cpu/cpu30/online
/sys/devices/system/cpu/cpu20/online
/sys/devices/system/cpu/cpu10/online
/sys/devices/system/cpu/cpu8/online
/sys/devices/system/cpu/cpu29/online
/sys/devices/system/cpu/cpu19/online
/sys/devices/system/cpu/cpu6/online
/sys/devices/system/cpu/cpu27/online
/sys/devices/system/cpu/cpu17/online
/sys/devices/system/cpu/cpu4/online
/sys/devices/system/cpu/cpu25/online
/sys/devices/system/cpu/cpu15/online
/sys/devices/system/cpu/cpu2/online
/sys/devices/system/cpu/cpu23/online
/sys/devices/system/cpu/cpu13/online
/sys/devices/system/cpu/cpu31/online
```

cpu0 online doesn't exist. is it why this case is needed?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +352,205 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
 }
 
-func (fs *realSysFs) IsCPUOnline(dir string) bool {
-	cpuPath := fmt.Sprintf("%s/online", dir)
-	content, err := ioutil.ReadFile(cpuPath)
+func (fs *realSysFs) IsCPUOnline(cpuPath string) bool {
+	return isCPUOnline(cpuPath)
+}
+
+// isCPUOnline verifies if a CPU in online. Takes path to /sys/devices/system/cpu/cpu*/ directory as a parameter.
```
bobbypage, line 359: aren't we actually checking `/sys/bus/cpu/devices/` which is `cpuBusPath`? 

what is difference between `/sys/bus/cpu/devices/` and `/sys/devices/system/cpu/cpu*/` ...?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +352,205 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+	const isOnline = 49
+	onlinePath := filepath.Join(cpuPath, "online")
+	onlineVal, err := ioutil.ReadFile(onlinePath)
+	// First, make sure that no error occurred.
 	if err != nil {
-		pathErr, ok := err.(*os.PathError)
-		if ok {
-			if errors.Is(pathErr.Unwrap(), os.ErrNotExist) && isZeroCPU(dir) {
+		// If error occurred and is os.PathError then it might be os.ErrNotExists.
+		pathErr, isPathError := err.(*os.PathError)
```
bobbypage, line 373: is it needed to do type assertion here? maybe can call os.IsNotExist directly, e.g. https://github.com/google/cadvisor/blob/master/machine/operatingsystem_unix.go#L41-L42

utils/sysfs/sysfs.go
```diff
@@ -335,25 +352,205 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+
+// isCPUOnline verifies if a CPU in online. Takes path to /sys/devices/system/cpu/cpu*/ directory as a parameter.
+func isCPUOnline(cpuPath string) bool {
+	if isx86() && isZeroCPU(cpuPath) {
+		// It *is* possible to enable CPU hotplug for CPU0 but it requires setting a boot flag.
+		// It is assumed it's a rare situation.
+		return true
+	}
+
+	const isOnline = 49
```
bobbypage, line 367: just for my own curiosity, why is 49? is is documented in kernel docs somewhere?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +352,205 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+	// Check if online file value is 1.
+	onlineVal = bytes.TrimSpace(onlineVal)
+	if len(onlineVal) == 0 || onlineVal[0] != isOnline {
+		klog.V(2).Infof("CPU %s is offline", cpuPath)
+		return false
+	}
+	return true
+}
+
+func isx86() bool {
```
bobbypage, line 398: i'm still a bit confused why this check is needed (see above comment), but if needed, wouldn't it better to use a build tag and have a amd64 and !not amd64 variant of the file or something like that?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +352,205 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
 func isZeroCPU(dir string) bool {
 	regex := regexp.MustCompile("cpu([0-9]*)")
 	matches := regex.FindStringSubmatch(dir)
 	return len(matches) == 2 && matches[1] == "0"
 }
+
+// isOnlineHotplug checks if CPU state matches online state from /sys/devices/system/cpu/hotplug/states.
+// It allows to determine if CPU is online on some ARM platforms such as RPi4.
+// See: https://www.kernel.org/doc/html/latest/core-api/cpu_hotplug.html#testing-of-hotplug-states
+func isOnlineHotplug(cpuPath string) bool {
```
bobbypage, line 416: just for my understanding, as i understand this function

1. looks at `cat /sys/devices/system/cpu/hotplug/states`, trying to find a line like

```
...
206: online
```

Then it checks `hotplug/state` on a cpu, e.g. 

```
cat /sys/devices/system/cpu/cpu0/hotplug/state
206
```

and checks if the number matches. What does "206" represent? if a cpu will be disabled is it expected the state not match anymore?

utils/sysfs/sysfs.go
bobbypage: if it can't change, I guess doesn't hurt to cache it. re using package level variable, I think it's fine, e.g. see this existing example: https://github.com/google/cadvisor/blob/master/machine/machine.go#L53

machine/machine.go
```diff
@@ -47,26 +42,19 @@ var (
...
-const sysFsCPUPhysicalPackageID = "physical_package_id"
-const sysFsCPUTopology = "topology"
-const memTypeFileName = "dimm_mem_type"
-const sizeFileName = "size"
+	machineArch = getMachineArch()
+	maxFreqFile = "/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq"
+)
 
 // GetPhysicalCores returns number of CPU cores reading /proc/cpuinfo file or if needed information from sysfs cpu path
 func GetPhysicalCores(procInfo []byte) int {
```
bobbypage, line 52: one additional question: I came across this function https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L813-L819 which also looks to get number of online cpus via calling sched_getaffinity(2). I'm not sure how that works with hotplug, but just curious if that would be helpful here...

utils/sysfs/sysfs.go
iwankgb: 49 is nothing more than 1, but as a byte rather than a string.

utils/sysfs/sysfs.go
iwankgb: In general you should not put cpu0 offline on x86 because it is used to boot the operating system. If it's unavailable then you might not be able to start your system. I am *extremely* surprised by what you shared about your VM: can you share your kernel boot params? Perhaps it's a case of disabled cpu0 and I bet you haven't altered your configuration.

utils/sysfs/sysfs.go
iwankgb: We are, you are right. `/sys/bus/cpu/devices/` consists of symlinks to `/sys/devices/system/cpu/cpu*/`, I think.

utils/sysfs/sysfs.go
iwankgb: Using build tag will make testing it a bit difficult unless we can test non-x86_64 build.

utils/sysfs/sysfs.go
iwankgb: On my RPi `206` represents online state but you mileage will vary: I bet it will be different on random x86 VM. This is why we need to parse `states` file first.

iwankgb: Digging a bit deeper: take a look at [`cpuhp_setup_state()`](https://elixir.bootlin.com/linux/latest/source/include/linux/cpuhotplug.h#L208) - it allows to register callbacks for entering and leaving particular CPU state. It use used across the kernel in more than 100 places. Entries in `states` file corresponds to second argument to this function. 
Some states are "generic" and are [predefined](https://elixir.bootlin.com/linux/latest/source/kernel/cpu.c#L1510).
Does the above answer your question?

machine/machine.go
iwankgb: I would dare to say that the function has an unfortunate name: it returns count of CPUs that current process can run on: they are undoubtedly online, but they might be affected by cpusets or isolcpus.

utils/sysfs/sysfs.go
bobbypage: This is just a standard ubuntu GCE VM, created as follows

```
$ gcloud beta compute instances create ubuntu-big --machine-type=e2-standard-32 --image-family=ubuntu-2004-lts --image-project=ubuntu-os-cloud --boot-disk-size=256GB --boot-disk-type=pd-ssd --zone=us-central1-c
```

```
$ uname -a
Linux ubuntu-big 5.4.0-1029-gcp #31-Ubuntu SMP Wed Oct 21 19:38:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

Here's KConfig from
```
sudo cat /boot/config-$(uname -r)
```
https://gist.github.com/bobbypage/3e65f6a6ad649f082760b9674bd02c5b

utils/sysfs/sysfs.go
bobbypage: got it :) might be more clear to define it maybe characters are or `const isOnline := byte('1')`

utils/sysfs/sysfs.go
iwankgb: Fixed.

utils/sysfs/sysfs.go
iwankgb: Fixed.

utils/sysfs/sysfs.go
iwankgb: I was, obviously, wrong. Fixed :)

iwankgb:
@bobbypage I think it's ready to be reviewed now :)

odinuge (commented):
Hey,

As @bobbypage said, I started looking at #2677 a while back trying to fix the same issue, but ended up stuck due to the amount of refactoring that was required to fix it. I will close my PR since this has better coverage.

When it comes to this change, I am not sure if I like the code variation between x86 and non-x86. I think a better approach is to use the information in the `/sys/devices/system/cpu/online` insted, even tho it would require some changes. By using that, we don't have to make guesses about what cpus are online when the files are missing, and it will also result in a single file read when looking for all available cores. Any thoughts on that?

Also, when it comes to adding all the sysfs-files to the repo, I'll defer that to some Googlers. However, it would be nice to also include a readme about how the data is dumped, and some more metadata about the system (eg. kernel, distro and the output of `uname -a`).

However thanks for the work and for all the testcases you created!

Nit: Would be really nice if you could rebase the changes to remove unnecessary merges and commits, so it would be easier to properly review all commits one by one, since this PR is so big.

iwankgb:
> When it comes to this change, I am not sure if I like the code variation between x86 and non-x86. I think a better approach is to use the information in the /sys/devices/system/cpu/online insted, even tho it would require some changes. By using that, we don't have to make guesses about what cpus are online when the files are missing, and it will also result in a single file read when looking for all available cores. Any thoughts on that?
I think it sounds reasonable, I will take a look.

utils/sysfs/sysfs.go
```diff
@@ -37,9 +38,16 @@ const (
...
 	s390xDevTree = "/etc" // s390/s390x changes
 
-	coreIDFilePath    = "/topology/core_id"
-	packageIDFilePath = "/topology/physical_package_id"
-	meminfoFile       = "meminfo"
+	coreIDFilePath       = "/topology/core_id"
+	packageIDFilePath    = "/topology/physical_package_id"
+	meminfoFile          = "meminfo"
+	sysFsCPUTopology     = "topology"
+	CPUPhysicalPackageID = "physical_package_id"
```
bobbypage, line 45: Would be great to leave some comments on what `physical_package_id` / `core_id` will represent, eg. xref: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu

```
core_id: the CPU core ID of cpu#. Typically it is the
		hardware platform's identifier (rather than the kernel's).
		The actual value is architecture and platform dependent.
```

```
	physical_package_id: physical package id of cpu#. Typically
		corresponds to a physical socket number, but the actual value
		is architecture and platform dependent.
```

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
-func (fs *realSysFs) IsCPUOnline(dir string) bool {
-	cpuPath := fmt.Sprintf("%s/online", dir)
-	content, err := ioutil.ReadFile(cpuPath)
+func (fs *realSysFs) IsCPUOnline(cpuPath string) bool {
+	return isCPUOnline(cpuPath)
+}
+
+// isCPUOnline verifies if a CPU in online. Takes path to /sys/devices/system/cpu/cpu*/ directory as a parameter.
+func isCPUOnline(cpuPath string) bool {
+	if isX86 && isZeroCPU(cpuPath) {
```
bobbypage, line 357: Can we add a comment that describes why the `isX86` check is needed?

As I understand the problem is that assumption that zero cpu is always online isn't true for some non x86 systems, e.g. from the rpi4 you collected:

```
$ cat utils/sysfs/testdata_rpi4/cpu0/online
cat: utils/sysfs/testdata_rpi4/cpu0/online: No such file or directory
```

side point: what is the reason that we need to special case cpu0 again? For example, even if cpu0 online file doesn't exist (like on rpi4), would you be able to rely on the backup hotplug check? It seems like it's there, for example:

```
$ cat utils/sysfs/testdata_rpi4/hotplug/states  | grep -E "(\d)*: online$"
206: online
$ cat utils/sysfs/testdata_rpi4/cpu0/hotplug/state
206
```

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
-				return true
-			}
+		if !os.IsNotExist(err) {
+			// If online file exists but reading failed we assume that CPU is offline.
+			klog.V(2).Infof("Cannot determine CPU %s online state, assuming offline", cpuPath)
+			return false
 		}
-		klog.Warningf("unable to read %s: %s", cpuPath, err.Error())
+		// If online file does not exists then we should check one more location.
+		klog.V(2).Infof("%s not found, checking %s directory", onlinePath, filepath.Join(cpuPath, "hotplug"))
```
bobbypage, line 374: regarding the log: `isOnlineHotplug` below actually checks  `filepath.Join(cpuPath, "hotplug", "state")` not `filepath.Join(cpuPath, "hotplug")` right? is the logged path incorrect?

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+	return len(uniques)
+}
+
+// GetMachineMemoryByType returns information about memory capacity and number of DIMMs.
+// Information is retrieved from sysfs edac per-DIMM API (/sys/devices/system/edac/mc/)
+// introduced in kernel 3.6. Documentation can be found at
+// https://www.kernel.org/doc/Documentation/admin-guide/ras.rst.
+// Full list of memory types can be found in edac_mc.c
+// (https://github.com/torvalds/linux/blob/v5.5/drivers/edac/edac_mc.c#L198)
+func GetMachineMemoryByType(edacPath string) (map[string]*v1.MemoryInfo, error) {
```
bobbypage, line 494: I know you mentioned you moved `GetMachineMemoryByType` since it makes more sense in `utils/sysfs`, but I would prefer if we can do that in a followup separate PR since this PR is already quite large and focused specifically on CPU topology / online info

utils/sysfs/sysfs_test.go
```diff
@@ -134,10 +134,178 @@ func TestIsCPUOnline(t *testing.T) {
...
+	// Test on arch other than x86.
+	origIsX86 := isX86
+	defer func() {
+		isX86 = origIsX86
+	}()
+	isX86 = false
+	hotplugStates = "./testdata_rpi4/hotplug/states"
+
+	count := GetUniqueCPUPropertyCount("./testdata_rpi4/", CPUPhysicalPackageID)
+	assert.Equal(t, 1, count)
```
bobbypage, line 230: can we check `CPUCoreID` for RaspberryPi4 as well like the other tests?

utils/sysfs/sysfs_test.go
```diff
@@ -134,10 +134,178 @@ func TestIsCPUOnline(t *testing.T) {
...
+	isX86 = true
+
+	count := GetUniqueCPUPropertyCount("./testdata_epyc7402/", CPUPhysicalPackageID)
+	assert.Equal(t, 1, count)
+
+	count = GetUniqueCPUPropertyCount("./testdata_epyc7402/", CPUCoreID)
+	assert.Equal(t, 24, count)
+}
+
+func TestGetUniqueCPUPropertyCountOnEpyc7402NoHT(t *testing.T) {
```
bobbypage, line 248: nit: `s/NoHT/NoHotPlug`

utils/sysfs/sysfs_test.go
```diff
@@ -134,10 +134,178 @@ func TestIsCPUOnline(t *testing.T) {
...
+}
+
+func TestGetUniqueCPUPropertyCountOnRaspberryPi4(t *testing.T) {
+	// Test on arch other than x86.
+	origIsX86 := isX86
+	defer func() {
+		isX86 = origIsX86
+	}()
+	isX86 = false
+	hotplugStates = "./testdata_rpi4/hotplug/states"
```
bobbypage, line 227: does `hotplugStates` also need to be reset like isX86 since it's a package var and being overriden here?

utils/sysfs/sysfs.go
bobbypage: I see interesting, thanks for the info!

utils/sysfs/sysfs.go
```diff
@@ -37,9 +38,16 @@ const (
 	ppcDevTree   = "/proc/device-tree"
 	s390xDevTree = "/etc" // s390/s390x changes
 
-	coreIDFilePath    = "/topology/core_id"
-	packageIDFilePath = "/topology/physical_package_id"
-	meminfoFile       = "meminfo"
+	coreIDFilePath       = "/topology/core_id"
```
bobbypage, line 41: nit: suggest to reorder the consts and list them something like this to make it a bit more clear

```
sysFsCPUTopology     = "topology"

CPUCoreID            = "core_id"
CPUPhysicalPackageID = "physical_package_id"

coreIDFilePath       = "/" + sysFsCPUTopology "/" + CPUCoreID
packageIDFilePath    = "/" + sysFsCPUTopology + "/" + CPUPhysicalPackageID
```

bobbypage (changes requested):

bobbypage:
Thanks @iwankgb , left some comments. 

@odinuge intersting point regarding potentially looking `/sys/devices/system/cpu/online`, it sounds much simpler. I wonder if it will handle the hotplug and strange cases like in RPI4 and other ARM devices. Thoughts @iwankgb?

machine/machine.go
bobbypage: I see, got it, thanks for the info.

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
 	}
 }
 
-func (fs *realSysFs) IsCPUOnline(dir string) bool {
-	cpuPath := fmt.Sprintf("%s/online", dir)
-	content, err := ioutil.ReadFile(cpuPath)
+func (fs *realSysFs) IsCPUOnline(cpuPath string) bool {
```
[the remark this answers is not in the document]
iwankgb: Well, `IsCPUOnline()` is still a wrapper but it allows us to optimize some cases of online CPU detection. I'm not sure if we can freely remove this function as it is part of an interface.

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+	// First, make sure that no error occurred.
 	if err != nil {
-		pathErr, ok := err.(*os.PathError)
-		if ok {
-			if errors.Is(pathErr.Unwrap(), os.ErrNotExist) && isZeroCPU(dir) {
-				return true
-			}
+		if !os.IsNotExist(err) {
+			// If online file exists but reading failed we assume that CPU is offline.
+			klog.V(2).Infof("Cannot determine CPU %s online state, assuming offline", cpuPath)
```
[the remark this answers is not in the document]
iwankgb: We now assume online - as lack on `online` file indicates that CPU hotplug is disabled.

utils/sysfs/sysfs.go
```diff
@@ -335,25 +348,198 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+		if err != nil {
+			klog.Errorf("Cannot open %s, number of unique %s  set to 0", propertyPath, propertyName)
+			return 0
+		}
+		packagePath := filepath.Join(sysCPUPath, sysFsCPUTopology, CPUPhysicalPackageID)
+		packageVal, err := ioutil.ReadFile(packagePath)
+		if err != nil {
+			klog.Errorf("Cannot open %s, number of unique %s  set to 0", packagePath, CPUPhysicalPackageID)
+			return 0
+		}
```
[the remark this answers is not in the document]
iwankgb: Well, it was a bug, apparently: #2783. Core IDs will be identical on each CPU in multi-socket server.

bobbypage:
hey @iwankgb is this ready for another round to review? wasn't sure if it's WIP still?

iwankgb:
@bobbypage not yet.

bobbypage:
@iwankgb np, let me know when it's ready for review.

iwankgb:
@bobbypage it took a lot of time but now this very PR is ready to re-review. I consider it working.

iwankgb:
@bobbypage I have also squashed all the commits into something more manageable :)

utils/sysfs/sysfs_x86.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sysfs
+
+var isX86 = true
```
bobbypage, line 19: nit: sorry to go back and forth on this, but maybe as another idea instead of the `isX86` package variable, consider passing `arch` as an argument in `IsCPUOnline`. That way in the tests you can easily just pass `amd64` or `arm64` and in the calling code of `IsCPUOnline` use `runtime.GOARCH` during runtime. Totally up to you if you want leave it as is or do that approach, just wanted to bring it up as another option.

utils/sysfs/sysfs.go
```diff
@@ -335,25 +346,154 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+	return &online, nil
+}
+
+func (c *cpuList) isOnline(cpu uint16) bool {
+	if c.hotplugDisabled {
+		return true
+	}
+	// It *is* possible to enable CPU hotplug for CPU0 but it requires setting a boot flag.
+	// It is assumed it's a rare situation, so we assume CPU0 is always online..
+	if isX86 && cpu == 0 {
```
bobbypage, line 442: can you note why this check is only for x86?

bobbypage (commented):
thanks for all your work on this, a few minor comments but overall looks great!

utils/sysfs/sysfs.go
```diff
@@ -335,25 +346,154 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
 	}
 }
 
-func (fs *realSysFs) IsCPUOnline(dir string) bool {
-	cpuPath := fmt.Sprintf("%s/online", dir)
-	content, err := ioutil.ReadFile(cpuPath)
+func (fs *realSysFs) IsCPUOnline(cpuPath string) bool {
```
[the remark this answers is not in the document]
iwankgb: Godoc is provided in interface definition (as it is in case of other functions).

utils/sysfs/sysfs_x86.go
iwankgb: At the moment I prefer to leave it as it is.

iwankgb:
@bobbypage, let me know when you review it - I will squash the commits.

utils/sysfs/sysfs.go
```diff
@@ -378,16 +389,20 @@ func getCPUID(dir string) (uint16, error) {
 	return 0, fmt.Errorf("can't get CPU ID from %s", dir)
 }
 
+// cpuList allows to determine quickly if CPU or particular ID is online.
 type cpuList struct {
-	online          map[uint16]interface{}
+	// online maps CPU ID to a struct for each online CPU.
+	online map[uint16]struct{}
+	// hotplugDisabled is true when kernel CPU hotplug is disable and all CPUs are assumed online.
 	hotplugDisabled bool
```
bobbypage, line 397: thanks for adding the hot plug check upfront in `IsCPUOnline`. Now that we have that check can we remove this and the duplicate check in `newCPUList`?

utils/sysfs/sysfs.go
bobbypage: thanks!

machine/topology_test.go
```diff
@@ -87,7 +89,9 @@ func TestSocketsReadingFromCpuBus(t *testing.T) {
 }
 
 func TestSocketsReadingFromWrongSysFs(t *testing.T) {
-	cpuBusPath = "./testdata/"           // overwriting package variable to mock sysfs
+	path, err := filepath.Abs("./testdata/sysfs_cpus/")
+	assert.NoError(t, err)
+	cpuBusPath = path                    // overwriting package variable to mock sysfs
```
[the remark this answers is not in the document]
bobbypage: thanks!

utils/sysfs/sysfs.go
```diff
@@ -335,25 +346,154 @@ func (fs *realSysFs) GetSystemUUID() (string, error) {
...
+		if err != nil {
+			return 0, err
+		}
+		return uint16(id), nil
+	}
+	return 0, fmt.Errorf("can't get CPU ID from %s", dir)
+}
+
+type cpuList struct {
+	online          map[uint16]interface{}
```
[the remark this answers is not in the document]
bobbypage: thanks!

iwankgb:
@bobbypage it would be great if we finished this PR. Kubernetes users area still complaining about problems it fixes.

bobbypage:
LGTM, looks great, please squash and we'll be good to go!

bobbypage (approved):

bobbypage:
thanks for all your hard work on this!! LGTM

PR #2749, opened 2020-12-03
Fix timeout flooding issue after containerd restart

### LINKED ISSUE #2748: cpu saturated after containerd.service restart
This issue was found during the investigation of https://github.com/kubernetes/kubernetes/issues/95727.

Description:
After cadvisor manager starts, if containerd got restarted, it will get into a [loop](https://github.com/containerd/containerd/blob/25daa7355cdc1334718ce42e887f1c367aff9d0e/pkg/dialer/dialer.go#L51-L64) and flooding timeout.

Reproduce:
To reproduce the issue, you just need to create a cadvisor manager and start it, then restart containerd. For simplicity, created a [repo](https://github.com/hanlins/k8s-95727), and you can reproduce the issue on a environment with containerd running.

hanlins:
Originally the gRPC backoff multiplier is not set so it got defaulted as zero. Once a connection timeout happens, gRPC exponential backoff will be triggered which results in setting the backoff durating to zero, causing retries flooding. Setting gRPC multiplier to 1.6 which is the default one from gRPC package will solve the issue.

Closes #2748

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -65,11 +65,10 @@ func Client(address, namespace string) (ContainerdClient, error) {
 		tryConn.Close()
 
 		connParams := grpc.ConnectParams{
-			Backoff: backoff.Config{
-				BaseDelay: baseBackoffDelay,
-				MaxDelay:  maxBackoffDelay,
-			},
+			Backoff: backoff.DefaultConfig,
 		}
+		connParams.Backoff.BaseDelay = baseBackoffDelay
+		connParams.Backoff.MaxDelay = maxBackoffDelay
 		gopts := []grpc.DialOption{
 			grpc.WithInsecure(),
 			grpc.WithContextDialer(dialer.ContextDialer),


hanlins:
The [fix](https://github.com/hanlins/k8s-95727/commit/f5e778f5fbfa7f591444b2dafaeecb827e2172e7) has been manually verified. You can use [this branch](https://github.com/hanlins/k8s-95727/tree/fix-dialer) to build a binary for verification.

dims:
Cc @bobbypage

bobbypage:
Thanks for the fix and the investigation here! I'm curious, it looks this dialer with used for a long time now.

The last change was https://github.com/google/cadvisor/commit/4cf2f01ac82fb67769ced1c633fa7bfce87e71df#diff-ad4070b5c5ba5108a8790984457dd485f6363be34312edf452967359bb1a195dR75 which changed from using from `dialer.Dialer` -> `dialer.ContextDialer`. 

Looking at containerd dialer (https://github.com/containerd/containerd/blame/master/pkg/dialer/dialer.go), `dialer.ContextDialer` looks to be a wrapper around `dialer.Dailer, so I think those should be the same?

Only interesting, the containerd client is using `dialer.ContextDailer`: https://github.com/containerd/containerd/blob/c8b14ae4c01e620dc84704dd4b6a080eed0dc62e/client.go#L123

Just trying to understand the issue, is #2748 tied to a specific cAdvisor / k8s release? I'm curious if you think this might be a problem with containerd's  `dialer.ContextDailer` which is too aggressive with restarts and perhaps that needs to be fixed in containerd?

hanlins:
Hi @bobbypage, I think the issue becomes obvious when k8s directly talk to containerd instead of docker shim (I tried to just restart docker and there's no such issue, [here](https://github.com/kubernetes/kubernetes/issues/95727#issuecomment-716990852)'s my findings). Also there's CVE for containerd recently so people have to restart containerd, so more people are experiencing the issue just recently. The issue was reported for `v1.19.3`, but I can reproduce it in latest master.

> Just trying to understand the issue, is #2748 tied to a specific cAdvisor / k8s release?

No, it seems I can reproduce it on multiple k8s versions.

> I'm curious if you think this might be a problem with containerd's `dialer.ContextDailer` which is too aggressive with restarts and perhaps that needs to be fixed in containerd?

I'm not confident on that, I'm leaning on there's a context misuse in `containerd`'s `ContextDialer`, in which the `deadline` is calculated only once, and then the [loop](https://github.com/containerd/containerd/blob/v1.4.3/pkg/dialer/dialer.go#L51-L64) will just flood the timeout (because the deadline has already reached and not updated anymore). Also, the [containerd client](https://github.com/google/cadvisor/blob/v0.38.5/container/containerd/client.go) is a singleton in cAdvisor repo, all these combination caused the goroutine leak which flood the error to the connection channel of that singleton.

I would say it's not realistic to change the containerd client from singleton to per-connection based, and there might not be a convincing reason to make the change in `containerd` repo as they're fine with the deadline assumptions made there. Considering the necessity and the release cycle, I think maybe it's better to just make the change in cadvisor without touching `containerd`, what do you think? @bobbypage

bobbypage:
Thanks @hanlins for all your hard work on this! 

For context, we chatted offline on slack to debug this further. Definitely an fun and interesting bug, requiring some deep diving into grpc backoff logic :) 

@hanlins can you please update the PR description with the new findings we discussed offline and the updated fix? If you can squash the outdated commits, it would be great as well. Thanks again for all your efforts on hunting down this issue!

hanlins:
@bobbypage thanks! It's great that we root caused this "gotcha", really worthy to share it with other folks, thanks for your helps!

bobbypage (approved):

bobbypage:
I've cut new cAdvisor releases ([`v0.37.3`](https://github.com/google/cadvisor/releases/tag/v0.37.3) and [`v0.38.6`](https://github.com/google/cadvisor/releases/tag/v0.38.6)) with this fix cherrypicked.

We should get this fix back into kubernetes.

1. k/k master to update cAdvisor to v0.38.6
2. k/k to release-1.20 to update cAdvisor to v0.38.6
3. k/k to release-1.19 to update cAdvisor to v0.37.3

PR #2751, opened 2020-12-04
Refactor process parsing to accommodate commands with spaces

### LINKED ISSUE #2750: Spaces in command name
According to `man ps`:
> The following user-defined format specifiers may contain spaces: args, cmd, comm, command, fname, ucmd, ucomm, lstart, bsdstart, start.

Therefore `ps` output parsing (using `strings.Fields()`) is broken and is not capable of handling such string:
```
root         107       2 21:34  0.0  0.0     0      0 I<   00:00:00 DWC Notificatio   3 -
```

### LINKED ISSUE #2770: memory cgroup controller disabled by default on some systems
On some systems (such as Raspberry Pi 4) memory cgroup controller is disabled by default. We need to check another controller to determine cgroup path rather than return `"/"`.

iwankgb:
Fixes #2750. Fixes #2770.

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -38,7 +38,7 @@ import (
 	"github.com/google/cadvisor/summary"
 	"github.com/google/cadvisor/utils/cpuload"
 
-	units "github.com/docker/go-units"
+	"github.com/docker/go-units"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 )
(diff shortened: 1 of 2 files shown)


Creatone:
@bobbypage

manager/container.go
```diff
@@ -331,20 +335,28 @@ func (cd *containerData) GetProcessList(cadvisorContainer string, inHostNamespac
...
+		var psr int
+		psr, err = strconv.Atoi(fields[fieldNo])
 		if err != nil {
-			return nil, fmt.Errorf("invalid pid %q: %v", fields[1], err)
+			fieldNo++
+			psr, err = strconv.Atoi(fields[fieldNo])
+			if err != nil {
+				return nil, fmt.Errorf("invalid pid %q: %v", fields[1], err)
+			}
+			cmd += fmt.Sprintf(" %s", fields[11])
```
bobbypage, line 350: why is this necessary?

manager/container.go
```diff
@@ -331,20 +335,28 @@ func (cd *containerData) GetProcessList(cadvisorContainer string, inHostNamespac
...
 		// convert to bytes
 		vs *= 1024
-		psr, err := strconv.Atoi(fields[11])
+		fieldNo := 11
+		var psr int
+		psr, err = strconv.Atoi(fields[fieldNo])
 		if err != nil {
-			return nil, fmt.Errorf("invalid pid %q: %v", fields[1], err)
+			fieldNo++
+			psr, err = strconv.Atoi(fields[fieldNo])
```
bobbypage, line 346: If I'm understanding the logic here, the idea is that the cmd should be at index 10 and the psr should be at index 11 ASSUMING there is no spaces in the cmd name (this was old logic).

The fix here is basically assume that cmd==index10 and psr==index11, but if you parse as index11 and get error from `strconv.Atoi` it means you have a space in the cmd name. In that case, it means psr is at some greater index than 11 and keep advancing `fieldNo` until you don't get an error from `strconv.Atoi`.

Am I understanding the workaround here correctly? If so, I feel like this a bit fragile... what if the command name has a space and the string after the space is a number, e.g. "mycmd 1". In that case, wouldn't you assume that "1" actually represents psr value, but actually it's just part of the command name and thus parse the output incorrectly?

manager/container_test.go
```diff
@@ -319,3 +319,19 @@ func TestOnDemandHousekeepingReturnsAfterStopped(t *testing.T) {
...
+	[]byte("root       15886       2 23:51  0.1  0.0     0      0 I    00:00:00 kworker/u8:3-ev   3 -\nroot       15887       2 23:51  0.0  0.0     0      0 I<   00:00:00 kworker/1:2H      1 -\nubuntu     15888    1804 23:51  0.0  0.0  2832  10176 R+   00:00:00 ps                1 8:devices:/user.slice,6:pids:/user.slice/user-1000.slice/session-3.scope,5:blkio:/user.slice,2:cpu,cpuacct:/user.slice,1:na"),
+	[]byte("root         104       2 21:34  0.0  0.0     0      0 I<   00:00:00 kthrotld          3 -\nroot         105       2 21:34  0.0  0.0     0      0 S    00:00:00 irq/41-aerdrv     0 -\nroot         107       2 21:34  0.0  0.0     0      0 I<   00:00:00 DWC Notificatio   3 -\nroot         109       2 21:34  0.0  0.0     0      0 S<   00:00:00 vchiq-slot/0      1 -\nroot         110       2 21:34  0.0  0.0     0      0 S<   00:00:00 vchiq-recy/0      3 -"),
+}
+
+func TestParseProcessList(t *testing.T) {
+	for i, ps := range psOutput {
+		t.Run(fmt.Sprintf("iteration %d", i), func(tt *testing.T) {
+			cd := &containerData{}
+			_, err := cd.parseProcessList("/", true, ps)
+			// checking *only* parsing errors - otherwise /proc would have to be emulated.
```
bobbypage, line 333: do we have some unit tests that check the actual parsing logic and ensure right data is returned after the parse?

bobbypage:
Left a few comments above. BTW, side question (out of scope for this PR): why are we even parsing the results from running `ps`? Wouldn't it make more sense to gather this data from `/proc` or something like that directly?

iwankgb:
I was as surprised as you are. Looking at what I did here makes me think it would be good idea to get rid of output parsing.

manager/container_test.go
iwankgb: Unfortunately not. Loop in `parseProcessList()` besides parsing out put of `ps`:
* reads from cgroup fs
* reads from proc fs

I will try to break it even more so that at least some of the code can be tested.

manager/container.go
iwankgb: Because I have forgotten that `strings.Join()` exists :)

manager/container.go
iwankgb: I think that parsing logic has been significantly improved and simplified now :)

manager/container_test.go
iwankgb: There is new shiny test function just below :)

iwankgb:
@bobbypage, I hope you will find this code less enigmatic and prone to error :)

manager/container.go
```diff
@@ -283,106 +285,127 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
...
-		percentCPU, err := strconv.ParseFloat(fields[4], 32)
-		if err != nil {
-			return nil, fmt.Errorf("invalid cpu percent %q: %v", fields[4], err)
-		}
-		percentMem, err := strconv.ParseFloat(fields[5], 32)
-		if err != nil {
-			return nil, fmt.Errorf("invalid memory percent %q: %v", fields[5], err)
-		}
-		rss, err := strconv.ParseUint(fields[6], 0, 64)
+		process, err := cd.parsePsLine(line, cadvisorContainer, inHostNamespace)
```
bobbypage, line 304: s/process/processInfo (just to be a bit more clear) :)

manager/container.go
```diff
@@ -283,106 +285,127 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
...
-				RSS:           rss,
-				VirtualSize:   vs,
-				Status:        fields[8],
-				RunningTime:   fields[9],
-				Cmd:           fields[10],
-				CgroupPath:    cgroupPath,
-				FdCount:       fdCount,
-				Psr:           psr,
-			})
+		if cd.isRoot() || cd.info.Name == process.CgroupPath {
```
bobbypage, line 322: I think the logic here has changed slightly regarding what processInfos are ultimately added to `processes` with respect to the the `cd.info.Name == process.CgroupPath` check

Old logic checked `cd.info.Name == cgroup`
https://github.com/google/cadvisor/blob/3066a8034f2eaf15f69bf1c867df5cd081cd538a/manager/container.go#L364

where `cgroup` was always populated:

https://github.com/google/cadvisor/blob/3066a8034f2eaf15f69bf1c867df5cd081cd538a/manager/container.go#L341

However, in the change here, `process.CgroupPath` will only be populated if the containerData is root:

https://github.com/google/cadvisor/blob/f34a89b2be44fe4915e7f827a4b60dcac9007230/manager/container.go#L393-L394

I think that might be why the old code (very confusingly) had both `cgroup` and `cgroupPath` where cgroupPath was what was set on the v2.ProcessInfo.CGroupPath struct:

https://github.com/google/cadvisor/blob/3066a8034f2eaf15f69bf1c867df5cd081cd538a/manager/container.go#L341-L353

As a result, I think this is currently only including processInfo's that are `isRoot()==true`  which is probably skipping more processInfos than we actually want.

manager/container.go
```diff
@@ -283,106 +285,127 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
...
+	if cd.isRoot() {
+		info.CgroupPath, err = cd.getCgroupPath(lastTwoFields[1])
+		if err != nil {
+			return nil, fmt.Errorf("could not parse cgroup path from %q: %v", lastTwoFields[1], err)
+		}
+	}
+
+	// Remove the ps command we just ran from cadvisor container.
+	// Not necessary, but makes the cadvisor page look cleaner.
+	if !inHostNamespace && cadvisorContainer == info.CgroupPath && info.Cmd == "ps" {
```
bobbypage, line 402: ditto re above comment `info.CgroupPath` will only be populated if `cd.isRoot()` unlike previous checke where cgroup was always populated

https://github.com/google/cadvisor/blob/3066a8034f2eaf15f69bf1c867df5cd081cd538a/manager/container.go#L341

https://github.com/google/cadvisor/blob/3066a8034f2eaf15f69bf1c867df5cd081cd538a/manager/container.go#L347

manager/container.go
```diff
@@ -283,106 +285,127 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
...
+		return nil, fmt.Errorf("invalid virtual size %q: %v", fields[7], err)
+	}
+	// convert to bytes
+	info.RSS *= 1024
+	info.VirtualSize *= 1024
+
+	// According to `man ps`: The following user-defined format specifiers may contain spaces: args, cmd, comm, command,
+	// fname, ucmd, ucomm, lstart, bsdstart, start.
+	// Therefore we need to be able to parse comm that consists of multiple space-separated parts.
+	info.Cmd = strings.Join(fields[10:len(fields)-2], " ")
```
bobbypage, line 385: good idea re. using `fields[10:len(fields)-2]`as cmd :)

manager/container.go
```diff
@@ -283,106 +285,127 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
 }
 
 func (cd *containerData) GetProcessList(cadvisorContainer string, inHostNamespace bool) ([]v2.ProcessInfo, error) {
-	// report all processes for root.
-	isRoot := cd.info.Name == "/"
-	rootfs := "/"
-	if !inHostNamespace {
-		rootfs = "/rootfs"
-	}
 	format := "user,pid,ppid,stime,pcpu,pmem,rss,vsz,stat,time,comm,psr,cgroup"
```
bobbypage, line 288: maybe move this up a constant near `expectedFields`, then it will be clear why `expectedFields = 13`

manager/container_test.go
bobbypage: awesome, thanks for adding tests here!

manager/container_test.go
```diff
@@ -319,3 +321,132 @@ func TestOnDemandHousekeepingReturnsAfterStopped(t *testing.T) {
...
+		nil,
+		&containerData{
+			info: containerInfo{ContainerReference: info.ContainerReference{Name: "/"}},
+		},
+	},
+	{
+		"process with space in name and no cgroup",
+		"root         107       2 21:34  0.0  0.1     3      4 I<   00:20:00 DWC Notificatio   3 -",
+		"/docker/cadvisor",
+		false,
```
bobbypage, line 379: nit: i think it would be a bit more readable if you included the struct names for each field inline here, since it can be a bit confusing what the struct fields map too.

bobbypage:
Thanks for your changes here, I think the updated parsing logic is a lot more clear and less error prone! Left a few smaller comments.

manager/container.go
iwankgb: I think I managed to fix the problem that you described. If you take a look at added test cases then you will see that:
* all processes (except `ps` are returned when root container if requested
* only processes belonging to requested container (with no cgroup information set) are returned when it is a non-root container

iwankgb:
@bobbypage, as you can see this PR fixes more than one issue now. I came across #2770 while trying to verify that process are really displayed as expected. They have not been, so I started debugging. I hope it's fine with you - these are still closely related changes to same file.

manager/container.go
```diff
@@ -195,20 +199,24 @@ func (cd *containerData) DerivedStats() (v2.DerivedStats, error) {
...
-	matches := cgroupPathRegExp.FindSubmatch([]byte(cgroups))
+	matches := cgroupMemoryPathRegExp.FindSubmatch([]byte(cgroups))
 	if len(matches) != 2 {
 		klog.V(3).Infof("failed to get memory cgroup path from %q", cgroups)
-		// return root in case of failures - memory hierarchy might not be enabled.
-		return "/", nil
+		// On some systems (e.g. Raspberry PI 4) cgroup memory controlled is disabled by default.
+		matches = cgroupCPUPathRegExp.FindSubmatch([]byte(cgroups))
+		if len(matches) != 2 {
+			// return root in case of failures - memory hierarchy might not be enabled.
```
bobbypage, line 215: add a klog here if you think this is worth logging? The above klog on 211 might be confusing since it will printed even if we find match by using cpu path regex

manager/container_test.go
```diff
@@ -474,3 +474,51 @@ func TestParsePsLine(t *testing.T) {
 		})
 	}
 }
+
+var cgroupCases = []struct {
+	name    string
```
bobbypage, line 479: can we add case here case where `cgroups==foobar` (some random string that doesn't match any of the regex / cases) and ensure that "/" is returned?

manager/container_test.go
```diff
@@ -474,3 +474,51 @@ func TestParsePsLine(t *testing.T) {
...
+	cgroups string
+	path    string
+}{
+	{
+		"no cgroup",
```
bobbypage, line 484: nit: add field names here and below

manager/container.go
```diff
@@ -48,8 +48,12 @@ var enableLoadReader = flag.Bool("enable_load_reader", false, "Whether to enable
 var HousekeepingInterval = flag.Duration("housekeeping_interval", 1*time.Second, "Interval between container housekeepings")
 
 // cgroup type chosen to fetch the cgroup path of a process.
-// Memory has been chosen, as it is one of the default cgroups that is enabled for most containers.
-var cgroupPathRegExp = regexp.MustCompile(`memory[^:]*:(.*?)[,;$]`)
+// Memory has been chosen, as it is one of the default cgroups that is enabled for most containers...
+var cgroupMemoryPathRegExp = regexp.MustCompile(`memory[^:]*:(.*?)[,;$]`)
```
bobbypage, line 52: random comment, out of scope for this PR: I wonder if this regex is really needed here. The format seems pretty easily parsable, 

e.g. from my system

```
11:pids:/system.slice/ssh.service,10:devices:/system.slice/ssh.service,4:cpu,cpuacct:/system.slice/ssh.service,2:memory:/system.slice/ssh.service,1:name=systemd:/system.slice/ssh.service,0::/system.slice/ssh.service
```

I think you could just split the string by commas and remove the leading `hierarchy-ID`, e.g. `11:`. We can also see which cgroup controller mounted by reading `/proc/cgroups`. Regex is totally fine for now (and what is used currently) but maybe worth a TODO as future improvement :)

manager/container.go
```diff
@@ -283,106 +288,130 @@ func (cd *containerData) getContainerPids(inHostNamespace bool) ([]string, error
...
+	// Do not report processes from other containers when non-root container requested.
+	if !cd.isRoot() && info.CgroupPath != cd.info.Name {
+		return nil, nil
+	}
+
+	// Remove cgroup information when non-root container requested.
+	if !cd.isRoot() {
+		info.CgroupPath = ""
+	}
+	return &info, nil
```
bobbypage, line 412: any reason to return pointer here? maybe just return by value, since the return value is dereferenced right away and appended to the processInfo's slice.

manager/container.go
bobbypage: Thanks for the fix, your change makes sense!

bobbypage:
Few small comments, but LGTM otherwise. Thanks again for all your work on this!

Can you please squash into maybe two commits, one for the changes of parsing and one for the memory cgroup fix?

manager/container.go
iwankgb: Yes. I moved logic for deciding whether `ProcessInfo` is of interest to caller: if `nil` is returned then it will not be appended to `processes` slice in `containerData.parseProcessList()`. Such approach keeps logic in one place and, what is even more important, make it testable. I think that `nil` is more meaningful then a zero value `ProcessInfo`.

manager/container.go
iwankgb: I have completely forgotten about it. Added!

manager/container.go
iwankgb: Unfortunately these two consts/variables are defined in separate functions.

manager/container_test.go
bobbypage: thanks!

manager/container.go
bobbypage: got it, that makes sense, thanks!

manager/container_test.go
bobbypage: thanks

bobbypage:
Thanks again for all your work on this! LGTM!

bobbypage (approved):

PR #2757, opened 2020-12-09
no NVIDIA devices found should be info level log

pacoxu:
I think https://github.com/google/cadvisor/commit/1baad3f20b5436ca9495fdfb1cca7a34398b34c6 changed the behavior.

Originally, when setup is called, no error log will be printed. 

I agree with the @dashpole  comments in https://github.com/google/cadvisor/pull/2477#discussion_r405731479

### DIFF
--- a/accelerators/nvidia.go
+++ b/accelerators/nvidia.go
@@ -29,6 +29,8 @@ import (
 	"github.com/google/cadvisor/stats"
 
 	"github.com/mindprince/gonvml"
+	"github.com/pkg/errors"
+
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


iwankgb:
@pacoxu I think it is already logged:
https://github.com/google/cadvisor/blob/ac63c56b6be017d23affe5b1409f2703921cb16a/accelerators/nvidia.go#L61

pacoxu:
@ iwankgb Removed the log line.

iwankgb:
@pacoxu what is goal of this PR now? Nothing is really changed.

pacoxu:
I may misunderstand here. 

I find this log is a warning log in recent released kubelet v1.20.0.

I think this is should be a info log.

```
Dec 17 11:22:41 daocloud kubelet[73512]: I1217 11:22:41.145694   73512 nvidia.go:61] NVIDIA setup failed: no NVIDIA devices found
```

iwankgb:
It seems to me that it is info level: https://github.com/google/cadvisor/pull/2757#issuecomment-741816220.

Log entry itself seems to point at it too: **I**1217 11:22:41.145694

PR #2767, opened 2020-12-17
Export entire smaps memory metrics instead of only referenced_memory

### LINKED ISSUE #2634: Account and export LazyFree memory by container
Golang does not "finally" hand memory back to the system any more on recent Linux kernels, which results in confusion on high memory usage/processes seemingly not freeing their memory any more (for example https://github.com/golang/go/issues/37585).

It would enhance monitoring experience if cadvisor would also export the amount of `LazyFree` memory that is already semi released; and can quickly reclaimed by either the process or Kernel if needed. The information required is stored in `/proc/<pid>/smaps`. This file is already parsed since a few weeks: https://github.com/google/cadvisor/commit/7ab5e27909c5b72363641ed6e683be42488e0365

Would you agree it is reasonable to expose these metrics? It seems to me most of the required code is already there, and all we need is some generalization of `referencedBytesStat` and related functions and a few new fields. The linked commit seems like a pretty comprehensive blueprint to me.

-- 
Jens Erat \<jens.erat@daimler.com\>, Daimler TSS GmbH  
[Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

JensErat:
#2495 introduced parsing smaps memory metrics, but only exposes referenced memory. As proposed in #2634, also other metrics in here are of interest, specifically LazyFree; but generally also the others seem of interest.

This commit replaces the referenced memory metrics by a generic smaps series of metrics. Runtime costs in cadvisor are not really impacted, as all the smaps parsing already happens anyway. The cardinality per container extends from 1 to 19 metrics, though. I think this is acceptable: also referenced memory scraping was disabled by default, so is scraping smaps.

Resolves #2634. It slightly extends the original proposal by simply exposing all metrics, but once I started work on the proposal it felt reasonable to do so. I'm fine with major changes, though!

While the code works (and the change is actually smaller than it looks, most of the lines only affect unit tests), I guess there is definitely more work to do.

- [ ] all unit tests run successfully, let's wait for the entire integration test suite
- [ ] do we want this approach at all?
- [ ] do we need downwards compatibility for `referenced_memory`?

As soon as there is agreement on the chosen approach, tasks needed to make merge-ready:

- [ ] some cleanup work, maybe extract (or get rid of) the array of smaps labels
- [ ] check whether documentation is required somewhere

This PR likely also affects #2715, which proposes optimizations to smaps data collection.

<sup>Jens Erat <jens.erat@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>

### DIFF
--- a/cmd/internal/storage/influxdb/influxdb_test.go
+++ b/cmd/internal/storage/influxdb/influxdb_test.go
@@ -231,7 +231,7 @@ func TestContainerStatsToPoints(t *testing.T) {
 
 	// Then
 	assert.NotEmpty(t, points)
-	assert.Len(t, points, 34+len(stats.Cpu.Usage.PerCpu))
+	assert.Len(t, points, 52+len(stats.Cpu.Usage.PerCpu))
 
 	// CPU stats
 	assertContainsPointWithValue(t, points, serCpuUsageTotal, stats.Cpu.Usage.Total)
@@ -276,7 +276,9 @@ func TestContainerStatsToPoints(t *testing.T) {
 	}
 
 	// Reference memory
-	assertContainsPointWithValue(t, points, serReferencedMemory, stats.ReferencedMemory)
+	for k, v := range stats.SmapsMemory {
+		assertContainsPointWithValue(t, points, serSmapsPrefix+k, v)
+	}
 
 	// Resource Control stats - memory bandwidth
 	for _, rdtMemoryBandwidth := range stats.Resctrl.MemoryBandwidth {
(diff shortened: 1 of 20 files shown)


cmd/internal/storage/influxdb/influxdb_test.go
```diff
@@ -359,8 +361,28 @@ func createTestStats() (*info.ContainerInfo, *info.ContainerStats) {
...
+			"Anonymous":       2,
+			"KernelPageSize":  3,
+			"LazyFree":        4,
+			"Locked":          5,
+			"MMUPageSize":     6,
+			"Private_Clean":   7,
+			"Private_Dirty":   8,
+			"Private_Hugetlb": 9,
+			"Pss":             10,
+			"Referenced":      11,
```
iwankgb, line 375: Removing `ContainerStats.ReferencedMemory` breaks backward compatibility. I'm afraid we will have to keep it, for a while at least.

JensErat: It might even be reasonable to exclude it from the smaps metrics, or keep it duplicate forever. From what I understand, all other smaps metrics do not get reset using this referenced-cycle-count-reset-logic, but I'll have to investigate further.

Adding the referenced metric again is a matter of minutes and adjusting some tests. I think some early feedback on the general approach (and matters like compatiblity) is very important though.

iwankgb:
@bobbypage I think it's something worth being discussed. As far as I can see in #2634 @dashpole was fine with introducing such changes. They look reasonable to me too if we keep backward compatibility.

JensErat:
@bobbypage @dashpole @iwankgb any news on how/whether to proceed? I was just puzzled once again because of high memory usage that boiled down to LazyFree -- even if you know about it you get surprised every now and then.

iwankgb:
It would be great to have `container_referenced_bytes` still available, I think. Some folks might be already using it. 
I'm just finishing very intense sprint at work and I should be able to be more active here again.

PR #2787, opened 2021-01-17
Support nfs in processMounts

gjkim42:
Fix kubenetes issue https://github.com/kubernetes/kubernetes/issues/98009

Helps with the situation when the kubelet starts with --root-dir=$NFS_DIR.


gjkim42:
/cc @bobbypage 
Would you take a look at this? Thanks.

fs/fs.go
```diff
@@ -170,7 +170,7 @@ func processMounts(mounts []*mount.Info, excludedMountpointPrefixes []string) ma
 	partitions := make(map[string]partition)
 
 	supportedFsType := map[string]bool{
-		// all ext systems are checked through prefix.
+		// all ext and nfs systems are checked through prefix.
```
bobbypage, line 173: is there any reason we can't just add `"nfs" : true` to `supportedFsType` as opposed to special casing it below?

I'm not clear why "ext" is special cased unlike all the other file system types here.

gjkim42: I think `"ext"` was special cased because there are a number of ext filesystems. (ext3, ext4...)
And there are a number of NFS too. (nfs3, nfs4...)

bobbypage: I see, that makes sense. 

I'm fine with either, but if you think the prefix check is easier, let's update the comment to make it clear why some fs type check via prefix vs some in `supportedFsType` (i.e. your explanation regarding number version of filesystem).

gjkim42: Updated :)

bobbypage: thanks!

bobbypage (approved):

PR #2796, opened 2021-02-01
Add cgroup_memory_migrate metric

kwisniewski98:
Signed-off-by: Wisniewski, Krzysztof2 <Krzysztof2.Wisniewski@intel.com>

### DIFF
--- a/info/v1/container.go
+++ b/info/v1/container.go
@@ -399,6 +399,10 @@ type MemoryStats struct {
 	HierarchicalData MemoryStatsMemoryData `json:"hierarchical_data,omitempty"`
 }
 
+type CPUSetStats struct {
+	MemoryMigrate uint64 `json:"memory_migrate"`
+}
+
 type MemoryNumaStats struct {
 	File        map[uint8]uint64 `json:"file,omitempty"`
 	Anon        map[uint8]uint64 `json:"anon,omitempty"`
@@ -957,6 +961,8 @@ type ContainerStats struct {
 
 	// Resource Control (resctrl) statistics
 	Resctrl ResctrlStats `json:"resctrl,omitempty"`
+
+	CpuSet CPUSetStats `json:"cpuset,omitempty"`
 }
 
 func timeEq(t1, t2 time.Time, tolerance time.Duration) bool {
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -455,6 +455,16 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 			},
 		}...)
 	}
+	if includedMetrics.Has(container.CPUSetMetrics) {
+		c.containerMetrics = append(c.containerMetrics, containerMetric{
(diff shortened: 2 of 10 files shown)


info/v1/container.go
```diff
@@ -395,10 +395,16 @@ type MemoryStats struct {
...
 	Failcnt uint64 `json:"failcnt"`
 
+	CpuSet CPUSetStats `json:"cpuset,omitempty"`
+
 	ContainerData    MemoryStatsMemoryData `json:"container_data,omitempty"`
 	HierarchicalData MemoryStatsMemoryData `json:"hierarchical_data,omitempty"`
 }
 
+type CPUSetStats struct {
+	Migrate uint64 `json:"memory_migrate"`
```
Creatone, line 405: Could you use `MemoryMigrate`?

Creatone (commented):
Small nit

Creatone (approved):

iwankgb (approved):
LGTM

info/v1/container.go
```diff
@@ -395,10 +395,16 @@ type MemoryStats struct {
 
 	Failcnt uint64 `json:"failcnt"`
 
+	CpuSet CPUSetStats `json:"cpuset,omitempty"`
```
bobbypage, line 398: curious to get your thoughts: does this make sense under `MemoryStats`? 

Since this is taken from the cpuset controller, maybe this makes sense to add a new `CpuSets` top level field under `ContainerStats` which will contain the relevant metrics from cpuset: https://man7.org/linux/man-pages/man7/cpuset.7.html

As I understand currently all the fields under `MemoryStats` actually come from the memory cgroup controller stats. 

Thoughts @kwisniewski98 @Creatone @iwankgb ?

Creatone: That will be a good place for other CPUSet metrics.

kwisniewski98: Done

metrics/prometheus.go
```diff
@@ -423,6 +423,14 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 					return metricValues{{value: float64(s.Memory.WorkingSet), timestamp: s.Timestamp}}
 				},
 			},
+			{
+				name:      "container_memory_migrate",
```
bobbypage, line 427: this is still under `if includedMetrics.Has(container.MemoryUsageMetrics)`.

I think since we're introducing new CPUSet category of metrics, makes sense to add a new [metric kind](https://github.com/google/cadvisor/blob/bc8b91889a41c6f8f7b12819824d197ae0d57b1f/container/factory.go#L45-L66) for CpuSet metrics instead of putting this under existing container memory usage metrics.

kwisniewski98: Done

bobbypage:
Thanks! LGTM

bobbypage (approved):

bobbypage (approved):

PR #2800, opened 2021-02-07
Update heuristic for container creation time

odinuge:
The file "cgroup.clone_children" is not a part of cgroup v2, so use
"cgroup.events" instead. This heuristic is quite bad, so also check the
cgroup directories to see.

The old metric based on this is also constantly increasing due to this, making it more or less useless:

Screenshot from 2021-02-07 17-06-27

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -86,24 +86,20 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
 	// Get the lowest creation time from all hierarchies as the container creation time.
 	now := time.Now()
 	lowestTime := now
-	for _, cgroupPath := range cgroupPaths {
-		fi, err := os.Stat(cgroupPath)
-		if err == nil && fi.ModTime().Before(lowestTime) {
-			lowestTime = fi.ModTime()
+	for _, cgroupPathDir := range cgroupPaths {
+		dir, err := os.Stat(cgroupPathDir)
+		if err == nil && dir.ModTime().Before(lowestTime) {
+			lowestTime = dir.ModTime()
 		}
-	}
-
-	for _, cgroupPath := range cgroupPaths {
 		// The modified time of the cgroup directory sometimes changes whenever a subcontainer is created.
 		// eg. /docker will have creation time matching the creation of latest docker container.
 		// Use clone_children/events as a workaround as it isn't usually modified. It is only likely changed
 		// immediately after creating a container. If the directory modified time is lower, we use that.
+		cgroupPathFile := path.Join(cgroupPathDir, "cgroup.clone_children")
 		if cgroups.IsCgroup2UnifiedMode() {
-			cgroupPath = path.Join(cgroupPath, "cgroup.events")
-		} else {
-			cgroupPath = path.Join(cgroupPath, "cgroup.clone_children")
+			cgroupPathFile = path.Join(cgroupPathDir, "cgroup.events")
 		}
-		fi, err := os.Stat(cgroupPath)
+		fi, err := os.Stat(cgroupPathFile)
 		if err == nil && fi.ModTime().Before(lowestTime) {
 			lowestTime = fi.ModTime()
 		}
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -30,6 +30,7 @@ import (
 	"github.com/karrick/godirwalk"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 
 	"k8s.io/klog/v2"
 )
@@ -68,6 +69,16 @@ func findFileInAncestorDir(current, file, limit string) (string, error) {
 	}
 }
 
+var bootTime = func() time.Time {
+	now := time.Now()
+	var sysinfo unix.Sysinfo_t
+	if err := unix.Sysinfo(&sysinfo); err != nil {
+		return now
+	}
+	sinceBoot := time.Duration(sysinfo.Uptime) * time.Second
+	return now.Add(-1 * sinceBoot).Truncate(time.Minute)
+}()
+
 func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoFactory, hasNetwork, hasFilesystem bool) (info.ContainerSpec, error) {
 	var spec info.ContainerSpec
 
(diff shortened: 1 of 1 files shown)


iwankgb (changes requested):
@odinuge, can you provide a test for your change, please?

odinuge:
> @odinuge, can you provide a test for your change, please?

Yeah! Thought I could see if you guys were interested before writing tests. This function doesn't have a test already, but can definitely add one.

iwankgb:
The chart the you added to this PR clearly show a bug: monotonically increasing start time of a running container :D

bobbypage:
Thanks @odinuge for the fix. Do you think this will fix https://github.com/kubernetes/kubernetes/issues/98494 as well?

I'm curious how well the ModTime of the cgroup path works as a heuristic there.

odinuge:
> Do you think this will fix kubernetes/kubernetes#98494 as well?

Ahh, I see. The current, and with my new patch, that would probably not be the case. In cadvisor, the `CreationTime` https://github.com/google/cadvisor/blob/013e451ee34b9ea516262ad0cdcd856f838c45a5/info/v2/container.go#L68-L69

clearly states that is it the time of the container creation (and that is not ==start time of a container). I think that the `CreationTime` should keep its value as long as the cgroup dir is not removed, and with that logic, this patch works.

However, if you want the start time of the CRI, I think we need another (and better) heuristic, preferably from CRI and not by reading cgroup files. I really don't think it is a good idea to track system daemon start times by reading mod times on cgroup files, compared to just providing it from the system or CRI. That is however another discussion.

> I'm curious how well the ModTime of the cgroup path works as a heuristic there as you proposed.

If we talk about the "container creation time", I think this patch is better than the old one, but I still think that this heuristic can never be 100% by nature, due to the fact that cgroups aren't designed to to this. One can look at the start time of the oldest proc in a cgroup, or something like that.


Also, started looking at tests, but not really sure how useful they will be. Guess I have to mock an fs, or use temp dirs.

bobbypage:
Thanks @odinuge. Agree, the goal here is to track CreationTime, not last start time. 

> However, if you want the start time of the CRI, I think we need another (and better) heuristic, preferably from CRI and not by reading cgroup files. I really don't think it is a good idea to track system daemon start times by reading mod times on cgroup files, compared to just providing it from the system or CRI. That is however another discussion.

Definitely, agree the tracking of that should be moved into CRI. I think the confusion stemed from the fact that the kubelet summary API calls this field `startTime` while it is in fact `CreationTime`. 

LGTM on your changes here. Only one question, did you have a chance to test out the `modTime` of the cgroup and to see how it if results in more or less stable `CreationTime`? I hope the results are a bit better than the graph you posted in https://github.com/google/cadvisor/pull/2800#issue-568998967 since we should expect that `CreationTime` should remain static for a running container :)

harche:
Thanks @odinuge, I see failures reported in `/stats/summary` test to with respect to container start time. I have documented it here, https://github.com/google/cadvisor/pull/2801#issuecomment-780392368

harche:
I just verified running `Summary API` test with this change, and it seem like this PR addresses the start time related failures in `/stats/summary` reported in https://github.com/google/cadvisor/pull/2801#issuecomment-780392368

Thanks @odinuge for addressing this.

harche:
LGTM

odinuge:
Thanks @bobbypage!

> LGTM on your changes here. Only one question, did you have a chance to test out the modTime of the cgroup and to see it if results in more or less stable CreationTime? I hope the results are a bit better than the graph you posted in #2800 (comment) since we should expect that CreationTime should remain static for a running container :)

Will test this later today and tomorrow, and report back as soon as I have more details.. :smile: 

> Definitely, agree the tracking of that should be moved into CRI. I think the confusion stemed from the fact that the kubelet summary API calls this field startTime while it is in fact CreationTime.

Yeah! If k8s wants a better "started" time, they could look into it separately. :+1: 

> @harche: Thanks @odinuge for addressing this.

Thanks! I am working with this becuase I want a more mature cgroup v2 support in Kubernetes so that I/we can start using it in production. :smile:

odinuge:
I have only had time to test with cgroup v2 (have no clusters with prometheus and cgroup v2), but it looks like it works well with there!

Pods on a given node:

Screenshot from 2021-02-22 09-36-33

The lowest line is `kubepods.slice`, and the timestamp corresponds to boot.

Here is a zoomed out one where one clearly can see the ever increasing metric, before it drops down with this patch:

Screenshot from 2021-02-22 09-41-22


----

The only issue I see so far is the root cgroup (only applies to cgroup v1, looks like this is correct for v2). You see it as the bottom line here, with a timestamp corresponding to the install date of the OS (I assume, since it is a date in early 2020). Maybe we can use boot time for the root cgroup?

Screenshot from 2021-02-22 09-36-46

harche:
> The only issue I see so far is the root cgroup (only applies to cgroup v1, looks like this is correct for v2). You see it as the bottom line here, with a timestamp corresponding to the install date of the OS (I assume, since it is a date in early 2020). Maybe we can use boot time for the root cgroup?

Makes sense to use boot time.

odinuge:
> Makes sense to use boot time.

Nice! Added a check to instead always use `max(creation_time, boot)` in the last commit. Does that look ok? The fix is working, and the root cgroup gets the boot time as "CreatedAt".

container/common/helpers.go
```diff
@@ -76,16 +77,36 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
...
 	lowestTime := now
 	for _, cgroupPath := range cgroupPaths {
-		// The modified time of the cgroup directory changes whenever a subcontainer is created.
+		fi, err := os.Stat(cgroupPath)
+		if err == nil && fi.ModTime().Before(lowestTime) {
+			lowestTime = fi.ModTime()
+		}
+	}
+
+	for _, cgroupPath := range cgroupPaths {
```
harche, line 86: Why are we iterating over `cgroupPaths` again when we already doing it at line 79? Can we not merge the two loops and just iterate once? The only difference that I see between the first loop and the second is you are updating `cgroupPath` variable based on if it's v1 or v2.

odinuge: I just thought it would be cleaner in order to make it simpler to understand the flow, but I can merge them. (If you are thinking about performance, I have no idea what the go compiler will do to cases like this). Agree that the duplicated logic is kinda stupid, so guess I can fix that anyway. Thank! :+1:

odinuge: Pushed a new change now. Does that look better @harche?

harche: Yes, thanks @odinuge

harche (approved):
LGTM

iwankgb (approved):

bobbypage:
LGTM

bobbypage (approved):

PR #2805, opened 2021-02-12
sched_getaffinity does not return number of online CPUs

### LINKED ISSUE #2792: cAdvisor running in docker ignore isolcpus
Hi,

I'm running cAdvisor in docker on machines with high CPU load.
I've configured the server kernel with isolcpus=1,2,3

I'm running cAdvisor with the following params:
```
docker run \
  --volume=/:/rootfs:ro \
  --volume=/sys:/sys:ro \
  --publish=8080:8080 \
  --detach=true \
  --name=cadvisor \
  --privileged \
  --device=/dev/kmsg \
  gcr.io/cadvisor/cadvisor:v0.37.0
```

I've noticed that cAdvisor still runs on all cpus (not only cpu 0).
If I run cAdvisor with docker --cpuset-cpus=0, then cAdvisor doesn't show info on the other cpus.
Also if I enforce cpu affinity with taskset (without the --cpuset-cpus flag) - cAdvisor will show me stats only on the allowed cpus.

What is the correct way to run cAdvisor on a specific cpu?
BTW, I can see that setting -max_procs to 1 doesn't affect the number of processes that cAdvisor uses.

Thanks,
Doron

iwankgb:
Fixes #2792

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -29,14 +29,13 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/cadvisor/container"
-	info "github.com/google/cadvisor/info/v1"
-	"golang.org/x/sys/unix"
-
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
-	fs2 "github.com/opencontainers/runc/libcontainer/cgroups/fs2"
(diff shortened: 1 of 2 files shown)


bobbypage (approved):

bobbypage:
LGTM, thanks for tracking this one down.

Issue #2813, opened 2021-02-26
OOM parsing incorrectly on 5.0 kernels and higher


kgtw:
OOM parsing was updated in https://github.com/google/cadvisor/pull/2418 to address the new format for kernels 5.0 and higher.

The new regex has a typo in it, where it expects `oom_memcg` to provide two values separated by a space. `oom_memcg=(.*) (.*)`
```
oom-kill:constraint=(.*),nodemask=(.*),cpuset=(.*),mems_allowed=(.*),oom_memcg=(.*) (.*),task_memcg=(.*),task=(.*),pid=(.*),uid=(.*)
```

As per the kernel commit, https://github.com/torvalds/linux/commit/f0c867d9588d9efc10d6a55009c9560336673369#diff-c14acc69885a24f354bc607686800e23, this was probably an accidental typo and should be a single value. `oom_memcg=(.*)`

```
oom-kill:constraint=(.*),nodemask=(.*),cpuset=(.*),mems_allowed=(.*),oom_memcg=(.*),task_memcg=(.*),task=(.*),pid=(.*),uid=(.*)
```

cc @dashpole as you were the original author and perhaps can confirm my suspicions here.

dashpole:
Yep, I think you are correct.

MadhavJivrajani:
Hi! Can I work on this?

PR #2817, opened 2021-03-01
Fix oomparser regex for kernels 5.0 and higher

kgtw:
#2418 introduced new regex for 5.0 kernels and higher, the regex had a typo and wasn't resolving the correct container names.

Fixes: https://github.com/google/cadvisor/issues/2813

### DIFF
--- a/utils/oomparser/oomparser.go
+++ b/utils/oomparser/oomparser.go
@@ -28,7 +28,7 @@ import (
 var (
 	legacyContainerRegexp = regexp.MustCompile(`Task in (.*) killed as a result of limit of (.*)`)
 	// Starting in 5.0 linux kernels, the OOM message changed
-	containerRegexp = regexp.MustCompile(`oom-kill:constraint=(.*),nodemask=(.*),cpuset=(.*),mems_allowed=(.*),oom_memcg=(.*) (.*),task_memcg=(.*),task=(.*),pid=(.*),uid=(.*)`)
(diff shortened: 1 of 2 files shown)


kgtw:
@googlebot I signed it!

iwankgb:
@kgtw would you be able to provide a unit test to save us from regression?

kgtw:
Hi @iwankgb - Thanks for the speedy review!

I believe the updated unit-test on [L84](https://github.com/google/cadvisor/pull/2817/files#diff-de0c174e6e7304b6ad5b97a8c32cf603f12c10972c4b6587b6fdbcc06b7d376aR84) should cover regressions, it was previously validating the bad regex.

iwankgb:
@kgtw will the regexp work for pre-5.0 messages too?

kgtw:
The regex is only for >5.0 kernels.

OOM messages on kernels lower than 5.0 have a different format and are handled by the legacy test case [L33-L59](https://github.com/google/cadvisor/blob/dcbeba1642afb4d533f585da0e6a98bd12d46695/utils/oomparser/oomparser_test.go#L33-L59) 

The existing functionality of `getContainerName()` has not changed, where it will try to parse the 5.0 regex, failing that it will fallback to legacy parsing.
https://github.com/google/cadvisor/blob/dcbeba1642afb4d533f585da0e6a98bd12d46695/utils/oomparser/oomparser.go#L73-L89

iwankgb (approved):
Thank you, sir!
@bobbypage you are still the only person that has power to LGTM it.

bobbypage:
thanks!

PR #2774, opened 2021-01-05
ADD "disable_specific_metrics" which can remove specific metric by name or regex

Alexbyy:
[#2769](https://github.com/google/cadvisor/issues/2769)
This param is a denyList which named  "disable_specific_metrics". Comma-separated list of metrics not to be enabled.It can remove specific metric by name or regex.

PR #2827, opened 2021-03-06
Revert "ADD "disable_specific_metrics" which can remove specific metric by name or regex"

bobbypage:
Reverts google/cadvisor#2774

This PR introduced some breaking changes due to changing the signature of `manager.go` `New()` which requires a new argument (deny list) and the additional necessity of passing denyList to `RegisterPrometheusHandler` and `NewPrometheusCollector` all of which are exported functions.

This forces users to downstream users to change their code for example: https://github.com/kubernetes/kubernetes/pull/99875#issuecomment-791827028

Taking a closer look I also noticed some other issues:

* denyList seems to be unused in `manager/manager.go`. After removing it, tests passed / everything still compiled fine. 
* even if we decide to make the change to the `RegisterPrometheusHandler` and `NewPrometheusCollector`, we should allow passing nil as the denyList. However, currently passing nil as denyList will result in panic since there isn't check that `denyList`  is nil.

Since we need to cut next cAdvisor for k8s release, I suggest we revert this change for now and revisit this later making the above fixes.

```
diff --git a/cmd/cadvisor.go b/cmd/cadvisor.go
index bef0a003..265796ba 100644
--- a/cmd/cadvisor.go
+++ b/cmd/cadvisor.go
@@ -194,7 +194,7 @@ func main() {

        collectorHttpClient := createCollectorHttpClient(*collectorCert, *collectorKey)

-       resourceManager, err := manager.New(memoryStorage, sysFs, housekeepingConfig, includedMetrics, &collectorHttpClient, strings.Split(*rawCgroupPrefixWhiteList, ","), *perfEvents, denyList)
+       resourceManager, err := manager.New(memoryStorage, sysFs, housekeepingConfig, includedMetrics, &collectorHttpClient, strings.Split(*rawCgroupPrefixWhiteList, ","), *perfEvents)
        if err != nil {
                klog.Fatalf("Failed to create a manager: %s", err)
        }
diff --git a/manager/manager.go b/manager/manager.go
index 0d72c27d..9f87278d 100644
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -35,9 +35,8 @@ import (
        "github.com/google/cadvisor/events"
-       "github.com/google/cadvisor/info/v2"
+       v2 "github.com/google/cadvisor/info/v2"
        "github.com/google/cadvisor/machine"
-       "github.com/google/cadvisor/metrics"
        "github.com/google/cadvisor/nvm"
        "github.com/google/cadvisor/perf"
        "github.com/google/cadvisor/resctrl"
@@ -147,7 +146,7 @@ type HouskeepingConfig = struct {
 }

 // New takes a memory storage and returns a new manager.
-func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig HouskeepingConfig, includedMetricsSet container.MetricSet, collectorHTTPClient *http.Client, rawContainerCgroupPathPrefixWhiteList []string, perfEventsFile string, denyList *metrics.DenyList) (Manager, error) {
+func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig HouskeepingConfig, includedMetricsSet container.MetricSet, collectorHTTPClient *http.Client, rawContainerCgroupPathPrefixWhiteList []string, perfEventsFile string) (Manager, error) {
        if memoryCache == nil {
                return nil, fmt.Errorf("manager requires memory storage")
        }
@@ -199,7 +198,6 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, houskeepingConfig
                maxHousekeepingInterval:               *houskeepingConfig.Interval,
                allowDynamicHousekeeping:              *houskeepingConfig.AllowDynamic,
                includedMetrics:                       includedMetricsSet,
-               DenyList:                              denyList,
                containerWatchers:                     []watcher.ContainerWatcher{},
                eventsChannel:                         eventsChannel,
                collectorHTTPClient:                   collectorHTTPClient,
@@ -259,7 +257,6 @@ type manager struct {
        allowDynamicHousekeeping bool
        includedMetrics          container.MetricSet
-       DenyList                 *metrics.DenyList
        containerWatchers        []watcher.ContainerWatcher
        eventsChannel            chan watcher.ContainerEvent
        collectorHTTPClient      *http.Client

```


iwankgb (approved):

Alexbyy:
I think this is a problem,because I think all of configs should be in one struct,but it isn't.For example,when we new an instance of manager,we need pass HouskeepingConfig and includedMetricsSet as argument to New().But I think both of   HouskeepingConfig  and 
includedMetricsSet is config.it should be in one struct.For that,when we add new configs,wo don't have to change the signature of manager.go New(),because we can add the new configs to the struct of configs.  @bobbypage @iwankgb

iwankgb:
@Alexbyy I think that @bobbypage is trying to stabilise cAdvisor API used in Kubernetes. It's a good idea but I also think that in long run Kubernetes must become independent of cAdvisor. There is some work around it happening but before if happens we will need to find a way of introducing your change unobtrusively.

PR #2828, opened 2021-03-06
Conditionally gathering FS usage metrics.

### LINKED ISSUE #1936: cannot disable ZFS metrics
I run cadvisor via a docker container like this:
```
  cadvisor:
    image: google/cadvisor:latest
    user: root
    privileged: true
    volumes:
      - /:/rootfs:ro
      - /cgroup:/cgroup:ro
      - /dev/zfs:/host/dev/zfs:ro
      - /proc:/host/proc:ro
      - /sys:/sys:ro
      - /var/lib/docker/:/var/lib/docker:ro
      - /var/run:/var/run:rw
    command:
      # tcp and udp create high CPU usage, disk does CPU hungry ``zfs list``
      - '--disable_metrics=tcp,udp,disk'
    restart: unless-stopped
    ports:
      - 9110:8080
```

cAdvisor does a load-heavy ``zfs list`` which is eating around 40-100% CPU time.
I want to disable it.
the disable-metrics=tcp,udb,disk doesn't seem to have any effect on this.

iwankgb:
Fixes #1936.

As far as I can tall all the `!= nil` has already been in place. Tested on:
* Docker with ZFS pool
* Docker with overlay

### DIFF
--- a/container/docker/factory.go
+++ b/container/docker/factory.go
@@ -347,23 +347,25 @@ func Register(factory info.MachineInfoFactory, fsInfo fs.FsInfo, includedMetrics
 	var (
 		thinPoolWatcher *devicemapper.ThinPoolWatcher
 		thinPoolName    string
+		zfsWatcher      *zfs.ZfsWatcher
 	)
-	if storageDriver(dockerInfo.Driver) == devicemapperStorageDriver {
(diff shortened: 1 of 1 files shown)


bobbypage:
LGTM, thanks for fix!

bobbypage (approved):

Issue #2622, opened 2020-07-17
Enabling container_referenced_memory metric cause metrics refresh slowdown


wacuuu:
I did some testing with big, memory intensive workloads. I used pmbench from here https://bitbucket.org/jisooy/pmbench/src/master/ and started to play with working set size around 1TB. I looked into how long will it take to execute h.libcontainerHandler.GetStats() function. My container agent is docker. Besides the workloads and cadvisor, nothing else was running on the node during tests. Command used to gather stats:

```
sudo ./cadvisor --disable_metrics=advtcp,process,sched,hugetlb,cpu_topology,resctrl,tcp,udp,percpu,accelerator,disk,diskIO,network --housekeeping_interval=5s --referenced_reset_interval=5 2>&1
```

Here are the results:

Number of containers | Working set size per container[GB] | Total ‘hot’ memory size[GB] | Getstats() execution time
-- | -- | -- | --
1 | 1024 | 1024 | 4.8s peak 9s
10 | 102.4 | 1024 | 650ms peak 1.2s
100 | 10.24 | 1024 | 500ms peak 1.1s

Last measurement may be a little bit off, as the machine got cpu overloaded by the number of containers. For comparison, with load as in the first test, GetStats execution without collecting container_referenced_bytes takes averages around 1ms CPU usage drops to couple of %.

It is important to note, that with container_referenced_bytes in the biggest single working set size cadvisor was able to consume a whole CPU core, with a hundred containers it used around 70% of one core(though as mentioned this result may be not accurate) and without it dropped to couple of %.

Two issues are related to this:

- Data refresh is slower, which is more of something to remember about than a real issue, as there is no real reason to gather stats in frequency much higher than scraping frequency(which reasonable value is usually set to couple of seconds)
- Real issue is that in this situation cadvisor consumes a lot of CPU. In case of the first test it was able fill one core fully. On production system it would be an issue. On the other hand deploying container with cpu limitation, as previous experiments show, always result in huge scraping slowdown. So the limitation of cpu usage should not come from resource limitation but from code optimization itself.

Therefore I think this is something at least worth addressing in documentation, that big working set sizes will slow down the referenced bytes readings, and as the code is sequential in this part, it will slow down statistics refreshment process.

dashpole:
@katarzyna-z

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2829, opened 2021-03-08
Expose OOM event count to prometheus

### LINKED ISSUE #1837: Getting docker container oom event.
Hello. 

Is there a way to get container oom event as a prometheus metric?
I am using cAdvisor to monitor docker containers with Prometheus and Grafana. I can see that there is a **oom_kill_event** in the API, but cannot find something similar in prometheus metrics.

kragniz:
Fixes #1837.

### DIFF
--- a/cmd/cadvisor.go
+++ b/cmd/cadvisor.go
@@ -112,6 +112,7 @@ var (
 		container.CPUTopologyMetrics:             struct{}{},
 		container.ResctrlMetrics:                 struct{}{},
 		container.CPUSetMetrics:                  struct{}{},
+		container.OOMMetrics:                     struct{}{},
 	}
 )
 
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/cadvisor/accelerators"
@@ -35,7 +36,7 @@ import (
 	"github.com/google/cadvisor/events"
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
-	"github.com/google/cadvisor/info/v2"
+	v2 "github.com/google/cadvisor/info/v2"
 	"github.com/google/cadvisor/machine"
 	"github.com/google/cadvisor/nvm"
 	"github.com/google/cadvisor/perf"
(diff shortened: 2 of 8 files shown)


kragniz:
I'm not convinced this is the best way to get the event data through, but seemed like the most straightforward. If this is an acceptable method, I can add tests and tidy this up further.

manager/manager.go
```diff
@@ -1237,6 +1237,24 @@ func (m *manager) watchForNewOoms() error {
...
+			if err != nil {
+				klog.Errorf("failed getting container info for %q: %v", oomInstance.ContainerName, err)
+				continue
+			}
+			if len(conts) != 1 {
+				klog.Errorf("Expected the request to match only one container")
+				continue
+			}
+			for _, cont := range conts {
+				cont.oomEvents++
```
iwankgb, line 1256: [It is not safe](https://golang.org/ref/mem) to increment this counter from one goroutine while another might read from it.

manager/manager.go
```diff
@@ -1237,6 +1237,24 @@ func (m *manager) watchForNewOoms() error {
...
 			}
+
+			// Count OOM events for later collection by prometheus
+			request := v2.RequestOptions{
+				IdType: v2.TypeName,
+				Count:  1,
+			}
+			conts, err := m.getRequestedContainers(oomInstance.ContainerName, request)
+			if err != nil {
+				klog.Errorf("failed getting container info for %q: %v", oomInstance.ContainerName, err)
```
iwankgb, line 1248: This is not an error if it's fine to `continue`.

manager/manager.go
```diff
@@ -1237,6 +1237,24 @@ func (m *manager) watchForNewOoms() error {
...
+				continue
+			}
+			if len(conts) != 1 {
+				klog.Errorf("Expected the request to match only one container")
```
iwankgb, line 1252: Same here - this is not an error.

manager/manager.go
kragniz: What change do you suggest? Replace the increment with `atomic.AddUint64`?

manager/manager.go
kragniz: Sure, what's the wanted log level? (`Warningf`?)

cmd/cadvisor.go
```diff
@@ -112,6 +112,7 @@ var (
 		container.CPUTopologyMetrics:             struct{}{},
 		container.ResctrlMetrics:                 struct{}{},
 		container.CPUSetMetrics:                  struct{}{},
+		container.OomMetrics:                     struct{}{},
```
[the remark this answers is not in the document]
iwankgb: We can't change names of existing exported fields, but we should not add new that repeat old errors :)

manager/manager.go
iwankgb: `klog.V(2).Info()` could make some sense, I think.

manager/manager.go
iwankgb: Or using a lock. Keep in mind that no matter what you do you will have to serialize reads too.

kragniz: Moved it to atomics

kragniz:
@iwankgb thanks for the reviews, I've updated with fixes

iwankgb (approved):

darkl0rd:
This metric (Docker) always returns 0 for me - see also #2622.

PR #2837, opened 2021-03-24
Fix read "max" value for cgroup v2

giuseppe:
and also simplify some other code to behave like cgroup v1

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -203,7 +203,10 @@ func readString(dirpath string, file string) string {
 
 func readUInt64(dirpath string, file string) uint64 {
 	out := readString(dirpath, file)
-	if out == "" || out == "max" {
+	if out == "max" {
+		return math.MaxUint64
+	}
+	if out == "" {
 		return 0
 	}
 
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -17,9 +17,9 @@ package common
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -48,25 +48,6 @@ func DebugInfo(watches map[string][]string) map[string][]string {
 	out["Inotify watches"] = lines
 
 	return out
-}
-
-// findFileInAncestorDir returns the path to the parent directory that contains the specified file.
-// "" is returned if the lookup reaches the limit.
-func findFileInAncestorDir(current, file, limit string) (string, error) {
-	for {
-		fpath := path.Join(current, file)
-		_, err := os.Stat(fpath)
-		if err == nil {
-			return current, nil
-		}
-		if !os.IsNotExist(err) {
-			return "", err
-		}
-		if current == limit {
-			return "", nil
-		}
-		current = filepath.Dir(current)
-	}
 }
 
 var bootTime = func() time.Time {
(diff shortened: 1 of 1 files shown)


giuseppe:
@odinuge PTAL

container/common/helpers.go
```diff
@@ -165,11 +146,7 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
 				spec.Memory.Reservation = readUInt64(memoryRoot, "memory.soft_limit_in_bytes")
 			}
 		} else {
-			memoryRoot, err := findFileInAncestorDir(memoryRoot, "memory.max", "/sys/fs/cgroup")
-			if err != nil {
-				return spec, err
-			}
-			if memoryRoot != "" {
+			if utils.FileExists(path.Join(memoryRoot, "memory.max")) {
```
odinuge, line 149: :+1:

container/common/helpers.go
```diff
@@ -226,7 +203,10 @@ func readString(dirpath string, file string) string {
 
 func readUInt64(dirpath string, file string) uint64 {
 	out := readString(dirpath, file)
-	if out == "" || out == "max" {
+	if out == "max" {
+		return math.MaxUint64
```
odinuge, line 207: nit: The max value for cgroup v1 is 9223372036854771712 (max signed integer rounded to nearest page). So to be correct it should be something like this I guess:

```suggestion
		return math.MaxInt64^0xFFF
```

odinuge (commented):
Just a small nit, otherwise lgtm

container/common/helpers.go
giuseppe: I am fine with changing it, but would it be a problem with cgroup v1?  `"max"` is used only on cgroup v2

odinuge: I don't remember where this value is used, but I just thought it was nice with the same value for both cgroup v2. If you try to write `MaxInt64^0xFFF  + 0x1` you get `max` on cgroup v2, and `9223372036854771712` on v1. 9223372036854771712 is also default on v1.

odinuge: Edit: This is used for more than the memory limits, and my above comments only apply to those. Hmm

giuseppe:
a similar fix for cpu specs: https://github.com/google/cadvisor/pull/2839

container/common/helpers.go
```diff
@@ -226,7 +203,10 @@ func readString(dirpath string, file string) string {
 
 func readUInt64(dirpath string, file string) uint64 {
 	out := readString(dirpath, file)
-	if out == "" || out == "max" {
+	if out == "max" {
```
iwankgb, line 206: Why have you decided to change behaviour for `out == ""`?

giuseppe: for `out == ""` I've left the same behavior there was before 2ccad4b42fe52b312f6d75a312c61a54fd123dc5

odinuge:
This change is probably related to:
https://github.com/google/cadvisor/blob/fa0733216494720e3fc301bdf046553f407a7131/manager/manager.go#L473-L479

Does that code do anything at all if this change merged?


Also, the description of this value is probably wrong, since it says -1 is default, but it is an unsigned value...

https://github.com/google/cadvisor/blob/fa0733216494720e3fc301bdf046553f407a7131/info/v2/container.go#L46-L48

giuseppe:
> [](/google/cadvisor/commit/fa0733216494720e3fc301bdf046553f407a7131)

yes it will be ignored.

And I think this is the expectation, at least from Kubernetes, since there is a e2e tests failing right now as the memory limit is being set while it is expected to be unlimited (`isMemoryUnlimited` in pkg/kubelet/stats/helper.go)

container/common/helpers.go
iwankgb: Of course, my bad!

iwankgb (approved):

PR #2839, opened 2021-03-25
helpers: fix reading cpu stats on cgroup v2

giuseppe:
read quota and limit from cpu.max, and convert cpu.weight to the range
expected with cgroup v1.


container/common/helpers.go
```diff
@@ -122,18 +123,43 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
...
+						spec.Cpu.Limit = limit
+					}
 				}
+				max := readString(cpuRoot, "cpu.max")
+				if max != "" {
+					splits := strings.SplitN(max, " ", 2)
+					if len(splits) != 2 {
+						klog.Errorf("GetSpec: Failed to parse CPUmax from %q", path.Join(cpuRoot, "cpu.max"))
+					} else {
+						spec.Cpu.Quota = parseUint64String(splits[0])
```
odinuge, line 158: https://github.com/google/cadvisor/blob/013e451ee34b9ea516262ad0cdcd856f838c45a5/info/v2/container.go#L39-L40

Ref. comment above and impl for v1, `spec.Cpu.Quota` should be `0` when not enabled/`max`.

container/common/helpers.go
```diff
@@ -122,18 +123,43 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
...
-			spec.Cpu.Period = readUInt64(cpuRoot, "cpu.cfs_period_us")
-			quota := readString(cpuRoot, "cpu.cfs_quota_us")
-
-			if quota != "" && quota != "-1" {
-				val, err := strconv.ParseUint(quota, 10, 64)
-				if err != nil {
-					klog.Errorf("GetSpec: Failed to parse CPUQuota from %q: %s", path.Join(cpuRoot, "cpu.cfs_quota_us"), err)
-				} else {
-					spec.Cpu.Quota = val
+			if !cgroups.IsCgroup2UnifiedMode() {
```
odinuge, line 126: nit: any reason for using `!cgroups.IsCgroup2UnifiedMode()` instead of just `cgroups.IsCgroup2UnifiedMode()`?

odinuge (commented):
Awesome! Was going to fix this, but didn't have time... :sweat_smile:  This fixes quota related metrics for cadvisor and k8s. :smile:

container/common/helpers.go
giuseppe: not really... going to fix it

container/common/helpers.go
giuseppe: thanks!  Fixed

container/common/helpers.go
```diff
@@ -122,17 +123,43 @@ func GetSpec(cgroupPaths map[string]string, machineInfoFactory info.MachineInfoF
...
+			} else {
+				spec.HasCpu = true
+				spec.Cpu.Limit = readUInt64(cpuRoot, "cpu.shares")
+				spec.Cpu.Period = readUInt64(cpuRoot, "cpu.cfs_period_us")
+				quota := readString(cpuRoot, "cpu.cfs_quota_us")
+
+				if quota != "" && quota != "-1" {
+					val, err := strconv.ParseUint(quota, 10, 64)
+					if err != nil {
+						klog.Errorf("GetSpec: Failed to parse CPUQuota from %q: %s", path.Join(cpuRoot, "cpu.cfs_quota_us"), err)
```
iwankgb, line 159: Consider using `%w` instead of `%s` to wrap error.

container/common/helpers.go
```diff
@@ -224,6 +251,31 @@ func readString(dirpath string, file string) string {
 	return strings.TrimSpace(string(out))
 }
 
+// Convert from [1-10000] to [2-262144]
+func convertCPUWeightToCPULimit(weight uint64) (uint64, error) {
+	if weight < 1 || weight > 10000 {
```
iwankgb, line 256: Can you introduce constants instead of using magic numbers, please? It would help me to understand what these two values represent.

container/common/helpers.go
```diff
@@ -224,6 +251,31 @@ func readString(dirpath string, file string) string {
...
+		return 0, fmt.Errorf("convertCPUWeightToCPULimit: invalid cpu weight: %v", weight)
+	}
+	return 2 + ((weight-1)*262142)/9999, nil
+}
+
+func parseUint64String(strValue string) uint64 {
```
iwankgb, line 262: Unit test would be appreciated.

iwankgb (commented):
My 3 cents. Do you think it would be possible to mock both version of cgroupfs and run create a test that would read from such a file system?

container/common/helpers.go
giuseppe: isn't that supported only with `fmt.Errorf`?

container/common/helpers.go
giuseppe: sure, going to clean it up

container/common/helpers.go
giuseppe: added

giuseppe:
> Do you think it would be possible to mock both version of cgroupfs and run create a test that would read from such a file system?

I guess it should be doable, I'll take a look

giuseppe:
> Do you think it would be possible to mock both version of cgroupfs and run create a test that would read from such a file system?

is it a blocker for this PR?

container/common/helpers.go
iwankgb: Yes, it is. My mistake, again.

iwankgb:
> is it a blocker for this PR?

It would be appreciated. If you think it's undoable or difficult to do (e.g. because of cAdvisor architecture then let me know, please).

giuseppe:
> It would be appreciated. If you think it's undoable or difficult to do (e.g. because of cAdvisor architecture then let me know, please).

thanks.  Tests added as you've suggested

iwankgb (approved):

PR #2853, opened 2021-04-26
Upgrade cAdvisor version in k8s deployment.

Creatone:

### DIFF
--- a/deploy/kubernetes/base/daemonset.yaml
+++ b/deploy/kubernetes/base/daemonset.yaml
@@ -20,11 +20,11 @@ spec:
         image: gcr.io/cadvisor/cadvisor:v0.39.0
         resources:
           requests:
-            memory: 1000Mi
-            cpu: 500m
+            memory: 400Mi
+            cpu: 400m
           limits:
             memory: 2000Mi
-            cpu: 1
+            cpu: 800m
         volumeMounts:
         - name: rootfs
           mountPath: /rootfs
--- a/deploy/kubernetes/base/daemonset.yaml
+++ b/deploy/kubernetes/base/daemonset.yaml
@@ -17,14 +17,14 @@ spec:
       serviceAccountName: cadvisor
       containers:
       - name: cadvisor
-        image: k8s.gcr.io/cadvisor:v0.30.2
+        image: gcr.io/cadvisor/cadvisor:v0.39.0
         resources:
           requests:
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/base/daemonset.yaml
```diff
@@ -17,14 +17,14 @@ spec:
       serviceAccountName: cadvisor
       containers:
       - name: cadvisor
-        image: k8s.gcr.io/cadvisor:v0.30.2
+        image: gcr.io/cadvisor/cadvisor:v0.39.0
         resources:
           requests:
-            memory: 200Mi
-            cpu: 150m
+            memory: 1000Mi
```
bobbypage, line 23: just curious, why is the bump in cpu and memory needed?

Creatone: Got this one:
```
[pszulik@node200 ~]$ curl 10.233.79.136:8080/metrics > out
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed connect to 10.233.79.136:8080; Connection refused

```
Lack of resources...

bobbypage: I see, sounds good to bump then. However, it seems like a large bump (requests 200Mi -> 1000Mi ; cpu 100m -> 500m). I don't have any objections but just curious how you picked those... or if it's just a rough guess :)

iwankgb: It has not been changed for a while but such a huge increase is a bit worrying to me too. @Creatone have you tried to set any lower value?

dashpole: Resources are going to vary based on the number of containers on the node.
You should also check to see if there were any additional --disable_metrics added during that time period that we need to ignore (if we are ignoring any).

Creatone: Can you now check out the values?

iwankgb:
@bobbypage there are few recently reported issues about performance. I think we can merge this PR and investigate reported issues.

iwankgb (approved):

bobbypage:
Sounds good, LGTM.

PR #2859, opened 2021-05-04
Add custom error type if device not found in partitions map

saschagrunert:
The new error type `ErrDeviceNotInPartisionsMap` can be used by external
consumers like Kubernetes to distinguish certain error paths.

Refers to https://github.com/kubernetes/kubernetes/pull/100448

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -19,6 +19,7 @@ package fs
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -55,6 +56,9 @@ const (
 
 // A pool for restricting the number of consecutive `du` and `find` tasks running.
 var pool = make(chan struct{}, maxConcurrentOps)
+
+// ErrDeviceNotInPartitionsMap is the error resulting if a device could not be found in the partitions map.
+var ErrDeviceNotInPartitionsMap = errors.New("could not find device in cached partitions map")
 
 func init() {
 	for i := 0; i < maxConcurrentOps; i++ {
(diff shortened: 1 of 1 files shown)


fs/fs.go
```diff
@@ -56,6 +57,9 @@ const (
 // A pool for restricting the number of consecutive `du` and `find` tasks running.
 var pool = make(chan struct{}, maxConcurrentOps)
 
+// ErrDeviceNotInPartisionsMap is the error resulting if a device could not be found in the partitions map.
+var ErrDeviceNotInPartisionsMap = errors.New("could not find device in cached partitions map")
```
bobbypage, line 61: nit: s/ErrDeviceNotInPartisionsMap/ErrDeviceNotInPartitionsMap

saschagrunert: Yep, fixed as suggested :+1:

bobbypage:
thanks! One more question, is there any reason for using `errors.Wrapf` and `github.com/pkg/errors` instead of native go %w wrapping?

saschagrunert:
> thanks! One more question, is there any reason for using `errors.Wrapf` and `github.com/pkg/errors` instead of native go `%w` wrapping?

It was just my personal convenience to use the package as I saw that it's already used in that project. Should I switch to `fmt` wrapping?

bobbypage:
I don't have a strong preference, but if the goal is to check for this error downstream in other projects (e.g. in k8s) it seems like it might be best to use standard go error wrapping.

iwankgb:
@saschagrunert, do you mind using stdlib error wrapping? We probably can get rid of `pkg/errors` from the project now.

saschagrunert:
> @saschagrunert, do you mind using stdlib error wrapping? We probably can get rid of `pkg/errors` from the project now.

Sure, changed the wrapping to use `%w`.

mrunalp:
lgtm

bobbypage:
LGTM, thanks!

bobbypage (approved):

Issue #2865, opened 2021-05-08
Perf: add group_id label


Creatone:
It must be possible to differ the same event metrics from different groups.

Perf events config group should be extended with id, e.g.

```json
{
  "core": {
    "events": [
      ["instructions", "cache-misses"],
      "cycles",
      {
        "group_id": "my_group",
        "group_events": ["instructions", "cycles"]
      }
    ]
  }
}
```
This would lead to:
- `["instructions", "cache-misses"]` -> `group_id="1"`
- `"cycles"` -> not sure if empty or `group_id="2"`
- the last one -> `group_id="my_group"`

@iwankgb

iwankgb:
Idea sounds reasonable to me. We should make sure that existing configs will still work.

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

PR #2866, opened 2021-05-10
Perf: Differ groups by the id.

Creatone:
Part of #2865

### DIFF
--- a/perf/uncore_libpfm_test.go
+++ b/perf/uncore_libpfm_test.go
@@ -137,7 +137,7 @@ func TestUncoreCollectorSetup(t *testing.T) {
 	assert.Equal(t, []string{"uncore_imc_1/cas_count_read"},
 		getMapKeys(collector.cpuFiles["1"]["uncore_imc_1"].cpuFiles))
 	assert.ElementsMatch(t, []string{"uncore_imc_0/cas_count_write", "uncore_imc_0/cas_count_read"},
-		getMapKeys(collector.cpuFiles["1"]["uncore_imc_0"].cpuFiles))
+		getMapKeys(collector.cpuFiles["3"]["uncore_imc_0"].cpuFiles))
 
 	// There are no errors.
 	assert.Nil(t, err)


eero-t:
Perf seems to support both metric/event group IDs, and cgroup IDs [1].  Which kind of perf group(s) this PR is about?

[1] perf-list:
```
EVENT GROUPS

       Perf supports time based multiplexing of events, when the number
       of events active exceeds the number of hardware performance
       counters. Multiplexing can cause measurement errors when the
       workload changes its execution profile.

       When metrics are computed using formulas from event counts, it is
       useful to ensure some events are always measured together as a
       group to minimize multiplexing errors. Event groups can be
       specified using { }.
```

perf-record:
```
       -e, --event=
           Select the PMU event. Selection can be:
...
           •   a group of events surrounded by a pair of brace
               ("{event1,event2,...}"). Each event is separated by
               commas and the group should be quoted to prevent the
               shell interpretation. You also need to use --group on
               "perf report" to view group events together.
...
       --group
           Put all events in a single event group. This precedes the
           --event option and remains only for backward compatibility.
...
       -G name,..., --cgroup name,...
           monitor only in the container (cgroup) called "name". This
           option is available only in per-cpu mode. The cgroup
           filesystem must be mounted. All threads belonging to
           container "name" are monitored when they run on the monitored
           CPUs. Multiple cgroups can be provided.
...
       If wanting to monitor, say, cycles for a cgroup and also for
       system wide, this command line can be used: perf stat -e cycles
       -G cgroup_name -a -e cycles.

       --namespaces
           Record events of type PERF_RECORD_NAMESPACES. This enables
           cgroup_id sort key.

       --all-cgroups
           Record events of type PERF_RECORD_CGROUP. This enables cgroup
           sort key.
```
perf-report:
```
       -s, --sort=
           Sort histogram entries by given key(s) - multiple keys can be
           specified in CSV format. Following sort keys are available:
           pid, comm, dso, symbol, parent, cpu, socket, srcline, weight,
           local_weight, cgroup_id.
...
           •   cgroup_id: ID derived from cgroup namespace device and
               inode numbers.

           •   cgroup: cgroup pathname in the cgroupfs.
...
       --group
           Show event group information together. It forces group output
           also if there are no groups defined in data file.

       --group-sort-idx
           Sort the output by the event at the index n in group. If n is
           invalid, sort by the first event. It can support multiple
           groups with different amount of events. WARNING: This should
           be used on grouped events.
```

Creatone:
@eero-t You're completely right. 
cAdvisor perf collector counts events for every cgroup_id (container). It's possible to match events in groups to be gathered at the same time.
The `group_id` label is needed to differ the same events gathered at different times in different groups.

eero-t:
> cAdvisor perf collector counts events for every cgroup_id (container). It's possible to match events in groups to be gathered at the same time.
> The `group_id` label is needed to differ the same events gathered at different times in different groups.

If it's about _per-cgroup_ `perf` data (instead of _per-event-group_ `perf` data), IMHO `group_id` should be be renamed to `cgroup_id`, to match `perf` documentation, and to clarify this confusion.  `GroupID` could then be renamed e.g. to `CgroupID`.

Creatone:
The `group_id` is not equal to `cgroup_id`. cAdvisor uses both parameters. Firstly differ events through the containers (cgroup_id [or "id" label in Prometheus format]), secondly through groups (group_id).

Issue #2867, opened 2021-05-10
GetRequestedContainersInfo failed as cache not synced


harche:
This issue is similar to https://github.com/google/cadvisor/issues/1970. While https://github.com/google/cadvisor/pull/1985 took care of that issue for the docker containers, that race condition still exists for non docker containers. 

We have seen [failures in CI](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/1121/pull-ci-openshift-cluster-monitoring-operator-master-e2e-aws-single-node/1383816160142692352) due to this existing race condition for the non docker containers. Typically we could see following messages in kubelet logs,

```
Apr 25 08:38:30.732009 ip-10-0-207-255 hyperkube[1580]: W0425 08:38:30.731960    1580 prometheus.go:1856] Couldn't get containers: partial failures: ["/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-podfb5503c7_7473_4520_9dad_541def516acc.slice": containerDataToContainerInfo: unable to find data in memory cache], ["/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-podfb5503c7_7473_4520_9dad_541def516acc.slice/crio-d15120b05438f37d9b5a8b7b6584b80aa6e6073e1dae498a6d67c312455fe0b7.scope": containerDataToContainerInfo: unable to find data in memory cache]
[...]

PR #2868, opened 2021-05-10
Do not return the error when hitting the cache not synced issue

harche:
Fixes: https://github.com/google/cadvisor/issues/2867

Signed-off-by: Harshal Patil <harpatil@redhat.com>

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -693,6 +693,10 @@ func (m *manager) GetRequestedContainersInfo(containerName string, options v2.Re
 	for name, data := range containers {
 		info, err := m.containerDataToContainerInfo(data, &query)
 		if err != nil {
+			if err == memory.ErrDataNotFound {
+				klog.Warningf("Error getting data for container %s because of race condition", name)
+				continue
+			}
 			errs.append(name, "containerDataToContainerInfo", err)
 		}
 		containersMap[name] = info


manager/manager.go
```diff
@@ -693,6 +693,10 @@ func (m *manager) GetRequestedContainersInfo(containerName string, options v2.Re
 	for name, data := range containers {
 		info, err := m.containerDataToContainerInfo(data, &query)
 		if err != nil {
+			if err == memory.ErrDataNotFound {
+				klog.Warningf("Error getting data for container %s because of race condition", name)
```
iwankgb, line 697: What makes you think that it's caused by race condition? Is there anything we could do to fix it?

harche: Sorry for the confusion, maybe I should have phrased it better. The race condition exists when the [metrics for all the containers are requested](https://github.com/google/cadvisor/blob/5d029c266a072305c0239990c714345b680f7d42/metrics/prometheus.go#L1864) by a prometheus metric. While the cdavisor is going over a loop of containers, there is a possibility that some of those containers might have been cleaned up from the system as well as the internal cache of the cadvisor. 

This problem had existed even earlier but it was fixed by this PR, https://github.com/google/cadvisor/pull/1985/ but only for the docker containers. That PR made sure that if for docker containers such cache miss happens instead of returning the error to the upper layers it will simply log it. For more info on the original issue, https://github.com/google/cadvisor/issues/1970

I simply had to add the same logic for other containers that already exists for the docker containers. So to maintain the consistency between the error messages I copied the error message returned by the [original implementation](https://github.com/google/cadvisor/pull/1985/files#diff-f679b61276b44dc3f4712950536707c0389ee4c536fc29837c0202c6e81c1bbfR633) for the docker containers.

iwankgb: Got it!

iwankgb (approved):

Issue #2785, opened 2021-01-11
Disk usage metrics for containerd


ribbybibby:
When switching from docker to containerd as my container runtime in Kubernetes, I noticed that `container_fs_usage_bytes` metrics were no longer being exported for my containers.

It looks like disk usage metrics aren't implemented for containerd, as noted by this comment: https://github.com/google/cadvisor/blob/v0.38.6/container/containerd/handler.go#L164-L165.

Disk usage is a pretty important metric to monitor, so I think, if possible, this should be added.

elcomtik:
I have experienced the same issue. 

> It looks like disk usage metrics aren't implemented for containerd, as noted by this comment: https://github.com/google/cadvisor/blob/v0.38.6/container/containerd/handler.go#L164-L165.

There is some conversation about these metrics https://github.com/containerd/containerd/issues/678. I suppose that contained provide this information.

yyrdl:
PR was submited  . https://github.com/google/cadvisor/pull/2872

jepio:
PR #2872 was closed, in favor of #2956, which was merged and subsequently reverted in #2964. The result is that these metrics are not available.

Is someone working on an alternative approach?

sarbajitc:
Is there any timeline for a fix of this issue?

baasumo:
+1 on looking for any update or timeline regarding this issue - these metrics are pretty important for observability and workload behavior.

fernandesnikhil:
Adding onto this ticket since we're blocked on switching to the `containerd` CRI without these metrics. We have alerting around ephemeral file system usage that would break if cAdvisor doesn't collect these from `containerd`.

snuggie12:
@bobbypage Do you have an update on this? Best I can follow is that there is a possibly-working version in the `containerd-cri` branch after #2966 was merged. However, it might be incomplete based on https://github.com/google/cadvisor/pull/2936#issuecomment-971619663?

Alternatively it seems like work has gone into not using cadvisor for container stats and k8s 1.23 has an alpha feature-gate which [uses the cri stats provider](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates) (`PodAndContainerStatsFromCRI`). Is the plan to put momentum into that instead? If so, do you know when it would go beta?

brandond:
The `PodAndContainerStatsFromCRI` feature-gate does not seem to work either; at least with containerd 1.6.4 the stats are still missing.

brandond:
[...]

PR #2872, opened 2021-05-19
support collecting FsUsageMetrics for containerd

yyrdl:
__What's added?__

   Collect container's disk usage metrics  for containerd .

__Related Issue:__ [Disk usage metrics for containerd #2785](https://github.com/google/cadvisor/issues/2785)

__Note:__

 This pr only report the usage and inode information powered by containerd's standard API. Some disk usage information is missing until there is way to get the id of the snapshot or  containerd enhances it's API.But for  now, we can report what we can get . 

We urgently need  the disk usage of the container，these information is ok for our scenario in Tencent ,although it's not complete.

For k8s users, cadvisor is integrated in kubelet , you need to apply these changes to your kubelet (modify cadvisor's code  in vendor ), and check the `New` method in `kubernetes/pkg/kubelet/cadvisor/cadvisor_linux.go`, if there is an conditional statement  like below:

```golang 
if usingLegacyStats  {
    includedMetrics[cadvisormetrics.DiskUsageMetrics] = struct{}{}
}
```
please add an new condition here，[Related discussion](https://github.com/kubernetes/kubernetes/pull/91366/files#diff-ca41d4fa89358cfbf866e71de1c54ab978a448e805495bf8bca923d85dae31fbR101)：

```golang 
if usingLegacyStats || utilfeature.DefaultFeatureGate.Enabled(kubefeatures.LocalStorageCapacityIsolation) {
    includedMetrics[cadvisormetrics.DiskUsageMetrics] = struct{}{}
}
```


yyrdl:
> @googlebot I signed it!

yyrdl:
@googlebot I fixed it.

bobbypage:
Thanks for PR! 

Can you clarify what cAdvisor prometheus metrics this PR aims to add to for containerd? Is it just `container_fs_usage_bytes`?

container/containerd/client.go
```diff
@@ -125,6 +130,21 @@ func (c *client) Version(ctx context.Context) (string, error) {
 	return response.Version, nil
 }
 
+func (c *client) ContainerFsUsage(ctx context.Context, snapshotter, snapshotkey string) (*common.FsUsage, error) {
```
bobbypage, line 133: does `ContainerFsUsage` need to be exported?

container/containerd/client.go
```diff
@@ -125,6 +130,21 @@ func (c *client) Version(ctx context.Context) (string, error) {
...
+	usage, err := c.snapshotsService.Usage(ctx, &snaptshotapi.UsageRequest{
```
bobbypage, line 134: how expensive is this?

will containerd snapshotter have to recaculate the usage every time this is called or does it cache the usage internally?

container/containerd/client.go
```diff
@@ -125,6 +130,21 @@ func (c *client) Version(ctx context.Context) (string, error) {
...
+		Snapshotter: snapshotter,
+		Key:         snapshotkey,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &common.FsUsage{
+		BaseUsageBytes:  uint64(usage.Size_),
+		TotalUsageBytes: uint64(usage.Size_),
```
bobbypage, line 143: is it correct that these are both set to `usage.Size`? 

BaseUsageBytes is "Number of bytes consumed by a container through its root filesystem." vs TotalusageBytes is "Total Number of bytes consumed by container." ?

bobbypage:
/cc @crosbymichael 

There was earlier slack thread about of FS Usage containerd metrics on cadvisor [xref](https://kubernetes.slack.com/archives/C0BP8PW9G/p1605899033213500)

container/common/fsHandler.go
```diff
@@ -37,15 +36,18 @@ type FsUsage struct {
 	InodeUsage      uint64
 }
 
+type FsUsageProvider interface {
+	Usage() (*FsUsage, error)
+	Targets() []string
```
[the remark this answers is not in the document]
yyrdl: added

container/common/fsHandler.go
```diff
@@ -58,51 +60,35 @@ const DefaultPeriod = time.Minute
...
-	if fh.extraDir != "" {
-		extraUsage, extraErr = fh.fsInfo.GetDirUsage(fh.extraDir)
+	usage, err := fh.usageProvider.Usage()
+
+	if err != nil {
+		return err
 	}
 
-	// Wait to handle errors until after all operartions are run.
+	// Wait to handle errors until after all operations are run.
```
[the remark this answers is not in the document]
yyrdl: removed

container/containerd/client.go
yyrdl: yes , this client  is delivered as interface, if this method is not exported, we can't get the fs usage.

```go
func newContainerdContainerHandler(
	client ContainerdClient,//  Here
	name string,
	machineInfoFactory info.MachineInfoFactory,
	fsInfo fs.FsInfo,
	cgroupSubsystems *containerlibcontainer.CgroupSubsystems,
	inHostNamespace bool,
	metadataEnvs []string,
	includedMetrics container.MetricSet,
) (container.ContainerHandler, error)
```

container/containerd/client.go
yyrdl: Containerd does not cache it  and will recaculate again when we send a request,but we update it at a fixed frequency, see `realFsHandler.trackUsage` in `cadvisor/container/common/fsHandler.go`, so I think it's ok .

And compare with the scenario of docker, it's same,the containerd only caculate the usage of upper(or diff)  directory,not the rootfs (merged).

container/containerd/client.go
yyrdl: Yes ,for containerd ,there is no log file(used to store logs from container's stdout or stderr) (At least ,I didn't find it T_T) . But for docker and crio , there are log files, so when collect fs usage, the usage of log file is counted.

By the way, kubelet storge container's  logs in another place (`/var/log/pods`), and it is counted by kubelet when it caculate the usage. see `makeContainerStats` in `kubelet/stats/cri_stats_provider.go`

```go
// NOTE: This doesn't support the old pod log path, `/var/log/pods/UID`. For containers
	// using old log path, empty log stats are returned. This is fine, because we don't
	// officially support in-place upgrade anyway.
	var (
		containerLogPath = kuberuntime.BuildContainerLogsDirectory(meta.GetNamespace(),
			meta.GetName(), types.UID(meta.GetUid()), container.GetMetadata().GetName())
		err error
	)
	result.Logs, err = p.getPathFsStats(containerLogPath, rootFsInfo)
	if err != nil {
		klog.Errorf("Unable to fetch container log stats for path %s: %v ", containerLogPath, err)
	}
	return result
```

yyrdl:
> Thanks for PR!
> 
> Can you clarify what cAdvisor prometheus metrics this PR aims to add to for containerd? Is it just `container_fs_usage_bytes`? What other filesystem metrics did you mention are still missing on containerd?

There are many metrics defined in [fs.FsStats](https://github.com/google/cadvisor/blob/master/info/v1/container.go#L732),at current ,we only get `Inode` ，`BaseUsage` and `Usage`. 

For prometheus metrics defined [here](https://github.com/google/cadvisor/blob/master/metrics/prometheus.go#L548) ,we only lost `container_fs_inodes_free `  and `container_fs_limit_bytes ` .

yyrdl:
And  thanks for your reply :)

yyrdl:
@bobbypage Hello, is there anything else to check？

container/containerd/client.go
yyrdl: Yes, there is a cache in it's CRI implemention . But `Usage` method will recaculate the usage every time, I have change it to `ContainerStats`

bobbypage:
Can you please rebase? There appears to be some go.mod changes which is causing `pull-cadvisor-e2e` to fail

container/containerd/client.go
```diff
@@ -28,20 +28,24 @@ import (
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/pkg/dialer"
 	ptypes "github.com/gogo/protobuf/types"
+	"github.com/google/cadvisor/container/common"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
```
bobbypage, line 34: I'm not sure if we can depend on k8s repos since cadvisor itself is imported into k8s which can result in an import loop. 

See https://github.com/google/cadvisor/pull/2726 and https://github.com/google/cadvisor/pull/2726#issuecomment-725363880

Do we need to import criapi here? Can we just talk to containerd directly?

yyrdl: It's ok ,  this package is just pb files.

And there are two way to get the diskusage metrics : 
* call snapshots.Usage()
* call cri.ContainerStats()

They are both grpc interface implemented by containerd , the difference is that `snapshots.Usage` didn‘t cache any thing, while containerd cri server will sync the disk usage metrics periodically, see [snapshotsSyncer.sync](https://github.com/containerd/containerd/blob/main/pkg/cri/server/snapshots.go#L71) ,in  order to avoid collecting disk usage metric repeatedly , we should use the cri method.

yyrdl:
> Can you please rebase? There appears to be some go.mod changes which is causing `pull-cadvisor-e2e` to fail

I have  rebased to master , but it still failed .
截屏2021-06-22 上午9 49 03
I found `ssh.go:116] failed to run SSH command: out: prow@34.122.70.58: Permission denied (publickey).`  in log file, does  this cause error ?

container/containerd/client.go
bobbypage: I see, thanks for the background. Unfortunately, I don't think the fact that this package is just pb files matters; I think it is general issue with dependency management in k8s => 

as @dims mentioned [here](https://github.com/google/cadvisor/pull/2726#issuecomment-725363880)

> "third party software can use staging modules IFF the third party software is not vendored back into k/k :)"

"k8s.io/cri-api/" is staging module and cadvisor itself is vendored back into k/k. 

@dims, please keep me honest here, but I don't think anything has changed here.

bobbypage: Will be able to switch to `snapshots.Usage()` instead? Can we maybe cache the results internally / only collect them during some periodic house keeping period?

bobbypage:
I think the issue from CI job is

```
I0622 01:37:49.137] go tidy changes are needed; please run make tidy
I0622 01:37:49.137] go.mod diff:
I0622 01:37:49.137] --- /tmp/tmp.dBhNDSSSVc	2021-06-22 01:37:48.564770598 +0000
I0622 01:37:49.137] +++ go.mod	2021-06-22 01:37:49.080807336 +0000
I0622 01:37:49.138] @@ -30,7 +30,7 @@
I0622 01:37:49.138]  	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110
I0622 01:37:49.138]  	golang.org/x/sys v0.0.0-20210426230700-d19ff857e887
I0622 01:37:49.138]  	google.golang.org/grpc v1.33.2
I0622 01:37:49.138] -	k8s.io/cri-api v0.21.2 // indirect
I0622 01:37:49.138] +	k8s.io/cri-api v0.21.2
I0622 01:37:49.138]  	k8s.io/klog/v2 v2.4.0
I0622 01:37:49.138]  	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
I0622 01:37:49.138]  )
I0622 01:37:49.139] go.sum diff:
I0622 01:37:49.139] --- /tmp/tmp.mYnk2jG8Wq	2021-06-22 01:37:48.568770882 +0000
I0622 01:37:49.139] +++ go.sum	2021-06-22 01:37:49.120810185 +0000
I0622 01:37:49.139] @@ -418,8 +418,8 @@
I0622 01:37:49.139]  github.com/kr/pretty v0.2.1/go.mod h1:ipq/a2n7PKx3OHsz4KJII5eveXtPO4qwEXGdVfWzfnI=
I0622 01:37:49.139]  github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
I0622 01:37:49.139]  github.com/kr/pty v1.1.5/go.mod h1:9r2w37qlBe7rQ6e1fg1S/9xpWHSnaqNdHD3WcMdbPDA=
I0622 01:37:49.139] -github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
I0622 01:37:49.140]  github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
I0622 01:37:49.140] +github.com/kr/text v0.2.0 h1:5Nx0Ya0ZqY2ygV366QzturHI13Jq95ApcVaJBhpS+AY=
I0622 01:37:49.140]  github.com/kr/text v0.2.0/go.mod h1:eLer722TekiGuMkidMxC/pM04lWEeraHUUmBw8l2grE=
I0622 01:37:49.140]  github.com/magiconair/properties v1.8.0/go.mod h1:PppfXfuXeibc/6YijjN8zIbojt8czPbwD3XqdrwzmxQ=
I0622 01:37:49.140]  github.com/mailru/easyjson v0.0.0-20190614124828-94de47d64c63/go.mod h1:C1wdFJiN94OJF2b5HbByQZoLdCWB1Yqtg26g4irojpc=
I0622 01:37:49.140] @@ -462,6 +462,7 @@
I0622 01:37:49.140]  github.com/mwitkow/go-conntrack v0.0.0-20161129095857-cc309e4a2223/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
I0622 01:37:49.141]  github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f/go.mod h1:ZdcZmHo+o7JKHSa8/e818NopupXU1YMK5fe1lsApnBw=
I0622 01:37:49.141]  github.com/ncw/swift v1.0.47/go.mod h1:23YIA4yWVnGwv2dQlN4bB7egfYX6YLn0Yo/S6zZO/ZM=
I0622 01:37:49.141] +github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e h1:fD57ERR4JtEqsWbfPhv4DMiApHyliiK5xCTNVSPiaAs=
I0622 01:37:49.141]  github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e/go.mod h1:zD1mROLANZcx1PVRCS0qkT7pwLkGfwJo4zjcN/Tysno=
I0622 01:37:49.141]  github.com/nxadm/tail v1.4.4/go.mod h1:kenIhsEOeOJmVchQTgglprH7qJGnHDVpk1VPCcaMI8A=
I0622 01:37:49.141]  github.com/oklog/ulid v1.3.1/go.mod h1:CirwcVhetQ6Lv90oh/F+FBtV6XMibvdAFo93nm5qn4U=
I0622 01:37:49.141] @@ -922,8 +923,8 @@
I0622 01:37:49.142]  gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
I0622 01:37:49.142]  gopkg.in/check.v1 v1.0.0-20141024133853-64131543e789/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
I0622 01:37:49.142]  gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
I0622 01:37:49.142] -gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 h1:YR8cESwS4TdDjEe65xsg0ogRM/Nc3DYOhEAlW+xobZo=
I0622 01:37:49.142]  gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
I0622 01:37:49.142] +gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f h1:BLraFXnmrev5lT+xlilqcH8XK9/i0At2xKjWk4p6zsU=
I0622 01:37:49.142]  gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
I0622 01:37:49.143]  gopkg.in/cheggaaa/pb.v1 v1.0.25/go.mod h1:V/YB90LKu/1FcN3WVnfiiE5oMCibMjukxqG/qStrOgw=
I0622 01:37:49.143]  gopkg.in/errgo.v2 v2.1.0/go.mod h1:hNsd1EY+bozCKY1Ytp96fpM3vjJbqLJn88ws8XvfDNI=
W0622 01:37:49.243] make: *** [Makefile:95: presubmit] Error 1
```

I think you need to run `make tidy` to update the go.mod files. However, we still have the issue with depending on staging module (https://github.com/google/cadvisor/pull/2872#discussion_r655783226) ...

container/containerd/client.go
yyrdl: If we just consider this feature in cadvisor independently , we can use `snapshots.Usage`, but for the sake of performance ,we should use `cri.ContainerStats` instead.  there is a picture which shows the relationshipt between cadvisor ,kubelet  and containerd in this situation :

diskmetrics

If we use `snapshots.Usage` in cadvisor, it will cause unnecessary performance loss.

yyrdl: I didn't find an another equivalent  library , can I copy that stuff over to cadvisor ？

bobbypage: I chatted with @dims offline and confirmed that importing `k8s.io/cri-api/` will result in circular dependency, so unfortunately we simply won't be able to do that.  

Do you have any idea for alternative idea? I understand `snapshots.Usage` is not cached, is there some other idea how to mitigate that? Can we maybe cache it internally? Or use/introduce some other containerd API?

yyrdl: Maybe we could use `snapshots.Usage`  and cache it internally.  And at the same time , modify kubelet , do not collect diskUsageMetrics by cadvisor  when it's containerd, and use the data returned by [runtimeService.ListContainerStats](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cri_stats_provider.go#L148).

But I don't known if it's ok . And I have sent an email to @Random-Liu  (who is familiar with this)  for more suggestion.

bobbypage: That sounds reasonable to me. Is the idea then that cAdvisor standalone would use `snapshots.Usage` but when used inside kubelet, we would disable using `snapshots.Usage` in cadvisor and instead rely on `runtimeService.ListContainerStats`. Feel free to cc me on the email: my mail is porterdavid@google.com

Random-Liu: > I chatted with @dims offline and confirmed that importing k8s.io/cri-api/ will result in circular dependency, so unfortunately we simply won't be able to do that.

Hm, I don't quite understand. The import dependency is at package level right? As long as `k8s.io/cri-api` doesn't directly/indirectly depend any part cadvisor repo, it seems safe to me.

The only issue I can imagine is k8s and cadvisor depends on different versions of CRI, in that case, it seems to be resolvable.

We already have this kind of dependency today IIRC: both cadvisor and kubelet depend on the same docker client library, and containerd libraries. CRI is just yet another dependency like this, and an easier to manage one because kubelet, cadvisor and CRI are all heavily maintained by K8s maintainers.

dims: @Random-Liu please have a chat with @liggitt and fill us back in here.

liggitt: > The import dependency is at package level not repo level right?

no, it's at the module level.

> As long as k8s.io/cri-api doesn't directly/indirectly depend on any part of cadvisor repo, it seems safe to me.

k8s.io/cri-api is published from k8s.io/kubernetes (canonical location at https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/cri-api)

If a dependency vendored into k8s.io/kubernetes depends on something published out of k8s.io/kubernetes, then it creates gridlock, where a change in k8s.io/cri-api could require a new level of cadvisor, which cannot be made simultaneously.

Random-Liu: Hm, but the `github.com/containerd/containerd` module depends on `k8s.io/cri-api`, but also being vendored into `k8s.io/kubernetes`. How does that work?

liggitt: The packages we vendor from github.com/containerd/containerd don't include the k8s.io/cri-api reference, and the version k/k currently vendors (v1.4.4) doesn't have a go module, (https://github.com/containerd/containerd/blob/v1.4.4/go.mod) so our vendor scripts are unaware of the circular dependency. If they were, they would forbid it.

liggitt: The github.com/containerd/containerd circular dependency back on a k/k staging module is a timebomb that will keep us from picking up newer versions :-/

Random-Liu: I believe what containerd does is correct. We should make sure our CRI API is safe to import for other projects in the community. :/

Is it possible to define a dedicated module for `k8s.io/cri-api/pkg/apis/runtime`? It is just api definition without much extra dependency.

yyrdl: Using CRI method is the better way ,but at current circular dependency is also a  real problem...  How about using `snapshots.Usage` for the time being ，and when `k8s.io/cri-api` is safe to import ,then switch to CRI?

bobbypage: > How about using snapshots.Usage for the time being ，and when k8s.io/cri-api is safe to import ,then switch to CRI?

This seems like the best (only) option for now so I agree -- we should do that. However, we should be careful to cache the `snapshots.Usage` info internally in cAdvisor since it's not cached as you mentioned earlier.

yyrdl: @bobbypage I have changed it to `snapshots.Usage`

container/common/fsHandler.go
```diff
@@ -37,15 +36,21 @@ type FsUsage struct {
...
 	sync.RWMutex
-	lastUpdate time.Time
-	usage      FsUsage
-	period     time.Duration
-	minPeriod  time.Duration
-	rootfs     string
-	extraDir   string
-	fsInfo     fs.FsInfo
+	lastUpdate    time.Time
+	usage         FsUsage
```
qiutongs, line 50: Is this`usage` the cache?

container/containerd/client.go
bobbypage: Thanks @yyrdl for the update! 

One thing I would like to understand before proceeding is the best path forward to integrate these changes back into kubelet. As I understand with this PR merged and cAdvisor vendored back into kubelet the story will be the following:

1. If using CRI runtime, kubelet will use `ListContainerStats` CRI method which containerd will use containerd's internal [`SnapshotSyncer`](https://github.com/containerd/containerd/blob/559f9a5b90073dc12005d1a611c97007db0be131/pkg/cri/server/service.go#L196-L200) which will internally call `Usage` on the snapshot [periodically](https://github.com/containerd/containerd/blob/559f9a5b90073dc12005d1a611c97007db0be131/pkg/cri/server/snapshots.go#L101) and cache the results. It looks like by default these stats are refreshed every [10 seconds] by containerd (https://github.com/containerd/containerd/blob/559f9a5b90073dc12005d1a611c97007db0be131/pkg/cri/config/config_unix.go#L95)
2. With this PR included, cAdvisor (vendored into kubelet) will also call `snapshots.Usage` (which is an expensive call). That value is not cached / shared with the CRI `SnapshotSyncer` method. How often will this happen (is this every house keeping period)?

as background, looks like `snapshot.Usage` implementation at least for [overlay snapshotter](https://github.com/containerd/containerd/blob/main/snapshots/overlay/overlay.go#L174) requires [walking the filesystem](https://github.com/containerd/containerd/blob/559f9a5b90073dc12005d1a611c97007db0be131/vendor/github.com/containerd/continuity/fs/du_unix.go#L51-L87) counting inodes and size by calling `stat` repeatedly  

As you mentioned previously, with this approach it sounds like we will collect the `snapshots.Usage` data twice from kubelet. So it sounds like we have the following options:

1. Collect the data twice (con: has potential perf impact -- i'm not sure how much -- this will be good to understand. Do you have any idea?)
2. Disable cAdvisor in kubelet from collecting container disk usage metrics. (pro: we avoid perf impact, con: kubelet /metrics/cadvisor endpoint will not report these metrics; they will only be available for users who run cAdvisor standalone/daemonset).

Do you have any thoughts on options there and perf impact? I think we should have clear idea how we integrate these changes back into kubelet before proceeding.

Random-Liu: Agree, `snapshot.Usage` is a heavy operation.

We should probably make sure it optional or query based. In the long run, this metrics should be covered by the metrics endpoint proposed in https://github.com/kubernetes/enhancements/issues/2371 based on my understanding.

yyrdl: * yes,cadvisor will call `snapshot.Usage` every house keeping period
* the perf impact depends on how many files  there. if there are too many files  ,the cost will be expensive

The metrics that the cadvisor will collect are configurable, but at current , we share a common configuration in kubelet,  how about  using  different configuration  according to the runtime ,because the situation is different between container runtimes .

And for metrics that is already returned by CRI , there is no need to collect it again in cadvisor.

container/common/fsHandler.go
yyrdl: yes

container/containerd/client.go
bobbypage: Makes sense! To confirm, the idea you're proposing is basically that kubelet will disable calling `snapshot.Usage` (i.e. option #2 in https://github.com/google/cadvisor/pull/2872#discussion_r675855527

Is there a way we can configure cAdvisor in kubelet to disable this collection? If not, do we need add something to cAdvisor to make this configurable?

yyrdl: No , the metric list declared in [kubelet's server](https://github.com/kubernetes/kubernetes/blob/fffaadc01331cca57cafa0fc066a2a3eec23acb8/pkg/kubelet/server/server.go#L357-L366) is the metrics that will export to user.  The source of the data is in `pkg/kubelet/stats`,  what we need to change is the data source. 

And cadvisor is configurable :

* [The place where we configure cadvisor in kubelet](https://github.com/kubernetes/kubernetes/blob/fffaadc01331cca57cafa0fc066a2a3eec23acb8/pkg/kubelet/cadvisor/cadvisor_linux.go#L85)
* Where cadvisor acts according to the configuration .
  * [for metrics from cgroup](https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L72)
  *  [for diskUsage metrics](https://github.com/google/cadvisor/blob/master/container/docker/handler.go#L241)

bobbypage: Got it. So does it mean that after this change is merged we will want to basically remove `container.DiskUsageMetrics` for containerd [1] in kubelet?

It would be good to understand what specific changes/patch we will need to make in kubelet as part of including this change. Would you be able to maybe draft the patch you're proposing? That way it will be easier to understand what we need to do to integrate these changes into kubelet.  

[1] https://github.com/kubernetes/kubernetes/blob/fffaadc01331cca57cafa0fc066a2a3eec23acb8/pkg/kubelet/cadvisor/cadvisor_linux.go#L85-L93

yyrdl: My pleasure ,I will do it later:)

bobbypage: Thanks @yyrdl!

Would you be able to draft a PR of the kubelet changes required prior to merging this PR? 

I would just like to ensure we know exactly what changes are required to kubelet first, prior to merging this PR. That way it's clear if we need to make any other cAdvisor/kubelet changes to fully integrate these changes since most important consumer of cAdvisor here will be kubelet.

yyrdl: It takes time to redesign the StatsProvider in kubelet , I am a little busy , but I will do it as soon as possible.

qiutongs: @yyrdl FYI, I work with @bobbypage and we are solving another cadvisor-containerd issue which needs CRI as well. To work around the circular dependency issue, current plan is integrating cadvisor with CRI in a separate branch. https://github.com/google/cadvisor/pull/2923

vsxen:
ping...

container/containerd/client.go
yyrdl: @qiutongs  good job!

qiutongs: @yyrdl I sent you a message on the slack and we can discuss more.

container/containerd/client.go
qiutongs: Shouldn't cadvisor take the log size of `/var/log/pods` in account when counting `TotalusageBytes`?

qiutongs:
@yyrdl On top of your PR, the fix was merged in `containerd-cri` branch. And https://github.com/google/cadvisor/pull/2956 will merge the fix to `master`.

Feel free to close this one.

qingguee:
Hi @yyrdl  and @qiutongs ,
Has the fix merged in v0.43.0? I deployed v0.43.0, but container_fs_use_bytes still not work on containerd POD.

vsxen:
Do you mean sanbox container（pause:3.2）？

qingguee:
@vsxen  No. I was cloned cadvisor code, ,change to tag v0.43.0, then docker build the cadvisor image. I deployed that image as daemonset on our K8sto check if the container_fs_use_bytes metrics works with containerd. 
The result I see, it still not works. But I thought this PR will fix this issue on containerd.

vsxen:
what is you containerd version?
I use containerd 1.4.8 & cadvisor 0.43.0, it works.

qingguee:
it's containerd://1.4.4
Maybe this is the problem. Thanks a lot for the info.

yyrdl:
Does your prometheus pull metrics  from the casvisor you deployed or still from kubelet ?

qingguee:
Hi @yyrdl , I have both. My v0.43.0 cadvisor was deployed as daemonset, and I use endpoints to collect those metrics. I use {job="cadvisor"} to filter out the kubelet advisor data. 
Anyway, I will check with my IT team, see if we can update to containerd 1.4.8, then I will double check. This may take time, but when it ready, I will update.

qingguee:
Hi @yyrdl 
After upgrade our cluster with containerd://1.4.12, I still can't see the fs usage for each container, I can only see some usage for node device.
```
cadvisor:v0.43.0

$ kubectl  get node -o wide
NAME          STATUS   ROLES       AGE   VERSION   KERNEL-VERSION      CONTAINER-RUNTIME
node-demo_0   Ready    edge,node   9d    v1.22.4   5.3.18-57-default   containerd://1.4.12
node-demo_1   Ready    edge,node   9d    v1.22.4   5.3.18-57-default   containerd://1.4.12
node-demo_2   Ready    node        9d    v1.22.4   5.3.18-57-default   containerd://1.4.12
node-demo_3   Ready    node        9d    v1.22.4   5.3.18-57-default   containerd://1.4.12
```

I deployed cadvisor as daemonset, my Prometheus collect metrics from those endpoint/POD. 

```
apiVersion: apps/v1 # for Kubernetes versions before 1.9.0 use apps/v1beta2
kind: DaemonSet
metadata:
  name: cadvisor
  namespace: cadvisor
spec:
  template:
    spec:
      containers:
      - name: cadvisor
        args:
          - --housekeeping_interval=10s                           # kubernetes default args
          - --enable_metrics=disk,diskIO,cpu                                   # only show stats for docker containers
          - --store_container_labels=true
          # - --whitelisted_container_labels=io.kubernetes.container.name, io.kubernetes.pod.name,io.kubernetes.pod.namespace
          # - --enable_metrics=app,cpu,disk,diskIO,memory,network,process
```

```
# HELP container_fs_usage_bytes Number of bytes that are consumed by the container on this filesystem.
# TYPE container_fs_usage_bytes gauge
container_fs_usage_bytes{container_label_adpbrlabelkey="",container_label_app="",container_label_app_kubernetes_io_component="",container_label_app_kubernetes_io_instance="",container_label_app_kubernetes_io_managed_by="",container_label_app_kubernetes_io_name="",container_label_app_kubernetes_io_version="",container_label_controller_revision_hash="",container_label_controller_uid="",container_label_demo_cm_mediator_access="",container_label_demo_data_message_bus_kf_client_access="",container_label_demo_sec_key_management_access="",container_label_helm_sh_chart="",container_label_io_cri_containerd_kind="",container_label_io_kubernetes_container_name="",container_label_io_kubernetes_pod_name="",container_label_io_kubernetes_pod_namespace="",container_label_io_kubernetes_pod_uid="",container_label_job_name="",container_label_k8s_app="",container_label_name="",container_label_pod_template_generation="",container_label_pod_template_hash="",container_label_statefulset_kubernetes_io_pod_name="",device="/dev",id="/",image="",name=""} 0 1646271995262
container_fs_usage_bytes{container_label_adpbrlabelkey="",container_label_app="",container_label_app_kubernetes_io_component="",container_label_app_kubernetes_io_instance="",container_label_app_kubernetes_io_managed_by="",container_label_app_kubernetes_io_name="",container_label_app_kubernetes_io_version="",container_label_controller_revision_hash="",container_label_controller_uid="",container_label_demo_cm_mediator_access="",container_label_demo_data_message_bus_kf_client_access="",container_label_demo_sec_key_management_access="",container_label_helm_sh_chart="",container_label_io_cri_containerd_kind="",container_label_io_kubernetes_container_name="",container_label_io_kubernetes_pod_name="",container_label_io_kubernetes_pod_namespace="",container_label_io_kubernetes_pod_uid="",container_label_job_name="",container_label_k8s_app="",container_label_name="",container_label_pod_template_generation="",container_label_pod_template_hash="",container_label_statefulset_kubernetes_io_pod_name="",device="/dev/rbd0",id="/",image="",name=""} 4.2053632e+07 1646271995262
container_fs_usage_bytes{container_label_adpbrlabelkey="",container_label_app="",container_label_app_kubernetes_io_component="",container_label_app_kubernetes_io_instance="",container_label_app_kubernetes_io_managed_by="",container_label_app_kubernetes_io_name="",container_label_app_kubernetes_io_version="",container_label_controller_revision_hash="",container_label_controller_uid="",container_label_demo_cm_mediator_access="",container_label_demo_data_message_bus_kf_client_access="",container_label_demo_sec_key_management_access="",container_label_helm_sh_chart="",container_label_io_cri_containerd_kind="",container_label_io_kubernetes_container_name="",container_label_io_kubernetes_pod_name="",container_label_io_kubernetes_pod_namespace="",container_label_io_kubernetes_pod_uid="",container_label_job_name="",container_label_k8s_app="",container_label_name="",container_label_pod_template_generation="",container_label_pod_template_hash="",container_label_statefulset_kubernetes_io_pod_name="",device="/dev/shm",id="/",image="",name=""} 0 1646271995262
```


So, is this issue fixed in 0.43.0? Or my configuration has issue?
This https://github.com/google/cadvisor/issues/2785#issuecomment-1020088355 ,  mentioned that it's not working. But as there is a lot different issue reporting this topic, it's quite confuse for me.  Could you help to provide the latest info?

yyrdl:
@qingguee  cadvisor is integrated in kubelet ,at current ,the fix of this PR is not merged for some package dependency problem , you can copy the code changed in this pr , and build  a custom kubelet  or cadvisor to solve your problem.

qingguee:
@yyrdl Thank you so much for the info. 
So, I get more clear for me, we need both kubelet fix and cadvisor fix.  But I don't think we could update our kubelet as we rent the cluster. 

Could you help to provide the link for package dependency problem, I'd like to follow up. When that solved, we will do upgrade accordingly.

yyrdl:
@qingguee  https://github.com/google/cadvisor/pull/2872#discussion_r655783226

PR #2885, opened 2021-06-07
Fix typo in test function name

eero-t:
2-letter typo fix is too small change to be copyrightable, so this "contribution" should not require CLA.

### DIFF
--- a/info/v2/conversion_test.go
+++ b/info/v2/conversion_test.go
@@ -30,7 +30,7 @@ var (
 	envs      = map[string]string{"foo": "bar"}
 )
 
-func TestContanierSpecFromV1(t *testing.T) {
+func TestContainerSpecFromV1(t *testing.T) {
 	v1Spec := v1.ContainerSpec{
 		CreationTime: timestamp,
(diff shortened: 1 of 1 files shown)


bobbypage:
I believe any change requires CLA, can you please sign it?

eero-t:
> I believe any change requires CLA, can you please sign it?

I'll have to ask our legal. It will likely take weeks, and I'm not sure what the outcome will be.

But it seems weird that you ask CLA for something that it does not apply to, as my change is too small to be copyrightable, and it is definitely something on which there could not be a patent for (another thing your CLA mentions).  :-)

Could you not make sure that your contributions rules are more reasonable and take into account whether contribution is something to which CLA can actually apply to?

Ps. I think minimum that can be copyrighted is about two words (trademark would be used for one word).  If you know of a jurisdiction where it is possible to have a copyright on swapping 2 letters, please let me know!

eero-t:
@googlebot I signed it!

Creatone (approved):
LGTM

PR #2889, opened 2021-06-10
Use "kubectl kustomize" instead of "kustomize build"

eero-t:
There is not anymore need to use a separate "kustomize" tool, one can use "kubectl".

It has supported "kustomize" command since v1.14 (released in March 2019).


Creatone (approved):
LGTM

PR #2898, opened 2021-06-24
Sort metric names in options for readability

eero-t:
List of metric names being generated automatically also avoids it accidentally getting out of date.

PS. Eventually I would like metrics to be sorted in all the lists based on their metric names for coder convenience too. It's harder to verify whether a metric is missing from some place in the cAdvisor code when metrics are not listed in same order, and they need to be added to so places.


Creatone (approved):
LGTM

PR #2902, opened 2021-06-29
Add 'enable_metrics' option

eero-t:
Looking at the cAdvisor issues, it's nowadays often used to get just some specific metrics [1].  Doing that with 'disable_metrics' option is bothersome because there's such a large number of options one needs to disable, and that set can grow on cAdvisor updates.

Doing that by specifying only metrics one wants with 'enable_metrics' is easier & hopefully future-proof.

Btw. Regarding toIncludedMetrics() function, I think it would be clearer just to inline container.AllMetrics.Difference() call instead of having that wrapper function for it.  Any comments on that, and whether such change should be another PR or part of this one?

[1] Eventually it would be nice to be able to disable all cAdvisor metrics, not just the ones on ignoreWhitelist, but that belongs definitely to another PR.

### DIFF
--- a/cmd/cadvisor.go
+++ b/cmd/cadvisor.go
@@ -95,6 +95,10 @@ var (
 		container.CPUSetMetrics:                  struct{}{},
 	}}
 
+	// Metrics to be enabled on top of metrics not in ignoreWhitelist.
+	// Used only if non-empty.
+	enableMetrics metricSetValue = metricSetValue{container.MetricSet{}}
+
 	// List of metrics that can be ignored.
 	ignoreWhitelist = container.MetricSet{
 		container.AcceleratorUsageMetrics:        struct{}{},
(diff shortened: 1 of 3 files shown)


Creatone:
https://github.com/eero-t/cadvisor/blob/8a49db247604926006ff3b841d5f9097698ad342/cmd/cadvisor.go#L140-L153

When set `enable_metrics`, you check the ignore metric list, which is wrong.

eero-t:
> When setting `enable_metrics` flag, you check the ignore metric list, which is wrong.

Nope, ignoreWhitelist is used as the list of metrics that user is shown, both for disabling & enabling, so obviously both disable & enable arguments need to be validated against that same list.  Use `-v 1` to verify what gets actually enabled as result.

(In later https://github.com/eero-t/cadvisor/commit/58e2d16326634cd3b18750131463d91168717126 change, for which I do not have PR yet, ignoreWhitelist is removed so that both enable & disable use AllMetrics instead.  Along with the removal of confusingly named toIncludedMetrics(), that makes the resulting logic clearer I think.)

PR #2904, opened 2021-07-02
prevent tcp metrics from being dropped when network is disabled

Serendo:
The network metrics should not be dropped when network metric is disabled. As there are multiple metric set besides network under ContainerStats.Network, like tcp advtcp udp. And I think maybe this include check function should be moved to the prometheus metric side since these metrics are already collected.

### DIFF
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -344,7 +344,7 @@ func (h *dockerContainerHandler) ContainerReference() (info.ContainerReference, 
 }
 
 func (h *dockerContainerHandler) needNet() bool {
-	if h.includedMetrics.Has(container.NetworkUsageMetrics) {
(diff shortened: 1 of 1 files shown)


Creatone:
`cri-o` and `containerd` handlers have the same issue.

Creatone:
What do you think about moving metric check to a separate function in `container/common/helpers.go`?

```go
// HasAnyNetworkMetrics check if there are network related metrics in MetricSet.
func HasAnyNetworkMetrics(metrics container.MetricSet) bool {
	return metrics.Has(container.NetworkUsageMetrics) ||
		metrics.Has(container.NetworkTcpUsageMetrics) ||
		metrics.Has(container.NetworkUdpUsageMetrics) ||
		metrics.Has(container.NetworkAdvancedTcpUsageMetrics)
	}
```

Using that will look like:
```go
func (h *dockerContainerHandler) needNet() bool {
	if common.HasAnyNetworkMetrics(h.includedMetrics) {
		return !h.networkMode.IsContainer()
	}
	return false
}
```

PR #2908, opened 2021-07-12
Depend only on containerd API and avoid vendoring unnecessary things from containerd

dims:
When cadvisor vendors in things from containerd, there is a dependency loop between cadvisor<->containerd<->kubernetes as containerd vendors in cri-api. So we have been unable to move to newer versions of containerd that use go modules. To help with this effort in 1.6, containerd folks will publish `containerd/containerd/api` as a separate go module. To prep for this, we can cleanup our usage of containerd now to depend only on the code under `containerd/containerd/api` by borrowing some code from containerd repository that essentially acts as a containerd client. 

When containerd 1.6 goes GA, we can switch over to the new `containerd/api` go module. Kubernetes also needs https://github.com/moby/moby/pull/42624 from moby/moby to completely eliminate this dependency loop unless/until we get rid of dockershim (WIP for dockershim removal is here - https://github.com/kubernetes/kubernetes/pull/97252). Either way we are in track to get rid of this dependency loop by kubernetes 1.24 🤞🏾 🙏🏾 

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -24,10 +24,10 @@ import (
 	containersapi "github.com/containerd/containerd/api/services/containers/v1"
 	tasksapi "github.com/containerd/containerd/api/services/tasks/v1"
 	versionapi "github.com/containerd/containerd/api/services/version/v1"
-	"github.com/containerd/containerd/containers"
-	"github.com/containerd/containerd/errdefs"
-	"github.com/containerd/containerd/pkg/dialer"
 	ptypes "github.com/gogo/protobuf/types"
+	"github.com/google/cadvisor/container/containerd/containers"
+	"github.com/google/cadvisor/container/containerd/errdefs"
+	"github.com/google/cadvisor/container/containerd/pkg/dialer"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
 )
--- a/container/containerd/client_test.go
+++ b/container/containerd/client_test.go
@@ -18,7 +18,7 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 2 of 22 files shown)


dims:
```
[dims@dims-a02 10:16] ~/go/src/github.com/google/cadvisor ⟩ tree -d vendor/github.com/containerd/
vendor/github.com/containerd/
├── console
├── containerd
│   └── api
│       ├── services
│       │   ├── containers
│       │   │   └── v1
│       │   ├── tasks
│       │   │   └── v1
│       │   └── version
│       │       └── v1
│       └── types
│           └── task
├── ttrpc
└── typeurl
```

eero-t:
I'm not dev, but I think description is misleading and it would be good to split this to several commits;  e.g. adding the containerd code to cadvisor, another changing usage of external containerd code to new internal copy, adding tests and updating go.mod/sum.

dims:
@eero-t thanks for the feedback, as noted in the title, it's a Work In Progress (WIP) PR.

Creatone:
You can simply change this to draft PR :)

dims:
you are right @Creatone. thanks for the tip.

dims:
github.com/containerd/containerd/api is now available ( https://github.com/containerd/containerd/blob/main/api/go.mod )

Now we wait for https://github.com/moby/moby/pull/42624 and https://github.com/moby/moby/pull/42623

bobbypage:
Discussed this with @dims offline. Long term ideally we can not copy these packages into cAdvisor and they will be an external module published by containerd we can consume. Until we get there this seems like the best option to remove k/k dependency loop.

bobbypage (approved):

PR #2909, opened 2021-07-13
fix when rootfs is empty, TotalUsageBytes will add extraUsage.Bytes on every update, lead to TotalUsageBytes keep increasing

Dingshujie:
fix when rootfs is empty, TotalUsageBytes will add extraUsage.Bytes on every update, lead to TotalUsageBytes keep increasing. 
this may cause kubelet to evict pods

### DIFF
--- a/container/common/fsHandler.go
+++ b/container/common/fsHandler.go
@@ -96,7 +96,12 @@ func (fh *realFsHandler) update() error {
 		fh.usage.TotalUsageBytes = rootUsage.Bytes
 	}
 	if fh.extraDir != "" && extraErr == nil {
-		fh.usage.TotalUsageBytes += extraUsage.Bytes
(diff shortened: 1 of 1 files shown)


bobbypage:
Thanks @Dingshujie. Is there a corresponding k/k issue for this? In what situation can rootfs be empty?

Dingshujie:
> Thanks @Dingshujie. Is there a corresponding k/k issue for this? Just to understand, in what situation can rootfs be empty?

@bobbypage 
docker use devicemapper or zfs as its storage driver,  rootfs can be empty.

there isn't has a corresponding k/k issue for this.

bobbypage:
Got it, thanks for the explanation.

bobbypage (approved):

bobbypage:
Let's wait for CI to get green...

PR #2910, opened 2021-07-15
Use AllMetrics for enable & disable lists instead of ignoreWhitelist

eero-t:
Use of AllMetrics instead of ignoreWhitelist, adds CpuUsageMetrics ("cpu") & AppMetrics ("app"), which are already checked by prometheus export code, to metrics that can be disabled/enabled from command line.

That also makes metricSetValue type redundant, as it can now be replaced with MetricSet use.  Removing it requires moving String() & Set() methods, used for options texts and validating given metrics options, to factory.go.

toIncludedMetrics() one-line wrapper is also removed, as inlining container.AllMetrics.Difference() call makes the code intent more explicit, and is about same amount of code.

### DIFF
--- a/container/factory.go
+++ b/container/factory.go
@@ -144,16 +144,6 @@ func (ms MetricSet) Difference(ms1 MetricSet) MetricSet {
 	return result
 }
 
-func (ms MetricSet) Append(ms1 MetricSet) MetricSet {
-	result := ms
-	for kind := range ms1 {
-		if !ms.Has(kind) {
-			result.Add(kind)
-		}
-	}
-	return result
-}
-
 // All registered auth provider plugins.
 var pluginsLock sync.Mutex
 var plugins = make(map[string]Plugin)
--- a/container/factory.go
+++ b/container/factory.go
@@ -16,6 +16,8 @@ package container
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 	"sync"
 
 	"github.com/google/cadvisor/fs"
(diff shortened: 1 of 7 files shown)


container/factory.go
```diff
@@ -108,6 +110,30 @@ func (ms MetricSet) Add(mk MetricKind) {
...
+	}
+	for _, metric := range strings.Split(value, ",") {
+		if AllMetrics.Has(MetricKind(metric)) {
+			(*ms).Add(MetricKind(metric))
+		} else {
+			return fmt.Errorf("unsupported metric %q specified", metric)
+		}
+	}
+	return nil
+}
```
iwankgb, line 135: Can you add a comment to `MetricSet` and state that the struct is not thread safe? And make sure that the struct is not accessed from multiple goroutines in cAdvisor?

eero-t:
Changes:
- Split toInclude() removal to a separate commit as it's unrelated change touching same lines of code
- Added commit clarifying MetricSet thread-safety (see commit message!)
- Related to that, added commit removing unused MetricSet.Append() method
- Added commit updating the documentation => this PR is now rebased on other PR that fixes documentation for an earlier merged PR

container/factory.go
eero-t: I added couple of commits clarifying & documenting MericSet thread-safety (removing unused Append() method, changing Add() method to private one and documenting Set() method usage).

Creatone (approved):
LGTM

PR #2918, opened 2021-08-09
cadvisor housekeeping to be static

sashankreddya:
Cadvisor adds jitter for each housekeeping(HK) interval, making the HK interval to be dynamic each time and can range anywhere between [(HK , 2HK)](https://github.com/google/cadvisor/blob/master/manager/container.go#L505). Even though the prometheus scraper can read the collection timestamp and do the delta, we have observed that our internal metrics system still needs to align these datapoints into the nearest bucket. As a result this introduced a 10-20% of variance(SPIKES) to the actual usage metrics making the data to be unreliable if we need to analyze at higher resolution (< 5 mins). 

After reading through original reason for introduction of [jitter](https://github.com/google/cadvisor/issues/1053), we could still achieve the same by distributing/drifting the HouseKeeping kick off for each container to be random for the first time(by adding jitter only for the first time HK) and making the subsequent HouseKeeping to be static (only HK, with no jitter). 
We have seen in our results that this still aligns with the expectation of solving [1053](https://github.com/google/cadvisor/issues/1053) as well as ensuring the datapoints are reliable for the usecase mentioned above.

Let us know your thoughts.

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -74,6 +74,7 @@ type containerData struct {
 	housekeepingInterval     time.Duration
 	maxHousekeepingInterval  time.Duration
 	allowDynamicHousekeeping bool
+        firstHousekeeping        bool
 	infoLastUpdatedTime      time.Time
 	statsLastUpdatedTime     time.Time
 	lastErrorTime            time.Time
@@ -111,9 +112,6 @@ type containerData struct {
 // to allow clients to avoid converging on periodic behavior.  If maxFactor is 0.0, a
 // suggested default value will be chosen.
 func jitter(duration time.Duration, maxFactor float64) time.Duration {
-	if maxFactor <= 0.0 {
-		maxFactor = 1.0
-	}
 	wait := duration + time.Duration(rand.Float64()*maxFactor*float64(duration))
 	return wait
 }
@@ -475,6 +473,7 @@ func newContainerData(containerName string, memoryCache *memory.InMemoryCache, h
 		klog.V(5).Infof("Failed to create summary reader for %q: %v", ref.Name, err)
 	}
 
+        cont.firstHousekeeping = true
 	return cont, nil
 }
 
@@ -501,8 +500,15 @@ func (cd *containerData) nextHousekeepingInterval() time.Duration {
 			}
 		}
 	}
+        // Drift the containers randomly so that the start time for each container is random. However make sure each container is then
+        // collected for the configured houseKeeping interval.
+        maxJitterFactor := 0.0
+        if cd.firstHousekeeping {
+            maxJitterFactor = 1.0
+            cd.firstHousekeeping = false
+        }
 
-	return jitter(cd.housekeepingInterval, 1.0)
+	return jitter(cd.housekeepingInterval, maxJitterFactor)
 }
 
 // TODO(vmarmol): Implement stats collecting as a custom collector.


sashankreddya:
> Cadvisor adds jitter for each housekeeping(HK) interval, making the HK interval to be dynamic each time and can range anywhere between [(HK , 2HK)](https://github.com/google/cadvisor/blob/master/manager/container.go#L505). Even though the prometheus scraper can read the collection timestamp and do the delta, we have observed that our internal metrics system still needs to align these datapoints into the nearest bucket. As a result this introduced a 10-20% of variance(SPIKES) to the actual usage metrics making the data to be unreliable if we need to analyze at higher resolution (< 5 mins).
> 
> After reading through original reason for introduction of [jitter](https://github.com/google/cadvisor/issues/1053), we could still achieve the same by distributing/drifting the HouseKeeping kick off for each container to be random for the first time(by adding jitter only for the first time HK) and making the subsequent HouseKeeping to be static (only HK, with no jitter).
> We have seen in our results that this still aligns with the expectation of solving [1053](https://github.com/google/cadvisor/issues/1053) as well as ensuring the datapoints are reliable for the usecase mentioned above.
> 
> Let us know your thoughts.
> 
> @Creatone



> Cadvisor adds jitter for each housekeeping(HK) interval, making the HK interval to be dynamic each time and can range anywhere between [(HK , 2HK)](https://github.com/google/cadvisor/blob/master/manager/container.go#L505). Even though the prometheus scraper can read the collection timestamp and do the delta, we have observed that our internal metrics system still needs to align these datapoints into the nearest bucket. As a result this introduced a 10-20% of variance(SPIKES) to the actual usage metrics making the data to be unreliable if we need to analyze at higher resolution (< 5 mins).
> 
> After reading through original reason for introduction of [jitter](https://github.com/google/cadvisor/issues/1053), we could still achieve the same by distributing/drifting the HouseKeeping kick off for each container to be random for the first time(by adding jitter only for the first time HK) and making the subsequent HouseKeeping to be static (only HK, with no jitter).
> We have seen in our results that this still aligns with the expectation of solving [1053](https://github.com/google/cadvisor/issues/1053) as well as ensuring the datapoints are reliable for the usecase mentioned above.
> 
> Let us know your thoughts.
> 
> @Creatone

@bobbypage could you take a look at this one.

bobbypage:
Hi @sashankreddya

I don't have the full context on introduction of jitter, but reading https://github.com/google/cadvisor/issues/1053 it seems like the goal there was to ensure that we spread house keeping for each container for every single house keeping attempt. 

As you mentioned, the change here will:

> and making the subsequent HouseKeeping to be static (only HK, with no jitter).

so as I understand, the result is that with this change, house keeping is static for 2nd+ house keeping for each container. This doesn't seem to be desirable, because my understanding is we want to spread out the house keeping intervals.

sashankreddya:
> Hi @sashankreddya
> 
> I don't have the full context on introduction of jitter, but reading #1053 it seems like the goal there was to ensure that we spread house keeping for each container for every single house keeping attempt.
> 
> As you mentioned, the change here will:
> 
> > and making the subsequent HouseKeeping to be static (only HK, with no jitter).
> 
> so as I understand, the result is that with this change, house keeping is static for 2nd+ house keeping for each container. This doesn't seem to be desirable, because my understanding is we want to spread out the house keeping intervals.

Hi @bobbypage , the original issue as mentioned in [#1053](https://github.com/google/cadvisor/issues/1053) conveys that jitter has been introduced to prevent reading Cgroup stats for all containers at the same time. 
The current PR ensures to achieve the same as well, however the only difference is that randomness is calculated only once for each container so that the start of the collection interval for each container is distributed to solve the problem of collecting all at the same time. We no longer need the subsequent intervals to also be random as the first time calculation will already be preserved.
cc: @dashpole

PR #2919, opened 2021-08-10
Fix the first container alias in containerd handler

qiutongs:
Signed-off-by: Qiutong Song <songqt01@gmail.com>

Unit Tests is Pending...

## Overview

In Prometheus and Statsd, they use the first container alias as the "container name" in the output metrics. However, docker handler and containerd handler have different logics for the alias. 

In docker, it is `k8s_<container-name>_<pod-name>_<namespace>_<pod-uid>_<restart-count>` but it is `<container id>` in containerd.

This PR is to make containerd behave the same way as docker.
- Add an extra flag to support both format for now
- Hardcoded restart count as zero if corresponding label is not present. This needs code change in kubelet.

## Local Test

```
sudo ./cadvisor
```

Query Prometheus endpont: `localhost:8080/metrics`

```
name="k8s_nginx-pod_nginx-pod_default_d91d6c3e-0378-432b-ae35-2af99c034874_0"
```
Similarly
```
sudo ./cadvisor --use_k8s_container_name_format=false
```

Output is 

```
name="c6a1aa99f14d3e57417e145b897e34961145f6b6f14216a176a34bfabbf79086"
```

### DIFF
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -17,6 +17,7 @@ package containerd
 
 import (
 	"encoding/json"
+	"flag"
 	"fmt"
 	"strings"
 	"time"
@@ -31,6 +32,9 @@ import (
 	info "github.com/google/cadvisor/info/v1"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
+
+// TODO: if it is decided that this is the only format, this flag can be removed.
+var useK8sContainerNameFormat = flag.Bool("use_k8s_container_name_format", true, "Use the k8s format: k8s_<container-name>_<pod-name>_<namespace>_<pod-uid>_<restart-count>")
 
 type containerdContainerHandler struct {
 	machineInfoFactory info.MachineInfoFactory
@@ -112,11 +116,22 @@ func newContainerdContainerHandler(
 		rootfs = "/rootfs"
 	}
 
+	aliases := []string{id, name}
+
+	if *useK8sContainerNameFormat && cntr.Labels["io.cri-containerd.kind"] == "container" {
+		restart := cntr.Labels["io.kubernetes.pod.restart"]
+		if restart == "" {
+			restart = "0"
+		}
+		containerNameConcat := "k8s_" + cntr.Labels["io.kubernetes.container.name"] + "_" + cntr.Labels["io.kubernetes.pod.name"] + "_" + cntr.Labels["io.kubernetes.pod.namespace"] + "_" + cntr.Labels["io.kubernetes.pod.uid"] + "_" + restart
+		aliases = []string{containerNameConcat, id, name}
+	}
+
 	containerReference := info.ContainerReference{
 		Id:        id,
 		Name:      name,
 		Namespace: k8sContainerdNamespace,
-		Aliases:   []string{id, name},
+		Aliases:   aliases,
 	}
 
 	libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootfs, int(taskPid), includedMetrics)


qiutongs:
@googlebot I signed it!

container/containerd/handler.go
```diff
@@ -112,11 +116,22 @@ func newContainerdContainerHandler(
...
 	}
 
+	aliases := []string{id, name}
+
+	if *useK8sContainerNameFormat && cntr.Labels["io.cri-containerd.kind"] == "container" {
+		restart := cntr.Labels["io.kubernetes.pod.restart"]
+		if restart == "" {
+			restart = "0"
+		}
+		containerNameConcat := "k8s_" + cntr.Labels["io.kubernetes.container.name"] + "_" + cntr.Labels["io.kubernetes.pod.name"] + "_" + cntr.Labels["io.kubernetes.pod.namespace"] + "_" + cntr.Labels["io.kubernetes.pod.uid"] + "_" + restart
```
SergeyKanzhelev, line 126: strings.Join looks cleaner like here https://github.com/kubernetes/kubernetes/blob/ab595f37d258b85291bb6f4f18a1649558346bc0/pkg/kubelet/dockershim/naming.go#L60-L80, but I don't think there is a difference

SergeyKanzhelev (approved):
please add tests

container/containerd/handler.go
qiutongs: Sounds good. Will change to strings.join

bobbypage: +1

bobbypage: Please also leave a comment where these labels are defined, i.e. https://github.com/kubernetes/kubernetes/blob/ab595f3/pkg/kubelet/types/labels.go#L24

container/containerd/handler.go
```diff
@@ -112,11 +116,22 @@ func newContainerdContainerHandler(
 		rootfs = "/rootfs"
 	}
 
+	aliases := []string{id, name}
+
+	if *useK8sContainerNameFormat && cntr.Labels["io.cri-containerd.kind"] == "container" {
+		restart := cntr.Labels["io.kubernetes.pod.restart"]
```
bobbypage, line 122: add note that this label is not populated and possibly TODO linking to issue to track this

container/containerd/handler.go
```diff
@@ -32,6 +33,9 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
+// TODO: if it is decided that this is the only format, this flag can be removed.
+var useK8sContainerNameFormat = flag.Bool("use_k8s_container_name_format", true, "Use the k8s format: k8s_<container-name>_<pod-name>_<namespace>_<pod-uid>_<restart-count>")
```
bobbypage, line 37: please update https://github.com/google/cadvisor/blob/master/docs/runtime_options.md with this flag

container/containerd/handler.go
```diff
@@ -112,11 +116,22 @@ func newContainerdContainerHandler(
 		rootfs = "/rootfs"
 	}
 
+	aliases := []string{id, name}
+
+	if *useK8sContainerNameFormat && cntr.Labels["io.cri-containerd.kind"] == "container" {
```
bobbypage, line 121: What about infra sandbox / pause containers?

container/containerd/handler.go
bobbypage: s/Use the k8s format: .../as the name and alias for the container

qiutongs:
I am hesitated about this path. Closed it for now.

PR #2921, opened 2021-08-16
Common flag env metadata whitelist

Colstuwjx:
This is a follow-up PR for #2330, after I proposed that PR, I've seen a contributor submitted PR #2857 in order to introduce containerd whitelist env flag.

Now I merged the containerd and docker env whitelist flags, and just a simple common flag: `env_metadata_whitelist`, the other two are deprecated while this PR merged.

BTW, I'm not sure about the state of the Mesos container, just keep the same interface signature as well :(


cmd/internal/container/mesos/handler_test.go
```diff
@@ -22,7 +22,6 @@ import (
 	containerlibcontainer "github.com/google/cadvisor/container/libcontainer"
 	"github.com/google/cadvisor/fs"
 	info "github.com/google/cadvisor/info/v1"
-	"github.com/mesos/mesos-go/api/v1/lib"
```
Colstuwjx, line 25: I found CI error after removed this line:

```
Error: vet: internal/container/mesos/handler_test.go:31:21: undeclared name: mesos
Error: make: *** [Makefile:71: vet] Error 2
```

But I have no idea about the reason. It seems that the mesos lib didn't exist anymore, see [mesos-go](https://github.com/mesos/mesos-go/tree/master/api).

Colstuwjx:
/cc @dashpole @bobbypage 

It seems no meaningful message to debug :(

```
+ echo 'Deleting /tmp/tmp.DrWtfAH0Ul...'
+ [[ 1001 -ne 0 ]]
+ sudo rm -rf /tmp/tmp.DrWtfAH0Ul
make: *** [Makefile:54: docker-test-integration] Error 143
```

eero-t:
> It seems no meaningful message to debug :(

This is the actual problem:
```
 Waiting for cAdvisor to start ...
!! cAdvisor exited unexpectedly with Exit 0
+ EXIT_VALUE=143
```
As there's no output, this could be early at startup?

Colstuwjx:
> As there's no output, this could be early at startup?

I found an issue that the integration tests are still using the deprecated flag `docker_env_metadata_whitelist`, I've been fixed them.

BTW, the integration tests should output the logs if there are errors, I'll try to propose another PR for fixing it.

bobbypage:
Thanks @Colstuwjx for the PR. Do you mind splitting the fix to integration test out into a separate PR so we can unblock this and other PRs in parallel? Thanks!

bobbypage:
This looks good overall. One thing I'm a bit worried about is removing the existing `--containerd_env_metadata_whitelist` and `--docker_env_metadata_whitelist` since it is a breaking change for users. Perhaps we can keep these flags around and just set them equal to `--env_metadata_whitelist` and leave a note that they are deprecated. Then in a release or two we can fully remove them. What do you think?

Colstuwjx:
> Thanks @Colstuwjx for the PR. Do you mind splitting the fix to integration test out into a separate PR so we can unblock this and other PRs in parallel? Thanks!

What do you mean? The integration test-related changes in this PR just changed the flag `docker_env_metadata_whitelist ` to `env_metadata_whitelist`. I think it's reasonable for including these changes inside this PR, otherwise, it would panic our integration tests as we've seen.

> Perhaps we can keep these flags around and just set them equal to --env_metadata_whitelist and leave a note that they are deprecated. Then in a release or two we can fully remove them.

I've been considered this path, it's ok to me, I'll add these deprecated flags back soon.

Colstuwjx:
Ping @bobbypage I've been added the deprecated flags back! Please take a review :)

container/factory.go
```diff
@@ -27,7 +27,7 @@ import (
 
 type ContainerHandlerFactory interface {
 	// Create a new ContainerHandler using this factory. CanHandleAndAccept() must have returned true.
-	NewContainerHandler(name string, inHostNamespace bool) (c ContainerHandler, err error)
+	NewContainerHandler(name string, metadataEnvs []string, inHostNamespace bool) (c ContainerHandler, err error)
```
bobbypage, line 30: can we rename `metadataEnvs` to `metadataEnvAllowList` or something like that? From the name, it's not clear that we are only going envs in `metadataEnvs`

metrics/prometheus.go
```diff
@@ -1856,6 +1856,9 @@ func BaseContainerLabels(whiteList []string) func(container *info.ContainerInfo)
 				set[ContainerLabelPrefix+k] = v
 			}
 		}
+		for k, v := range container.Spec.Envs {
```
bobbypage, line 1859: Is this change needed? It seems like this is change in behavior of adding env to Prometheus labels. I'm not sure how this change is relevant here?

bobbypage:
Thanks for updating @Colstuwjx. Overall LGTM, minor nit about naming and also question around why we need to change the labels provided in Prometheus. Thanks!

container/factory.go
Colstuwjx: Sure.

metrics/prometheus.go
Colstuwjx: > It seems like this is change in behavior of adding env to Prometheus labels.

Yes. You could get to know the background from [ PR 2330 #issue-comment ](https://github.com/google/cadvisor/pull/2330#issuecomment-545656563). PR #2330 is original to add container whitelisted envs to prom labels, and @dashpole  think it'd better to unify the flags, so I proposed this PR righ now!

If you think it'd better to separate these changes, and make the env changes an individual PR, I'm ok with it :)

bobbypage: Got it, thanks for the explanation.

bobbypage:
Please rebase to fix conflicts and update as per https://github.com/google/cadvisor/pull/2921#discussion_r692644778. Thanks!

Colstuwjx:
Hi @bobbypage 

Now I revised the PR changes according to your suggestions! Please take a look again.
Once this PR gets merged, I will take another PR for adding container envs as labels in the prom metrics.

PR #2920, opened 2021-08-12
add flag for max_housekeeping_interval

rphillips:
This allows Kubelet to pass down the max_housekeeping_interval to cadvisor the same way housekeeping_interval is set. This will be necessary to decrease load on the system in single node environments.

PR #2925, opened 2021-08-20
Refactor max_housekeeping_interval to move in manager package

harche:
As per the discussion https://github.com/google/cadvisor/pull/2920#issuecomment-902406524, this PR refactors `max_housekeeping_interval` to move it from `cmd` package to `manager` package so that along with the command line kubelet can also override it. 

Potential users of this refactor - https://github.com/kubernetes/kubernetes/pull/104339 

I have tested this refactoring with cadvisor command, 
```bash
[harshal@localhost cadvisor]$ ./cadvisor -h 
Usage of ./cadvisor:
</snip>
 -max_housekeeping_interval duration
        Largest interval to allow between container housekeepings (default 1m0s)
</snip>
```
So this isn't breaking existing flag. 

Plus it works with kubelet too,
```bash
[harshal@localhost kubernetes]$ _output/bin/kubelet -h | grep max-house
      --max-housekeeping-interval duration                       Largest interval to allow between container housekeepings (default 1m0s)
[harshal@localhost kubernetes]$ 
```

/cc @bobbypage @rphillips @pxaws

Signed-off-by: Harshal Patil <harpatil@redhat.com>

### DIFF
--- a/cmd/cadvisor.go
+++ b/cmd/cadvisor.go
@@ -26,7 +26,6 @@ import (
 	"sort"
 	"strings"
 	"syscall"
-	"time"
 
 	cadvisorhttp "github.com/google/cadvisor/cmd/internal/http"
 	"github.com/google/cadvisor/container"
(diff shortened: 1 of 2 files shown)


rphillips:
@bobbypage what do you think of this solution? Not a breaking change to Kubelet.

bobbypage:
I like this solution -- I think it's the best path forward. LGTM.

bobbypage (approved):

PR #2930, opened 2021-08-27
feat: add container whitelisted envs as metrics labels.

Colstuwjx:
Signed-off-by: colstuwjx <colstuwjx@gmail.com>

It's a follow-up PR for #2921 and #2330.

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -1836,7 +1836,7 @@ func DefaultContainerLabels(container *info.ContainerInfo) map[string]string {
 }
 
 // BaseContainerLabels returns a ContainerLabelsFunc that exports the container
-// name, first alias, image name as well as white listed label values.
+// name, first alias, image name as well as all its white listed env and label values.
 func BaseContainerLabels(whiteList []string) func(container *info.ContainerInfo) map[string]string {
 	whiteListMap := make(map[string]struct{}, len(whiteList))
 	for _, k := range whiteList {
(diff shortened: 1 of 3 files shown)


metrics/prometheus.go
```diff
@@ -1856,6 +1856,9 @@ func BaseContainerLabels(whiteList []string) func(container *info.ContainerInfo)
 				set[ContainerLabelPrefix+k] = v
 			}
 		}
+		for k, v := range container.Spec.Envs {
+			set[ContainerEnvPrefix+k] = v
+		}
```
iwankgb, line 1861: A test case would be greatly appreciated.

iwankgb (changes requested):

metrics/prometheus.go
Colstuwjx: Hi @iwankgb, after diving into the code, I found that the env is filtered while manage call container handler, e.g. containerd is at [here](https://github.com/google/cadvisor/blob/aa0afbe/container/containerd/handler.go#L146).

As a result, I added test case `TestPrometheusCollectorWithWhiteList` to cover `BaseContainerLabels` test. If there are any suggestions, pls let me know!

Colstuwjx: BTW, I found the handler part already did have some test cases about whitelist envs, e.g. see docker [tests](https://github.com/google/cadvisor/blob/aa0afbee6da04e8d8a214333bf2747f0ae27f7a0/container/docker/handler_test.go#L121).

So... I think it's enough.

bobbypage:
LGTM, thanks.

Issue #2932, opened 2021-08-30
cAdvisor constantly polls cgroup data for metrics that have been disabled


eero-t:
This is continuation on https://github.com/google/cadvisor/issues/2897 issue.  After that, cAdvisor code was change to support enabling/disabling all metrics, including "cpu" & "app" ones in https://github.com/google/cadvisor/pull/2910.

Setup:
*  Ubuntu 20.04 host (= cgroup v1)
*  Few days old git head of runC & cAdvisor (and drm-tip kernel)

Use-case:
*  Start cAdvisor with only one metric enabled: `cadvisor -enable_metrics hugetlb`

Expected output:
* Only the enabled metrics are queried (in this case, "hugetlb")

Actual output:
* cAdvisor constantly queries "cpu" & "pids" cgroups

These are the opened file & dir names from 1 min of tracing:
```
# strace -f -e openat -p $(pidof cadvisor) 2> trace.txt
^C
# grep openat trace.txt | grep -v "openat resumed" | cut -d'"' -f2 | sed 's%^.*/%%' | sort | uniq -c | sort -nr
   2002 cpu.shares
   2002 cpu.cfs_quota_us
   2002 cpu.cfs_period_us
   1976 pids.max
    134 system.slice
...
```

eero-t:
@Creatone does this look as OK fix for this: https://github.com/eero-t/cadvisor/commit/e57ee468522bf9a70705a3819e8d7f4fcb57003e

At least in my own testing it fixed the redundant polling.

Creatone:
@eero-t Could you made PR :)?

eero-t:
@Creatone https://github.com/google/cadvisor/pull/2940

PR #2940, opened 2021-09-16
Disable polling of CPU & PIDs cgroups when they are not needed

eero-t:
https://github.com/google/cadvisor/pull/2910 made CPU usage metrics configurable from CLI, and process metrics were that already earlier.  Prometheus code skips reporting them when they are disabled, and polling should be skipped too.

Fixes: https://github.com/google/cadvisor/issues/2932


Creatone (approved):
LGTM

PR #2946, opened 2021-09-29
Prepend rootfs to the paths in containerd handler

qiutongs:
Signed-off-by: Qiutong Song <songqt01@gmail.com>

## Overview

Previous commit has a bug of checking wrong path. The problem was the path should be prepended with rootfs if cadvisor runs within container.
```
Failed to create existing container: /kubepods/besteffort/pod3fd5ec35-da3b-4174-8cb6-d710b7c483b3/ffd567befc2a43b587af98008130f195fb6f0693e8807e93df085252e3576369: stat failed on /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/177/fs with error: no such file or directory
```

## Testing

```
make build
make test
```

```
make release

VERSION=v0.36.0 # use the latest release version from https://github.com/google/cadvisor/releases
sudo docker run \
  --volume=/:/rootfs:ro \
  --volume=/sys:/sys:ro \
  --publish=8080:8080 \
  --detach=true \
  --name=cadvisor \
  --privileged \
  --device=/dev/kmsg \
  gcr.io/cadvisor/cadvisor:$VERSION
```

- Verified FS metrics are properly reported
- Verified no error logs

### DIFF
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -18,6 +18,7 @@ package containerd
 import (
 	"encoding/json"
 	"fmt"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bobbypage:
LGTM

PR #2963, opened 2021-10-18
Support listening on TLS

frebib:
Allows specifying a certificate&key pair by file paths, as well as an
optional CA file. Enforce/allow mutual-TLS handshake behaviour too.

Signed-off-by: Joe Groocock <me@frebib.net>

---

I appreciate that this implementation is rather rudimentary, and I don't expect it to be accepted immediately but I figured it was a good start and a discussion point.

### DIFF
--- a/cmd/cadvisor.go
+++ b/cmd/cadvisor.go
@@ -16,8 +16,10 @@ package main
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"os"
@@ -57,6 +59,11 @@ var httpAuthFile = flag.String("http_auth_file", "", "HTTP auth file for the web
 var httpAuthRealm = flag.String("http_auth_realm", "localhost", "HTTP auth realm for the web UI")
 var httpDigestFile = flag.String("http_digest_file", "", "HTTP digest file for the web UI")
 var httpDigestRealm = flag.String("http_digest_realm", "localhost", "HTTP digest file for the web UI")
+
+var tlsCertPath = flag.String("tls_cert_path", "", "TLS certificate file path to serve cadvisor HTTPS with. Must be specified along with -tls_key_path")
+var tlsKeyPath = flag.String("tls_key_path", "", "TLS key file path to serve cadvisor HTTPS with. Must be specified along with -tls_cert_path")
+var tlsCAPath = flag.String("tls_ca_path", "", "TLS certificate authority file path. Used to verify TLS clients connecting to cadvisor. System CA roots will be used if this is not specified.")
+var tlsClientAuth = flag.String("tls_client_auth", "require", "TLS authentication mode, must be one of request, optional, requireany, require or none.")
 
 var prometheusEndpoint = flag.String("prometheus_endpoint", "/metrics", "Endpoint to expose Prometheus metrics on")
 
@@ -155,6 +162,9 @@ func main() {
 		klog.Fatalf("Failed to register HTTP handlers: %v", err)
 	}
 
+	// Generate tls.Config if it was requested by flags
+	tlsConfig := createTLSConfig(*tlsCertPath, *tlsKeyPath, *tlsCAPath, *tlsClientAuth)
+
 	containerLabelFunc := metrics.DefaultContainerLabels
 	if !*storeContainerLabels {
 		whitelistedLabels := strings.Split(*whitelistedContainerLabels, ",")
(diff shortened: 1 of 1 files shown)


frebib:
Bump?

cmd/cadvisor.go
```diff
@@ -55,6 +57,11 @@ var httpAuthRealm = flag.String("http_auth_realm", "localhost", "HTTP auth realm
 var httpDigestFile = flag.String("http_digest_file", "", "HTTP digest file for the web UI")
 var httpDigestRealm = flag.String("http_digest_realm", "localhost", "HTTP digest file for the web UI")
 
+var tlsCertPath = flag.String("tls_cert_path", "", "TLS certificate file path")
+var tlsKeyPath = flag.String("tls_key_path", "", "TLS key file path")
+var tlsCAPath = flag.String("tls_ca_path", "", "TLS certificate authority file path")
+var tlsClientAuth = flag.String("tls_client_auth", "require", "TLS authentication mode, must be one of request, optional, requireany, require or none")
```
iwankgb, line 63: Am I right that:
* `tlsCertPath` and `tlsKeyPath` are used for HTTPS purposed, and
* `tlsCAPath` and `tlsClientAuth` are used for authentication purposes?

If so it would be great to clearly state it on the flag usage description.

cmd/cadvisor.go
```diff
@@ -153,6 +160,53 @@ func main() {
...
+	case "requireany":
+		tlsConfig.ClientAuth = tls.RequireAnyClientCert
+	case "require":
+		tlsConfig.ClientAuth = tls.RequireAndVerifyClientCert
+	case "none":
+		tlsConfig.ClientAuth = tls.NoClientCert
+	default:
+		klog.Fatalf("Invalid tls_client_auth: %s", *tlsClientAuth)
+	}
+
```
iwankgb, line 209: Extracting this whole part to a separate function would make code a bit more readable. You would just need to return `*tls.Config`.

cmd/cadvisor.go
```diff
@@ -176,7 +230,23 @@ func main() {
...
 
 	addr := fmt.Sprintf("%s:%d", *argIP, *argPort)
-	klog.Fatal(http.ListenAndServe(addr, rootMux))
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	// Wrap in a TLS listener if cert/key was specfied
+	if *tlsCertPath != "" {
```
iwankgb, line 239: I would rather check for `tlsConfig != nil` (assuming that it's of type `*tls.Config`).

iwankgb (changes requested):

cmd/cadvisor.go
```diff
@@ -153,6 +160,53 @@ func main() {
 		klog.Fatalf("Failed to register HTTP handlers: %v", err)
 	}
 
+	// Load TLS server configuration
+	var tlsConfig tls.Config
+	if *tlsCertPath != "" && *tlsKeyPath != "" {
+		if *tlsCertPath == "" || *tlsKeyPath == "" {
+			klog.Fatal("Both tls_cert_path and tls_key_path are required at the same time")
+		}
```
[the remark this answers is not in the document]
frebib: Refactored the order of these checks. I think it's probably better now?

cmd/cadvisor.go
frebib: Done. Lmk what you think of the wording. I don't words good

cmd/cadvisor.go
frebib: Done

cmd/cadvisor.go
frebib: Done

frebib:
Is there some documentation or help pages I should update to go along with this change?

frebib:
/retest

Test failures seem irrelevant. It was working fine before.
Bump on this

frebib:
🤷🏻  It fails to build locally for me on `master`- it's not the fault of this PR though.

cmd/cadvisor.go
```diff
@@ -55,6 +57,11 @@ var httpAuthRealm = flag.String("http_auth_realm", "localhost", "HTTP auth realm
 var httpDigestFile = flag.String("http_digest_file", "", "HTTP digest file for the web UI")
 var httpDigestRealm = flag.String("http_digest_realm", "localhost", "HTTP digest file for the web UI")
 
+var tlsCertPath = flag.String("tls_cert_path", "", "TLS certificate file path to serve cadvisor HTTPS with. Must be specified along with -tls_key_path")
+var tlsKeyPath = flag.String("tls_key_path", "", "TLS key file path to serve cadvisor HTTPS with. Must be specified along with -tls_cert_path")
+var tlsCAPath = flag.String("tls_ca_path", "", "TLS certificate authority file path. Used to verify TLS clients connecting to cadvisor. System CA roots will be used if this is not specified.")
+var tlsClientAuth = flag.String("tls_client_auth", "require", "TLS authentication mode, must be one of request, optional, requireany, require or none.")
```
Ichabond, line 63: Defaulting this flag to `require` implies TLS to be mandatory. While https://github.com/google/cadvisor/pull/2963/commits/346cad48eceb7eab40962b5a5583f62f0b835813#diff-cec39746c40e86227962aa52ed9ac22cf95c1504cef42cb16c0dd5c16a8cf6caR308 makes it clear that's not the case, I think this is a logic bug.
If this flag is `require`, the program should terminate if the other required parameters aren't provided (so fail close, not fail open). As such, defaulting this flag to `require` at the same time as adding the feature is a (imo) bad idea.

cmd/cadvisor.go
```diff
@@ -153,6 +160,9 @@ func main() {
 		klog.Fatalf("Failed to register HTTP handlers: %v", err)
 	}
 
+	// Generate tls.Config if it was requested by flags
```
Ichabond, line 163: Comment is inaccurate. Implies the existence of a conditional that isn't there. Instead, consider something like:
`createTLSConfig returns nil if arguments are empty strings`

iwankgb (commented):
See: https://github.com/google/cadvisor/pull/2963/files#r921991414, please.

frebib:
Needs rework

frebib:
Wow? Stale bot? Really?

PR #2964, opened 2021-10-18
Revert "Integrate CRI and containerd file system stats "

bobbypage:
Reverts google/cadvisor#2956

We found that integrating the CRI via copying the proto definition will fail once vendoring into k/k with runtime panic

```
panic: proto: duplicate enum registered: runtime.v1alpha2.Protocol
```

As a result, reverting this change.


luarx:
Are you looking for an alternative solution? 👀

PR #2966, opened 2021-10-19
Remove copied CRI and depend on CRI module

qiutongs:
The copied CRI code failed build in K8s. So we decided to give up that way. In containerd-cri branch, it doesn't need copied CRI anymore.


build/boilerplate/boilerplate.py
```diff
@@ -95,7 +95,7 @@ def file_passes(filename, refs, regexs):
 def file_extension(filename):
     return os.path.splitext(filename)[1].split(".")[-1].lower()
 
-skipped_dirs = ['Godeps', 'vendor', 'third_party', '_gopath', '_output', '.git', 'cri-api']
+skipped_dirs = ['Godeps', 'vendor', 'third_party', '_gopath', '_output', '.git']
```
kolyshkin, line 98: You can remove Godeps while at it. Obsoleted since commit 236f41d419d72d9469b018843cda9635ecf328a1

qiutongs: ok. I can remove it.

PR #2969, opened 2021-10-21
cmd: enable linter, fix warnings

kolyshkin:
Apparently, golangci-lint (and thus `make lint`, `make presubmit`, and various CI jobs)
was not checking any code in cmd.

Make it do so, and fix found warnings.

**This requires careful review, as some changes are not trivial. Please review commit by commit.**

### DIFF
--- a/cmd/internal/api/handler.go
+++ b/cmd/internal/api/handler.go
@@ -135,10 +135,6 @@ func writeResult(res interface{}, w http.ResponseWriter) error {
 }
 
 func streamResults(eventChannel *events.EventChannel, w http.ResponseWriter, r *http.Request, m manager.Manager) error {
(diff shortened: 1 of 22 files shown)


cmd/internal/api/handler.go
```diff
@@ -151,7 +147,7 @@ func streamResults(eventChannel *events.EventChannel, w http.ResponseWriter, r *
 	enc := json.NewEncoder(w)
 	for {
 		select {
-		case <-cn.CloseNotify():
+		case <-r.Context().Done():
```
bobbypage, line 150: this seems right me, i.e. switch to context.Done()...

bobbypage:
LGTM, big thanks @kolyshkin, all the changes here look good. Thanks for your help cleaning it up and fixing all of the linter issues.

PR #2974, opened 2021-10-22
Optimize PrometheusCollector scrape efficiency.

bwplotka:
Fixes https://github.com/kubernetes/kubernetes/issues/104459

This might end up as some reusable framework for CRI and others, but for now attempting to optimize cadvisor in a bespoke fashion.

CPU profile: https://share.polarsignals.com/e46845c/
Memory profile: https://share.polarsignals.com/9cb2d48/

Related to https://github.com/prometheus/client_golang/discussions/917

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/metrics/cache/cache_test.go
+++ b/metrics/cache/cache_test.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package cache_test
 
 import (
--- a/metrics/cache/cache.go
+++ b/metrics/cache/cache.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package cache
 
 import (
--- a/metrics/cache/cache_bench_test.go
+++ b/metrics/cache/cache_bench_test.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package cache
 
 import (
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package metrics
 
 import (


metrics/prometheus.go
```diff
@@ -1771,50 +1779,52 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+//}
+
+// Collect fetches latest statistics about containers and exposes them to Prometheus handler that converts them
+// to expected Prometheus format.
+// Based on option it
+// * ask the latest from provided services.
+// * returns cached results if until time expires.
+//
+// It implements prometheus.rawCollector.
+func (c *PrometheusCollector) Collect() []*dto.MetricFamily {
```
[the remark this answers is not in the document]
bwplotka: Right. This is because we need to use this collector in new `promhttp.TransactionalHandler` similar how i implemented it on cadvisor. Looks like this has to be changed on kubelet too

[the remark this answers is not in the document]
ehashman: I did a bunch of digging and the fix won't be quite as trivial or straightforward as it was in this PR. This is because in k8s interaction with Prometheus client libraries goes through the component-base metrics libraries, and it's not clear to me what the best way forward is. I pinged @logicalhan to get some feedback.

xref:
- https://github.com/kubernetes/kubernetes/blob/70e7876bef5815284b43bb1fe20b55453899d721/pkg/kubelet/server/server.go#L360
- https://github.com/kubernetes/kubernetes/blob/47c63a39ed6866c1dc45eb9ed6c4a78fec8474c6/staging/src/k8s.io/component-base/metrics/registry.go#L114-L116

Annoyingly, while it would be nice to stop using k8s' deprecated `RawMustRegister`, it looks like that is still used in a couple of other places, so we can't just get rid of it and define something better for this use case:

https://cs.k8s.io/?q=RawMustRegister&i=nope&files=&excludeFiles=&repos=kubernetes/kubernetes

metrics/prometheus.go
```diff
@@ -83,40 +98,33 @@ type containerMetric struct {
...
 
-func (cm *containerMetric) desc(baseLabels []string) *prometheus.Desc {
-	return prometheus.NewDesc(cm.name, cm.help, append(baseLabels, cm.extraLabels...), nil)
-}
-
 // ContainerLabelsFunc defines all base labels and their values attached to
 // each metric exported by cAdvisor.
 type ContainerLabelsFunc func(*info.ContainerInfo) map[string]string
 
 // PrometheusCollector implements prometheus.Collector.
```
logicalhan, line 105: This isn't true anymore right?

metrics/prometheus.go
```diff
@@ -1771,50 +1779,52 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
 
-// Describe describes all the metrics ever exported by cadvisor. It
-// implements prometheus.PrometheusCollector.
-func (c *PrometheusCollector) Describe(ch chan<- *prometheus.Desc) {
-	c.errors.Describe(ch)
-	for _, cm := range c.containerMetrics {
-		ch <- cm.desc([]string{})
+		session.MustAddMetric(
+			"container_scrape_error", "1 if there was an error while getting container metrics, 0 otherwise",
+			nil, nil, prometheus.GaugeValue, float64(errorsGauge), nil,
```
ehashman, line 1820: Spent some time today trying to get this working properly in k8s.

I am finding that I keep running into errors... If I don't set `IdType: cadvisorv2.TypeName` in the cadvisor options, I get

```
ehashman@fedora:~/src/k8s$ curl http://127.0.0.1:8001/api/v1/nodes/127.0.0.1/proxy/metrics/cadvisor
# HELP cadvisor_version_info A metric with a constant '1' value labeled by kernel version, OS version, docker version, cadvisor version & cadvisor revision.
# TYPE cadvisor_version_info gauge
cadvisor_version_info{cadvisorRevision="",cadvisorVersion="",dockerVersion="",kernelVersion="5.10.0-9-amd64",osVersion="Debian GNU/Linux 11 (bullseye)"} 1
# HELP container_scrape_error 1 if there was an error while getting container metrics, 0 otherwise
# TYPE container_scrape_error gauge
container_scrape_error 1
```

and error in the logs
```
W1124 16:45:01.711801 1279370 prometheus.go:1815] Couldn't get containers: invalid request type ""
```

If I do set it, I get

```
ehashman@fedora:~/src/k8s$ curl http://127.0.0.1:8001/api/v1/nodes/127.0.0.1/proxy/metrics/cadvisor
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "error trying to reach service: EOF",
  "reason": "ServiceUnavailable",
  "code": 503
}
```

Not sure what I'm doing wrong.

k8s WIP PR is https://github.com/kubernetes/kubernetes/pull/106334

And the patch I'm applying on top of the PR to get to that second error case is

```diff
diff --git a/pkg/kubelet/server/server.go b/pkg/kubelet/server/server.go
index ce3fc942e49..be9cb5e2603 100644
--- a/pkg/kubelet/server/server.go
+++ b/pkg/kubelet/server/server.go
@@ -378,7 +378,16 @@ func (s *Server) InstallDefaultHandlers() {
                includedMetrics[cadvisormetrics.AcceleratorUsageMetrics] = struct{}{}
        }
 
-       br.MustRegisterRaw(metrics.NewPrometheusCollector(prometheusHostAdapter{s.host}, containerPrometheusLabelsFunc(s.host), includedMetrics, clock.RealClock{}))
+       cadvisorOpts := cadvisorv2.RequestOptions{
+               IdType:    cadvisorv2.TypeName,
+               Count:     1,
+               Recursive: true,
+       }
+
+       containersCollector := metrics.NewPrometheusCollector(prometheusHostAdapter{s.host}, containerPrometheusLabelsFunc(s.host), includedMetrics, clock.RealClock{})
+       containersCollector.SetOpts(cadvisorOpts)
+       br.MustRegisterRaw(containersCollector)
+
        br.MustRegister(metrics.NewPrometheusMachineCollector(prometheusHostAdapter{s.host}, includedMetrics))
        s.restfulCont.Handle(cadvisorMetricsPath,
                promhttp.HandlerForTransactional(br, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}),

```

bwplotka:
Updated for the newest client_golang changes (https://github.com/prometheus/client_golang/pull/929)

Should be ready for review. There are some [observations](https://github.com/prometheus/client_golang/pull/929#issuecomment-1023077450) for further improvements, but otherwise, all races should be fixed and we should see significant improvement in allocs and cpus.

TODO still:
* Fix a few failing tests.
* Benchmark
* Run with Kubelet to and measure/ check functionally.

cc @ehashman @dgrisonnet 

FYI: I have to take 2w off due to post knee surgery recovery, but anyone is welcome to continue on this!

bwplotka:
Ofc @dgrisonnet said, move to pure watcher based approach would be extremely beneficial, but it needs huge changes, so keeping that off for now.

cmd/internal/http/handlers.go
```diff
@@ -92,31 +93,41 @@ func RegisterHandlers(mux httpmux.Mux, containerManager manager.Manager, httpAut
...
-		r := prometheus.NewRegistry()
-		r.MustRegister(
-			metrics.NewPrometheusCollector(resourceManager, f, includedMetrics, clock.RealClock{}, opts),
-			machineCollector,
-			goCollector,
-			processCollector,
-		)
-		promhttp.HandlerFor(r, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}).ServeHTTP(w, req)
+		opts.Count = 1        // We only want the latest datapoint.
+		opts.Recursive = true // Get all child containers.
```
ehashman, line 122: These options don't appear to be provided to the nameTypeGatherer anymore; previously we'd plug these in via the NewPrometheusCollector.

ehashman:
I have updated my k/k PR https://github.com/kubernetes/kubernetes/pull/107960/files#diff-c3e7f724d1b0dcee40df80716ed57d90d2649710150fa92bf9822bdad35e0429 which integrates this change + the required changes to rebase on top of runc 1.1.0.

I still am not getting any metrics... any time I try to hit the kubelet endpoint, I get an empty response and a corresponding logline of:

```
I0210 15:37:17.477351  121888 httplog.go:131] "HTTP" verb="GET" URI="/metrics/cadvisor" latency="91.027µs" userAgent="curl/7.74.0" audit-ID="" srcIP="127.0.0.1:51322" resp=0
```

bwplotka:
Ack @ehashman taking a look today/tomorrow (:

cmd/internal/http/handlers.go
bwplotka: got it now, you mean this 

image

bwplotka: You comment was on point, we DO forget using `UpdateOnMaxAge` here (and in our k8s impl). My bad, fixed (:

bwplotka:
This PR is ready to go IMO.

All benchmarked. With kubelet we see significantly less allocations and CPU usage (https://github.com/kubernetes/kubernetes/pull/108194)

bwplotka:
cc @ehashman @dgrisonnet

bwplotka:
On kubelet side I see only one discrepancy between this PR and without this PR:

image

Worth double checking before merging. Otherwise no blocker.

bwplotka:
Not sure what pull-cadvisor-e2e job fail means: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/google_cadvisor/2974/pull-cadvisor-e2e/1496837320517619712/

bwplotka:
TODO: Clean up the commits.

metrics/cache/cache_test.go
```diff
@@ -0,0 +1,252 @@
+package cache_test
+
+import (
+	"strings"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+//
```
bwplotka, line 9: uncomment

saswatamcode: Done in https://github.com/google/cadvisor/pull/2974/commits/a587512437b8dda97203ffee4757a2b9ea563c44.

tossmilestone:
@bwplotka hi, is this pr still in progress? We have encountered the probelm either.

PR #2981, opened 2021-10-27
Pinned tag using SHA

nathan-415:
Pinned actions by SHA instead of a tag because tags can be moved while SHA can’t.

https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies


bobbypage:
Thank you @nathan-415 for the contribution but as mentioned I don't think this will provide significant value, so will close this PR.

PR #3001, opened 2021-11-02
cgroupv2: fix swap accounting

bobrik:
Since runsc v1.0.0-rc94 `stats.MemoryStats.SwapUsage` on cgroupv2 includes memory usage:

* https://github.com/opencontainers/runc/commit/b99ca25ad0f

This results in wrong metrics:

```
container_memory_rss{id="/system.slice/qs-dnsv2.service"} 9.173716992e+09 1635890983582
container_memory_swap{id="/system.slice/qs-dnsv2.service"} 1.249011712e+10 1635890983582
container_memory_usage_bytes{id="/system.slice/qs-dnsv2.service"} 1.249011712e+10 1635890983582
```

Here my service dosn't have any swapped pages, yet it reports all memory as swapped.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -806,7 +806,7 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	if cgroups.IsCgroup2UnifiedMode() {
 		ret.Memory.Cache = s.MemoryStats.Stats["file"]
 		ret.Memory.RSS = s.MemoryStats.Stats["anon"]
-		ret.Memory.Swap = s.MemoryStats.SwapUsage.Usage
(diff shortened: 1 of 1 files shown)


bobrik:
I'm not really sure if it's something that needs to be fixed or if I'm expected to do the math myself at query time.

bobrik:
The metric is correct on cgroupv1, so I think the right way is to merge this fix for cgroupv2.

bobbypage:
makes sense, LGTM. 

/cc @kolyshkin who made the runc change

kolyshkin:
Yeah this is weird. cgroup v1 reports mem+swap in this field, so we had to make cgroup v2 report the same. Not sure if there's a better way to address this.

The patch LGTM though.

bobbypage:
Thanks, LGTM!

bobbypage (approved):

PR #3006, opened 2021-11-04
deployment: Add helm chart.

Creatone:


Creatone:
@bobbypage PTAL

deploy/helm/templates/daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+          {{if .Values.perf}}- --perf_events_config=/etc/config/perf-config.json{{end}}
+          {{if .Values.disable_metrics}}- --disable_metrics={{ join "," .Values.disable_metrics}}{{end}}
+          {{if .Values.enable_metrics}}- --enable_metrics={{ join "," .Values.enable_metrics}}{{end}}
+          {{if .Values.port}}- --port={{.Values.port}}{{end}}
+          {{if .Values.docker_only}}- --docker_only={{.Values.docker_only}}{{end}}
+        securityContext:
+          {{- toYaml .Values.securityContext | nindent 10 }}
+        resources:
+          requests:
+            memory: 400Mi
```
bobbypage, line 39: why did we pick these values?

deploy/helm/templates/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
bobbypage, line 1: is there a way to disable creating service? I would imagine not everyone wants to create a nodeport, so maybe should be configurable?

deploy/helm/values.yaml
```diff
@@ -0,0 +1,39 @@
+enable_metrics:
+  - perf_event
```
bobbypage, line 2: should we enable these by default? maybe they should be opt-in similar as normal cAdvisor?

deploy/helm/templates/daemonset.yaml
Creatone: Source:  https://github.com/google/cadvisor/blob/master/deploy/kubernetes/base/daemonset.yaml#L23

deploy/helm/values.yaml
Creatone: Yes, I agree, they should be opt-in similar to normal cAdvisor.

Creatone:
@bobbypage @iwankgb @dims @mrunalp PTAL

bobbypage:
I would prefer we avoid adding additional deployment modes into cAdvisor itself that we need to support long term (we already support basic kustomize (https://github.com/google/cadvisor/tree/master/deploy/kubernetes). Users can always create external repos for helm charts using the existing GCR images, and it looks like already a few exist: https://artifacthub.io/packages/helm/ckotzbauer/cadvisor

What do you think?

PR #3007, opened 2021-11-06
Fix missing metrics on cgroups v2

mhermida:
Path lookups fail in the cgroups v2 cgroupPaths map (common/helpers.go, getSpecInternal) that only has key "" and some metrics are missing (cpu cfs related ones, for instance).

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -195,7 +195,6 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// Processes, read it's value from pids path directly
-	//pidsRoot, ok := cgroupPaths["pids"]
 	pidsRoot, ok := getControllerPath(cgroupPaths, "pids", cgroup2UnifiedMode)
 	if ok {
 		if utils.FileExists(pidsRoot) {
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -105,7 +105,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// CPU.
-	cpuRoot, ok := cgroupPaths["cpu"]
+	cpuRoot, ok := getControllerPath(cgroupPaths, "cpu", cgroup2UnifiedMode)
 	if ok {
 		if utils.FileExists(cpuRoot) {
 			if cgroup2UnifiedMode {
@@ -152,7 +152,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 
 	// Cpu Mask.
 	// This will fail for non-unified hierarchies. We'll return the whole machine mask in that case.
-	cpusetRoot, ok := cgroupPaths["cpuset"]
+	cpusetRoot, ok := getControllerPath(cgroupPaths, "cpuset", cgroup2UnifiedMode)
 	if ok {
 		if utils.FileExists(cpusetRoot) {
 			spec.HasCpu = true
@@ -167,7 +167,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// Memory
-	memoryRoot, ok := cgroupPaths["memory"]
+	memoryRoot, ok := getControllerPath(cgroupPaths, "memory", cgroup2UnifiedMode)
 	if ok {
 		if cgroup2UnifiedMode {
 			if utils.FileExists(path.Join(memoryRoot, "memory.max")) {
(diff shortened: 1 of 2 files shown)


container/common/helpers.go
```diff
@@ -195,7 +195,8 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// Processes, read it's value from pids path directly
-	pidsRoot, ok := cgroupPaths["pids"]
+	//pidsRoot, ok := cgroupPaths["pids"]
```
bobbypage, line 198: remove this line

bobbypage:
LGTM on the change overall, I tested out cadvisor on cgroupv2 image without this change and added some debug logging of the `cgroupsPath` and they looked like this:

```
cgroupsPath:map[:/sys/fs/cgroup/system.slice/systemd-update-utmp.service]
cgroupsPath:map[:/sys/fs/cgroup/system.slice/containerd.service]
cgroupsPath:map[:/sys/fs/cgroup/system.slice/containerd.service]
cgroupsPath:map[:/sys/fs/cgroup/system.slice/docker.socket]
```

so as a result, none of the subcontrollers were found. So I think change here is appropriate, on cgroupv2 we will only have a single cgroupsPath with key `""` and from there we can get the individual controller files.

kolyshkin:
Overall SGTM, although I don't like the complication of having to add `cgroup2UnifiedMode` to every call. Perhaps this can be dropped, and `getControllerPath` should determine whether we use v2 all by itself (the call is very cheap as we cache it).

I'm on PTO, let me take a closer look later this week.

mhermida:
Hi @bobbypage, @kolyshkin
thanks for your reviews and comments.

>  ...although I don't like the complication of having to add cgroup2UnifiedMode to every call. 

yes, I don't like it much either, but the only way of avoiding it I found was to call `cgroups.IsCgroup2UnifiedMode()` within `getControllerPath` and I wasn't very convinced about that one too, so whatever you think is best.

giuseppe:
the `cgroups.IsCgroup2UnifiedMode()` result is memoized, so it should be fine to call it from `getControllerPath()`

mhermida:
Thanks @giuseppe!!
The thing dropping the parameter `cgroup2UnifiedMode` from `getControllerPath` (which could also be dropped from `getSpecInternal`) and calling `cgroups.IsCgroup2UnifiedMode()` from the functions is that testing becomes a bit more complex as we need to adapt the behaviour of  `cgroups.IsCgroup2UnifiedMode()` to cgroups v1 or cgroups v2 depending on which case are the tests for. Not sure how to do that...

giuseppe:
I am fine with the current version.  Deferring to @bobbypage and @kolyshkin :)

bobbypage:
ack, I'm happy with change here too, we can also refactor the cgroupv2 check. Thank you for the fix.

bobbypage (approved):

kolyshkin:
> The thing dropping the parameter `cgroup2UnifiedMode` from `getControllerPath` (which could also be dropped from `getSpecInternal`) and calling `cgroups.IsCgroup2UnifiedMode()` from the functions is that testing becomes a bit more complex as we need to adapt the behaviour of `cgroups.IsCgroup2UnifiedMode()` to cgroups v1 or cgroups v2 depending on which case are the tests for. Not sure how to do that...

I see two approaches.

The first one would be to check whether you have v1 or v2 right inside the test, and only use test cases for the available version. This relies on the fact that CI is executed in both cgroup v1 and cgroup v2 environments (which AFAIK is not entirely true at the moment, but this can and should be fixed). The upside is it's trivial to implement, the downside you need different test environments (which, to my mind, is OK).

The alternative, somethat more complicated, is to add something like

```go
// isV2 can be overridden by some unit tests.
var isV2 = cgroups.IsCgroup2UnifiedMode
```

to the code, and use the new `isV2()` everywhere you need it.

Now, you can override it on a per-test basis, something like (note this is entirely untested):

```go
func TestThis(t *testing.T) {
  isV2Saved := isV2
  t.Cleanup(func() {isV2 = isV2Saved})

  // Tests for cgroup v1.
  isV2 = func() bool {return false}
  ....
 
  // Tests for cgroup v2.
  isV2 = func() bool {return true} 
  ....
```
 
The downsides are (1) this brings in some complexity into the test code and (2) you can'd execute such tests in parallel (if you want to, this calls for more complexity).

Overall, I'd choose the first solution, but I'm OK with the code as it is.

kolyshkin:
> This relies on the fact that CI is executed in both cgroup v1 and cgroup v2 environments (which AFAIK is not entirely true at the moment, but this can and should be fixed).

Filed https://github.com/google/cadvisor/issues/3009. Should be easy to implement.

PR #3008, opened 2021-11-08
container/libcontainer: Add container mountpoints counter.

Creatone:


container/libcontainer/handler.go
```diff
@@ -297,12 +298,25 @@ func processStatsFromProcs(rootFs string, cgroupPath string, rootPid int) (info.
 				socketCount++
 			}
 		}
+		mountpointPath := path.Join(rootFs, "/proc", pid, "mountpoint")
+		fileReader, err := os.Open(mountpointPath)
```
bobbypage, line 302: curious how expensive this is, i.e. calling /proc/pid/mountpoint for every single pid?

Creatone:
It shouldn't be counted by default.

PR #3013, opened 2021-11-15
fix /metrics/cadvisor when one container scrape error may lead to missing all container metrics

Dingshujie:
``` golang
	containers, err := c.infoProvider.GetRequestedContainersInfo("/", c.opts)
	if err != nil {
		c.errors.Set(1)
		klog.Warningf("Couldn't get containers: %s", err)
                return
	}
```

when GetRequestedContainersInfo return error, collectContainersInfo log a warning and return, ignore all contaienr info return by function, may lead /metrics/cadvisor lack of all container metrics data.  it seems  reasonable.


Dingshujie:
@bobbypage can you review this PR?

PR #3029, opened 2021-12-13
fix: atomic operations on 32-byte device

jonyhy96:
xref: https://github.com/kubernetes/kubernetes/issues/106977

prevent panic on atomic operations

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -64,6 +64,7 @@ type containerInfo struct {
 }
 
 type containerData struct {
+	oomEvents                uint64
 	handler                  container.ContainerHandler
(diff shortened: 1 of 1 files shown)


soider:
@jonyhy96 fyi, I tried using the same patch in k0sproject release for 1.23 (https://github.com/k0sproject/k0s/pull/1340/commits/953909e91663b6a167945284b40af44f59a1441a) it seems to be working on armv7, at least kubelet doesn't panic and generally passed our test suite.

brandond:
Will be using this for K3s as well.

iwankgb (approved):

PR #3030, opened 2021-12-14
Fix process metrics collection in cgroup v2

### LINKED ISSUE #3026: Missing process metrics in cgroup v2
We are running cadvisor in our k8s cluster following the example setup, but process metrics such as container_sockets and container_processes are missing.  Our node's OS is [Flatcar 2983.2.1](https://www.flatcar-linux.org/releases/#release-2983.2.1) and it has already migrated to the cgroup V2 from version 2969.0.0. 

I investigated cadvisor code and found out that cadvisor didn't call processStatsFromProcs because it tried to get CPU subsystem path and use it but there's no such pash in cgroup v2. I think cadvisor needs to use the unified cgroup path in cgroup v2 environment.
https://github.com/google/cadvisor/blob/v0.43.0/container/libcontainer/handler.go#L173

cAdvisor v0.43.0
Node OS: Flatcar 2983.2.1
Container runtime: containerd 1.5.6

Manifest
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    seccomp.security.alpha.kubernetes.io/pod: docker/default
  labels:
    app: cadvisor
  name: cadvisor
  namespace: monitoring
spec:
  selector:
    matchLabels:
      app: cadvisor
      name: cadvisor
  template:
    metadata:
      labels:
        app: cadvisor
        name: cadvisor
    spec:
      automountServiceAccountToken: false
      containers:
      - args:
        - --housekeeping_interval=10s
        - --enable_metrics=app,cpu,cpuLoad,disk,diskIO,memory,network,process
        - --containerd=/var/run/k8s-containerd.sock
        - --store_container_labels=false
        - --whitelisted_container_labels=io.kubernetes.container.name,io.kubernetes.pod.name,io.kubernetes.pod.namespace
        image: quay.io/cybozu/cadvisor:0.43.0.1
        name: cadvisor
        ports:
        - containerPort: 8080
          hostPort: 9110
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 800m
            memory: 2000Mi
          requests:
            cpu: 400m
            memory: 400Mi
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /rootfs
          name: rootfs
          readOnly: true
        - mountPath: /var/run
          name: var-run
          readOnly: true
        - mountPath: /sys
          name: sys
          readOnly: true
        - mountPath: /var/lib/docker
          name: docker
          readOnly: true
        - mountPath: /dev/disk
          name: disk
          readOnly: true
      hostPID: true
      serviceAccountName: cadvisor
      terminationGracePeriodSeconds: 30
      tolerations:
      - key: cke.cybozu.com/role
        operator: Equal
        value: storage
      volumes:
      - hostPath:
          path: /
        name: rootfs
      - hostPath:
          path: /var/run
        name: var-run
      - hostPath:
          path: /sys
        name: sys
      - hostPath:
          path: /var/lib/docker
        name: docker
      - hostPath:
          path: /dev/disk
        name: disk
```

ysksuzuki:
This PR fixes #3026 to use a unified root path to collect process stats from procs if cadvisor is running in cgroup v2.

Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -105,7 +105,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// CPU.
-	cpuRoot, ok := getControllerPath(cgroupPaths, "cpu", cgroup2UnifiedMode)
+	cpuRoot, ok := GetControllerPath(cgroupPaths, "cpu", cgroup2UnifiedMode)
 	if ok {
 		if utils.FileExists(cpuRoot) {
 			if cgroup2UnifiedMode {
@@ -152,7 +152,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 
 	// Cpu Mask.
 	// This will fail for non-unified hierarchies. We'll return the whole machine mask in that case.
-	cpusetRoot, ok := getControllerPath(cgroupPaths, "cpuset", cgroup2UnifiedMode)
+	cpusetRoot, ok := GetControllerPath(cgroupPaths, "cpuset", cgroup2UnifiedMode)
 	if ok {
 		if utils.FileExists(cpusetRoot) {
 			spec.HasCpu = true
@@ -167,7 +167,7 @@ func getSpecInternal(cgroupPaths map[string]string, machineInfoFactory info.Mach
 	}
 
 	// Memory
-	memoryRoot, ok := getControllerPath(cgroupPaths, "memory", cgroup2UnifiedMode)
+	memoryRoot, ok := GetControllerPath(cgroupPaths, "memory", cgroup2UnifiedMode)
 	if ok {
 		if cgroup2UnifiedMode {
 			if utils.FileExists(path.Join(memoryRoot, "memory.max")) {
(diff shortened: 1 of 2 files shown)


jepio (approved):
I tested your changes and they work correctly. LGTM.

bobbypage:
Thank you for the fix! 

Do you think we can use this helper function instead? 

https://github.com/google/cadvisor/blob/5a288945e65a91ddf5e0212d35d0c08d40316b82/container/common/helpers.go#L220-L231

It might be need to exported since it's a different package, but I think it would be nice to use this helper function to avoid duplicating this logic... it's already used in a few places to get cgroup controller path:

https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/google/cadvisor%24%405a28894+getControllerPath&patternType=literal

ysksuzuki:
Thank you for your comment!

> Do you think we can use this helper function instead?

Yes, I think we should use this helper function. I modified this patch and confirmed that it's working correctly in my environment.

bobbypage (approved):

bobbypage:
LGTM, thanks!

PR #3041, opened 2022-01-04
Upgrade github.com/opencontainers/runc to version 1.0.3

dims:
Don't think we end up using the code ... but still good thing to keep up with the jones'es. containerd updated it as well here - https://github.com/containerd/containerd/blob/main/script/setup/runc-version

Also the same for image-spec to v1.0.2

- https://github.com/advisories/GHSA-v95c-p5hm-xq8f
- https://github.com/advisories/GHSA-77vh-xpmg-72qh


bobbypage (approved):

bobbypage:
Thanks! from https://github.com/opencontainers/runc/releases/tag/v1.0.3, looks like a few metric related fixes

> Fixed returning error from GetStats when hugetlb is unsupported (which causes
excessive logging for kubernetes) (#3295)

PR #3042, opened 2022-01-05
backport v0.39: fix: atomic operations on 32-byte device

bobbypage:
Backport https://github.com/google/cadvisor/pull/3029 to v0.39 cAdvisor 

xref: kubernetes/kubernetes#106977


bobbypage:
closing, after closer look, `oomEvents` was only added in v0.40, so issue only affects 0.40+ and does not need to be backport to v0.39.

PR #3043, opened 2022-01-06
manager: fix determining selfContainer name

jepio:
I found that on cgroup v1 `selfContainer` was not being determined correctly, due to two issues, both of which this PR fixes.


bobbypage:
LGTM, thank you for the fix.

bobbypage (approved):

PR #3046, opened 2022-01-09
Bumping up minimum version of go to a supported version

iwankgb:
Hey, long time, no see.

This is an RFC to some extent: go 1.13 is not supported since 1.15 (August 2020). Should we still indicated 1.13 as minimum required golang version? I don't think so:
* there has been bunch of CVEs published since 08/2020
* I think it's good idea to force users to use supported compiler version

@bobbypage, what do you think? Kubernetes requires 1.16 since [March 2021](https://github.com/kubernetes/kubernetes/commit/3c2824e3dbed6069dabddac7816239de9cb2a6ee).

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -1,6 +1,6 @@
 module github.com/google/cadvisor/cmd
 
-go 1.13
+go 1.16
 
 // Record that the cmd module requires the cadvisor library module.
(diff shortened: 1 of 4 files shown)


bobbypage (approved):

bobbypage:
LGTM, thanks!

PR #3031, opened 2021-12-14
deps: update runc to 1.1.0-rc.1

kolyshkin:
runc 1.1.0-rc.1 was just released. This PR updates vendored runc/libcontainer
and modifies the code according to the changes in the libcontainer API.

Since the final 1.1.0 is supposed to be very similar (if not identical), I guess this could
be merged as is (if no issues are found).

The changes to the code are:

 - resctrl: use intelrtd.Root instead of GetIntelRdtPath (which was
   removed);

 - resctrl: test: set cgroups.TestMode = true in TestGetPids (since
   by default cgroup drivers check they open cgroupfs files);

 - container/libcontainer/helpers: modify NewCgroupManager for the
   updated fs[2].NewManager API.

PR #3048, opened 2022-01-18
vendor: bump runc to 1.1.0

kolyshkin:
Brought to you by

	go get -d github.com/opencontainers/runc@latest
	go mod tidy
	cd cmd
	go get -d github.com/opencontainers/runc@latest
	go mod tidy


kolyshkin:
@bobbypage can you cut a release once this PR lands, so I can do the same bump in kubernetes?

kolyshkin:
@bobbypage PTAL 🙏🏻

bobbypage:
> @bobbypage can you cut a release once this PR lands, so I can do the same bump in kubernetes?

Usually we cut new cAdvisor release prior to k/k release, so I was planning on cutting next 0.44 release closer to when k8s 1.24 will be released. Can we update runc in k/k without also updating cAdvisor tag or is that needed?

kolyshkin:
> I think we should be able to update update runc in k/k without also updating cAdvisor tag or is that not the case?

Due to API changes in runc, I think it is not possible. Currently, in https://github.com/kubernetes/kubernetes/pull/107149 I have updated cadvisor to commit a3b5f4f6c501d5a4ac67530bc973a05c1ebd7d8c, but an unreleased version does not look good in go.mod.

kolyshkin:
>> I think we should be able to update update runc in k/k without also updating cAdvisor tag or is that not the case?

> Due to API changes in runc, I think it is not possible.

To clarify -- attempting this will result in compilation errors.

bobbypage:
ah, thank you for clarification, it makes sense due to the changes you made in https://github.com/google/cadvisor/pull/3031

I will cut a new release then to unblock updating runc update in k/k. Thanks!

Issue #3057, opened 2022-02-08
resctrl do not handle the orphan mon_groups 


srl11:
resctrl will create mon_group for container if user set resctrl_interval(!=0), and clear it when container is destroyed. 
But, cadvisor won't clear up the mon_group of the container which is destoryed during cadvisor restart.
So should we check the orphan mon_group, and clear it up?

**releated version**: v0.43.0 v0.42.0 v0.41.0 ([commit](https://github.com/google/cadvisor/commit/80e65740c169abb5097d848f0b44debd0fa20876))

**My Solution**:
if resctrl_interval!=0, cache the mon_group created by cadvisor with timestamp, and clear up the expired mon_group.

srl11:
another comment for resctrl [comment](https://github.com/google/cadvisor/commit/80e65740c169abb5097d848f0b44debd0fa20876#r66221164)

srl11:
Is that appropriate to deal with the issue in this way? Could you please help me to review it[PR](https://github.com/google/cadvisor/pull/3058)? Thanks. 🌹 @Creatone @katarzyna-z @dashpole

Creatone:
Maybe the simplest solution is to clear all "cadvisor" prefixed mon groups during setup?

srl11:
> Maybe the simplest solution is to clear all "cadvisor" prefixed mon groups during setup?

It's the simplest solution. But when cadvisor crash frequently， it will lead to clear-up and re-create mon-group frequently.  I'm not sure whether it will cost a lot for kernel as I'm not familar with the mon_group in the kernel 😭 .

Creatone:
Frequently crashing is not expected behavior. 
I understand that corner case, we fix the same thing previously with a simplest solution in [intel/workload-collocation-agent](https://github.com/intel/workload-collocation-agent)

Creatone:
@srl11 Could you check if that commit is sufficient for you? https://github.com/Creatone/cadvisor/commit/3e2feae6363d7f511e127ff14da19d5db96bd98b

srl11:
你的邮件我已收到，谢谢你的来件，我会尽快查看。

PR #3058, opened 2022-02-09
feat: clear up the orphan mon_groups which are created by cadvisor

srl11:
Revival of [#3057](https://github.com/google/cadvisor/issues/3057), handling the orphan mon_groups.
if resctrl_interval!=0, cache the mon_group created by cadvisor, which has cadvisor monGroupPrefix, and clear up the expired mon_group when a new container is being creating.

@Creatone


srl11:
你的邮件我已收到，谢谢你的来件，我会尽快查看。

Creatone:
Why do you keep mon group names? All groups have "cadvisor" prefix.

srl11:
> "cadvisor" prefix.

All groups have "cadvisor" prefix, but some of them are still working(by cadvisor, for some case it may used by other program).  We can use the name to quickly determine whether the mon_group is out of date, related container has exited,  and should be removed.

PR #3083, opened 2022-03-23
Support for exposing PSI metrics 

### LINKED ISSUE #3052: Exopse Pressure Stall Information as metrics
[Pressure Stall Information](https://www.kernel.org/doc/html/latest/accounting/psi.html) is exposed per cgroup in cgroupv2. It's a good way to understand contention due to lack of resources ( cpu, memory, io ). For example

```
# /sys/fs/cgroup/system.slice/cpu.pressure
some avg10=0.00 avg60=0.00 avg300=0.00 total=306212315
full avg10=0.00 avg60=0.00 avg300=0.00 total=246733962
```

It would be great to expose this data source in cadvisor as metrics.

dqminh:
Fix #3052 

This depends on https://github.com/opencontainers/runc/pull/3358, so it should not be merged as-is, but we can review the structure and how metrics are exposed.


szuecs:
@bobbypage does this need to wait longer, anything missing?
I ask because there is an internal request to support this feature. :)

@dqminh likely it needs a rebase, maybe you can do it :)

szuecs:
@bobbypage not sure should I take over the PR?
@dqminh thoughts?

SuperQ:
Fyi, it's unnecessary and not in best practices to expose the pre-computed averages in Prometheus.

Prometheus can compute arbitrary averages from the Total data.

nmlc:
Any news here? Seems like [runc PSI pr](https://github.com/opencontainers/runc/pull/3900) got merged

metrics/prometheus.go
```diff
@@ -1768,6 +1768,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 		})
 	}
 
+	if includedMetrics.Has(container.PSITotalMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:        "container_cpu_psi_total_seconds",
```
SuperQ, line 1774:
```suggestion
				name:        "container_pressure_cpu_seconds_total",
```

metrics/prometheus.go
```diff
@@ -1768,6 +1768,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				help:        "Total time spent under cpu pressure in seconds.",
+				valueType:   prometheus.CounterValue,
+				extraLabels: []string{"kind"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					return getPSIValues(s, &s.Cpu.PSI, "total")
+				},
+			}, {
+				name:        "container_memory_psi_total_seconds",
```
SuperQ, line 1782:
```suggestion
				name:        "container_pressure_memory_seconds_total",
```

metrics/prometheus.go
```diff
@@ -1768,6 +1768,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+				help:        "Total container time spent under memory pressure in seconds.",
+				valueType:   prometheus.CounterValue,
+				extraLabels: []string{"kind"},
+				getValues: func(s *info.ContainerStats) metricValues {
+					return getPSIValues(s, &s.Memory.PSI, "total")
+				},
+			}, {
+				name:        "container_io_psi_total_seconds",
```
SuperQ, line 1790:
```suggestion
				name:        "container_pressure_io_seconds_total",
```

metrics/prometheus.go
```diff
@@ -1768,6 +1768,64 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
+					}
+				},
+			}
+		}
+		for _, controller := range []string{"cpu", "memory", "io"} {
+			for _, window := range []string{"10", "60", "300"} {
+				c.containerMetrics = append(c.containerMetrics, makePSIAvgMetric(controller, window))
+			}
+		}
+	}
```
SuperQ, line 1827: These metrics are unnecessary in Prometheus as we can compute averages from the counters. Please remove them to avoid excess metric cardinality.

```suggestion
```

metrics/prometheus.go
```diff
@@ -2060,3 +2118,23 @@ func getMinCoreScalingRatio(s *info.ContainerStats) metricValues {
...
+	case "avg60":
+		v = append(v, metricValue{value: psi.Some.Avg60, timestamp: s.Timestamp, labels: []string{"some"}})
+		v = append(v, metricValue{value: psi.Full.Avg60, timestamp: s.Timestamp, labels: []string{"full"}})
+	case "avg300":
+		v = append(v, metricValue{value: psi.Some.Avg300, timestamp: s.Timestamp, labels: []string{"some"}})
+		v = append(v, metricValue{value: psi.Full.Avg300, timestamp: s.Timestamp, labels: []string{"full"}})
+	case "total":
+		// total is measured as microseconds
+		v = append(v, metricValue{value: float64(time.Duration(psi.Some.Total)*time.Microsecond) / float64(time.Second), timestamp: s.Timestamp, labels: []string{"some"}})
+		v = append(v, metricValue{value: float64(time.Duration(psi.Full.Total)*time.Microsecond) / float64(time.Second), timestamp: s.Timestamp, labels: []string{"full"}})
```
SuperQ, line 2137: Note for CPU, we don't need to expose "full". In practice, I've only found the "some" metrics to be useful. The "some" value is a superset of "full". IMO we should just include that to reduce the cardinality.

See the [PSI docs](https://docs.kernel.org/accounting/psi.html):

> CPU full is undefined at the system level, but has been reported since 5.13, so it is set to zero for backward compatibility.

SuperQ (changes requested):

szuecs:
@bobbypage can you set ok to test and do a review?
Would be great to see this in cadvisor

metrics/prometheus.go
ein-stein-chen: I would like to raise my opinion for exposing "CPU full" metrics on a container level.

While "CPU full" is undefined at the system level, it reports values for other cgroups:
```console
$ cat /sys/fs/cgroup/user.slice/cpu.pressure
some avg10=0.00 avg60=0.05 avg300=0.20 total=68139509
full avg10=0.00 avg60=0.00 avg300=0.00 total=40148380
```

Exposing both might be useful to differentiate between cgroups that try to execute more work then compute time available to them (indicated through "CPU some") and cgroups which are fully blocked/stalled (maybe because of other cgroups and/or process priority, …; indicated through "CPU full").

SuperQ: Interesting, thanks for this info. Are there kernel docs that document the per-cgroup behavior?

IMO this should be a separate metric name, rather than a label. The reason is that since `some` is a inclusive of `full`, doing something like `sum(rate(container_pressure_cpu_seconds_total[1m]))` would be confusing.

I would suggest these two metric names:
* `container_pressure_cpu_seconds_total`
* `container_pressure_cpu_full_seconds_total`

metrics/prometheus.go
tomershafir: it depends on the use case, for example psi avg10 provides new info for a regular 1m scrape interval

bobbypage:
Waiting for new runc release to include https://github.com/opencontainers/runc/pull/3900/commits/1aa7ca804641ebd22b3abea68bf834a80d259cda

go.mod
```diff
@@ -33,8 +33,10 @@ require (
...
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110
-	golang.org/x/sys v0.0.0-20211116061358-0a5406a5449c
+	golang.org/x/sys v0.0.0-20220310020820-b874c991c1a5
 	google.golang.org/grpc v1.33.2
 	k8s.io/klog/v2 v2.4.0
 	k8s.io/utils v0.0.0-20211116205334-6203023598ed
 )
+
+replace github.com/opencontainers/runc => github.com/dqminh/runc v0.0.0-20220513155811-6414629ada8a
```
dims, line 42: Please remove this and use the upstream `runc` as-is.

zouyee:
https://github.com/opencontainers/runc/pull/4221

zouyee:
> @bobbypage not sure should I take over the PR? @dqminh thoughts?

@szuecs take over ?

SuperQ:
I would be happy to take over this change, I would like to make sure it aligns well with best practices. As it is, it does not.

szuecs:
There is only a release candidate so I would wait until there's a proper release.
I don't mind that someone else would do it.

akgoel18:
@szuecs Any update on this ? I'm asking this is a internally heavily requested feature. :)

szuecs:
@akgoel18 please check upstream project release cycle, thanks

pacoxu:
https://github.com/opencontainers/runc/pull/3900 is merged in runc 1.2.0

pacoxu:
https://github.com/opencontainers/runc/releases/tag/v1.2.0 runc v1.2.0 is released today.

Do you have time to bump runc?

dims:
@pacoxu when containerd 2.0 gets out runc used with it will be 1.2.0, so till we have only 1.6/1.7 containerd with k8s, we should stick to older runc(s) both as binary as well as vendoring

haircommander:
@dims why do cadvisor/k8s and containerd need to have an in sync runc version?

dims:
> @dims why do cadvisor/k8s and containerd need to have an in sync runc version?

bad things have happened before @haircommander example see https://github.com/opencontainers/runc/issues/3849

alexandremahdhaoui:
@dims @haircommander to address and prevent the issue with "out of sync" runc versions we could add a CI check.

NB: I wrote a [tool](https://github.com/kubernetes-sigs/controller-runtime/blob/main/hack/tools/cmd/gomodcheck/main.go) ([usage](https://github.com/kubernetes-sigs/controller-runtime/blob/main/.gomodcheck.yaml)) in kubernetes-sigs/container-runtime to verify and ensure specified go modules stays in sync with some upstream modules. I'd be glad to open a PR here or in k/k to add a check.

metrics/prometheus.go
rexagod: I believe this may be resolved since the `10/60/300` averages are [inherent](https://github.com/rexagod/runc/blob/edd00eb3cbd223ab565c1fed975a69260be26d02/libcontainer/cgroups/stats.go#L35-L40) to the exposed PSI data?

rexagod:
Since the [backports](https://github.com/containerd/containerd/pull/10877#issuecomment-2455335173) are in now, I think we can go ahead with this (and incorporating @SuperQ's suggestions above)?

SuperQ:
I might go ahead and make a fork/alternative to this specific PR.

enp0s3:
@SuperQ Hi, do you plan to take this over?

SuperQ:
I was hoping I would have time over the holidays. But ended up working on other projects. I could take a look soon. But as-is I don't think this PR should be merged.

xinau:
@SuperQ I would be happy to help on the development of this feature. Is it basically a rebase of this PR + resolving the review comments? Or is there more to be done?

rexagod:
@xinau Opening a PR with those suggestions should be a good starting point, any additional changes may be made incrementally upon review.

xinau:
I've opened #3649 with the requested rebase and addressed the review comments.

PR #3089, opened 2022-04-02
api/v2: document and consume count=-1

stevekuznetsov:
The underlying code already knows how to consume count=-1 when fetching
all data points. This patch allows the server to accept such a value
from the client and pass it through.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

cc @mrunalp

### DIFF
--- a/cmd/internal/api/versions.go
+++ b/cmd/internal/api/versions.go
@@ -547,11 +547,14 @@ func GetRequestOptions(r *http.Request) (v2.RequestOptions, error) {
 	}
 	count := r.URL.Query().Get("count")
 	if len(count) != 0 {
-		n, err := strconv.ParseUint(count, 10, 32)
+		n, err := strconv.Atoi(count)
 		if err != nil {
 			return opt, fmt.Errorf("failed to parse 'count' option: %v", count)
 		}
-		opt.Count = int(n)
+		if n < -1 {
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Ah, this has Prow!

/cc @mrunalp 
/assign @mrunalp

dims (approved):
LGTM

cmd/internal/api/versions.go
```diff
@@ -547,11 +547,14 @@ func GetRequestOptions(r *http.Request) (v2.RequestOptions, error) {
 	}
 	count := r.URL.Query().Get("count")
 	if len(count) != 0 {
-		n, err := strconv.ParseUint(count, 10, 32)
+		n, err := strconv.Atoi(count)
 		if err != nil {
 			return opt, fmt.Errorf("failed to parse 'count' option: %v", count)
```
ncdc, line 552: Worth wrapping the error with %w?

stevekuznetsov: Happy to do more work for this later, but this PR was an attempt to be surgical and just get the use-case working. I don't know that this HTTP API has a good set of tests as it is ...

mrunalp:
Looks fine to me.
@bobbypage ptal.

bobbypage:
LGTM

dims (approved):
LGTM

PR #3100, opened 2022-04-18
Adding allowed_device flag for diskIO Metrics

rushilenekar20:
Hey ,
I am suggesting a flag for diskIO metrics named as -allowed_diskIO_devices 

**Problem** : Requirement for only allowed-devices flag in diskIO.

**Description :**
In our organisation we faced a problem where Operating System create some devices at runtime whose names are unknown to us . So we were unable to ignored those devices .

As negetive lookahead is not supported in Go language because it conflicts with the O(n)-time guarantees of the library.so, we were unable to ignored those devices.

I suggest to have a flag for only allowed-devices for diskIO metrics .

Here are some screenshots of the solution for this problem

1) without using allowed_diskIO_device flag
Screenshot (13)

there are multiple device sda , dm , sr0 .Lets assume sda and sr0 devices were unknown to us and they are created at runtime and we dont want to show metrics of those devices . we only want to show metrics for dm devices.

2) allowed-device flag
Screenshot (12)

3) Use case
Screenshot (14)

only dm devices are allowed

here , i also want to know why it showing metrics for blank devices?

### DIFF
--- a/utils/sysinfo/sysinfo.go
+++ b/utils/sysinfo/sysinfo.go
@@ -15,6 +15,7 @@
 package sysinfo
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


rushilenekar20:
Hey @tallclair , Could you help me for this PR ?

PR #3103, opened 2022-04-23
Retry when we get a unknown status from containerd task API

dims:
Let's check if `response.Process.Status` is `tasktypes.Status_UNKNOWN` and ensure we retry under this condition as it is most probably a context timeout on containerd side as mentioned here:

https://github.com/kubernetes/kubernetes/pull/109608#issuecomment-1106653373

got some inspiration from:

https://github.com/containerd/containerd/blob/3633cae64b0dc854764f3f121d6739d5c05b5c2a/container.go#L388-L405

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -24,6 +24,7 @@ import (
 	containersapi "github.com/containerd/containerd/api/services/containers/v1"
 	tasksapi "github.com/containerd/containerd/api/services/tasks/v1"
 	versionapi "github.com/containerd/containerd/api/services/version/v1"
+	tasktypes "github.com/containerd/containerd/api/types/task"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/google/cadvisor/container/containerd/containers"
 	"github.com/google/cadvisor/container/containerd/errdefs"
@@ -113,6 +114,9 @@ func (c *client) TaskPid(ctx context.Context, id string) (uint32, error) {
 	})
 	if err != nil {
 		return 0, errdefs.FromGRPC(err)
+	}
+	if response.Process.Status == tasktypes.StatusUnknown {
+		return 0, errdefs.ErrUnknown
 	}
 	return response.Process.Pid, nil
 }


dims:
trying this in k/k WIP PR - https://github.com/kubernetes/kubernetes/pull/109628

dims:
holding steady 🟢 so far

image

dims:
24 green runs so far 

image

dims:
Ran 3 copies of this code in PRs against k/k:
- https://github.com/kubernetes/kubernetes/pull/109628
- https://github.com/kubernetes/kubernetes/pull/109629
- https://github.com/kubernetes/kubernetes/pull/109631

image

Added a separate config for the images used in cadvisor testing: 
- https://github.com/kubernetes/test-infra/pull/26072

Let's see if can land this.

container/containerd/client.go
```diff
@@ -114,6 +115,9 @@ func (c *client) TaskPid(ctx context.Context, id string) (uint32, error) {
 	if err != nil {
 		return 0, errdefs.FromGRPC(err)
 	}
+	if response.Process.Status == tasktypes.StatusUnknown {
```
SergeyKanzhelev, line 118: with this we are not getting into the fast backoff retry mechanism, we are failing to create a handler and re-attempt to create it at the next cycle. WOuld it be better to first retry in the small cycle inside the handler creation logic?

bobbypage:
Thank you Dims for the fix!

Just a quick question, just to confirm how does the retry logic work here in case we return error from `TaskPid`? Because AFAIU cAdvisor receives inotify events when the cgroup is created, and then tries to create the handler. But if handler creation fails, where does the retry come in if there are no more inotify events for ContainerAdd?

xref: https://github.com/google/cadvisor/blob/master/container/raw/watcher.go

container/containerd/client.go
dims: @SergeyKanzhelev return `ErrUnknown` here will end up with a retry in the loop here:

https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/google/cadvisor/container/containerd/handler.go#L96-L110

it loops there at a max of 5 times. before this PR we would not have looped at all.

bobbypage:
LGTM

container/containerd/client.go
bobbypage: That's a great point. Should we consider to remove the `!errdefs.IsNotFound(err)` check from https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/google/cadvisor/container/containerd/handler.go#L105 and retry on _all_ errors?

SergeyKanzhelev: I think this is the best course of action. Since the retry period is small, this should be fine.

dims:
closing in favor of https://github.com/google/cadvisor/pull/3105

PR #3105, opened 2022-04-25
Retry containerd task api request when task is in unknown state

bobbypage:
In some cases, it appears that containerd task api can return a zero pid
when the task is not ready yet. In these cases, the process task state
is reported to be in unknown state. In this case, we should retry the
containerd task api request as the task can take some time to move away
from unknown state and be initialized.

Signed-off-by: David Porter <porterdavid@google.com>

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -16,6 +16,7 @@ package containerd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sync"
@@ -24,6 +25,7 @@ import (
 	containersapi "github.com/containerd/containerd/api/services/containers/v1"
 	tasksapi "github.com/containerd/containerd/api/services/tasks/v1"
 	versionapi "github.com/containerd/containerd/api/services/version/v1"
+	tasktypes "github.com/containerd/containerd/api/types/task"
 	ptypes "github.com/gogo/protobuf/types"
 	"github.com/google/cadvisor/container/containerd/containers"
 	"github.com/google/cadvisor/container/containerd/errdefs"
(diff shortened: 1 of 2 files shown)


container/containerd/client.go
```diff
@@ -44,6 +46,10 @@ type ContainerdClient interface {
 	Version(ctx context.Context) (string, error)
 }
 
+var (
+	ErrTaskIsInUnknownState = errors.New("containerd task is in unknown state") // used when containerd task api is in nUnknownState
```
SergeyKanzhelev, line 50:
```suggestion
	ErrTaskIsInUnknownState = errors.New("containerd task is in unknown state") // used when containerd task api is in Unknown State
```

container/containerd/handler.go
```diff
@@ -101,10 +102,13 @@ func newContainerdContainerHandler(
 		if err == nil {
 			break
 		}
-		retry--
-		if !errdefs.IsNotFound(err) || retry == 0 {
+
+		isRetriableError := errdefs.IsNotFound(err) || errors.Is(err, ErrTaskIsInUnknownState)
```
SergeyKanzhelev, line 106:
```suggestion
                We retry is task wasn't yet created or when it was created, but not started yet
		isRetriableError := errdefs.IsNotFound(err) || errors.Is(err, ErrTaskIsInUnknownState)
```

SergeyKanzhelev (approved):
/lgtm

container/containerd/handler.go
bobbypage: Added comment!

container/containerd/client.go
bobbypage: Fixed, thanks!

mikebrow (approved):
LGTM

ruiwen-zhao:
LGTM

PR #3131, opened 2022-06-23
update go.mod to 1.18

akhilerm:
- update go.mod in root and cmd/ to 1.18
- update the dependencies in go.mod files

Signed-off-by: Akhil Mohan <makhil@vmware.com>


bobbypage:
Golang has to moved to 1.19, this is obsolete now.

akhilerm:
@bobbypage Shall I raise a PR to update the go.mod file  in the root and cmd/ directories? I see that in https://github.com/google/cadvisor/pull/3151 its not done.

bobbypage:
Sure, we can raise the version in the go.mod. I believe that is used as the "The minimum version of Go required by module" and I don't think we need to make it 1.19, likely 1.18 will work too.

akhilerm:
Ok. Currently the go.mod version is at 1.16. I think that needs to be updated to 1.18, if that works. Also the dependents also needs to be updated as in this PR.

bobbypage:
Sounds good to update go.mod.

I'm a bit careful to update dependencies, because we vendor cAdvisor in kubelet and there can be some issues if we always update to latest version of things in cadvisor. 

Is there anything we need to update specifically? If we are using the same dep as in k/k but an older version, it totally makes sense to update the dep to same version in cAdvisor as is used in k/k.

I would prefer avoid bumping deps to higher versions than k/k unless we have some need to.

PR #3134, opened 2022-07-08
[Cherrypick to release-0.39]  Retry containerd task api request when task is in unknown state

bobbypage:
Cherrypick https://github.com/google/cadvisor/pull/3105 to release-0.39

In some cases, it appears that containerd task api can return a zero pid
when the task is not ready yet. In these cases, the process task state
is reported to be in unknown state. In this case, we should retry the
containerd task api request as the task can take some time to move away
from unknown state and be initialized.

Signed-off-by: David Porter <porterdavid@google.com>

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -16,6 +16,7 @@ package containerd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sync"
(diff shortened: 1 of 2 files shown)


bobbypage:
pull-cadvisor-e2e is failing because it's using go1.18 on older release branch which used to use golang1.17

PR #3138, opened 2022-07-19
fix error type in load errno

BSWANG:
Fix invalid type error on kernel netlink return `errno` not zero. Should use *int32 to load `errno` value.

```
W0719 20:21:38.057540 3689835 container.go:586] Failed to update stats for container "/system.slice/kubelet.service": failed to get load stat for "/system.slice/kubelet.service" - path "/sys/fs/cgroup/system.slice/kubelet.service", error binary.Read: invalid type int32
```


aoxn:
+1

PR #3143, opened 2022-07-28
Support running integration tests on cgroupv2 hosts

bobbypage:
Adds a few fixes to the integration tests to support running on cgroupv2. Now running `make docker-test-integration` on a cgroupv2 host should work fine.

### DIFF
--- a/build/integration.sh
+++ b/build/integration.sh
@@ -72,6 +72,22 @@ while [ "$(curl -Gs http://localhost:8080/healthz)" != "ok" ]; do
   sleep 1
 done
 
+if [[ "${DOCKER_IN_DOCKER_ENABLED:-}" == "true" ]]; then
+  # see https://github.com/moby/moby/blob/master/hack/dind
+  # cgroup v2: enable nesting
+  if [ -f /sys/fs/cgroup/cgroup.controllers ]; then
+    echo ">> configuring cgroupsv2 for docker in docker..."
+    # move the processes from the root group to the /init group,
+    # otherwise writing subtree_control fails with EBUSY.
+    # An error during moving non-existent process (i.e., "cat") is ignored.
+    mkdir -p /sys/fs/cgroup/init
+    xargs -rn1 < /sys/fs/cgroup/cgroup.procs > /sys/fs/cgroup/init/cgroup.procs || :
+    # enable controllers
+    sed -e 's/ / +/g' -e 's/^/+/' < /sys/fs/cgroup/cgroup.controllers \
+      > /sys/fs/cgroup/cgroup.subtree_control
+  fi
+fi
+
 echo ">> running integration tests against local cAdvisor"
 if ! [ -f ./api.test ] || ! [ -f ./healthz.test ]; then
   echo You must compile the ./api.test binary and ./healthz.test binary before
--- a/integration/tests/api/docker_test.go
+++ b/integration/tests/api/docker_test.go
@@ -25,6 +25,7 @@ import (
 	v2 "github.com/google/cadvisor/info/v2"
 	"github.com/google/cadvisor/integration/framework"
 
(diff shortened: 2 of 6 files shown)


bobbypage:
@tstapler ptal

integration/tests/api/docker_test.go
```diff
@@ -211,7 +212,15 @@ func TestDockerContainerSpec(t *testing.T) {
 	assert := assert.New(t)
 
 	assert.True(containerInfo.Spec.HasCpu, "CPU should be isolated")
-	assert.Equal(cpuShares, containerInfo.Spec.Cpu.Limit, "Container should have %d shares, has %d", cpuShares, containerInfo.Spec.Cpu.Limit)
+	if cgroups.IsCgroup2UnifiedMode() {
+		// see https://github.com/google/cadvisor/blob/24e7a9883d12f944fd4403861707f4bafcaf4f3d/container/common/helpers.go#L249-L260
```
tstapler, line 216: Could you add more context to this comment. Is it pointing out how shares are calculated?

If so, perhaps replace `// see` with `// see how cpu shares are calculated`

build/integration.sh
```diff
@@ -56,6 +56,22 @@ while [ "$(curl -Gs http://localhost:8080/healthz)" != "ok" ]; do
...
+if [[ "${DOCKER_IN_DOCKER_ENABLED:-}" == "true" ]]; then
+  # see https://github.com/moby/moby/blob/master/hack/dind
+  # cgroup v2: enable nesting
+  if [ -f /sys/fs/cgroup/cgroup.controllers ]; then
+    echo ">> configuring cgroupsv2 for docker in docker..."
+    # move the processes from the root group to the /init group,
+    # otherwise writing subtree_control fails with EBUSY.
+    # An error during moving non-existent process (i.e., "cat") is ignored.
+    mkdir -p /sys/fs/cgroup/init
+    xargs -rn1 < /sys/fs/cgroup/cgroup.procs > /sys/fs/cgroup/init/cgroup.procs || :
```
tstapler, line 68: What does this xargs block do?

I can see that `-r` is do not run the command if there is no stdin, and `-n1 is max args 1?` but I don't see a command that its executing.

bobbypage: I copied this over from https://github.com/moby/moby/blob/master/hack/dind :)

So the idea here, is that for docker in docker for cgroupv2 we need to be move all the processes in a child cgroup or else we will hit the runc/cgroupv2 limitation mentioned here - https://github.com/containerd/containerd/issues/6659#issuecomment-1064754423

The goal here is to first create a new cgroup (`mkdir -p /sys/fs/cgroup/init`). Then we want to copy all of pids one by one from `/sys/fs/cgroup/cgroup.procs` to the new child cgroup `/sys/fs/cgroup/init/cgroup.procs` (the xargs `xargs -rn1` ensures it's split one by one). 

There is no command being passed to xargs, so as a result it will simply echo each pid (see https://unix.stackexchange.com/questions/110134/what-does-xargs-do-if-its-used-without-any-parameter)

integration/tests/api/docker_test.go
bobbypage: updated comment

bobbypage:
Updated based on your feedback, ptal again

build/integration.sh
tstapler: Neat, so xargs allows you to copy the file but throw an error if the source doesn't have any lines in it.

The new comment is helpful thanks!

integration/tests/api/docker_test.go
tstapler: What a beautiful comment :D, thanks.

tstapler (approved):
Thanks for adding the extra context

lgtm

PR #3158, opened 2022-08-17
chore: update go.mod to 1.18

akhilerm:
- update go.mod in root and cmd/ to 1.18
- versions of dependencies are not updated. only the indirect dependencies are added to the go.mod file

Signed-off-by: Akhil Mohan <makhil@vmware.com>


akhilerm:
PTAL @dims @bobbypage

akhilerm:
@bobbypage @dims Can you take a look at this PR?

bobbypage:
Sorry for delay, I don't fully understand why there are so many changes to the `go.mod` file? What set of commands were used to update it?

akhilerm:
Since go1.17, indirect dependencies are kept in separate `require` block. Ref: https://go.dev/doc/go1.17#go-command
Only `go mod tidy` has been run after updating the go version.

PR #3166, opened 2022-09-13
Update README.md

miavi:
A bug has been detected when running cadvisor on a debian 11 host and we want to retrieve "container_process"


bobbypage:
> A bug has been detected when running cadvisor on a debian 11 host

Why is this needed if we already mount sys? `--volume=/sys:/sys:ro`

miavi:
In debían 11 if no mount cfgroup(sys/fs/cgroup:/sys/fs/cgroup:ro )not
exposed procesos docker in cadvisor

In debían 10 working whitout mount sys/fs/cgroup:/sys/fs/cgroup:ro

everything seems to be a bug in the latest kernel debian versions.

PR #3178, opened 2022-09-26
fix some typos

cuishuang:


cmd/internal/http/handlers.go
```diff
@@ -31,7 +31,6 @@ import (
 	auth "github.com/abbot/go-http-auth"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"k8s.io/klog/v2"
```
bobbypage, line 34: Removing this breaks the build

PR #3194, opened 2022-11-04
[WIP] go.mod: update to new compute/metadata module

aramase:
Signed-off-by: Anish Ramasekar <anish.ramasekar@gmail.com>

The new module has 0 external dependencies which helps simplify dep trees. (xref: https://github.com/kubernetes/kubernetes/issues/113366)

Marking it WIP because we need to update transitive deps.


liggitt:
hmm, this pulls in an updated genproto that triggers the explosion in https://github.com/kubernetes/kubernetes/issues/113366... probably best to not merge this until that is resolved or it will be difficult for kubernetes to pick this up

aramase:
`master` branch is now using `compute/metadata` module https://github.com/google/cadvisor/blob/fbd519ba03978d54cb54ea7ed8ab9d6e3dd64590/go.mod#L8.

This PR is no longer required. Closing it!

PR #3200, opened 2022-11-21
fix: vulnerabilities in dependency package hints

fengshunli:
delete redundant code and fix high-risk vulnerabilities indicated @bobbypage 
image


cmd/go.mod
```diff
@@ -37,3 +37,5 @@ require (
 	k8s.io/klog/v2 v2.80.1
 	k8s.io/utils v0.0.0-20211116205334-6203023598ed
 )
+
+replace github.com/aws/aws-sdk-go => github.com/fengshunli/aws-sdk-go v0.0.0-20221121041908-199b8c2a907a
```
bobbypage, line 41: I don't think we should point to a fork, is there a new aws-sdk-go version we should update to instead?

fengshunli: aws-sdk-go did not merge the pr I submitted, and all the latest versions will also have this problem. I made a transformation based on the code of aws-sdk-go see https://github.com/aws/aws-sdk-go/pull/4557

bobbypage: Please work with aws-sdk to fix the issue, we should only use the official module, not forks. Additionally, CVE-2020-8911 does not impact cAdvisor, it is specific to s3 crypto based on https://github.com/advisories/GHSA-f5pg-7wfw-84q9 which is not used by cAdvisor.

PR #3201, opened 2022-11-22
Migrate to AWS Go SDK v2

skitt:
Version 2 of the SDK splits features across different Go modules. See
https://aws.github.io/aws-sdk-go-v2/docs/migrating/ for the migration
guide.

Signed-off-by: Stephen Kitt <skitt@redhat.com>


dims:
@skitt landing this in cadvisor will make it hard to integrate a new version of cadvisor in kubernetes. Unless of course we have a draft PR in k8s that can do the same change in dependency in kubernetes as well. (we don't want to have 2 versions of a cloud provider library/dependency in k8s).

skitt:
> @skitt landing this in cadvisor will make it hard to integrate a new version of cadvisor in kubernetes. Unless of course we have a draft PR in k8s that can do the same change in dependency in kubernetes as well. (we don't want to have 2 versions of a cloud provider library/dependency in k8s).

Ah yes, good point… I’ll look into the k8s situation and report back.

skitt:
@dims now that the AWS dependency is gone from K8s, this should be OK; is it?

dims:
> @dims now that the AWS dependency is gone from K8s, this should be OK; is it?

yep!

iwankgb:
@skitt, you may need to merge `master` here.

skitt:
> @skitt, you may need to merge `master` here.

Done, I’ve included #3553 to be able to build locally, but I think that warrants a separate review.

iwankgb:
You need to run `go mod tidy`, I believe (https://github.com/google/cadvisor/actions/runs/9747285841/job/26899619315?pr=3201)

skitt:
> You need to run `go mod tidy`, I believe (https://github.com/google/cadvisor/actions/runs/9747285841/job/26899619315?pr=3201)

I do indeed, in `cmd`, thanks for the reminder!

iwankgb (approved):

PR #3206, opened 2022-12-01
Drop accelerator metrics and nvidia implementation

liggitt:
This capability is unused as of Kubernetes 1.25+ (see https://github.com/kubernetes/kubernetes/pull/114204#issuecomment-1332756032 and https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1867-disable-accelerator-usage-metrics)

That means nvidiaManager was always a no-op implementation in Kubernetes 1.25+

This drops the nvidia implementation, collector, and AcceleratorUsageMetrics stats type, which also allows dropping the github.com/mindprince/gonvml dependency, which some license scanners were unhappy with (xref https://groups.google.com/g/kubernetes-sig-architecture/c/dsMVIdPPUK8/m/u3ZjJtcnBwAJ?utm_medium=email&utm_source=footer)

cc @dashpole @bobbypage


bobbypage:
Thanks @liggitt, I agree it makes sense that we can drop accelerator metrics after k8s has deprecated in-tree accelerator metrics and drop this dependency. 

For cAdvisor users, we can recommend users to use https://github.com/NVIDIA/dcgm-exporter for accelerator metrics.

bobbypage (approved):

liggitt:
thanks, let's plan to tag cadvisor and bump in k/k early in the 1.27 cycle

bobbypage:
Sounds, good, will do!

Issue #3207, opened 2022-12-02
Memory.Reservation is using `memory.high` but kube is using `memory.min` as request


pacoxu:
https://github.com/google/cadvisor/blob/a52ec5d60cf70b22f8b6d204780aec7a222cf6bb/container/common/helpers.go#L175

Below is logic of Kuberentes. The KEP is https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2570-memory-qos.
```
		memoryRequest := container.Resources.Requests.Memory().Value()
		memoryLimit := container.Resources.Limits.Memory().Value()
		if memoryRequest != 0 {
			unified[cm.MemoryMin] = strconv.FormatInt(memoryRequest, 10)
		}

		// If container sets limits.memory, we set memory.high=pod.spec.containers[i].resources.limits[memory] * memory_throttling_factor
		// for container level cgroup if memory.high>memory.min.
		// If container doesn't set limits.memory, we set memory.high=node_allocatable_memory * memory_throttling_factor
		// for container level cgroup.
		memoryHigh := int64(0)
		if memoryLimit != 0 {
			memoryHigh = int64(float64(memoryLimit) * m.memoryThrottlingFactor)
		} else {
			allocatable := m.getNodeAllocatable()
			allocatableMemory, ok := allocatable[v1.ResourceMemory]
			if ok && allocatableMemory.Value() > 0 {
				memoryHigh = int64(float64(allocatableMemory.Value()) * m.memoryThrottlingFactor)
			}
		}
		if memoryHigh > memoryRequest {
			unified[cm.MemoryHigh] = strconv.FormatInt(memoryHigh, 10)
		}
```
https://github.com/pacoxu/kubernetes/blob/f0d1a39e8f22e5dfbaf601832650375b6001a43e/pkg/kubelet/kuberuntime/kuberuntime_container_linux.go#L104-L126

PR #3208, opened 2022-12-02
use memory.min for reservation memory instead of high

pacoxu:
high is throttling bar

Fixes https://github.com/google/cadvisor/issues/3207

File | Description
-- | --
memory.min | memory.min specifies a minimum amount of memory the cgroup must always retain, i.e., memory that can never be reclaimed by the system. If the cgroup's memory usage reaches this low limit and can’t be increased, the system OOM killer will be invoked. We map it to requests.memory.
memory.max | memory.max is the memory usage hard limit, acting as the final protection mechanism: If a cgroup's memory usage reaches this limit and can't be reduced, the system OOM killer is invoked on the cgroup. Under certain circumstances, usage may go over the memory.high limit temporarily. When the high limit is used and monitored properly, memory.max serves mainly to provide the final safety net. The default is max. We map it to limits.memory as consistent with existing memory.limit_in_bytes for cgroups v1.
memory.low | memory.low is the best-effort memory protection, a "soft guarantee" that if the cgroup and all its descendants are below this threshold, the cgroup's memory won't be reclaimed unless memory can’t be reclaimed from any unprotected cgroups. Not yet considered for now.
memory.high | memory.high is the memory usage throttle limit. This is the main mechanism to control a cgroup's memory use. If a cgroup's memory use goes over the high boundary specified here, the cgroup’s processes are throttled and put under heavy reclaim pressure. The default is max, meaning there is no limit. We use a formula to calculate memory.high depending on limits.memory/node allocatable memory and a memory throttling factor.

This was added in #2309.
/cc @dashpole  @giuseppe


giuseppe (approved):
LGTM

pacoxu:
The test failure is fixed.

bobbypage:
kick CI

bobbypage:
LGTM on the change

bobbypage:
CI is being a bit weird

```
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
+ delete
Deleting /tmp/tmp.NnvwDGEsX6...
+ echo 'Deleting /tmp/tmp.NnvwDGEsX6...'
+ [[ 1001 -ne 0 ]]
+ sudo rm -rf /tmp/tmp.NnvwDGEsX6
make: *** [Makefile:45: docker-test-integration] Error 1
```

bobbypage:
Can you please rebase the PR against master? https://github.com/google/cadvisor/pull/3224 should fix the CI issue

pacoxu:
> Can you please rebase the PR against master? #3224 should fix the CI issue

Rebased, the failed test passes now.

bobbypage:
Thanks!

bobbypage (approved):

PR #3209, opened 2022-12-05
Add support for ppc64le architecture

mkumatag:
This PR will add support for the linux/ppc64le OS/architecture


bobbypage:
I would prefer we avoid adding new architectures that we can't easily test in CI. Otherwise, when we eventually do a release we may find an architecture does not build but it will only be appear during release time.

mkumatag:
> I would prefer we avoid adding new architectures that we can't easily test in CI. Otherwise, when we eventually do a release we may find an architecture does not build/work but it will only be appear during release time.

one another way is to include the image build step in the PR to confirm it builds for all the architectures we are shipping, I have added a GitHub action for the same but failing some DNS issue needs more debugging.

PR #3213, opened 2022-12-11
Update dependencies - go.mod/sum both in root and cmd/ directories

dims:


dims:
We haven't done an across the board updates in a while. So proposing this PR. All our tests here work fine as well as k/k.

bobbypage:
Thanks Dims for update and testing it in k/k

bobbypage (approved):

PR #3239, opened 2023-01-23
Change cAdvisor version in Readme.md and running.md

cornelius-99:
Simply copy pasting the command provided in README.md and running.md did not work for me.
Would be easier for other people wo try out cAdvisor to have a running copy paste command


Creatone:
As the v0.47.0 image is missing some features due to issue #3237, I would rather wait for a fixed release and put it in the README.

cornelius-99:
@Creatone Is it ready to be merged now?

PR #3244, opened 2023-02-15
detect hung filesystems

tzneal:
I'm curious on thoughts for this approach.  I've been looking at https://github.com/kubernetes/kubernetes/issues/101622 and this change plus a few on the kubelet side are enough to improve things so that kubelet doesn't hang on startup or go NotReady when an NFS filesystem in use by a pod hangs.

### DIFF
--- a/utils/path.go
+++ b/utils/path.go
@@ -14,11 +14,35 @@
 
 package utils
 
-import "os"
+import (
+	"context"
+	"errors"
+	"os"
+	"os/exec"
+	"time"
+)
 
+// FileExists returns true if the file exists
 func FileExists(file string) bool {
 	if _, err := os.Stat(file); err != nil {
 		return false
 	}
 	return true
+}
+
+// FilesystemHung returns true if the filesystem appears to be hung.  This is determined by running the `stat` command
+// against the path provided and waiting up to the specified timeout for the command to complete.  If it does not
+// complete, the command is killed and the function returns false. If the `stat` command is not found on the system,
+// the function always returns false.
+func FilesystemHung(path string, timeout time.Duration) bool {
+	stat, err := exec.LookPath("stat")
+	if err != nil {
+		return false
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// we don't actually care if the command suceeds or fails, just if it hangs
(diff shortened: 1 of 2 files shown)


utils/path.go
```diff
@@ -14,11 +14,38 @@
...
 	}
 	return true
 }
+
+// FilesystemHung returns true if the filesystem appears to be hung.  This is determined by running the `stat` command
+// against the path provided and waiting up to the specified timeout for the command to complete.  If it does not
+// complete, the command is killed and the function returns false. If the `stat` command is not found on the system,
+// the function always returns false.
+func FilesystemHung(path string, timeout time.Duration) bool {
+	stat, err := exec.LookPath("stat")
```
tzneal, line 38: I thought about doing a `sync.Once` one-time check for stat and logging if it doesn't exist to avoid the `exec.LookPath` every time and make users aware that hung filesystem checking is disabled if `stat` is missing.

dims: Could we do this programatically instead of calling an external binary?

https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/hostutil/hostutil_linux.go#L286-L290

tzneal: It doesn't, I wrote it so if it's missing it will continue to work as is.  An external binary is needed because AFAIK that's the only good way to detect this.  If you use a system call from within cadvisor, it blocks and you can't stop it.  If you launch a goroutine, you end up leaking them as they don't return and I *think* each will also consume an OS thread since its blocked on a syscall.  With the external process, the SIGKILL will interrupt the hung system call.

tzneal: > Could we do this programatically instead of calling an external binary?

Not that I'm aware of, could attempt to catch signals and signal ourself, but it seems error prone.  The intent is to move the hang to a process that we can then interrupt.

dims: you are right... now i remember

utils/path.go
```diff
@@ -14,11 +14,35 @@
...
+	if err != nil {
+		return false
```
dims, line 40: So here ... if `stat` executable is missing, we are declaring that the file system is hung... is that right?

tzneal: No, if it's missing err will be non-nil causing FilesystemHung to always return false.

dims: 👍🏾

PR #3246, opened 2023-02-18
Enabling integration tests on armv7 and aarch64

iwankgb:
I would like to have tests running for `aarch64` and `armv7`.


iwankgb:
Running unit test takes *ages* and the list of executed tests seems to be extremely short. If I can't find a way for caching images, then I will close the PR 🤷

.github/workflows/test.yml
```diff
@@ -47,7 +47,50 @@ jobs:
         make docker-test-integration
     - name: Upload cAdvisor log file
       uses: actions/upload-artifact@v1
-      if: failure()
+      if: success() || failure()
       with:
-        name: cadvisor.log
-        path: ${{ github.workspace }}/go/src/github.com/google/cadvisor/cadvisor.log
+        name: cadvisor-${{ matrix.distro }}-${{ matrix.arch }}-${{ matrix.go-versions}}-${{ github.run_number }}-${{ github.run_id }}.log
```
Creatone, line 52: There is no `matrix.distro` and `matrix.arch` in `test-integration` job.
This leads to e.g.
```
Uploading artifact 'cadvisor---1.20-1899-4410275830.log' from '/home/runner/work/cadvisor/cadvisor/cadvisor.log' for run #1793
```
Please provide arch variable. Could you use platform instead of distro?

Creatone: Why do you want to keep logs when everything seems to be working?

.github/workflows/test.yml
```diff
@@ -47,7 +47,50 @@ jobs:
...
+        path: ${{ github.workspace }}/cadvisor.log
+  test-integration-non-amd64:
+    strategy:
+      matrix:
+        go-versions: [ 1.19, 1.18 ]
+        arch: [ aarch64, armv7 ]
+        distro: [ ubuntu20.04 ]
```
Creatone, line 59: Could you use the `platform` variable like in test, and test-integration jobs?
Or even stay with distro but let's have it the same for all jobs :)

.github/workflows/test.yml
```diff
@@ -47,7 +47,50 @@ jobs:
...
+      if: success() || failure()
       with:
-        name: cadvisor.log
-        path: ${{ github.workspace }}/go/src/github.com/google/cadvisor/cadvisor.log
+        name: cadvisor-${{ matrix.distro }}-${{ matrix.arch }}-${{ matrix.go-versions}}-${{ github.run_number }}-${{ github.run_id }}.log
+        path: ${{ github.workspace }}/cadvisor.log
+  test-integration-non-amd64:
+    strategy:
+      matrix:
+        go-versions: [ 1.19, 1.18 ]
```
Creatone, line 57: Other jobs use 1.20 and 1.19. Why this scenario is different?

iwankgb: It's just old :P

PR #3257, opened 2023-02-25
Replacing readlink with io_uring in processStatsFromProcs()

### LINKED ISSUE #3233: processStatsFromProcs High CPU Usage with High Number of File Descriptors
On our nodes with a large number of pods (in excess of 150 in most cases) which may exhibit a large number of open file descriptors, we see that the kubelet running on these nodes is often using anywhere between 0.5-2 threads out of 24.

When flame-graphing the kubelet process we can see that the majority of the time is spent in the `processStatsFromProcs` function:

kubelet-flame-graph

This is due to the fact that the function performs a system call per `/proc/<pid>/fd` directory (readdir) and a system call per file descriptor listed in this directory (readlink). If there are a large number of open file descriptors and a large number of process running on these nodes, the overheads of these system calls add a lot of impact to the process.

Is there a way to optimise this function to reduce the overall CPU usage of the kubelet?

iwankgb:
Fixes #3233.

The idea is to replace `readlink()` with approach that would utilise much smaller number of syscalls.


iwankgb:
`readlink` is not supported `io_uring` operation.

Issue #3263, opened 2023-03-04
resctrl system remove mode&size check


Rouzip:
In the current resctrl system, we have files called `mode` and `size` like below which we don't need to check. So we can remove useless check in `findGroup`.

![resctrl file system](https://s1.ax1x.com/2023/03/03/ppANoWQ.png)

PR #3264, opened 2023-03-04
reduce useless check in resctrl file system

Rouzip:
Fix https://github.com/google/cadvisor/issues/3263

In the resctrl file system, there will be automatically created mode and size files, so we can remove the useless check for them in the `findGroup` function.


Rouzip:
Hello, can you help review my pr? @Creatone

Creatone:
Is it ready for review?

PR #3278, opened 2023-03-22
fix container_oom_events_total always returns 0.

### LINKED ISSUE #3015: container_oom_events_total always returns 0
Running Docker (swarm), when OOM events occur the counter never increases. For reference, the node-exporter metric (node_vmstat_oom_kill) does increase.

Running cAdvisor v0.43.0.

chengjoey:
fix #3015 
In a Kubernetes pod, if a container is OOM-killed, it will be deleted and a new container will be created. Therefore, the `container_oom_events_total` metric will always be 0. this pr refactor the collector of oom events, and retain the deleted container oom information for a period of events. And add flag `oom_event_retain_time` to decide how long the oom event will be keep, default is 5 minutes

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -37,6 +37,7 @@ import (
 	info "github.com/google/cadvisor/info/v1"
 	v2 "github.com/google/cadvisor/info/v2"
 	"github.com/google/cadvisor/machine"
+	"github.com/google/cadvisor/metrics"
 	"github.com/google/cadvisor/nvm"
 	"github.com/google/cadvisor/perf"
 	"github.com/google/cadvisor/resctrl"
@@ -144,6 +145,8 @@ type Manager interface {
 	AllPodmanContainers(c *info.ContainerInfoRequest) (map[string]info.ContainerInfo, error)
 
 	PodmanContainer(containerName string, query *info.ContainerInfoRequest) (info.ContainerInfo, error)
+
+	GetOOMInfos() map[string]*oomparser.ContainerOomInfo
 }
 
 // Housekeeping configuration for the manager
@@ -153,7 +156,9 @@ type HousekeepingConfig = struct {
 }
 
 // New takes a memory storage and returns a new manager.
-func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, HousekeepingConfig HousekeepingConfig, includedMetricsSet container.MetricSet, collectorHTTPClient *http.Client, rawContainerCgroupPathPrefixWhiteList, containerEnvMetadataWhiteList []string, perfEventsFile string, resctrlInterval time.Duration) (Manager, error) {
+func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, HousekeepingConfig HousekeepingConfig, includedMetricsSet container.MetricSet,
+	collectorHTTPClient *http.Client, rawContainerCgroupPathPrefixWhiteList, containerEnvMetadataWhiteList []string,
+	perfEventsFile string, resctrlInterval time.Duration, f metrics.ContainerLabelsFunc, oomRetainDuration *time.Duration) (Manager, error) {
 	if memoryCache == nil {
 		return nil, fmt.Errorf("manager requires memory storage")
 	}
@@ -208,6 +213,9 @@ func New(memoryCache *memory.InMemoryCache, sysfs sysfs.SysFs, HousekeepingConfi
 		collectorHTTPClient:                   collectorHTTPClient,
 		rawContainerCgroupPathPrefixWhiteList: rawContainerCgroupPathPrefixWhiteList,
 		containerEnvMetadataWhiteList:         containerEnvMetadataWhiteList,
+		oomInfos:                              map[string]*oomparser.ContainerOomInfo{},
+		containerLabelFunc:                    f,
+		oomRetainDuration:                     oomRetainDuration,
 	}
 
 	machineInfo, err := machine.Info(sysfs, fsInfo, inHostNamespace)
@@ -247,6 +255,7 @@ type namespacedContainerName struct {
 }
 
 type manager struct {
+	oomInfos                 map[string]*oomparser.ContainerOomInfo
 	containers               map[namespacedContainerName]*containerData
 	containersLock           sync.RWMutex
 	memoryCache              *memory.InMemoryCache
@@ -271,6 +280,8 @@ type manager struct {
 	rawContainerCgroupPathPrefixWhiteList []string
 	// List of container env prefix whitelist, the matched container envs would be collected into metrics as extra labels.
 	containerEnvMetadataWhiteList []string
+	containerLabelFunc            metrics.ContainerLabelsFunc
+	oomRetainDuration             *time.Duration
 }
 
 func (m *manager) PodmanContainer(containerName string, query *info.ContainerInfoRequest) (info.ContainerInfo, error) {
@@ -318,7 +329,7 @@ func (m *manager) Start() error {
 		return err
 	}
 	klog.V(2).Infof("Starting recovery of all containers")
-	err = m.detectSubcontainers("/")
+	err = m.detectSubContainers("/")
 	if err != nil {
 		return err
 	}
@@ -340,6 +351,7 @@ func (m *manager) Start() error {
 	quitUpdateMachineInfo := make(chan error)
 	m.quitChannels = append(m.quitChannels, quitUpdateMachineInfo)
 	go m.updateMachineInfo(quitUpdateMachineInfo)
+	go m.cleanUpOomInfos()
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


chengjoey:
/assign @iwankgb @kragniz 

Is it feasible to keep the container metrics with oomkilled without deleting them?
please take a look

iwankgb:
What happens when PID 1 forks another process and the forked process get OOM-killed?

chengjoey:
> What happens when PID 1 forks another process and the forked process get OOM-killed?

The forked process that was OOM-killed can still read relevant log information from /dev/kmsg. It should still be possible to associate with the corresponding container.

szuecs:
@chengjoey what happens if a container is killed every second?
Do I understand correctly that we keep creating new container metrics or would this counter increase to >1?
In case we would create new container metrics, then I would call this feature a memory leak, because every container start will create a new set of metrics and now we would store them forever as far as I understand. (I am not very familiar with the code)

ishworgurung:
> memory leak [ ... ]

@szuecs In what way would it be a _memory leak_ ?

szuecs:
@ishworgurung maybe the wording is not correct, but it will increase memory overtime, which is never GCed and finally cadvisor get oom. As far as I understand.
Increasing the counter is great, though. Having old metrics forever is likely an issue.
Maybe it's also not part of this PR, feel free to ignore this.

chengjoey:
> @ishworgurung maybe the wording is not correct, but it will increase memory overtime, which is never GCed and finally cadvisor get oom. As far as I understand. Increasing the counter is great, though. Having old metrics forever is likely an issue. Maybe it's also not part of this PR, feel free to ignore this.

hi @szuecs @ishworgurung , I have made modifications in this PR, putting the oom event metric information in a separate map, and adding the flag `oom_event_retain_time` to configure the retention time. Oom metric that exceeds this time will still be deleted to prevent memory leaks.

@iwankgb could you please task a review when you have time

dims:
@chengjoey please resolve merge conflicts:
image

chengjoey:
> @chengjoey please resolve merge conflicts: image

Thanks @dims , pr has been rebased

taraspos:
Hey!
just wanted to bump this issue again. Would be great to get it fixed.

tsipo:
Hi @pschichtel and others.
I was testing [this issue](https://github.com/google/cadvisor/issues/3015) on Kubernetes using a small image I have built - see [here](https://github.com/densify-dev/oomkill). I have noticed that when I use that tool in a **forked** mode - [this](https://github.com/densify-dev/oomkill/blob/main/examples/pod-oom-forked.yaml) and [this](https://github.com/densify-dev/oomkill/blob/main/examples/pod-sidecar-oom-forked.yaml) use-cases, I go get `container_oom_events_total == 1` as the container continues to live enough time after the OOMKill for `cAdvisor` to be scraped. If I change the `AFTER_FORK_INTERVAL` env var to 0, and the container exits immediately, I gon't get `container_oom_events_total == 1` as the container is de-registered immediately from `cAdvisor`.
FYI.

frittentheke:
@dims @iwankgb @kragniz may I kindly ask for an update on this issue?
Monitoring OOM events seems to be in a quite confusing slightly messy state ...

I was discussion the monitoring aspects of OOM events in Kubernetes with e.g. @dgrisonnet in https://github.com/kubernetes/kubernetes/issues/69676#issuecomment-2250844336 and was pointed at https://github.com/kubernetes/kubernetes/pull/108004 which supposedly enabled Kubelet to expose the OOM metrics from Cadvisor. But in the end those need to be available, which, looking at this very issue / PR here, seems to be buggy still?

chengjoey:
now, the main reason is probably the container OOM. Kubelet will kill the container and create a new one, so the historical OOM events are lost. I think the current PR idea is correct. Maybe there are some things I missing.

frittentheke:
@dims may I kindly ask again if there is any way forward with this PR? Google's Autopilot paper (https://research.google/pubs/autopilot-workload-autoscaling-at-google-scale/) talks at length about how OOM events are used to rightsize application resource requests - so having Kubernetes do it's best to provide these metrics and events should be obvious.
Please kindly also see my references in https://github.com/google/cadvisor/pull/3278#issuecomment-2292982675.

manager/manager.go
```diff
@@ -1247,7 +1314,9 @@ func (m *manager) watchForNewOoms() error {
 				continue
 			}
 			for _, cont := range conts {
-				atomic.AddUint64(&cont.oomEvents, 1)
+				if err := m.addOrUpdateOomInfo(cont, oomInstance.TimeOfDeath); err != nil {
+					klog.Errorf("failed to add OOM info for %q: %v", oomInstance.ContainerName, err)
+				}
```
dgrisonnet, line 1319: This is dependent on the existence of container: https://github.com/google/cadvisor/blob/0b6dfebf06a55e606935c6aad60b20b49b846bdc/manager/manager.go#L1307-L1311

If we want to address the current dependency issue of the metric on the container lifecycle, I think we shouldn't rely on the containers seen by the manager at all and rather just produce a metrics based on the info that cadvisor parses via the oom parser.

metrics/prometheus.go
```diff
@@ -1832,24 +1822,21 @@ func (c *PrometheusCollector) collectContainersInfo(ch chan<- prometheus.Metric)
 		}
 	}
 
+	if c.includedMetrics.Has(container.OOMMetrics) {
+		for _, oomInfo := range c.infoProvider.GetOOMInfos() {
+			labels, values := genLabelValues(rawLabels, oomInfo.MetricLabels)
```
dgrisonnet, line 1827: rawLabels also depends on the presence of the container. Instead of using the default labels that we can't always rely on with OOM kills, it might be better to define new labels specific to the OOM metrics

manager/manager.go
pschichtel: shouldn't it be possible to track the necessary information before containers are killed to have them available independent of the container lifecycle?

dgrisonnet: It is possible but that would require to cache all the container data.
For a separate reason, there was an attempt to create one in https://github.com/google/cadvisor/pull/2974, but I don't know if any of the maintainers would be willing to pull on the trigger on such a big change. At least it wasn't conceivable in the other PR due to the memory implications of such a change.
Also Kubernetes is and has been to trying to move away from cadvisor in favor of [container-runtime metrics](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2371-cri-pod-container-stats/README.md) which doesn't go in the favor of such an investment in cadvisor.

pschichtel: > It is possible but that would require to cache all the container data.

true, but not _all_ container data, just the bare-minimum to be able to produce useful metrics

fabiand:
Anything we can do in order to move this forward?

frittentheke:
There is at least some movement in regards to OOM detection and parsing ... (https://github.com/google/cadvisor/issues/3715) @jesseward

fabiand:
Off-topic @frittentheke your's is one of the best nick's I've seen in a while.

dims:
needs a rebase please

ringerc:
Would it be feasible to consume the `containerd` event stream to get this information and increment the counter?

If I tail `crictl events` and filter it with

```
jq '
  .
  |select(.containerEventType=="CONTAINER_STOPPED_EVENT")
  |del(.podSandboxStatus)
  |.containersStatuses|=[
    .[]
    |select(.reason=="OOMKilled")
    |del(.mounts,.user,.imageId,.resources,.logPath,.annotations,.imageRef,.image)
  ]
'
```

produces events like

```
{
  "containerId": "dbc4b8a02e5b1bf90229b82500b4b46006e168487870e14f1a7f778a4b2a7cbd",
  "containerEventType": "CONTAINER_STOPPED_EVENT",
  "createdAt": "1767061346879749884",
  "containersStatuses": [
    {
      "id": "ce9e36fedc7c0b5b6b435b267cac397c53d4a7d3df3bb2154cd9de7e905d9bfe",
      "metadata": {
        "name": "oomtest",
        "attempt": 1
      },
      "state": "CONTAINER_EXITED",
      "createdAt": "1767061323866327412",
      "startedAt": "1767061324060497453",
      "finishedAt": "1767061325952232903",
      "exitCode": 137,
      "reason": "OOMKilled",
      "message": "",
      "labels": {
        "io.kubernetes.container.name": "oomtest",
        "io.kubernetes.pod.name": "oomtest-86889bb6d5-tnrqh",
        "io.kubernetes.pod.namespace": "default",
        "io.kubernetes.pod.uid": "dd9e8754-f83a-4eab-b242-cee7138411b2"
      }
    },
    {
      "id": "dbc4b8a02e5b1bf90229b82500b4b46006e168487870e14f1a7f778a4b2a7cbd",
      "metadata": {
        "name": "oomtest",
        "attempt": 2
      },
      "state": "CONTAINER_EXITED",
      "createdAt": "1767061344714647928",
      "startedAt": "1767061344889964019",
      "finishedAt": "1767061346792728130",
      "exitCode": 137,
      "reason": "OOMKilled",
      "message": "",
      "labels": {
        "io.kubernetes.container.name": "oomtest",
        "io.kubernetes.pod.name": "oomtest-86889bb6d5-tnrqh",
        "io.kubernetes.pod.namespace": "default",
        "io.kubernetes.pod.uid": "dd9e8754-f83a-4eab-b242-cee7138411b2"
      }
    }
  ]
}
```

when run with this test deployment spec https://github.com/google/cadvisor/issues/3015#issuecomment-3698036945 

This should provide the information `containerd` needs to maintain a counter, right?

IMO it doesn't actually make sense to dimension this metric with the container's `id` at all; that makes it difficult to query, as it's necessary to compute the increase of all the counters, then sum them. Instead, this metric's current name should be deprecated, and a new metric defined with dimensions `pod, container`, omitting the `id` and `name` that vary per-container-instance. The event stream has labels for `io.kubernetes.pod.name` and `io.kubernetes.container.name`, so that should be sufficient to associate them with a metric owned by the `Pod` and dimensioned by container-name.

ahippler:
Hi. Any chance that this will be fixed? I would really like to use the metric.

frittentheke:
@dims may I kindly ask what the reason for closing this one was?
Is this already addressed elsewhere?

* https://github.com/google/cadvisor/pull/3889
* https://github.com/google/cadvisor/pull/3716

the state of OOM events is still puzzling to me.

Issue #3281, opened 2023-03-24
How to close UI web page of cadvisor service?


wuyanfeiwork:
1. How to close the UI web page of cadvisor service?

2. In addition, I want to allow only Prometheus to access the `/metrics` address after closing the UI web page. Is there any related option to configure?

wuyanfeiwork:
As shown in the figure below, how to close this UI interface?

![image](https://user-images.githubusercontent.com/104751498/227775981-8472fef0-0aba-42ac-9a48-45827a344406.png)

chengjoey:
https://github.com/google/cadvisor/blob/b4c955f149c0ed838998c486aeac0cf87f66fa3b/cmd/cadvisor.go#L47-L76

there are all cadvisor flags, it seems that there is no option to close the UI interface.

chengjoey:
I try to add a flag to control whether to enable webUI, and edit cadvisor daemonset container command like this:
![image](https://user-images.githubusercontent.com/30427474/230322665-e3f95310-fb07-4764-a195-041b4989c13c.png)

after set flag `disable_web_ui` to true,The final effect is:

webUI is not found:
![image](https://user-images.githubusercontent.com/30427474/230323070-dce2a7e2-c098-4dfc-b301-73c403084c72.png)

The metrics route works fine:
![image](https://user-images.githubusercontent.com/30427474/230323270-a5ca3b2e-65a9-40eb-8c74-19deb150b0ee.png)


@wuyanfeiwork Is this you want?

wuyanfeiwork:
> I try to add a flag to control whether to enable webUI, and edit cadvisor daemonset container command like this: ![image](https://user-images.githubusercontent.com/30427474/230322665-e3f95310-fb07-4764-a195-041b4989c13c.png)
> 
> after set flag `disable_web_ui` to true,The final effect is:
> 
> webUI is not found: ![image](https://user-images.githubusercontent.com/30427474/230323070-dce2a7e2-c098-4dfc-b301-73c403084c72.png)
> 
> The metrics route works fine: ![image](https://user-images.githubusercontent.com/30427474/230323270-a5ca3b2e-65a9-40eb-8c74-19deb150b0ee.png)
> 
> @wuyanfeiwork Is this you want?
Yes, this effect can also meet the needs, thank you.

wuyanfeiwork:
> 

Oh, you started cadvisor in k8s, while I tested it using command line in the terminal and didn't use the `--disable_web_ui=true` option.

[...]

PR #3289, opened 2023-04-06
chore: support disable web ui

chengjoey:
In many cases, it is only necessary to collect metrics, and the webUI is not needed, so is it possible to add a switch to support turning off the webUI?


chengjoey:
#3281

iwankgb (changes requested):
If you set both `enableProfiling` and `disableWebUI` then you gonna end up without profiling capabilities despite of enabling them. I don't think it is expected behavior.

PR #3301, opened 2023-05-03
Request the pid of another container if current pid is not longer valid

harche:
When using CRI-O as the container runtime, the infrastructure container can be dropped. In this scenario, CRI-O provides cadvisor with the process ID (PID) of one of the containers within the pod to collect network statistics. However, if the particular container exits, cadvisor is left with an invalid PID. To address this issue, this PR makes cadvisor to switch to another PID from the same pod, ensuring the continuous collection of network statistics. 

Pod used to validate this change, 

```yaml 
apiVersion: v1
kind: Pod
metadata:
  name: init-sleep-pod
spec:
  initContainers:
  - name: init-sleep
    image: busybox
    command: ["sh", "-c", "sleep 60"]
  containers:
  - name: forever-sleep
    image: busybox
    command: ["sh", "-c", "while true; do sleep 1; done"]
```

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -305,14 +305,20 @@ func (h *crioContainerHandler) GetStats() (*info.ContainerStats, error) {
 	if err != nil {
 		return stats, err
 	}
-	// Clean up stats for containers that don't have their own network - this
-	// includes containers running in Kubernetes pods that use the network of the
-	// infrastructure container. This stops metrics being reported multiple times
-	// for each container in a pod.
-	if !h.needNet() {
-		stats.Network = info.NetworkStats{}
-	}
 
+	if !h.needNet() {
+		// Clean up stats for containers that don't have their own network - this
+		// includes containers running in Kubernetes pods that use the network of the
+		// infrastructure container. This stops metrics being reported multiple times
+		// for each container in a pod.
+		stats.Network = info.NetworkStats{}
+	} else if len(stats.Network.Interfaces) == 0 {
+		// No network related information indicates that the pid of the
+		// container is not longer valid and we need to ask crio to
+		// provide the pid of another container from that pod
+		h.pidKnown = false
+		return stats, nil
+	}
 	// Get filesystem stats.
 	err = h.getFsStats(stats)
 	if err != nil {


container/crio/handler.go
```diff
@@ -305,6 +305,14 @@ func (h *crioContainerHandler) GetStats() (*info.ContainerStats, error) {
...
 		return stats, err
 	}
+
+	if len(stats.Network.Interfaces) == 0 {
+		// No network related information indicates that the pid of the
+		// container is not longer valid and we need to ask crio to
+		// provide the pid of another container from that pod
+		h.pidKnown = false
+		return stats, nil
+	}
```
haircommander, line 315: I think we should double check `h.needNet()` here, or else we may needlessly recall into cri-o because a container didn't provide the network stats

haircommander: we can just combine this check as  `else if len(...)` below

harche: @haircommander Thanks, good point. Updated.

container/crio/handler.go
```diff
@@ -305,11 +305,18 @@ func (h *crioContainerHandler) GetStats() (*info.ContainerStats, error) {
...
-	// infrastructure container. This stops metrics being reported multiple times
-	// for each container in a pod.
-	if !h.needNet() {
+
+	if h.needNet() && len(stats.Network.Interfaces) == 0 {
+		// No network related information indicates that the pid of the
+		// container is not longer valid and we need to ask crio to
+		// provide the pid of another container from that pod
+		h.pidKnown = false
+		return stats, nil
```
haircommander, line 314: I would prefer these two conditions are switched so we don't have to call needsNet twice:

```
if !h.needNet() {
....
} else if len(...) {
....
}
```

harche: @haircommander Done. Thanks.

haircommander:
/lgtm

thanks!

dims (approved):
LGTM

bobbypage:
lgtm, thanks!

bobbypage:
Backports:

* v0.47 - https://github.com/google/cadvisor/pull/3313 - released in https://github.com/google/cadvisor/releases/tag/v0.47.2
* v0.46 - https://github.com/google/cadvisor/pull/3320 - released in https://github.com/google/cadvisor/releases/tag/v0.46.1
* v0.45 - https://github.com/google/cadvisor/pull/3321 - released in https://github.com/google/cadvisor/releases/tag/v0.45.1
* v0.44 - https://github.com/google/cadvisor/pull/3322 - released in https://github.com/google/cadvisor/releases/tag/v0.44.2

Issue #3116, opened 2022-05-12
Move to GCP Prow CI instead of Kubernetes's Prow CI


BenTheElder:
xref: https://github.com/kubernetes/test-infra/issues/12863

It would be great if we could move this project to use https://github.com/GoogleCloudPlatform/oss-test-infra instead of https://github.com/kubernetes/test-infra since it is a Google project and not part of the Kubernetes project.

In the long term, we are migrating Kubernetes's CI to community managed infrastructure and those resources are allocated for the Kubernetes project.

A few projects have moved already, as you can see in https://github.com/kubernetes/test-infra/issues/12863

For googlers: see [go/oss-prow-onboard](https://goto.google.com/oss-prow-onboard)

BenTheElder:
cc @dims @bobbypage https://github.com/kubernetes/test-infra/issues/12863#issuecomment-1526126716

bobbypage:
Thanks for bringing this up.

Currently, cAdvisor is using a mix of github actions and prow. The main thing we use prow for is for running the integration tests. 

There are two jobs defined in [prow](https://github.com/kubernetes/test-infra/blob/a479c206796fee9ded9793bfffcd1ca93b87b35f/config/jobs/cadvisor/cadvisor.yaml): 

1. pull-cadvisor-e2e
2. ci-cadvisor-e2e

`build/prow_e2e.sh` is invoked by the node e2e runner, followed by `build/integration.sh`, see [here](https://github.com/kubernetes/kubernetes/tree/4115b2b1803fad5233921b168b46ba346734f228#L44-L75).

The github action `test-integration` currently runs `make docker-test-integration`, which will end up running `build/integration-in-docker.sh` which runs the same integration tests but inside a dockerized container. 

As a result, if we need to drop prow support, I think it may be simpler to simply drop all the prow jobs and only rely on github actions, since we are already running the integartion tests there.

Thoughts @dims @iwankgb @Creatone?

iwankgb:
I think that using GitHub Actions is a good idea. Everything else will be more or less Google-specific and more difficult to grasp for non-Googlers. I'm not sure if we can use free runners in long run, as they can stop being free or their limitations (e.g. number of parallel builds) may start biting us.

BenTheElder:
> As a result, if we need to drop prow support, I think it may be simpler to simply drop all the prow jobs and only rely on github actions, since we are already running the integration tests there. I don't see any reason we need to depend on prow.

[...]

PR #3302, opened 2023-05-04
Optimize network metrics collection

kolyshkin:
Apparently, we collect network stats for all containers, and then discarding some or most of it:

 - for docker, we collect and discard stats for containers which share netns with another container (which is rare I guess);

 - for both crio and containerd, we collect and discard stats for containers that are not infra (sandbox, pod, pause) containers (which is very common).

Instead of reading and parsing a bunch of files in /proc/PID/net and when removing those, let's set things up in a way so we don't collect useless stats in the first place.

This should improve performance, memory usage, and ease the load on garbage collection.

### DIFF
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -126,7 +126,14 @@ func newContainerdContainerHandler(
 		Aliases:   []string{id, name},
 	}
 
-	libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootfs, int(taskPid), includedMetrics)
+	// Containers that don't have their own network -- this includes
+	// containers running in Kubernetes pods that use the network of the
+	// infrastructure container -- does not need their stats to be
+	// reported. This stops metrics being reported multiple times for each
+	// container in a pod.
+	metrics := common.RemoveNetMetrics(includedMetrics, cntr.Labels["io.cri-containerd.kind"] != "sandbox")
+
+	libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootfs, int(taskPid), metrics)
 
 	handler := &containerdContainerHandler{
 		machineInfoFactory:  machineInfoFactory,
@@ -134,7 +141,7 @@ func newContainerdContainerHandler(
 		fsInfo:              fsInfo,
 		envs:                make(map[string]string),
 		labels:              cntr.Labels,
-		includedMetrics:     includedMetrics,
+		includedMetrics:     metrics,
 		reference:           containerReference,
 		libcontainerHandler: libcontainerHandler,
 	}
@@ -164,22 +171,12 @@ func (h *containerdContainerHandler) ContainerReference() (info.ContainerReferen
 	return h.reference, nil
 }
 
-func (h *containerdContainerHandler) needNet() bool {
-	// Since containerd does not handle networking ideally we need to return based
-	// on includedMetrics list. Here the assumption is the presence of cri-containerd
-	// label
-	if h.includedMetrics.Has(container.NetworkUsageMetrics) {
-		//TODO change it to exported cri-containerd constants
-		return h.labels["io.cri-containerd.kind"] == "sandbox"
-	}
-	return false
-}
-
 func (h *containerdContainerHandler) GetSpec() (info.ContainerSpec, error) {
 	// TODO: Since we dont collect disk usage stats for containerd, we set hasFilesystem
 	// to false. Revisit when we support disk usage stats for containerd
 	hasFilesystem := false
-	spec, err := common.GetSpec(h.cgroupPaths, h.machineInfoFactory, h.needNet(), hasFilesystem)
+	hasNet := h.includedMetrics.Has(container.NetworkUsageMetrics)
+	spec, err := common.GetSpec(h.cgroupPaths, h.machineInfoFactory, hasNet, hasFilesystem)
 	spec.Labels = h.labels
 	spec.Envs = h.envs
 	spec.Image = h.image
@@ -203,13 +200,6 @@ func (h *containerdContainerHandler) GetStats() (*info.ContainerStats, error) {
 	stats, err := h.libcontainerHandler.GetStats()
 	if err != nil {
 		return stats, err
-	}
-	// Clean up stats for containers that don't have their own network - this
-	// includes containers running in Kubernetes pods that use the network of the
-	// infrastructure container. This stops metrics being reported multiple times
-	// for each container in a pod.
-	if !h.needNet() {
-		stats.Network = info.NetworkStats{}
 	}
 
 	// Get filesystem stats.
--- a/container/docker/handler.go
+++ b/container/docker/handler.go
@@ -33,7 +33,6 @@ import (
 	"github.com/google/cadvisor/zfs"
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 
-	dockercontainer "github.com/docker/docker/api/types/container"
 	docker "github.com/docker/docker/client"
 	"golang.org/x/net/context"
 	"k8s.io/klog/v2"
@@ -71,9 +70,6 @@ type dockerContainerHandler struct {
 
 	// Image name used for this container.
 	image string
-
-	// The network mode of the container
-	networkMode dockercontainer.NetworkMode
 
 	// Filesystem handler.
 	fsHandler common.FsHandler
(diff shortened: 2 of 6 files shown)


kolyshkin:
@bobbypage PTAL

Also, is there anything I can do to skip waiting for ok-to-test? The bot [says I need to join kubernetes org](https://github.com/google/cadvisor/pull/3302#issuecomment-1533945766) but I am already a member.

bobbypage:
Thanks @kolyshkin for the investigation here and perf fixes. Let me take a closer look. 

> Also, is there anything I can do to skip waiting for ok-to-test? The bot says I need to join kubernetes org but I am already a member.

Hmm, I'm not sure, I was under the assumption that all k8s members could skip the ok-to-test, but there may be some differences for non k8s repos...

container/containerd/handler.go
```diff
@@ -164,22 +172,33 @@ func (h *containerdContainerHandler) ContainerReference() (info.ContainerReferen
...
 	}
-	return false
+
+	// This is not a pod/sandbox container; remove network metrics.
+
+	// Check if there is anything we can remove.
+	if !metrics.HasAny(container.AllNetworkMetrics) {
+		return metrics
+	}
+
```
bobbypage, line 191: Is there any reason we need to first check if the network metrics are included and only if true, then we call `metrics.Difference`? Shouldn't it be safe to always call `metric.Difference` (if the network metrics are not included, it should be a no-op)?

Or is this check more a performance optimization to avoid creating a new `MetricSet`?

container/docker/handler.go
```diff
@@ -217,7 +216,6 @@ func newDockerContainerHandler(
 		Namespace: DockerNamespace,
 	}
 	handler.image = ctnr.Config.Image
-	handler.networkMode = ctnr.HostConfig.NetworkMode
```
bobbypage, line 220: why was this removed?

bobbypage:
A couple small questions but changes largely LGTM. Has this been tested locally? It would be nice to test this end to end, by integrating into k8s and ensuring that the network metrics are still present.

We have an e2e test in k8s, `summary_test` that verifies network metrics (xref: https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/summary_test.go#L208-L218).

One idea, can you maybe open a WIP PR to k/k that vendors this copy of cAdvisor (with the changes in the PR) and then use that WIP PR to run pull job in k/k to verify network metrics are still working correctly?

container/containerd/handler.go
kolyshkin: Yes, this is a performance optimization -- why bothering with copying if we don't have to? The downside is code looks a bit less compact than it could be -- but still pretty readable.

container/docker/handler.go
kolyshkin: This is because the only user of `h.networkMode` was `func (h *dockerContainerHandler) needNet() bool`.

That method is now gone, and so the field is unused.

kolyshkin:
Sure; https://github.com/kubernetes/kubernetes/pull/117833

BenTheElder:
> Also, is there anything I can do to skip waiting for ok-to-test? The bot https://github.com/google/cadvisor/pull/3302#issuecomment-1533945766 but I am already a member.

It's actually checking if you you're a member of the organization the PR is opened to (so github.com/google), but the join link is mis-configured to Kubernetes's.

The first part of the bot message is correct:
> I'm waiting for a [google](https://github.com/orgs/google/people) member to verify that this patch is reasonable to test.

But the join link is wrongly configured: https://github.com/kubernetes/test-infra/blob/62ef55338ead7466542e0c20d226c695f97ac3f4/config/prow/plugins.yaml#L21

Prow had support for setting a trusted organization but that's been long deprecated in favor of "is a member of the organization that owns the repo || is a collaborator on the repo" (which can be restricted to only org members by config).

cAdvisor as a Google project should stop using Kubernetes's instance and use the Google OSS instance anyhow (#3116), now that we have distinct instances and dedicated funding for Kubernetes (well, for years now ... it's one of the few projects lagging on switching).

iwankgb:
Could you provide tests for `maybeRemoveNet()` to avoid weird regression in future, please?

container/containerd/handler.go
kolyshkin: > Could you provide tests for `maybeRemoveNet()` to avoid weird regression in future, please?

Done

kolyshkin:
OK, I did a rewrite to avoid code repetition, added a unit tests for RemoveNetMetrics (as requested by @iwankgb), improved some comments and hopefully made the whole thing more readable.

https://github.com/kubernetes/kubernetes/pull/117833 is also updated.

iwankgb:
@bobbypage if you don't have more comments then I'm going to merge it tomorrow.

kolyshkin:
Whoops, fixed the logic (that was inverted for containerd and crio). 😊 

```diff
diff --git a/container/containerd/handler.go b/container/containerd/handler.go
index 626d4fd3..57a2d82c 100644
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -131,7 +131,7 @@ func newContainerdContainerHandler(
        // infrastructure container -- does not need their stats to be
        // reported. This stops metrics being reported multiple times for each
        // container in a pod.
-       metrics := common.RemoveNetMetrics(includedMetrics, cntr.Labels["io.cri-containerd.kind"] == "sandbox")
+       metrics := common.RemoveNetMetrics(includedMetrics, cntr.Labels["io.cri-containerd.kind"] != "sandbox")
 
        libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootfs, int(taskPid), metrics)
 
diff --git a/container/crio/handler.go b/container/crio/handler.go
index e945b790..a6832518 100644
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -154,7 +154,7 @@ func newCrioContainerHandler(
        // infrastructure container -- does not need their stats to be
        // reported. This stops metrics being reported multiple times for each
        // container in a pod.
-       metrics := common.RemoveNetMetrics(includedMetrics, cInfo.Labels["io.kubernetes.container.name"] == "POD")
+       metrics := common.RemoveNetMetrics(includedMetrics, cInfo.Labels["io.kubernetes.container.name"] != "POD")
 
        libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootFs, cInfo.Pid, metrics)
 
```

kolyshkin:
> Whoops, fixed the logic (that was inverted for containerd and crio).

OTOH this gave a way to check if the test mentioned in https://github.com/google/cadvisor/pull/3302#issuecomment-1535391559 is working.

Indeed it does (from https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/117833/pull-kubernetes-node-e2e-containerd/1655703979042017280/):

> E2eNode Suite: [It] [sig-node] Summary API [NodeConformance] when querying /stats/summary should report resource usage through the stats api expand_less | 3m21s
> -- | --
> { failed [FAILED] Timed out after 180.001s. Expected     <string>: Summary to match fields: { [.Pods[summary-test-836::stats-busybox-0].Network: 	Expected 	    <*v1alpha1.NetworkStats \| 0x0>: nil 	not to be <nil>, .Pods[summary-test-836::stats-busybox-1].Network: 	Expected 	    <*v1alpha1.NetworkStats \| 0x0>: nil 	not to be <nil>] } In [It] at: test/e2e_node/summary_test.go:332 @ 05/08/23 22:58:22.455 }

kolyshkin:
With the change from https://github.com/google/cadvisor/pull/3302#issuecomment-1540577804 the e2e tests are now passing in https://github.com/kubernetes/kubernetes/pull/117833 (this time I ran crio-e2e as well).

> pull-kubernetes-node-crio-e2e — Job succeeded.
> pull-kubernetes-node-e2e-containerd — Job succeeded.

bobbypage:
all updates LGTM!

thank you for testing in k/k and glad to see it was helpful and caught a potential issue! :)

bobbypage:
@iwankgb please take a final look and merge if it's all ok from your side. LGTM from me.

kolyshkin:
> thank you for testing in k/k and glad to see it was helpful and caught a potential issue! :)

I actually found the bug while re-reviewing this PR first, and then took a look at test results. In retrospect, it's good that we validated that the e2e test works.

haircommander:
here's a follow up to this we can make for cri-o https://github.com/google/cadvisor/pull/3592

PR #3304, opened 2023-05-08
Updating minimum Go version to 1.19

iwankgb:
Reasoning behind this proposal:
* Kubernetes requires `1.20` on `main` anyway
* We run tests on `1.19` and `1.20` only
* Versions in `/` and `/cmd` shall be the same.

Note for the reviewers: take a look at `utils.sysfs.toFileInfo()` - this the only addition/change that was not performed using replace functionality in my IDE.


container/docker/handler_test.go
```diff
@@ -38,13 +37,13 @@ func TestStorageDirDetectionWithOldVersions(t *testing.T) {
 
 func TestStorageDirDetectionWithNewVersions(t *testing.T) {
 	as := assert.New(t)
-	testDir, err := ioutil.TempDir("", "")
+	testDir, err := os.MkdirTemp(os.TempDir(), "")
```
kolyshkin, line 40: it's better to use `t.TemDir()` here.

fs/fs_test.go
```diff
@@ -119,14 +118,14 @@ func TestDirDiskUsage(t *testing.T) {
 	as := assert.New(t)
 	fsInfo, err := NewFsInfo(Context{})
 	as.NoError(err)
-	dir, err := ioutil.TempDir(os.TempDir(), "")
+	dir, err := os.MkdirTemp(os.TempDir(), "")
```
kolyshkin, line 121: better to use `t.TempDir()` here.

utils/sysfs/sysfs.go
```diff
@@ -279,7 +290,23 @@ func (fs *realSysFs) GetNetworkStatValue(dev string, stat string) (uint64, error
 
 func (fs *realSysFs) GetCaches(id int) ([]os.FileInfo, error) {
 	cpuPath := fmt.Sprintf("%s%d/cache", cacheDir, id)
-	return ioutil.ReadDir(cpuPath)
+	dir, err := os.ReadDir(cpuPath)
```
kolyshkin, line 293: I know this is an incompatible change, but it would make sense to convert this to return `[]os.DirEntry` instead.

Otherwise, practically, what happens here, is a `stat(2)` call to every entry, which is kinda expensive. More to say, it seems the only user of this function is in utils/sysinfo ([I checked](https://sourcegraph.com/search?q=context:global+sysfs.GetCaches+-file:%28%5E%7C/%29vendor/+&patternType=standard&sm=1&groupBy=repo)), so while this is a public API, apparently it has no users.

iwankgb: There are three other uses of `toFileInfo()` that would have to be fixed. While I understand you concern and I believe that it's valid, I'm hesitant to do this as part of this PR and I would prefer to address it in future.

Creatone (commented):
LGTM

Will you handle the @kolyshkin proposal on the `t.TempDir`?

Creatone (approved):

iwankgb:
I did in https://github.com/google/cadvisor/commit/814bb2719c1bcdd2ba8b2eaa47f78acd3836449f

resctrl/utils_test.go
```diff
@@ -383,7 +382,7 @@ func TestGetPids(t *testing.T) {
 
 func TestGetAllProcessThreads(t *testing.T) {
 	mockProcFs := func() string {
-		path, _ := ioutil.TempDir("", "proc")
+		path, _ := os.MkdirTemp("", "proc")
```
Creatone, line 385: Why not use here a `t.TempDir()`?

Creatone:
> I did in [814bb27](https://github.com/google/cadvisor/commit/814bb2719c1bcdd2ba8b2eaa47f78acd3836449f)

I miss that, sorry

bobbypage:
delayed lgtm, thank you for the fixes!

PR #3308, opened 2023-05-09
Add crio client timeout

VicThomas-Medallia:
Running cAdvisor in our Kubernetes clusters that use the CRI-O container runtime, we observed that for a small number of nodes, the cAdvisor pod's cAdvisor container was in a crashloop due to startup probe failure. Upon remote debugging of such containers, we found that they were stuck doing an HTTP GET via crio.sock for a given container, in the following stack trace:
```
runtime.gopark(proc.go:382)
runtime.selectgo(select.go:327)
net/http.(*persistConn).roundTrip(transport.go:2638)
net/http.(*Transport).roundTrip(transport.go:603)
net/http.(*Transport).RoundTrip(roundtrip.go:17)
net/http.send(client.go:252)
net/http.(*Client).send(client.go:176)
net/http.(*Client).do(client.go:716)
net/http.(*Client).Do(client.go:582)
github.com/google/cadvisor/container/crio.(*crioClientImpl).ContainerInfo(client.go:136)
github.com/google/cadvisor/container/crio.newCrioContainerHandler(handler.go:109)
github.com/google/cadvisor/container/crio.(*crioFactory).NewContainerHandler(factory.go:75)
github.com/google/cadvisor/container.NewContainerHandler(factory.go:257)
github.com/google/cadvisor/manager.(*manager).createContainerLocked(manager.go:913)
github.com/google/cadvisor/manager.(*manager).createContainer(manager.go:900)
github.com/google/cadvisor/manager.(*manager).detectSubcontainers(manager.go:1104)
github.com/google/cadvisor/manager.(*manager).Start(manager.go:300)
main.main(cadvisor.go:166)
runtime.main(proc.go:250)
runtime.goexit(asm_amd64.s:1598)
runtime.newproc(<autogenerated>:1)
```
It turned out that such containers for which the HTTP GET via crio.sock was hanging were containers for pods stuck in terminating. A co-worker reproduced the hang problem using `crictl` and submitted [this issue for CRI-O](https://github.com/cri-o/cri-o/issues/6865).
However, another way to avoid this problem is to allow for a timeout for the crio client. That is what this pull request does.

We are currently running a fork containing the code in this pull request in our Kubernetes clusters. Using a new `--crio_client_timeout` flag to specify a timeout, it successfully bypasses the problem, allowing iteration over the set of detected containers to continue.

The default behavior -- when no `--crio_client_timeout` flag is used -- remains as is. That is, the default behavior continues to be no timeout.


VicThomas-Medallia:
It is good news that CRI-O has been fixed! Nonetheless, I still think this pull request has merit, as a way to defensively protect against any similar issue occurring the future and to ensure that those who cannot upgrade their CRI-O version on a timely basis can avoid the problem now.

SergeyKanzhelev:
@rphillips can you please review?

rphillips:
/lgtm

should be no change of behavior by default.

VicThomas-Medallia:
Thank you @SergeyKanzhelev and @rphillips.

SergeyKanzhelev (approved):
/lgtm

Issue #3314, opened 2023-05-26
CVE-2022-23648 found in v0.44.1


matttrach:
Would it be possible to upgrade the `containerd/containerd` dependency in the `release-v0.44` branch for next patch release?
Trivy is finding CVE-2022-23648 in Kubelet on release `kubernetes/kubernetes:v1.24.14`.

Sorry if this is a duplicate, or if this is not feasible, we are attempting to reduce CVEs as much as possible.

Even while these CVEs may not be easily exploitable, we are attempting to meet compliance goals and guidelines.

<details>
<Summary>Trivy details</Summary>

- I build kubelet using https://github.com/rancher/image-build-kubernetes/blob/v1.24.14-rke2r1-build20230518/Dockerfile
- then copied it out of the build container, and pushed it to its own Docker/scratch container to scan just the bin
- here is the result:

```
2023-05-25T19:39:42.559Z        INFO    Detecting gobinary vulnerabilities...

kubelet (gobinary)

Total: 3 (HIGH: 2, CRITICAL: 1)

┌──────────────────────────────────┬────────────────┬──────────┬────────────────────────────────────┬───────────────────────────────────┬──────────────────────────────────────────────────┐
│             Library              │ Vulnerability  │ Severity │         Installed Version          │           Fixed Version           │                      Title                       │
├──────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────┤
│ github.com/containerd/containerd │ CVE-2022-23648 │ HIGH     │ v1.4.12                            │ 1.4.13, 1.5.10, 1.6.1             │ containerd: insecure handling of image volumes   │
│                                  │                │          │                                    │                                   │ https://avd.aquasec.com/nvd/cve-2022-23648       │
├──────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────┤
│ github.com/emicklei/go-restful   │ CVE-2022-1996  │ CRITICAL │ v2.9.5+incompatible                │ 2.16.0+incompatible               │ Authorization Bypass Through User-Controlled Key │
│                                  │                │          │                                    │                                   │ https://avd.aquasec.com/nvd/cve-2022-1996        │
├──────────────────────────────────┼────────────────┼──────────┼────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────────────┤
[...]

PR #3319, opened 2023-05-30
upgrading containerd to resolve CVE-2022-23648 in release-v0.44

matttrach:
This is attempting to resolve https://github.com/google/cadvisor/issues/3314.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.15
 	github.com/aws/aws-sdk-go v1.35.24
 	github.com/blang/semver v3.5.1+incompatible
-	github.com/containerd/containerd v1.4.12
+	github.com/containerd/containerd v1.4.13
 	github.com/containerd/ttrpc v1.0.2
 	github.com/containerd/typeurl v1.0.2
 	github.com/docker/distribution v2.8.0+incompatible // indirect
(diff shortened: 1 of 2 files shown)


dims:
@matttrach please sign the google CLA as well

matttrach:
CLA signed

matttrach:
The CI error appears to be related to the deprecation of `go get` in v1.18, the CI fails to install the new mod version and the code. I was able to trace the code to the below link:
https://github.com/google/cadvisor/blob/release-v0.44/build/assets.sh#L33

matttrach:
The CI is pulling in the latest version of go in the link below, seems like it should downgrade to the go.mod version (1.16)
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/google_cadvisor/3319/pull-cadvisor-e2e/1664304522698166272/#1:build-log.txt%3A142

iwankgb:
I don't think that we are going to merge anything to 0.44 - it was released more than a year ago. @bobbypage?

iwankgb:
@dims, is this something that Kubernetes needs?

dims:
k/k does not vendor `containerd/containerd` in its go.mod so probably not

iwankgb:
Closing per @dims comment.

iwankgb:
Sorry, I was not aware of discussion in #3314.

iwankgb:
This is more complicated, apparently. You would have to cherry-pick 8794bc923a3cb7439a4299592a2cf96c88b05568 too (it changes container registry hostname and no tests will pass without this change). I consider your concern quite valid, but we will need @bobbypage to release a new version as we have no automated release process in place.

iwankgb:
According to CVE description:

> A bug was found in containerd where containers launched through containerd’s CRI implementation with a specially-crafted image configuration could gain access to read-only copies of arbitrary files and directories on the host.

cAdvisor launches no containers, so this fix is irrelevant.

PR #3322, opened 2023-06-01
[0.44] Request the pid of another container if current pid is not longer valid

harche:
This PR backports the changes from https://github.com/google/cadvisor/pull/3301 to 0.44

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -305,14 +305,20 @@ func (h *crioContainerHandler) GetStats() (*info.ContainerStats, error) {
 	if err != nil {
 		return stats, err
 	}
-	// Clean up stats for containers that don't have their own network - this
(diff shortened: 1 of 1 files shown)


iwankgb:
0.44 was released more than a year ago and I don't think we are keen to keep updating it. What do you think, @bobbypage?

iwankgb:
@dims, is this something that Kubernetes needs?

bobbypage:
cadvisor 0.44 is used in k8s 1.24 which will be EoL soon. @harche do you think we still need 0.44 support here or are the other cherrypicks sufficient?

harche:
This bug significantly impacts network metrics, as they disappear whenever any container in the pod undergoes a restart. Therefore, to address this issue, we are working to backport the fix to all releases that have not yet reached their end-of-life status.

PR #3326, opened 2023-06-13
Resctrl collector fixes

JulSenko:
Resctrl collector fixes:
* Resume stat collection if collector err was resolved. From the code looks like it was intended to do so, but `enabled=true` was missing.
* Rely on `libcontainer` funcs to collect pids/threads instead of executing `ps` call which is very expensive.
* Collect threads instead of pids. During collector startup we collect threads, so for the following checks we should also collect threads, not pids. Pid approach woks on relatively simple processes, but for any complex one pid list will never be the same as the original thread list, thus resulting in err on every collector run.

### DIFF
--- a/resctrl/utils.go
+++ b/resctrl/utils.go
@@ -77,6 +77,9 @@ var (
 		modeFileName:     {},
 		sizeFileName:     {},
 	}
+
+	errNotEnoughPIDs = fmt.Errorf("there should be all pids in group")
+	errTooManyPIDs   = fmt.Errorf("group should have container pids only")
 )
 
 func Setup() error {
@@ -105,7 +108,7 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
 		return rootResctrl, nil
 	}
 
-	pids, err := getContainerPids()
+	pids, err := getPids(containerName)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


resctrl/utils.go
```diff
@@ -105,7 +108,7 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
 		return rootResctrl, nil
 	}
 
-	pids, err := getContainerPids()
+	pids, err := getPids(containerName)
```
iwankgb, line 111: Is `getContainerPids()` used (it does not seem to be)? If not, would it be possible to drop the argument?

resctrl/utils.go
```diff
@@ -114,60 +117,79 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
...
+		processPath = "/rootfs/proc"
+	}
+	var processThreads []string
+	for _, pid := range pids {
+		processThreads, err = getAllProcessThreads(filepath.Join(processPath, strconv.Itoa(pid), processTask))
+		if err != nil {
+			return "", err
+		}
+	}
+
```
iwankgb, line 130: Can you explain what are the benefits of this approach? As far as I understand (and I might be wrong) you extracted obtaining task IDs to separate loop, but the result will be the same.

iwankgb:
@JulSenko, can you merge `master` to your branch, please?

resctrl/utils.go
JulSenko: Now threads are extracted earlier and passed as param to the `findGroup` func, which previously accepted pids resulting in the incomplete list

resctrl/utils.go
```diff
@@ -190,20 +212,16 @@ func getPids(containerName string) ([]int, error) {
...
 		return processThreads, err
 	}
 
 	for _, dir := range threadDirs {
-		pid, err := strconv.Atoi(dir.Name())
-		if err != nil {
-			return nil, fmt.Errorf("couldn't parse %q dir: %v", dir.Name(), err)
-		}
-		processThreads = append(processThreads, pid)
+		processThreads = append(processThreads, dir.Name())
```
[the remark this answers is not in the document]
JulSenko: pids used to be ints and were passed to the external `ps` call eventually. Since `ps` is replaced by `libcontainer` function call which accepts strings, we no longer need this additional `Atoi` conversion.

resctrl/utils.go
```diff
@@ -233,7 +251,7 @@ func findGroup(group string, pids []string, includeGroup bool, exclusive bool) (
 	for _, path := range availablePaths {
 		groupFound, err := arePIDsInGroup(path, pids, exclusive)
 		if err != nil {
-			return "", err
+			return path, err
```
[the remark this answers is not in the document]
JulSenko: This is not ideal, but we need path value for additional check later if error type is `errNotEnoughPIDs` or `errTooManyPIDs`. E.g.: group could have changed completely and we need to recreate it or that path is correct, but some threads died / were spawned anew.

resctrl/utils.go
JulSenko: It should be, but it touches quite a bunch of definitions. I'll follow up on this if that works for you:)

resctrl/utils.go
```diff
@@ -114,60 +117,79 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
...
+			return "", fmt.Errorf("couldn't find monitoring group matching %q container: %v", containerName, err)
 		}
 
-		// Add own prefix and use `-` instead `/`.
-		// e.g. my/container -> cadvisor-my-container
-		properContainerName := fmt.Sprintf("%s-%s", monGroupPrefix, strings.Replace(containerName, "/", "-", -1))
-		monGroupPath = filepath.Join(controlGroupPath, monitoringGroupDir, properContainerName)
+		rmErr := os.Remove(monGroupPath)
+		if rmErr != nil && !os.IsNotExist(rmErr) {
+			return "", fmt.Errorf("couldn't clean up monitoring group matching %q container: %v", containerName, rmErr)
```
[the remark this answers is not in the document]
JulSenko: Is `Errorf` not enough? Do you have in mind specific error type?  Please let me know how to improve this error

resctrl/utils.go
```diff
@@ -114,60 +117,79 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
...
+		}
+		if existingPath != monGroupPath {
+			rmErr = os.Remove(existingPath)
+			if rmErr != nil && !os.IsNotExist(rmErr) {
+				return "", fmt.Errorf("couldn't clean up monitoring group matching %q container: %v", containerName, rmErr)
```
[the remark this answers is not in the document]
JulSenko: Do you have in mind specific error type?  Please let me know how to improve this error

resctrl/utils.go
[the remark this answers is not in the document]
iwankgb:
```suggestion
			return "", fmt.Errorf("couldn't clean up monitoring group matching %q container: %w", containerName, rmErr)
```

resctrl/utils.go
```diff
@@ -113,60 +116,80 @@ func prepareMonitoringGroup(containerName string, getContainerPids func() ([]str
...
-					return "", fmt.Errorf("coudn't assign pids to %q container monitoring group: %w", containerName, err)
+			err = intelrdt.WriteIntelRdtTasks(monGroupPath, treadInt)
+			if err != nil {
+				secondError := os.Remove(monGroupPath)
+				if secondError != nil {
+					return "", fmt.Errorf(
+						"coudn't assign pids to %q container monitoring group: %w \n couldn't clear %q monitoring group: %v",
+						containerName, err, containerName, secondError)
 				}
+				return "", fmt.Errorf("coudn't assign pids to %q container monitoring group: %w", containerName, err)
```
iwankgb, line 192: Correct me if I'm wrong but:
* The Old Way:
  1. Iterate over all `pid`s. 
  2. Fetch all the `tid`s for a `pid`
  3. Iterate over all `tid`s 
  4. Write the `tid` to monitoring group.
  5. Exit `tid`s loop.
  6. Exit `pid`s loop.
* The New Way:
  1. Iterate over all `pid`s.
  2. Save all the `tid`s for a `pid` to a variable.
  3. Exit `pid`s loop.
  4. Iterate over all `tid`s.
  7. Write the `tid` to monitoring group.
  8. Exit `tid`s loop.

Functionally these two approaches are identical. In the PR description you wrote:

> Collect threads instead of pids. During collector startup we collect threads, so for the following checks we should also collect threads, not pids. Pid approach woks on relatively simple processes, but for any complex one pid list will never be the same as the original thread list, thus resulting in err on every collector run.

, but threads (`tid`s) have always been collected and written to monitoring group.

I might be missing something but I can't understand where the bug that you are trying to fix is. I will appreciate more detailed explanation that will help me to understand your reasoning. A test case failing with The Old Way and passing with The New Way would be perfect!

JulSenko: Sorry for abandoning this PR for, well, quite a lot, but I'm back. And since the resctrl collector hasn't changed much since then, we hopefully can continue discussing this issue. 
I need to brush my memory a bit, but if I remember correctly, it goes like this: at the start up we collect tids. Then, we collect pids during every run and expect them to match tids. We observed many cases where pids are constantly changing - new ones are popping up and old ones finish and exit. As a result, disappearing pids are being flagged as errors and collector exits. Thus goes this fix - it accounts those changing pid numbers and also continues monitoring if errors got resolved on the next run. I'll try adding a test case for this.

Also, if you feel like there is a lot going in this PR, I can split the changes.

I also no longer work for Uber, can I continue using this branch? I'll fill in the new agreement.

PR #3348, opened 2023-07-20
Bringing number of syscalls down in fs.getDistStatsMap()

### LINKED ISSUE #3082: Collecting Filesystem metrics makes extremely high CPU
PR #2768 fixed docker can collect FileSystem metrics. But in the scenario of multiple devices, per seconds(`housekeeping time`) collect FileSystem metrics causes kubelet CPU usage extremely high and impacts the application responsiveness.
This should be optimized or fixed in cadvisor.
CPU profiling data shows the hot code is here:
20220323-163351(WeLinkPC)
https://github.com/google/cadvisor/blob/dca17e6b6a784d101b157e9ad06a3db02687adb7/container/docker/handler.go#L402-L430
https://github.com/google/cadvisor/blob/dca17e6b6a784d101b157e9ad06a3db02687adb7/fs/fs.go#L451-L472

With `kubelet` (version 1.22.1) and about 202 block device, Probably consumed **40%CPU**
image
image

When I try to disable collecting FileSystem metrics, like this, only consumed **20%CPU**, `kubelet` restore normal consumption.
```
 // Docker does not impose any filesystem limits for containers. So use capacity as limit. 
 for _, fs := range mi.Filesystems { 
 	if fs.Device == device { 
 		limit = fs.Capacity 
 		fsType = fs.Type 
 		fsInfo = nil 
 		break 
 	} 
 } 
```
image
And normal pprof like this:
normal

iwankgb:
Fixes #3082: instead of using buffer scanner we deal with new line manually.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -19,8 +19,9 @@
 package fs
 
 import (
-	"bufio"
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path"
@@ -56,6 +57,9 @@ const (
 
 // A pool for restricting the number of consecutive `du` and `find` tasks running.
 var pool = make(chan struct{}, maxConcurrentOps)
+
+// bufSize control size of read buffer in getDiskStatsMap()
+var bufSize = 5000
 
 func init() {
 	for i := 0; i < maxConcurrentOps; i++ {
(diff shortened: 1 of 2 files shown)


iwankgb:
@Creatone @bobbypage - review will be appreciated.

iwankgb:
@Creatone @bobbypage :pleading_face:

PR #3351, opened 2023-07-21
fix performance degradation of NFS

### LINKED ISSUE #3349: Performance degradation of NFS after upgrading Kubernetes from 1.20 to 1.22
After upgrading our Kubernetes cluster from version 1.20 to 1.22, we have observed a version discrepancy, which is manifested in the form of a decrease in NFS performance. After some investigation, we suspect it might be related to this change: 
https://github.com/google/cadvisor/pull/2787

To provide a brief description of our environment, our cluster has nearly 10,000 Pods, all of which are using the same NAS service via PV. On a single node, we can have more than 400 nfs mountpoints. Currently, the metrics collection includes NFS mount points, leading to a large number of statfs operations and consequently, slow NFS metadata operations.

Would it be possible to adjust the metrics collection for mountpoints so that data is only collected once for the same device?

fusida:
One NFS is mounted to multiple Pods and statistics are requested only once.

fix #3349

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -386,6 +386,7 @@ func (i *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, error
 	if err != nil {
 		return nil, err
 	}
+	nfsInfo := make(map[string]Fs, 0)
 	for device, partition := range i.partitions {
 		_, hasMount := mountSet[partition.mountpoint]
 		_, hasDevice := deviceSet[device]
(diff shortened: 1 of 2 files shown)


fs/fs.go
```diff
@@ -410,10 +411,21 @@ func (i *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, error
 			default:
 				var inodes, inodesFree uint64
 				if utils.FileExists(partition.mountpoint) {
+					if strings.HasPrefix(partition.fsType, "nfs") {
```
iwankgb, line 414: Could you move NFS handling to dedicated case statement? I think it would increase code readability.

fusida: @iwankgb done

iwankgb (approved):

PR #3358, opened 2023-07-26
remove the check for the existence of NFS files

### LINKED ISSUE #3349: Performance degradation of NFS after upgrading Kubernetes from 1.20 to 1.22

fusida:
after #3351, we find there are still many requests for nfs stats in some case.
through analysis and debugging, we found the FileExists function will cause manay unnecessary requests.
so we suggesst remove the check for the existence of NFS files and that reduced requests to nas significantly.

fix #3349 
fix #3351

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -413,10 +413,6 @@ func (i *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, error
 				// if /dev/zfs is not present default to VFS
 				fallthrough
 			case NFS.String():
-				if !utils.FileExists(partition.mountpoint) {
-					klog.V(4).Infof("the file system type is %s, partition mountpoint does not exist: %v", partition.fsType, partition.mountpoint)
-					break
-				}
 				devId := fmt.Sprintf("%d:%d", partition.major, partition.minor)
 				if v, ok := nfsInfo[devId]; ok {
 					fs = v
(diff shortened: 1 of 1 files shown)


fusida:
@iwankgb PTAL, thanks a lot～

fs/fs.go
```diff
@@ -413,17 +413,17 @@ func (i *RealFsInfo) GetFsInfoForPath(mountSet map[string]struct{}) ([]Fs, error
...
-				}
 				devId := fmt.Sprintf("%d:%d", partition.major, partition.minor)
 				if v, ok := nfsInfo[devId]; ok {
 					fs = v
 					break
 				}
 				var inodes, inodesFree uint64
 				fs.Capacity, fs.Free, fs.Available, inodes, inodesFree, err = getVfsStats(partition.mountpoint)
+				if err != nil {
+					klog.V(4).Infof("the file system type is %s, partition mountpoint does not exist: %v", partition.fsType, partition.mountpoint)
```
iwankgb, line 424: Can you do something along the lines:
```suggestion
					klog.V(4).Infof("the file system type is %s, partition mountpoint does not exist: %v, error: %s", partition.fsType, partition.mountpoint, err)
```
I don't think we should lose the error causing the message to be logged.

fusida: done

iwankgb (approved):

PR #3361, opened 2023-08-07
Adjust the log level of Initialize Plugins

yangjunmyfm192085:
Adjust the log level of Initialize Plugins to make it easier for users to analyze the registration status of Plugins.
Now the default log level of plugin register failure is 5. When plugin register fails, the error info cannot be recorded from the logs, and the issue cannot be analyzed.


dims (approved):
LGTM

PR #3364, opened 2023-08-08
Update genproto dependency to isolated submodule

liggitt:
**What type of PR is this?**

Dependency update

**What this PR does / why we need it**:

Updates genproto to use the isolated submodule. This reduces overall dependency count and avoids future issues with ambiguous imports. See https://github.com/googleapis/go-genproto/issues/1015 for details.


liggitt:
cc @bobbypage

iwankgb (approved):

PR #3371, opened 2023-08-10
fix: fix bug in perf module

Rouzip:
1. Fix ioctl enable since we use perf group context
2. Group leader and other events use same pid(cgroup fd).

### DIFF
--- a/perf/collector_libpfm.go
+++ b/perf/collector_libpfm.go
@@ -209,11 +209,11 @@ func (c *collector) setup() error {
 
 		// Group is prepared so we should reset and enable counting.
 		for _, fd := range leaderFileDescriptors {
-			err = c.ioctlSetInt(fd, unix.PERF_EVENT_IOC_RESET, 0)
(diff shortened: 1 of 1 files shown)


iwankgb:
@Rouzip have you come across an issue with perf in cAdvisor?

Rouzip:
> @Rouzip have you come across an issue with perf in cAdvisor?

Sorry for the late response. I am developing a similar tool like cAdvisor perf module. In this [document](https://man7.org/linux/man-pages/man2/perf_event_open.2.html), it says:
> pid == -1 and cpu >= 0
              This measures all processes/threads on the specified CPU.
              This requires CAP_PERFMON (since Linux 5.8) or
              CAP_SYS_ADMIN capability or a
              /proc/sys/kernel/perf_event_paranoid value of less than 1.

> PERF_EVENT_IOC_ENABLE
              This enables the individual event or event group specified
              by the file descriptor argument.
              If the PERF_IOC_FLAG_GROUP bit is set in the ioctl
              argument, then all events in a group are enabled, even if
              the event specified is not the group leader (but see
              BUGS).


So it think we need to change it in cAdvisor.

Rouzip:
hello, anything I can do there? @iwankgb

PR #3373, opened 2023-08-11
vendor: bump runc to 1.1.9

haircommander:
to fix an issue with cgroupv2 memory accounting not being accurate

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -83,7 +83,7 @@ require (
 	github.com/mrunalp/fileutils v0.5.0 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.0.2 // indirect
-	github.com/opencontainers/runc v1.1.7 // indirect
+	github.com/opencontainers/runc v1.1.9 // indirect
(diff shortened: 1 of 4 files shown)


kolyshkin:
@haircommander I think you forgot to run `go mod tidy`.

haircommander:
hm, running it doesn't change the files there

bobbypage:
@haircommander can you please rebase to include https://github.com/google/cadvisor/pull/3380? That should fix the issues with the CI. Thank you!

haircommander:
yeah I've rebased and definitely have the fixes. Any hints here?

bobbypage:
Can you run `make tidy`?

haircommander:
ah hah! that was it, thank you

bobbypage:
Great! Can you please also remove the change to the `.gitignore`? I don't think that is needed.

haircommander:
sure, though I added it because the vendor directory is generated when we run `go mod vendor` but not checked into the repo...

bobbypage:
> sure, though I added it because the vendor directory is generated when we run `go mod vendor` but not checked into the repo...

gotcha, we don't vendor in cadvisor, so there shouldn't be a vendor directory and running `go mod vendor` shouldn't be necessary.

PR #3379, opened 2023-08-14
Fix lint

rphillips:


bobbypage:
ah, thanks for the fix! I didn't see, but got the fix in https://github.com/google/cadvisor/pull/3380

PR #3383, opened 2023-08-17
vendor: bump runc to 1.1.9

haircommander:
to fix an issue with cgroupv2 memory accounting not being accurate


iwankgb:
I'm afraid we have to stick to 1.7 until Kubernetes updates its [dependency version](https://github.com/kubernetes/kubernetes/blob/d5a42e2c22162895a17fdc29127f3cc3d56e3ad0/go.mod#L56).

iwankgb (changes requested):
Pending update in Kubernetes.

haircommander:
WDYT @iwankgb ?

iwankgb:
Kubernetes depends on cAdvisor and I"m hesitant to introduce changes that would affect their go.mod.

iwankgb:
This is why i decided to block the PR: i should have been clearer explaining my reasoning.

haircommander:
ah I see, but I am trying to bump it there too https://github.com/kubernetes/kubernetes/pull/119909 I figured this would be a good first step

haircommander:
@iwankgb @bobbypage WDYT?

bobbypage:
Thanks @haircommander for the PR? Why do we need backport this PR to v0.47? Can we just merge this in master and use the next cut version of cAdvisor in the next k8s?

haircommander:
I believe it's worth bumping the older versions to get https://github.com/kubernetes/kubernetes/issues/118916 fixes in the k8s release branches to more accurately describe the memory usage on v2

haircommander:
WDYT @bobbypage

bobbypage:
Thanks @haircommander. Where we plan to merge the fix for kubernetes/kubernetes#118916? If it's only master branch, then we don't need to backport. Or do we plan to merge it to both 1.27 + 1.28 and master? cAdvisor v0.47 is used in k/k in v1.27 and v1.28, so if we backport it here, we should ensure make the corresponding changes in k/k in v1.27 and v1.28.

bobbypage:
sg, let's merge this than, and I will followup to create a release.

ritazh:
> sg, let's merge this than, and I will followup to create a release.

Can we also get this into k8s 1.25 and 1.26? Thank you!

RDPzero:
I would appreciate to see this on k8s, since it solves a problem I'm facing for a long time.

Are you going to cut a patch release from 0.47 @bobbypage ?

aritraghosh:
@bobbypage  When is the next release scheduled for,  We would love to have it  to merge https://github.com/kubernetes/kubernetes/pull/119909

@haircommander

aritraghosh:
@haircommander Can this change also be cherry picked to 1.26, 1.27 and 1.28?  This change significantly reduces the memory reported on cgroup v2 to match v1

cgroupv2

ritazh:
> sg, let's merge this than, and I will followup to create a release.

@bobbypage thanks for merging the PR! Can you please share when the next release is scheduled for so we can back port this change to kubernetes? Thanks in advance!

xref: https://github.com/kubernetes/kubernetes/pull/121637#issuecomment-1823563472

bobbypage:
Thanks @ritazh for the ping.

Before we backport this change, I would ensure have the full clarity on the fix. Specifically there are still some open questions in https://github.com/opencontainers/runc/pull/3933#issuecomment-1771110499 if this is the right fix.

PR #3395, opened 2023-09-13
allow for cadvisor to detect disk stats when the read-only and writeable layers are on separate disks.

kannon92:
https://github.com/kubernetes/enhancements/issues/4191

We are working on splitting the image disk and we realize that some upstream changes are needed in cadvisor in order to account for crio splitting the containers between a read-only and writeable layer.  

This is only mergeable once the K8s KEP is approved and the CRIO changes are merged.  

Pre Reqs:

- [CRIO](https://github.com/cri-o/cri-o/pull/7401)
- [KEP-PR](https://github.com/kubernetes/enhancements/pull/4198)
- [k8s/k8s](https://github.com/kubernetes/kubernetes/pull/120616)

Note we vendored these cadisvor changes in the k8s/k8s PR to test this idea out.

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -43,6 +43,7 @@ const (
 	LabelSystemRoot          = "root"
 	LabelDockerImages        = "docker-images"
 	LabelCrioImages          = "crio-images"
+	LabelCrioContainers      = "crio-containers"
 	DriverStatusPoolName     = "Pool Name"
 	DriverStatusDataLoopFile = "Data loop file"
 )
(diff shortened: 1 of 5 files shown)


haircommander:
nit: I think the title could be a touch more informative

fs/fs.go
```diff
@@ -295,11 +296,38 @@ func (i *RealFsInfo) addDockerImagesLabel(context Context, mounts []*mount.Info)
...
+	// Everything will be stored in crio-images
+	if context.Crio.ImageStore == "" {
+		labelCrioImageOrContainers = LabelCrioImages
+	}
 	if context.Crio.Root != "" {
 		crioPath := context.Crio.Root
 		crioImagePaths := map[string]struct{}{
 			"/": {},
 		}
+		for _, dir := range []string{"devicemapper", "btrfs", "aufs", "overlay", "zfs"} {
```
haircommander, line 310: why not use StorageDriver as well instead of iterating through all of them?

kannon92:
> nit: I think the title could be a touch more informative

updated.

fs/fs.go
kannon92: Can you link or explain more on this?

haircommander: the [storage driver](https://github.com/google/cadvisor/pull/3395/files#diff-f427f5cea39f949a5e7e26217e7d444dbafef6923fbd4d10e2012b8ffecf00b2R45) field will be one of these values, and will also be the dir that's being used. instead of iterating through the options, I think you could just use that value

kannon92: I see!

kannon92: Pushed a change for this.

dims (approved):
LGTM

kannon92:
@dims can you advise on how to fix this CI? It looks like I have a bunch of jobs that are stuck.

dims:
@kannon92 try quashing the commits into one and pushing an update. given there is a `ok-to-test` label now, let's hope that succeeds.

kannon92:
> @kannon92 try quashing the commits into one and pushing an update. given there is a `ok-to-test` label now, let's hope that succeeds.

Thank you for the suggestion. Did exactly that we will see.

dims:
i see a bunch kick off after i hit `Approve and Run` https://github.com/google/cadvisor/pull/3395/checks

kannon92:
@dims sorry to bother you so much. Do you know who/how to get this merged?

Looks like all the jobs finished but PR requires something with write access to merge.

dims:
@kannon92 we'll need a review from @bobbypage

iwankgb (approved):
Sorry for the delay

bobbypage:
lgtm, but I'm curious about if we need separate changes for containerd as the changes are specific to cri-o?

kannon92:
> lgtm, but I'm curious about if we need separate changes for containerd as the changes are specific to cri-o?

Im not sure of the work needed for containerd to be honest. We have discussed with containerd about this feature but im not sure if/when they may implement support for it. So it’s hard to say what work will be needed. Currently containerd will be not supported for this feature due to lack of support for this feature in general (no way to separate writeable / read only layers).

PR #3396, opened 2023-09-15
container/libcontainer: Improve limits file parsing perf 

brancz:
Only max open files is ever parsed from limits files, therefore this
change optimizes for that case.

Benchmark:

```
$ benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/google/cadvisor/container/libcontainer
cpu: AMD Ryzen 5 3400GE with Radeon Vega Graphics
                    │   old.txt    │               new.txt               │
                    │    sec/op    │   sec/op     vs base                │
ProcessLimitsFile-8   85.012µ ± 1%   1.324µ ± 0%  -98.44% (p=0.000 n=10)
```

On a GKE v1.27.4 production cluster, this code path used roughly 1.5% of
the total kubelet CPU usage, and at 98.44% improvement this likely
results in at least a 1.5% CPU reduction (perhaps even more since also
less garbage is produced to be collected by the GC).

This change came out of a live stream that we do on a weekly basis trying to improve performance of popular open source software. Episode is here: https://www.youtube.com/watch?v=ricEoAdJNHc

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -39,7 +39,6 @@ import (
 )
 
 var (
-	whitelistedUlimits      = [...]string{"max_open_files"}
 	referencedResetInterval = flag.Uint64("referenced_reset_interval", 0,
 		"Reset interval for referenced bytes (container_referenced_bytes metric), number of measurement cycles after which referenced bytes are cleared, if set to 0 referenced bytes are never cleared (default: 0)")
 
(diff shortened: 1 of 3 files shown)


dashpole:
@bobbypage @SergeyKanzhelev

bobbypage:
Thanks for the performance improvement, looks very reasonable to me, LGTM!

brancz:
What do we need to do to get this merged? :)

SergeyKanzhelev (approved):
/lgtm

Looks fine with me. I doubt we will hit any instances where case sensitive check vs case insensitive check will matter

@bobbypage if you have concerns, please comment. Otherwise let's merge

PR #3406, opened 2023-10-06
container/common: Replace readUInt64 with optimized version

brancz:
This patch replaces the `readUint64` function with one that optimizes
for reading uint64s from files only. It prevents unnecessary calls to
`.Stat` because a valid uint64 can be 20 characters at most anyway. It
also avoids a few allocations.

Overall, benchmarking has shown to reduce the total CPU time by ~15%,
which given a total usage of roughly 3% in the entire kubelet, this
means we're reducing the kubelet baseline CPU by about ~0.45%.

Original profiling data from the Kubelet filtered down to only show `readUInt64`: https://pprof.me/5dfadaf/?filter_by_function=readuint64&search_string=readuint64

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -15,7 +15,9 @@
 package common
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"math"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


container/common/helpers.go
```diff
@@ -294,6 +297,41 @@ func readUInt64(dirpath string, file string) uint64 {
...
+	buf := make([]byte, i)
+	n, err := io.ReadFull(f, buf)
+	if err != nil {
+		if err != io.EOF && err != io.ErrUnexpectedEOF {
+			klog.Warningf("readNBytesFromFile: Failed to read %q: %s", file, err)
+		}
+	}
+
+	return bytes.TrimSpace(buf[:n])
+}
```
iwankgb, line 333: Can I have uni tests for this function, please?

PR #3410, opened 2023-10-10
container: skip checking for files in non-existent directories.

filippog:
Inspired by profile data from @brancz at https://pprof.me/5dfadaf/ and confirmed via `strace`. On cgroup v1 systems about half of Stat calls turn out to be for files within paths that don't exist.

I have also added very basic benchmarks for getSpecInternal.


dims:
@filippog please see broken test

image

filippog:
> @filippog please see broken test
> 
> image

Thank you! Fixed

iwankgb (approved):

PR #3412, opened 2023-10-10
manager: require higher verbosity level for container info misses

### LINKED ISSUE #3407: manager.go:694] Error getting data for container / because of race condition
Running cadvisor like:

```
# VERSION=v0.47.2 # use the latest release version from https://github.com/google/cadvisor/releases
sudo docker run \
  --volume=/:/rootfs:ro \
  --volume=/sys:/sys:ro \
  --publish=8080:8080 \
  --privileged \
  --device=/dev/kmsg \
  gcr.io/cadvisor/cadvisor:$VERSION \
  --docker_only=true \
  --disable_root_cgroup_stats=true
```

Gives error log every 15s:

> manager.go:694] Error getting data for container / because of race condition

Setting `--disable_root_cgroup_stats=false` make this error log go away.

hhromic:
This PR increases the required verbosity level for logging container stats info misses such as:
```
manager.go:694] Error getting data for container / because of race condition
```
When not scrapping root container metrics (using `-disable_root_cgroup_stats`), the above log message is repeated for every metrics scrape request. The error in this case is expected due to to the container stats info not having stats for `/`.

Fixes #3407 
Similar to #3341

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -698,7 +698,7 @@ func (m *manager) GetRequestedContainersInfo(containerName string, options v2.Re
 		info, err := m.containerDataToContainerInfo(data, &query)
 		if err != nil {
 			if err == memory.ErrDataNotFound {
-				klog.Warningf("Error getting data for container %s because of race condition", name)
+				klog.V(4).Infof("Error getting data for container %s because of race condition", name)
 				continue
 			}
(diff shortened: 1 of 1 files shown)


dims (approved):
LGTM

PR #3327, opened 2023-06-14
Fix flags typo

caiocfer:
Found typo when building cAdvisor that was preventing Perf event metrics from being available.
And tags were passed to build as "-tags=-tags=libfpm,netgo,libipmctl".

PR #3424, opened 2023-10-31
Disable go tags due to breaking the build

bobbypage:


bobbypage:
See https://github.com/google/cadvisor/pull/3327#issuecomment-1786606482

Issue #3430, opened 2023-11-23
Cryptic Load Reader error with binary.Read "invalid type" is likely due to a single error buffer type in the code


ShadowJonathan:
This has taken me ages to track down, but you probably have a binary.Read bug right here:

https://github.com/google/cadvisor/blob/2129e1c4fae8567551d1ac252c7a1dec27617e21/utils/cpuload/netlink/netlink.go#L210

I'm getting spammed (and see issues on getting spammed with) `error binary.Read: invalid type int32`, i've been looking in the entire codebase for this single type, since i know that it has caused this problem, and this is the only instance in the netlink path.

For reference, i'm tracking down why enabling `--enable_load_reader=true` and then deploying cadvisor as a daemonset creates these log messages.

I've seen https://github.com/google/cadvisor/issues/3242 also refer to this, in their case it was simply disabling the load reader that worked for them, but I'd like to have the metrics that're associated with it.

ShadowJonathan:
The reason this codepath got hit in the first place is https://github.com/google/cadvisor/issues/3137

teapot9:
I'm also encountering this issue on all my Alpine Linux hosts running cadvisor.

When using `-enable_load_reader=true`, I get the `error binary.Read: invalid type int32` error:
```
W0502 14:22:50.404350       1 container.go:586] Failed to update stats for container "/docker/60adec7ce829952bf1f26a9c022d51368e2bda7274f35f0ac1d22657938fcc88": failed to get load stat for "/docker/60adec7ce829952bf1f26a9c022d51368e2bda7274f35f0ac1d22657938fcc88" - path "/sys/fs/cgroup/docker/60adec7ce829952bf1f26a9c022d51368e2bda7274f35f0ac1d22657938fcc88", error binary.Read: invalid type int32
```

When using `-enable_load_reader=false`, I don't have the error, however the load metric `container_cpu_load_average_10s` is always zero:
```
container_cpu_load_average_10s{...,id="/docker/60adec7ce829952bf1f26a9c022d51368e2bda7274f35f0ac1d22657938fcc88",...} 0 1746195530574
```

dims:
Closing this issue as it has been open since before January 1, 2024. If this is still relevant, please feel free to reopen or create a new issue.

dgrisonnet:
I ran into this issue as well after checking whether we could remove this metric or not from kubelet: https://github.com/kubernetes/kubernetes/pull/134981#issuecomment-3842236167

[...]

PR #3431, opened 2023-11-24
Fix errno reference

ShadowJonathan:
Fixes https://github.com/google/cadvisor/issues/3430

This makes the error propagate through properly, tested it on my production cluster


bobbypage:
Please sign the CLA. thanks!

PR #3445, opened 2024-01-08
feat: add metrics for total_active_file and total_inactive_file memory

jrcichra:
The goal of this PR is to have additional cAdvisor metrics which
expose total_active_file and total_inactive_file.

Today working_set_bytes subtracts total_inactive_file in its calculation,
but there are situations where exposing these metrics directly is valuable.

For example, two containers sharing files in an emptyDir increases total_active_file over time. This is not tracked in the working_set memory.

Exposing total_active_file and total_inactive_file to the user
allows them to subtract out total_active_file or total_inactive_file
if they so choose in their alerts.

In the case of prometheus with a thanos sidecar, working_set can give
a false sense of high memory usage. The kernel counts thanos reading prometheus written files as "active_file" memory.
In that situation, a user may want to exclude active_file from their ContainerLowOnMemory alert.

Relates to: kubernetes/kubernetes#43916

### DIFF
--- a/cmd/internal/storage/bigquery/bigquery.go
+++ b/cmd/internal/storage/bigquery/bigquery.go
@@ -50,6 +50,10 @@ const (
 	colMemoryUsage string = "memory_usage"
 	// Working set size
 	colMemoryWorkingSet string = "memory_working_set"
+	// Total active file size
+	colMemoryTotalActiveFile string = "memory_total_active_file"
+	// Total inactive file size
+	colMemoryTotalInactiveFile string = "memory_total_inactive_file"
 	// Container page fault
 	colMemoryContainerPgfault string = "memory_container_pgfault"
 	// Constainer major page fault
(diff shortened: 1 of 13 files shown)


jrcichra:
CLA check is passing now.

jrcichra:
@iwankgb Thanks for the ok-to-test. This is ready for review.

SuperQ:
Hmm, this appears to be doing a calculation in cAdvisor. In Prometheus we prefer to expose raw metrics so that the user can do whatever math they want in PromQL.

IMO we should get rid of "working set" and simply expose all the byte gauges like `total_active_file`.

jrcichra:
> Hmm, this appears to be doing a calculation in cAdvisor. In Prometheus we prefer to expose raw metrics so that the user can do whatever math they want in PromQL.
> 
> IMO we should get rid of "working set" and simply expose all the byte gauges like `total_active_file`.

That would be good too. For the use case I'm describing, the solution would be to have a combination of three cgroup metrics:

1) `memory.usage_in_bytes`
2) `total_active_file`
3) `total_inactive_file`
4) ...all other cgroup raw memory metrics...

Then we could build a prometheus query + alert for `memory_usage_in_bytes` - `total_active_file` - `total_inactive_file`.

Then perhaps we do a similar calculation for kubelet calculating memory pressure instead of `working_set`. That would be done here:

https://github.com/kubernetes/kubernetes/blob/a309fadbac3339bc8db9ae0a928a33b8e81ef10f/pkg/kubelet/eviction/memory_threshold_notifier.go#L92

jrcichra:
@SuperQ I've updated the PR to add raw metrics for `total_active_file` and `total_inactive_file` so the users (alert authors, kubelet, etc) can make their own memory calculations (e.g `container_memory_usage_bytes - container_memory_total_active_file_bytes - container_memory_total_inactive_file_bytes`).

jrcichra:
@iwankgb would you mind taking a look at the updated code? Thank you!

dongjiang1989 (approved):
LGTM

jrcichra:
Thanks for reviewing @dongjiang1989 . It looks like I still need 1 approving review.

iwankgb (approved):

Issue #3446, opened 2024-01-10
libcontainer/handler.go scanInterfaceStats() may have potential memory leak problem


cl0udee:
When deploying thousands of pods using Kubernetes, I noticed that the memory consumption of kubelet is unexpectedly high, reaching several tens of gigabytes. This surprised me greatly. To analyze the memory usage, I utilized the go tool pprof and found that a specific line of code within `cadvisor/container/libcontainer/handler.go` (line 510) in the `scanInterfaceStats()` function had a significant memory footprint. 

```
devName := fields[0]
```

This line of code references `fields[0]`, which suggests that fields might not be garbage collected for a significant period of time. Therefore, I suspect there might be a potential memory leak issue. Has anyone encountered similar issues? I'm unsure if my analysis is correct, and I hope someone can assist me.

PR #3447, opened 2024-01-10
fix: libcontainer/handler.go potential memory leak

cl0udee:
fix issue: #3446 

Use the `strings.Clone` function to make `devName` no longer reference `fields[0]`.


bobbypage:
Can you please explain why this fixes the memory leak? Why does cloning the string fix the issue?

cl0udee:
> Can you please explain why this fixes the memory leak? Why does cloning the string fix the issue? Can you provide a pprof maybe?

I did perform pprof analysis, and you can refer to this link #3446 . However, I have lost the original pprof file. Here is a possible explanation: When there are a large number of pods running in the system, the system load increases rapidly, leading to slower garbage collection. The `devName` parameter continues to reference the original `fields` array, causing memory accumulation and further exacerbating the system load. However, when I choose to use `strings.Clone` for copying, the original `fields` can be quickly garbage collected. This phenomenon can be observed well in the pprof analysis.

cl0udee:
@bobbypage Hi, I hope everything is going well. Just wanted to check in and see if there's any update on the progress of this Pull Request. Thank you!

PR #2957, opened 2021-10-11
crio: filter out systemd related components

rphillips:
In Openshift we are [seeing](https://bugzilla.redhat.com/show_bug.cgi?id=1978528#c4) systemd components get added to housekeeping within the Kubelet using crio. This patch filters the `system-systemd` namespace for containers.

cc @harche

PR #3457, opened 2024-01-19
Reduce kubelet logs 'Failed to create existing container' when kubelet is using crio

### LINKED ISSUE #3456: cadvisor print a lot of 'manager.go:1116 Failed to create existing container …' logs.
We are using crio as the container runtime in a Kubernetes environment, and we have noticed that kubelet is printing a lot of  **'manager.go:1116] Failed to create existing container …'** logs.

lanxin_20240119105523

Crio and the Kubernetes community also have related issues:
https://github.com/kubernetes/kubernetes/issues/120988
https://github.com/cri-o/cri-o/issues/5490
Although there have been modifications to the issue 5490 in crio, but the problem has not been fixed, and the issue still persists.

kubernetes version: v1.25.8
crio version: 1.25.4

dsxing:
fix #3456

lanxin_20240119105118

when We found that these logs are related to crio sandbox containers. The root cause is that when a pod is created, there will be a sandbox container folder in the cgroup file. When cadvisor scans the cgroup directory to identify containers, it recognizes the sandbox container. It then uses CanHandleAndAccept to determine that it needs to retrieve container information from crio using the id parameter. However, when attempting to get container information from crio via grpc, crio does not query the sandbox container, resulting in a 404 error(https://github.com/cri-o/cri-o/blob/main/server/inspect.go#L68). so cadvisor **createContainer** failed and print the logs periodically(https://github.com/google/cadvisor/blob/master/manager/manager.go#L1116) 

We think that this situation should be handled in the crio factory, filtering out sandbox containers. similar to this pr:https://github.com/google/cadvisor/pull/2957 for https://bugzilla.redhat.com/show_bug.cgi?id=1978528#c4

### DIFF
--- a/container/crio/factory.go
+++ b/container/crio/factory.go
@@ -32,6 +32,9 @@ import (
 // The namespace under which crio aliases are unique.
 const CrioNamespace = "crio"
 
+// The namespace suffix under which crio aliases are unique.
+const CrioNamespaceSuffix = ".scope"
+
 // The namespace systemd runs components under.
 const SystemdNamespace = "system-systemd"
 
@@ -114,15 +117,20 @@ func (f *crioFactory) CanHandleAndAccept(name string) (bool, bool, error) {
 		// TODO(runcom): should we include crio-conmon cgroups?
 		return false, false, nil
 	}
-	if !strings.HasPrefix(path.Base(name), CrioNamespace) {
-		return false, false, nil
-	}
 	if strings.HasPrefix(path.Base(name), SystemdNamespace) {
 		return true, false, nil
+	}
+	if !strings.HasPrefix(path.Base(name), CrioNamespace) {
+		return false, false, nil
 	}
 	// if the container is not associated with CRI-O, we can't handle it or accept it.
 	if !isContainerName(name) {
 		return false, false, nil
+	}
+
+	if !strings.HasSuffix(path.Base(name), CrioNamespaceSuffix) {
+		// this mean it's a sandbox container
+		return true, false, nil
 	}
 	return true, true, nil
 }
--- a/container/crio/factory_test.go
+++ b/container/crio/factory_test.go
@@ -20,6 +20,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type canHandleAndAccept struct {
+	canHandle bool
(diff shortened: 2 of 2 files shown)


iwankgb (approved):

iwankgb (approved):

dims:
@dsxing please see https://github.com/cri-o/cri-o/issues/8034 - this commit caused a regression when systemd is absent, We are trying to see if we should revert this PR

dims:
xref: https://github.com/google/cadvisor/pull/3565

haircommander:
in the case of systemd cgroup manager in cri-o and when we drop the infra container (as in, we "create" it but it's not a running container because it's not technically required), we create a dummy cgroup so cadvisor recognizes the pod and can report networking metrics. 

I think this is actually a bug in crio if we're not correctly finding the infra container.  we should be adding it to the container server even though we've dropped it. Do you have a succinct reproducer @dsxing ?

dsxing:
> @dsxing please see [cri-o/cri-o#8034](https://github.com/cri-o/cri-o/issues/8034) - this commit caused a regression when systemd is absent, We are trying to see if we should revert this PR

I'm sorry for submitted an incorrect PR. I believe we should revert this PR.

dsxing:
> in the case of systemd cgroup manager in cri-o and when we drop the infra container (as in, we "create" it but it's not a running container because it's not technically required), we create a dummy cgroup so cadvisor recognizes the pod and can report networking metrics.
> 
> I think this is actually a bug in crio if we're not correctly finding the infra container. we should be adding it to the container server even though we've dropped it. Do you have a succinct reproducer @dsxing ?

I'm sorry for being away for so long. The initial issue was that after restarting crio, kubelet would print a large number of "Failed to create existing container" logs(see https://github.com/kubernetes/kubernetes/issues/115609 ; https://github.com/kubernetes/kubernetes/issues/120988). To reduce these log messages, I submitted this PR. However, after further analyzing the root cause, I discovered that it was an issue in crio. In the end, I submitted a PR(https://github.com/cri-o/cri-o/pull/7726) that fundamentally resolved the problem, so I believe we should revert this PR. I apologize for any other issues this may have caused.

PR #3497, opened 2024-03-13
use pids.current rather than getting len of pids from cgroup.procs

kannon92:
While I was investigating https://github.com/kubernetes/kubernetes/issues/115215, I discovered that at the pod cgroup level cgroup.pods doesn't quite do what we need for process count.

Looking at cgroups `pids.current` seems to correctly report the process count.

Note this impacts cadvisor stats provider in kubelet because we are reporting processes at the pod level and we need to look at the correct cgroup for pods.

CAdvisor is correctly reporting this at the container cgroup level.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -299,8 +299,9 @@ func processStatsFromProcs(rootFs string, cgroupPath string, rootPid int) (info.
 		}
 	}
 
+	processCount := common.ReadUInt64(cgroupPath, "pids.current")
 	processStats := info.ProcessStats{
-		ProcessCount: uint64(len(pids)),
+		ProcessCount: processCount,
 		FdCount:      fdCount,
 		SocketCount:  socketCount,
 	}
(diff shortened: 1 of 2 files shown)


container/libcontainer/handler.go
```diff
@@ -299,8 +299,9 @@ func processStatsFromProcs(rootFs string, cgroupPath string, rootPid int) (info.
 		}
 	}
 
+	processCount := common.ReadUInt64(cgroupPath, "pids.current")
```
dims, line 302: Can you please add a specific method in helpers.go instead of exposing the method `readUInt64`?

kannon92: I exposed it because we already had this function elsewhere. I can copy it and bring it here also.

kannon92: I realized I didn't need to export readString but I still need to export this.

I exported this from common so that I could use it in libcontainer.

kannon92: @dims wdyt?

I can copy duplicate ReadUInt64 into libcontainer if we don't want to export the function. It just means we will have the same function in two locations.

kannon92:
@haircommander pointed out https://github.com/kubernetes/kubernetes/pull/124101 as a fix without changes in cadvisor.

/close

Issue #3495, opened 2024-03-11
container_fs_usage_bytes metrics miss pod/container label


tripitakav:
kubectl version: v1.24.15-eks
contained: 1.6.19

chengjoey:
https://github.com/google/cadvisor/blob/570d8a7eea0a7669d0a1d024190010a200bbef7b/container/containerd/handler.go#L174-L185

containerd not support fs disk usage yet

chengjoey:
when use containerd cri, we could get mountinfo from `proc/self/mountinfo` like this:
```
2826 28 0:478 / /run/containerd/io.containerd.runtime.v2.task/k8s.io/02f14690a20e38c89d9f1c251b2b043c542e49d9aa62baa3408fb3e51d369819/rootfs rw,relatime shared:811 - overlay overlay rw,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66230/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66153/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66152/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66151/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/59780/fs,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66239/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/66239/work,index=off
```

Therefore, when we query the disk usage of contianerd, we can use this directory`/run/containerd/io.containerd.runtime.v2.task/k8s.io/02f14690a20e38c89d9f1c251b2b043c542e49d9aa62baa3408fb3e51d369819/rootfs`

chengjoey:
get disk usage by containerd rootfs:
![image](https://github.com/google/cadvisor/assets/30427474/d677de47-2ea2-4f49-a6e3-ec763046295e)

chengjoey:
#3502

wsszh:
> when use containerd cri, we could get mountinfo from `proc/self/mountinfo` like this:
> 
> ```
[...]

PR #3502, opened 2024-03-19
support disk usage metrics for containerd factory

chengjoey:
query rootfs dir from contianerd status and get disk usage stat by rootfs dir `/run/containerd/io.containerd.runtime.v2.task/${contianerid}/rootfs`

set `hasFileSystem` to true for containerd when included disk usage metrics
added common fs handler for container rootfs


chengjoey:
[container_fs_usage_bytes metrics miss pod/container label when useing containerd](https://github.com/google/cadvisor/issues/3495)

container/docker/handler_test.go
```diff
@@ -169,7 +169,7 @@ func TestAddDiskStatsCheck(t *testing.T) {
 	fileSystems := []fs.Fs{fileSystem}
 
 	var fsStats info.FsStats
-	addDiskStats(fileSystems, nil, &fsStats)
+	fs.AddDiskStats(fileSystems, nil, &fsStats)
```
iwankgb, line 172: This test should be moved to `fs` module.

container/docker/handler_test.go
```diff
@@ -203,7 +203,7 @@ func TestAddDiskStats(t *testing.T) {
 	fileSystems := []fs.Fs{fileSystem}
 
 	// Act
-	addDiskStats(fileSystems, &fsInfo, &fsStats)
+	fs.AddDiskStats(fileSystems, &fsInfo, &fsStats)
```
iwankgb, line 206: Ditto.

container/containerd/client.go
```diff
@@ -130,6 +146,23 @@ func (c *client) TaskPid(ctx context.Context, id string) (uint32, error) {
...
 }
 
+func (c *client) RootfsDir(ctx context.Context) (string, error) {
+	r, err := c.runtimeService.Status(ctx, &runtimeapi.StatusRequest{Verbose: true})
+	if err != nil {
+		return "", err
+	}
+	configStr := r.Info["config"]
+	config := make(map[string]interface{})
+	if err := json.Unmarshal([]byte(configStr), &config); err != nil {
```
iwankgb, line 156: Could you unmarshal config into a struct? It will be easier to grasp what values are interesting.

iwankgb (changes requested):

container/docker/handler_test.go
chengjoey: done~

container/docker/handler_test.go
chengjoey: done~

container/containerd/client.go
chengjoey: The original definition of config is [here](https://github.com/containerd/containerd/blob/c17839bdc7bf8699badf796078afc91fd54de6ac/internal/cri/config/config.go#L411-L424)

I extracted the struct in config.go to container/containerd/config

silence8013:
Will this commit continue? Looking forward to it~

duanhongyi:
I am looking forward to this feature merging.

jianxiaoguo:
I eagerly anticipate the completion of this feature.

EamonZhang:
This feature is particularly important and I'm looking forward to it

wsszh:
Hi, I think the disk usage of a container should include the usage of log files in /var/log/pods.

chengjoey:
> Hi, I think the disk usage of a container should include the usage of log files in /var/log/pods.

hi @wsszh , The disk usage calculation of contianerd is similar to that of docker and podman. I think that several container factories currently do not include `/var/log/pods` in the calculation. This may be because cadvisor is more about container-level calculations rather than pods. It is also difficult to know which directory should be calculated because it is not fixed in /var/log/pods. although I think it should be calculated, the specific directory of the log cannot be obtained from the info interface of cri.

chengjoey:
> Hi, I think the disk usage of a container should include the usage of log files in /var/log/pods.

if a container should include the usage of log files, I would like to submit another PR to modify the container runtimes such as podman, docker, containerd, etc. This PR only does usage metric for containerd

wsszh:
> > Hi, I think the disk usage of a container should include the usage of log files in /var/log/pods.
> 
> hi @wsszh , The disk usage calculation of contianerd is similar to that of docker and podman. I think that several container factories currently do not include `/var/log/pods` in the calculation. This may be because cadvisor is more about container-level calculations rather than pods. It is also difficult to know which directory should be calculated because it is not fixed in /var/log/pods. although I think it should be calculated, the specific directory of the log cannot be obtained from the info interface of cri.

I see, and of course this PR is helpful enough for most circumstances!
PS: 老哥的博客地址好像访问不了了，方便给个联系方式向你请教学习吗~

chengjoey:
> PS: 老哥的博客地址好像访问不了了，方便给个联系方式向你请教学习吗~

wechat: backend-development

kxbin:
great work，Hope to speed up to merge in master

melikeiremguler:
Hi, will this pull request solve [this problem ](https://github.com/google/cadvisor/issues/2785#issuecomment-1204837267) ?

chengjoey:
> Hi, will this pull request solve [this problem ](https://github.com/google/cadvisor/issues/2785) ?

@melikeiremguler , Yes, this PR is to solve the issue of 2785.

gigilee:
Is this PR going to be merged?

Imlosing:
Looking forward to the early merge.

Imlosing:
> Looking forward to the early merge.

Or is there any alternative solution now?

kam1kaze:
@iwankgb How could we speed up merge process of this code? We are using it on production and it works perfect.

Imlosing:
> @iwankgb How could we speed up merge process of this code? We are using it on production and it works perfect.

@kam1kaze I also merged this PR and tried to use it, but I found that the actual value collected was greater than the writable layer capacity. Have you encountered this problem?

kamikaze:
please

desert3agle:
is it still in work or has been abandoned?

ttc0419:
@iwankgb could you take a look again?

till:
@dims maybe you can help review this? And re-open the ticket

dims:
@till if/when it gets rebased/updated the PR author can reopen this PR

container/containerd/client.go
```diff
@@ -166,6 +180,19 @@ func (c *client) Version(ctx context.Context) (string, error) {
 	return response.Version, nil
 }
 
+func (c *client) RootfsDir(ctx context.Context) (string, error) {
+	r, err := c.runtimeService.Status(ctx, &runtimeapi.StatusRequest{Verbose: true})
```
dims: please find another way. we can't be creating circular dependencies

mindw:
still relevant.

hugoprudente:
Extremely relevant

wsszh:
I recommend this project to everyone if you are waiting for this issue to be solved: https://github.com/utilitywarehouse/kube-summary-exporter

PR #3517, opened 2024-04-17
Add descriptive information for HousekeepingInterval and add a mechanism to lower the HousekeepingInterval 

dsxing:
In my testing Kubernetes environment, I observed that if the HousekeepingInterval of cadvisor is set very high by using --housekeeping-interval, for example, greater than 2 minutes, then obtaining container data via kubelet returns null.
image

Upon analyzing the cAdvisor code, I found that this is due to the fact that the **RecentStats** of containers only contain **one** data  in cadvisor's timed_store. point. _https://github.com/google/cadvisor/blob/master/manager/manager.go#L529_;
Consequently, it becomes impossible to calculate the CpuStats of containers， resulting in a nil value for stat.CpuInst(_https://github.com/google/cadvisor/blob/master/info/v2/conversion.go#L192, https://github.com/google/cadvisor/blob/master/info/v2/conversion.go#L234_);

When kubelet aggregates metrics data for all containers on a node, it filters container data. _https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cadvisor_stats_provider.go#L97_； Due to **stat.CpuInst=nil**，kubelet considers the container **Terminated**,_https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/cadvisor_stats_provider.go#L399_,resulting in a null value for 'containers' in the final metric data.

The reason why RecentStats of containers in cAdvisor's timed_store only contain one data point is because the HousekeepingInterval set by --housekeeping-interval exceeds the expiration period of data in timed_store， In kubelet, this expiration period is defined as 2 minutes by default so the memoryCache's maxAge is 2minutes. _https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cadvisor/cadvisor_linux.go#L55_；Therefore, each time housekeeping polls to write data to timed_store, **previous data has already expired and been deleted**._https://github.com/google/cadvisor/blob/master/utils/timed_store.go#L78_

Based on this analysis, I believe we should provide more detailed instructions in the flag to guide users to set it correctly. Additionally, since there is no enforced validation for the value set by --housekeeping-interval, could we introduce logic to proactively lower the HousekeepingInterval in the nextHousekeepingInterval function?


dsxing:
@iwankgb Is it possible to handle the code logic in this way, or should we just add a note for clarification?

PR #3522, opened 2024-04-24
sysfs: trim spaces in device hidden check

clwluvw:
The hidden file might have an endline, so the check for equals to 1 would not be true. Trimming the spaces would help to cover this case.

### DIFF
--- a/utils/sysfs/sysfs.go
+++ b/utils/sysfs/sysfs.go
@@ -229,10 +229,8 @@ func (fs *realSysFs) IsBlockDeviceHidden(name string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("failed to read %s: %w", devHiddenPath, err)
 	}
-	if string(hidden) == "1" {
-		return true, nil
-	}
-	return false, nil
+
+	return strings.TrimSpace(string(hidden)) == "1", nil
 }
 
 func (fs *realSysFs) GetBlockDeviceScheduler(name string) (string, error) {


iwankgb (approved):

clwluvw:
Hi @iwankgb - Can you please help here with why the CI is not happy?

ringerc:
If you see errors like `Failed to get disk map: open /sys/block/nvme0c0n1/dev: no such file or directory` but

```
# cat  /sys/block/nvme0c0n1/hidden    
1
```

then this might well be the problem. Check:

```
# od -x /sys/block/nvme0c0n1/hidden
0000000 0a31
0000002
```

Here, my `hidden` files have a leading linefeed, which means they won't match `== "1"`.

clwluvw:
> Here, my `hidden` files have a leading linefeed, which means they won't match `== "1"`.

This PR must actually have addressed such. Are you running a version >= v0.51.0?

ringerc:
@clwluvw Yes - I've been able to verify that Red Hat OpenShift  (v4.18) is currently shipping a `kubelet` `openshift-kubelet-4.18.0-202505060638.p0.gaf98ede.assembly.stream.el9` that embeds an older fork of cadvisor based on v0.49.0 (https://github.com/openshift/google-cadvisor/tree/v0.49.0-openshift-4.17-3) that omits this fix.

I raised https://github.com/openshift/google-cadvisor/pull/30 to ask them to backport the fix. Thanks very much for writing the fix.

I don't think this fix is entirely comprehensive - see https://github.com/google/cadvisor/issues/3693 for details. But it will prevent problems with hidden devices as are used in multi-path I/O systems, so it'll solve the immediate problem.

PR #3552, opened 2024-06-23
Expose Podman version info

riyad:
This exposes the Podman version and it's Docker API version. The fields were there, but they were always showing "Unknown".

This is how it looks after the patch:
cAdvisor - Podman versions

### DIFF
--- a/container/podman/podman.go
+++ b/container/podman/podman.go
@@ -108,7 +108,15 @@ func Status() (v1.DockerStatus, error) {
 		return v1.DockerStatus{}, err
 	}
 
-	return docker.StatusFromDockerInfo(*podmanInfo)
+	status, err := docker.StatusFromDockerInfo(*podmanInfo)
+	if err != nil {
+		return v1.DockerStatus{}, err
+	}
+
+	status.Version, _ = VersionString()
+	status.APIVersion, _ = APIVersionString()
+
+	return status, nil
 }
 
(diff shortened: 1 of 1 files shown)


iwankgb:
@riyad, you need to sign Google CLA.

container/podman/podman.go
```diff
@@ -108,7 +108,15 @@ func Status() (v1.DockerStatus, error) {
...
 	}
 
-	return docker.StatusFromDockerInfo(*podmanInfo)
+	status, err := docker.StatusFromDockerInfo(*podmanInfo);
+	if err != nil {
+		return v1.DockerStatus{}, err
+	}
+
+	status.Version, _ = VersionString()
+	status.APIVersion, _ = APIVersionString()
```
iwankgb, line 117: Can you log these errors, please?

riyad: I was ignoring them, because both fields are set to "Unknown" by  `docker.StatusFromDockerInfo()` above.
Similarly both functions also return "Unknown" in case of an error.
I've pushed https://github.com/google/cadvisor/pull/3552/commits/31c860f4a30cf4ab72218a0eefd3c5be3ef1e60b with error handling to make this behavor more explicit.
Please have a look. :smiley:

Creatone:
@riyad please rebase to current master

iwankgb (approved):

PR #3555, opened 2024-06-26
Add uninterruptible metric load d

j4ckstraw:
Currently container_cpu_load_average_10s only trace running tasks, we need uninterruptible metric to find out offline load too. Here I add a metric named load.d

see https://github.com/torvalds/linux/blob/master/kernel/cgroup/cgroup-v1.c#L724C1-L743C3

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -73,6 +73,7 @@ type containerData struct {
 	loadReader               cpuload.CpuLoadReader
 	summaryReader            *summary.StatsSummary
 	loadAvg                  float64 // smoothed load average seen so far.
+	loadDAvg                 float64 // smoothed load.d average seen so far.
 	housekeepingInterval     time.Duration
 	maxHousekeepingInterval  time.Duration
 	allowDynamicHousekeeping bool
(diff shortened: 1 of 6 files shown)


manager/container.go
```diff
@@ -633,6 +634,14 @@ func (cd *containerData) updateLoad(newLoad uint64) {
 	}
 }
 
+func (cd *containerData) updateLoadD(newLoad uint64) {
+	if cd.loadDAvg < 0 {
```
iwankgb, line 638: How is this variable initialized with negative value?

j4ckstraw: fixed.  loadDAvg initialized with -1 like loadAvg does.

iwankgb (approved):

iwankgb:
@j4ckstraw you may need to merge `master`, I think.

j4ckstraw:
> @j4ckstraw you may need to merge `master`, I think.

done. thank you for your comment.

j4ckstraw:
thanks @wenjianhn for the idea of load.d

PR #3565, opened 2024-07-11
Revert "reduce_logs_for_kubelet_use_crio"

### LINKED ISSUE #3538: K8s/cri-o broken for cgroup-driver != systemd
This:

https://github.com/google/cadvisor/blob/54dff2b8ccb147747d814b0ff3b4a3256dc569c0/container/crio/factory.go#L131-L134

checks for `.scope` suffix, but that is unique for cgroup-driver=systemd ([systemd.scope.5](https://www.man7.org/linux/man-pages/man5/systemd.scope.5.html)).

This causes K8s systems with `cri-o` to fail getting pod/container metrics in K8s v1.30+ if `--cgroup-driver=cgroupfs` is used (which is the default).

Ref https://github.com/cri-o/cri-o/issues/8034

uablrek:
This reverts commit b407dcd7447be18a4e83c1fa4973469dacc32572.

Fixes #3538

There may be a better way by detecting if cgroups are managed by systemd, but I have not enough knowledge of cadvisor, and not much time. So, the easy way out is to revert the commit causing the problem.

Further, I don't even know how to test it. I use K8s with cri-o, but cadvisor is vendored in K8s, so I don't know how to make K8s use this branch (I know howto build and test K8s from master though)

### DIFF
--- a/container/crio/factory.go
+++ b/container/crio/factory.go
@@ -32,9 +32,6 @@ import (
 // The namespace under which crio aliases are unique.
 const CrioNamespace = "crio"
 
-// The namespace suffix under which crio aliases are unique.
-const CrioNamespaceSuffix = ".scope"
-
 // The namespace systemd runs components under.
 const SystemdNamespace = "system-systemd"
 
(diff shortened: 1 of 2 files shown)


dims:
xref: https://github.com/google/cadvisor/pull/3457

uablrek:
I have verified that the revert fixes the problem by copying `cadvisor/container/crio/factory.go` from this PR to K8s `vendor/github.com/google/cadvisor/container/crio/factory.go`.
```
Server Version: v1.31.0-alpha.3.185+2d4514e1690dc5-dirty
crio version 1.29.2
```
My metrics test works. They fail for crio without this PR (but works with containerd)

iwankgb (approved):

Issue #3464, opened 2024-01-31
Segmentation violation in podman/factory.go


crrrazyden:
Context:

Ubuntu 22.04
Docker 24.0.5 (installed, containers present)
Podman 3.4.4
Cadvisor image v.0.48-test

Launching the advisor:
```bash
podman run \
  --pod=${POD_NAME} \
  --volume=/:/rootfs:ro \
  --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
  --volume=/var/run/podman:/var/run/podman:ro \
  --volume=/dev/disk:/dev/disk:ro \
  --volume=/dev/rbd:/dev/rbd:ro \
  --detach=true \
  --name=cadvisor \
  --privileged=true \
  --device=/dev/kmsg \
  --ulimit=host \
  --memory=512Mi \
  --cpus=0.5 \
  gcr.io/cadvisor/cadvisor-amd64:0.48-test \
  -listen_ip "127.0.0.1" \
  -port 8081
```

After several hours of work:
```
I0130 19:09:41.796163       1 manager.go:234] Version: {KernelVersion:5.15.0-78-generic ContainerOsVersion:Alpine Linux v3.16 DockerVersion: DockerAPIVersion: CadvisorVersion:0.48-test CadvisorRevision:d19df4f3}
I0130 19:09:41.798298       1 factory.go:219] Registration of the mesos container factory failed: unable to create mesos agent client: failed to get version
I0130 19:09:41.798460       1 factory.go:219] Registration of the containerd container factory failed: unable to create containerd client: containerd: cannot unix dial containerd api service: dial unix /run/containerd/containerd.sock: connect: no such file or directory
I0130 19:09:41.866494       1 factory.go:219] Registration of the crio container factory failed: Get "http://%2Fvar%2Frun%2Fcrio%2Fcrio.sock/info": dial unix /var/run/crio/crio.sock: connect: no such file or directory
I0130 19:09:41.867083       1 factory.go:219] Registration of the docker container factory failed: failed to validate Docker info: failed to detect Docker info: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
I0130 19:09:42.265240       1 plugin.go:89] Registering Podman factory
I0130 19:09:42.515637       1 factory.go:221] Registration of the podman container factory successfully
I0130 19:09:42.515691       1 factory.go:55] Registering systemd factory
[...]

PR #3566, opened 2024-07-17
address another nil ptr check in podmanFactory.CanHandleAndAccept()

char8:
We ran into the same issue as #3464, but adding the check against `cntr.State` was insufficient to prevent the nil ptr panic.

The `ContainerJSON` type returned by `InspectContainer` embeds a ptr to `ContainerJSONBase`, and it seemed this ptr was nil in the return, leading cadvisor to crash loop on startup.

I haven't looked into the specific JSON response podman is returning to trigger this, so there's potentially a bug in the podman API itself where it returned a nil error and an empty payload in response to the inspect call.

I do see the podman daemon returning 500 error codes for some inspect queries, and it appears `validateResponse` (container/podman/podman.go:41) doesn't handle 5xx status codes? so that may be the underlying issue here - podman returns a 5xx and empty body, and the json decoder unmarshals it to a nil.

Happy to add handling for 5xx to this PR as an added measure if that is appropriate.

### DIFF
--- a/container/podman/factory.go
+++ b/container/podman/factory.go
@@ -94,7 +94,7 @@ func (f *podmanFactory) CanHandleAndAccept(name string) (handle bool, accept boo
 	if err != nil {
 		return false, true, fmt.Errorf("error inspecting container: %v", err)
 	}
-	if ctnr.State == nil || !ctnr.State.Running {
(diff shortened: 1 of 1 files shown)


iwankgb (approved):

char8:
CLA signed - sorry for the delay

PR #3570, opened 2024-07-26
Update docker/docker dependency

ptodev:
This is needed so that we can [update OTel in Alloy](https://github.com/grafana/alloy/pull/1362).


ptodev:
Apologies, wrong repo :)

PR #3582, opened 2024-08-27
container/crio: fix network metrics collection logic

### LINKED ISSUE #3577: POD Networking metrics missing with crio-1.30 and kubelet-1.30
It seems that https://github.com/google/cadvisor/commit/eac1257f76a4a55bb2bf836f41a577a2fcb148a4 breaks network metric collection. At least, with crio-1.30 with defaults.

If I apply the simple diff, I can get container metrics again:

```diff
--- a/vendor/github.com/google/cadvisor/container/crio/handler.go
+++ b/vendor/github.com/google/cadvisor/container/crio/handler.go
@@ -155,6 +155,7 @@ func newCrioContainerHandler(
        // reported. This stops metrics being reported multiple times for each
        // container in a pod.
        metrics := common.RemoveNetMetrics(includedMetrics, cInfo.Labels["io.kubernetes.container.name"] != "POD")
+       metrics = includedMetrics
 
        libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootFs, cInfo.Pid, metrics)
 ```

The command used to test this:

```
curl -s -H "Authorization: bearer $TOKEN" -k https://192.168.111.122:10250/metrics/cadvisor | grep "^container_network_transmit_packets_total"
```

Perhaps, cri-o does not have or keep a `POD` named container?

sohankunkerkar:
Fixes #3577 
This change modifies the logic for gathering network metrics in the CRI-O handler to ensure that metrics are collected from all containers. Since all containers share the same network namespace, gathering metrics from every container is crucial to account for scenarios where the infra container may have empty network metrics.

### Future improvements:
We can use a caching mechanism to maintain a designated container per pod. This cache will allow us to track which container to use for metrics collection efficiently. If the designated container dies, we should dynamically find another running container within the same pod.

### DIFF
--- a/container/crio/handler.go
+++ b/container/crio/handler.go
@@ -148,13 +148,13 @@ func newCrioContainerHandler(
 		Namespace: CrioNamespace,
 	}
 
-	// Find out if we need network metrics reported for this container.
-	// Containers that don't have their own network -- this includes
-	// containers running in Kubernetes pods that use the network of the
-	// infrastructure container -- does not need their stats to be
-	// reported. This stops metrics being reported multiple times for each
-	// container in a pod.
-	metrics := common.RemoveNetMetrics(includedMetrics, cInfo.Labels["io.kubernetes.container.name"] != "POD")
+	// In a Kubernetes pod, all containers share the same network namespace.
+	// If the infrastructure container has empty network metrics, the CRI-O
+	// handler uses a running container in the pod to gather the necessary
+	// metrics. Therefore, it is essential to collect network metrics from
+	// all containers to ensure that if there is any running container in
+	// the pod, we capture the required network metrics for accurate reporting.
+	metrics := includedMetrics
 
 	libcontainerHandler := containerlibcontainer.NewHandler(cgroupManager, rootFs, cInfo.Pid, metrics)
 


sohankunkerkar:
cc @haircommander @mrunalp

haircommander:
LGTM. thanks! @iwankgb PTAL

kannon92:
Can you add a Fixes to the description so we correctly link what issue this solves?

kannon92:
Fixes https://github.com/google/cadvisor/issues/3577.

kolyshkin:
So, if there are many containers in the pod, we collect stats for them all (and they are the same)?

haircommander:
> So, if there are many containers in the pod, we collect stats for them all (and they are the same)?

ah I was slightly wrong about how we're falling back to the pod. I think there's another bug here. The pod cgroup instance should have the PID of one of the containers (reported by cri-o) and should replace that pid if the container exits. We shouldn't need to collect the network metrics on pod level for them to show up...

haircommander:
/close

I believe https://github.com/google/cadvisor/commit/ca820b635076e6d7bfb85b39202836157966cb7b fixes this better

iwankgb:
I don't think that removing network stats filtering completely is an acceptable solution. I also believe that @kolyshkin is right: same metrics might be reported for more than one container.

PR #3592, opened 2024-09-12
crio: only collect required metrics for infra container

haircommander:
as we'll only use the network metrics from it


container/crio/handler.go
```diff
@@ -148,13 +148,22 @@ func newCrioContainerHandler(
...
 	}
 
-	// Find out if we need network metrics reported for this container.
-	// Containers that don't have their own network -- this includes
-	// containers running in Kubernetes pods that use the network of the
-	// infrastructure container -- does not need their stats to be
-	// reported. This stops metrics being reported multiple times for each
-	// container in a pod.
-	metrics := common.RemoveNetMetrics(includedMetrics, cInfo.Labels["io.kubernetes.container.name"] != "POD")
+	nonNetworkMetrics := common.RemoveNetMetrics(includedMetrics, true)
```
kolyshkin, line 151: While at it, you may remove the second argument of `RemoveNetMetrics` now.

haircommander: it looks like containerd/docker use them

haircommander:
@iwankgb @bobbypage @mrunalp @kolyshkin PTAL

container/crio/handler.go
```diff
@@ -148,13 +148,22 @@ func newCrioContainerHandler(
...
+	// If it's a non-infra container, then we can exclude network metrics,
+	// because in kubernetes a container always shares the network namespace
+	// with the infra container.
+	metrics := nonNetworkMetrics
+	// If this is the infra container cgroup, then we only need to collect network metrics
+	// as the rest will be discarded.
+	if cInfo.Labels["io.kubernetes.container.name"] == "POD" {
+		// Only include the non-non-network metrics. This way, we can tell if there are no network metrics
+		metrics = includedMetrics.Difference(nonNetworkMetrics)
```
bitoku, line 160: Would adding `Intersect` in MetricsSet make more sense?

haircommander:
I'm not actually positive this helps
/hold
for now

iwankgb (approved):

iwankgb (changes requested):
Blocking per @haircommander's comment.

haircommander:
come to think of it, this will improve some, but it won't materially improve, we can keep or drop this, I think it techincally will save a bit of time but not really enough to notice

Issue #3606, opened 2024-10-16
Kubernetes cadvisor test suite is failing


kannon92:
https://github.com/kubernetes/kubernetes/issues/127818

Please take a look when you can and see what we should do about this job.

dims:
xref: https://github.com/google/cadvisor/pull/3608

kannon92:
@dims fixed this.

/close

PR #3608, opened 2024-10-18
Avoid linting in Prow CI jobs

dims:
Linting is failing in prow CI jobs. We should not be running linters in periodic CI jobs, 

By definition in Makefile, `all` is defined as:
```
all: presubmit build test
```

Since linters SHOULD be run in presubmits, let's skip them in https://testgrid.k8s.io/sig-node-cadvisor#cadvisor-e2e by limiting the targets to `build test`


dims:
xref: https://github.com/google/cadvisor/issues/3606

PR #3609, opened 2024-10-18
Drop unused setNetworkStats

dims:
Avoid referring to the things in the library's root directory: `"github.com/opencontainers/runc/libcontainer"` since we never even threaded in support for network stats. We are creating `libcontainer.Stats` with cgroup stats but network stats are empty.

https://github.com/google/cadvisor/blame/96c346ed6af33ecc856f82dfd9ef6b0fd6b66949/container/libcontainer/handler.go#L92-L95

PS: This will eventually help us minimize the things we use in runc.


container/libcontainer/handler.go
```diff
@@ -939,8 +913,5 @@ func newContainerStats(libcontainerStats *libcontainer.Stats, includedMetrics co
 			setCPUSetStats(s, ret)
 		}
 	}
-	if len(libcontainerStats.Interfaces) > 0 {
-		setNetworkStats(libcontainerStats, ret)
```
cwangVT, line 943: Just to make myself more clear about the removal of `setNetworkStats`.

Here, we believe it is safe to remove the function `setNetworkStats` because:

1. It only appears in this line in the old code, and
2. It actually never be called, because the `libcontainerStats` was created in https://github.com/google/cadvisor/pull/3609/files#diff-28d7cc5aa0eb19db1d3a6d3d16080b4a1940e691ec501ccbd2af273de8034508L92:
```
	libcontainerStats := &libcontainer.Stats{
		CgroupStats: cgroupStats,
	}
```
So the `libcontainerStats.Interfaces` in this line should always be empty, which makes the `if condition` always false. 

Is my understanding correct?

cc: @bobbypage

dims: Exactly @cwangVT ! bingo!

bobbypage: ack, it makes sense. Thanks @dims for the fix and @cwangVT for reviewing.

bobbypage:
LGTM

cwangVT (approved):

PR #3620, opened 2024-11-05
Update to dependency-free compute/metadata

liggitt:


liggitt:
cc @dims @bobbypage

liggitt:
done in https://github.com/google/cadvisor/pull/3622

PR #3626, opened 2024-11-15
machine: use `unix.ByteSliceToString`

tklauser:
Use the convenience function `ByteSliceToString` from `x/sys/unix` to convert `Utsname` members to string.

### DIFF
--- a/machine/info.go
+++ b/machine/info.go
@@ -15,7 +15,6 @@
 package machine
 
 import (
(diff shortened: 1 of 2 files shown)


tklauser:
Not sure why this was closed. I think the change would still apply. Reopened as https://github.com/google/cadvisor/pull/3775.

Issue #3528, opened 2024-05-09
To reduce the lock pressure on the kernfs file system, we should reduce the frequency of cgroup reads.


xigang:
Intro:
We call cadvisor subcontainers API to obtain the data of all cgroup containers. With the increase of QPS, the frequency of cgroup read will increase, which will aggravate the pressure of kernfs file system lock and even cause machine hang.

improve:
Do we need to read the cgroup less frequently when we call `GetInfo` to fetch `ContainerInfo` data by caching it?

https://github.com/google/cadvisor/blob/54dff2b8ccb147747d814b0ff3b4a3256dc569c0/manager/manager.go#L543C19-L543C47
```
func (m *manager) containerDataToContainerInfo(cont *containerData, query *info.ContainerInfoRequest) (*info.ContainerInfo, error) {
	// Get the info from the container.
	cinfo, err := cont.GetInfo(true)  
	if err != nil {
		return nil, err
	}

	stats, err := m.memoryCache.RecentStats(cinfo.Name, query.Start, query.End, query.NumStats)
	if err != nil {
		return nil, err
	}

	// Make a copy of the info for the user.
	ret := &info.ContainerInfo{
		ContainerReference: cinfo.ContainerReference,
		Subcontainers:      cinfo.Subcontainers,
		Spec:               m.getAdjustedSpec(cinfo),
		Stats:              stats,
	}
	return ret, nil
}
```

If the call `GetInfo (shouldUpdateSubcontainers bool)` function `shouldUpdateSubcontainers` parameter is set to true, The `updateSpec()` function is called on each request to read the cgroup data, which increases the pressure on kernsfs Lock.

https://github.com/google/cadvisor/blob/54dff2b8ccb147747d814b0ff3b4a3256dc569c0/manager/container.go#L172
```
func (cd *containerData) GetInfo(shouldUpdateSubcontainers bool) (*containerInfo, error) {
	// Get spec and subcontainers.
	if cd.clock.Since(cd.infoLastUpdatedTime) > 5*time.Second || shouldUpdateSubcontainers {
		err := cd.updateSpec()
		if err != nil {
			return nil, err
		}
		if shouldUpdateSubcontainers {
			err = cd.updateSubcontainers()
			if err != nil {
				return nil, err
			}
		}
		cd.infoLastUpdatedTime = cd.clock.Now()
	}
	cd.lock.Lock()
	defer cd.lock.Unlock()
	cInfo := containerInfo{
[...]

PR #3633, opened 2024-12-08
Reduce cgroups read frequency to avoid kernel kernfs clock pressure

xigang:
Fixes #https://github.com/google/cadvisor/issues/3528

When `GetInfo` is called with false parameter, it now reads cgroups data once every 5 seconds and gets data from containerData cache at other times. This change reduces the frequency of reading cgroups to prevent potential Linux 
machine hangs caused by kernel kernfs clock pressure.

The change affects the `containerDataToContainerInfo` method in manager.go which is used to retrieve container information. By caching and reusing the cgroups data between reads, we reduce system load while still maintaining 
reasonably up-to-date container metrics.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -541,8 +541,10 @@ func (m *manager) GetContainerInfoV2(containerName string, options v2.RequestOpt
 }
 
 func (m *manager) containerDataToContainerInfo(cont *containerData, query *info.ContainerInfoRequest) (*info.ContainerInfo, error) {
-	// Get the info from the container.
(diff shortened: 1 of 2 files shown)


xigang:
@iwankgb Is there an issue with this workflow? It runs fine on my local setup.

xigang:
@dims For approve.

dims:
still need this @xigang ?

xigang:
@dims This PR is still needed, but the testing hasn't been completed yet.

PR #3639, opened 2024-12-22
Add more memory cgroup metrics

major1201:
Currently memory metrics are not enough for us to debug in various scenarios in which we need more clues.  I added several new memory metrics:

- container_memory_shmem
- container_memory_dirty
- container_memory_writeback
- container_memory_unevictable
- container_memory_sock


major1201:
@dims Hi, could you give a reason of the close?

dims:
@major1201 there is no active set of maintainers working on cadvisor

Issue #3081, opened 2022-03-17
Metric `container_memory_working_set_bytes` includes `slab` reclaimable memory


cyrus-mc:
I ran into a somewhat unique situation in which a POD had very high slab memory - high as in 1.1GB worth. In terms of anonymous and active file the memory usage was only around 25MB. Working set calculate shows 1.1GB due to the fact that slab reclaimable memory isn't subtracted from `workingSet` calculation.

Working set calculation

```
  ret.Memory.Usage = s.MemoryStats.Usage.Usage
  workingSet := ret.Memory.Usage
  if v, ok := s.MemoryStats.Stats[inactiveFileKeyName]; ok {
    if workingSet < v {
      workingSet = 0
    } else {
      workingSet -= v
    }
  }
  ret.Memory.WorkingSet = workingSet
```

Where `MemoryStats.Usage.Usage` is the value from `memory.current` (cgroup v2) or `memory.usage_in_bytes` (cgroup v1). Memory statistics (memory.stat) contains the following fields:

```
anon 663552
file 10313728
kernel_stack 49152
...
inactive_anon 573440
active_anon 32768
inactive_file 5066752
active_file 5246976
unevictable 0
slab_reclaimable 1232589368
slab_unreclaimable 128408
slab 1232717776
...
```

Of which `slab_reclaimable` is memory that can be reclaimed by the OS when needed. Should we be subtracting this value when calculating `workingSet`?

bwplotka:
Good question! I don't want to overcrowd this issue, but why we don't subtract `inactive_anon` as well? 

Rationale: https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt#:~:text=inactive_anon%09%2D%20%23%20of%20bytes%20of%20anonymous%20and%20swap%20cache%20memory%20on%20inactive%0A%09%09LRU%20list.

cyrus-mc:
@bwplotka I can understand why `inactive_anon` isn't subtracted since most container clusters don't run with swap so that memory can't be swapped out and thus is part of the working set. 

Is `slab_reclaimable` the same? Since it is more of a cache thing it can be reclaimed by the OS when it needs memory.

bwplotka:
Yea, agree,  something is off, but for me, it's not really slab. I can reproduce this problem with a large number of open file descriptors. The WSS shows quite large memory usage:

WSS: 
[...]

Issue #3286, opened 2023-03-31
Proposal to update container_memory_usage_bytes to Cache+RSS


HonakerM:
## TLDR

Hello, I would like to propose updating `container_memory_usage_bytes` from reading `memory.usage_in_bytes`/`memory.current` to calculating it manually from the values located in `memory.stat` (specifically Cache+RSS). The reason being that there can be a discrepancy  with usage_in_bytes that's exacerbated on multi-core systems.

### Metrics
My original investigation started when I was trying to debug the memory usage of an unrelated application running in a Kubernetes pod and was watching the `container_memory_working_set_bytes` metric as suggested by [the kubernetes docs](https://kubernetes.io/docs/tasks/debug/debug-cluster/resource-metrics-pipeline/#memory). I wanted to find out the source of this value which lead me to [cadvisor GetStats](https://github.com/google/cadvisor/blob/ce07bb28eadc18183df15ca5346293af6b020b33/container/libcontainer/handler.go#L76) which told me it was `container_memory_usage_bytes - inactive file`. My question then became where does `container_memory_usage_bytes ` come from? The cadvisor code calls out to [runc's GetStats](https://github.com/opencontainers/runc/blob/a187c84e42015402263167a2c1007b5c62d494f3/libcontainer/cgroups/fs/memory.go#L142) which answered my question: the information is gathered from files in the `/sys/fs/cgroup/memory` directory specifically `memory.stat` and either  `memory.usage_in_bytes` or `memory.current` depending on cgroup version.

### Linux Kernel
Throughout this research I ran into a few other cadvisor memory issues like [#3197](https://github.com/google/cadvisor/issues/3197) and [#3081](https://github.com/google/cadvisor/issues/3081) which discuss these values and what we should be subtracting. This made me assume that `usage_in_bytes` could be calculated from `stat` file so now I was curious about the calculation. That lead me to section[5.5 in the Kernel Docs](https://docs.kernel.org/admin-guide/cgroup-v1/memory.html?highlight=cgroups#usage-in-bytes) which says the following:

>### 5.5 usage_in_bytes
>
[...]

PR #3641, opened 2025-01-05
exclude cgroups slab_reclaimable from container_memory_working_set_bytes

CharlieR-o-o-t:
Related to issue: #3081

"slab_reclaimable" should be excluded from container_memory_working_set_bytes. 

"slab_reclaimable" is only unused cached kernel structured, which can be freed on mem pressure.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -848,6 +848,12 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 			workingSet -= v
 		}
 	}
+
+	// Exclude "slab_reclaimable" from workingSet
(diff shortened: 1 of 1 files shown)


asbachb:
@cwangVT do you have any opinion on that change?

fabiand:
@dims it would be great if you could give this PR a look )I think there have been similar ones in the past).

For context see https://github.com/google/cadvisor/issues/3081#issuecomment-2490952322 , similar #3286

tldr; today, workingset contains reclaimable memory, which then in turn is leading to incorrect decisions.

CharlieR-o-o-t:
Guys, PR 7 months old. Is anybody able to take a look?

Issue #3632, opened 2024-12-06
How to generate snapshot tests for `build/config/plain.sh`?


cafkafk:
I'm working on a PR adding more of the cgroup metrics to cadvisor (since k8s may e.g. OOM from kernel memory exhaustion without ever reaching memlimits, this is super useful in large production environments).

However, I'm stuck trying to regenerate the snapshot tests that seem to be run through `build/config/plain.sh`.

It is possible that I'm missing something obvious, but I haven't had any luck so far finding any place describing how to do this.

If someone explains it, I can probably add documentation to `docs/development/` in a PR.

cafkafk:
I'll close for now since I'll not work more on this. If anyone wants to take over please open a separate issue.

PR #3642, opened 2025-01-06
feat: group memory.stats sock metric

cafkafk:
This adds the cgroup stat `sock` from the `memory.stats` metric to
cAdvisor.

The motivation is that we've seen numerous examples at DBC Digital of
application developers creating applications that exhaust socket memory,
e.g. by accidentally creating too many TCP connections and not closing
them, or keeping around a few large allocations, or many other such
issues.

Because cAdvisor currently doesn't report socket memory usage, this has
been hard to monitor, and will only be observed when the OOM killer is
reached.

By adding this metric, it will be possible to proactively handle socket
memory exhaustion (which is really kernel memory exhaustion), before it
becomes a potential incident, and to create alerting and enhance

Signed-off-by: Christina Sørensen <chrs@dbc.dk>

---

Notice: I've been unable to figure out how to regenerate the snapshot tests,
I've opened an issue #3632 for this, but have yet to recieve any replies.

I'm hoping making this PR will bring more attention to this change, so it can
recieve feedback.

### DIFF
--- a/cmd/internal/storage/influxdb/influxdb.go
+++ b/cmd/internal/storage/influxdb/influxdb.go
@@ -68,6 +68,8 @@ const (
 	serMemorySwap string = "memory_swap"
 	// Size of memory mapped files in bytes
 	serMemoryMappedFile string = "memory_mapped_file"
+	// Size of socket memory in bytes
+	serMemorySocket string = "memory_socket"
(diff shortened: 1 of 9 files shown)


cafkafk:
I'm no longer available to work on this.

PR #3649, opened 2025-01-26
Add Pressure Stall Information Metrics

xinau:
issues: https://github.com/google/cadvisor/issues/3052, https://github.com/google/cadvisor/pull/3083, https://github.com/kubernetes/enhancements/issues/4205

This change adds metrics for pressure stall information, that indicate
why some or all tasks of a cgroupv2 have waited due to resource
congestion (cpu, memory, io). The change exposes this information by
including the _PSIStats_ of each controller in it's stats, i.e.
_CPUStats.PSI_, _MemoryStats.PSI_ and _DiskStats.PSI_.

The information is additionally exposed as Prometheus metrics. The
metrics follow the naming outlined by the prometheus/node-exporter,
where stalled eq full and waiting eq some.

```
container_pressure_cpu_stalled_seconds_total
container_pressure_cpu_waiting_seconds_total
container_pressure_memory_stalled_seconds_total
container_pressure_memory_waiting_seconds_total
container_pressure_io_stalled_seconds_total
container_pressure_io_waiting_seconds_total
```

This change is a rebase and resolve of the comments the work done in #3083.


xinau:
@rexagod, @SuperQ Could you please give this a review and advise me how to get this change merged.

cmd/go.mod
```diff
@@ -129,3 +129,4 @@ require (
 	google.golang.org/protobuf v1.36.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
```
SuperQ, line 132: Extraneous extra line.

```suggestion
```

metrics/prometheus.go
```diff
@@ -1746,6 +1746,54 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
...
 	}
 
+	if includedMetrics.Has(container.PressureMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:      "container_pressure_cpu_stalled_seconds_total",
+				help:      "Total time duration no tasks in the container could make progress due to CPU congestion.",
+				valueType: prometheus.CounterValue,
+				getValues: func(s *info.ContainerStats) metricValues {
+					return metricValues{{value: float64(s.Cpu.PSI.Full.Total) / 1000.0 / 1000.0, timestamp: s.Timestamp}}
```
SuperQ, line 1756: There's a helper function, `func asNanosecondsToSeconds(v uint64) float64`, let's add a similar one for this conversion.

```go
// asMicrosecondsToSeconds converts microseconds into a float64 representing seconds.
func asMicrosecondsToSeconds(v uint64) float64 {
  return float64(v) / float64(time.Microsecond)
}
```

Then we can simplify to:

```suggestion
					return metricValues{{value: asMicrosecondsToSeconds(s.Cpu.PSI.Full.Total), timestamp: s.Timestamp}}
```

SuperQ (changes requested):

SuperQ:
Looking great so far, the metric names and other conventions look fine.

metrics/prometheus.go
```diff
@@ -1746,6 +1746,54 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 		})
 	}
 
+	if includedMetrics.Has(container.PressureMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:      "container_pressure_cpu_stalled_seconds_total",
```
SuperQ, line 1752: According to [the docs](https://facebookmicrosites.github.io/cgroup2/docs/pressure-metrics.html), there is no "Full" for CPU. Just "Some". We can drop this metiric as it will always be zero.

metrics/prometheus.go
xinau: Done.

metrics/prometheus.go
xinau: Great catch.

xinau:
@SuperQ Thanks for the quick review. I've added the improvements.

metrics/prometheus.go
```diff
@@ -33,6 +33,11 @@ import (
 // asFloat64 converts a uint64 into a float64.
 func asFloat64(v uint64) float64 { return float64(v) }
 
+// asMicrosecondsToSeconds converts nanoseconds into a float64 representing seconds.
+func asMicrosecondsToSeconds(v uint64) float64 {
+	return float64(v) / float64(time.Millisecond)
```
SuperQ, line 38: I wonder if we should not use `time` constants, as this is slightly confusing having `time.Millisecond` here.

```suggestion
	return float64(v) / 1e6
```

xinau: Seems sensible, I've updated both functions.

metrics/prometheus.go
SuperQ: So, I may be wrong about this. The documentation that I can find has conflicting information. In the other PR, [this comment](https://github.com/google/cadvisor/pull/3083#discussion_r1377422274) suggests that `full` is valid for cgroup CPU pressure.

I can confirm this locally:

```console
[/sys/fs/cgroup/user.slice] $ cat cpu.pressure 
some avg10=0.00 avg60=0.00 avg300=0.07 total=425967721
full avg10=0.00 avg60=0.00 avg300=0.00 total=177284511
```

I also checked the kernel docs, but it still isn't clear. :confused:

SuperQ (approved):

xinau:
@SuperQ I'm going take a look at the CPU PSI metrics again today. It seems that the CPU PSI full metric can be neq 0. I've stumbled upon this while reading https://github.com/kubernetes/enhancements/issues/5062

xinau:
@SuperQ I'm going to re-add the CPU full metric, as it's actively being reported by the kernel for cgroups.

```
* Naturally, the FULL state doesn't exist for the CPU resource at the
* system level, but exist at the cgroup level, means all non-idle tasks
* in a cgroup are delayed on the CPU resource which used by others outside
* of the cgroup or throttled by the cgroup cpu.max configuration.
```
See
https://lore.kernel.org/all/20210303034659.91735-2-zhouchengming@bytedance.com/
https://lore.kernel.org/all/20220408121914.82855-1-zhouchengming@bytedance.com/

rexagod:
Thank you for your work (and investigation) on this, @xinau!

Not sure but after a quick look I can see we dropped [`container_%s_psi_avg%s_ratio`](https://github.com/google/cadvisor/pull/3083/files#diff-583dd1a38478c42e7ee4f90a9c3dfb5fd8a07b82f57d4ed24fa6a98a5951a4e7R1804) here, was this intentional?

SuperQ:
@rexagod Yup. With Prometheus we can derive arbitrary averages as they're just `rate(container_..._total[Xm])`.

xinau:
@rexagod, @SuperQ all good from my side now.

rexagod (approved):
@dims Could you please approve the pending workflow here, or ping someone who could? The patch builds on top of the original PR while additionally following the community guidelines, and looks good to go in.

pacoxu:
> @dims Could you please approve the pending workflow here, or ping someone who could? The patch builds on top of the original PR while additionally following the community guidelines, and looks good to go in.

Kindly ping @dims.

dims:
this is waiting for google maintainers like @cwangVT - i help take care of dependencies mostly (not features!). Also the Github hooks are broken, so the prow based ci jobs don't really work :(

metrics/prometheus.go
```diff
@@ -33,9 +33,14 @@ import (
...
 func asFloat64(v uint64) float64 { return float64(v) }
 
+// asMicrosecondsToSeconds converts nanoseconds into a float64 representing seconds.
+func asMicrosecondsToSeconds(v uint64) float64 {
+	return float64(v) / 1e6
+}
+
 // asNanosecondsToSeconds converts nanoseconds into a float64 representing seconds.
 func asNanosecondsToSeconds(v uint64) float64 {
-	return float64(v) / float64(time.Second)
```
haircommander, line 38: the old way reads as slightly easier to understand to me. What motivates this change?

metrics/prometheus.go
haircommander: this is only true for the top level cgroup, but not for children AFAIU https://docs.kernel.org/accounting/psi.html#pressure-interface

metrics/prometheus.go
```diff
@@ -1746,6 +1751,54 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 		})
 	}
 
+	if includedMetrics.Has(container.PressureMetrics) {
+		c.containerMetrics = append(c.containerMetrics, []containerMetric{
+			{
+				name:      "container_pressure_cpu_stalled_seconds_total",
```
haircommander, line 1757: I'm trying to reason about which metrics make sense to emit. It's interesting we emit the totals but don't emit the 10/60/300... I almost wonder if we should add the structure pieces first and discuss actual metrics we emit after..

metrics/prometheus.go
```diff
@@ -1746,6 +1751,54 @@ func NewPrometheusCollector(i infoProvider, f ContainerLabelsFunc, includedMetri
 		})
 	}
 
+	if includedMetrics.Has(container.PressureMetrics) {
```
haircommander, line 1754: I can almost see these as also being nested under cpu/memory/disk metrics. I am not sure if there is precedent for this, but maybe add both a PressureMetrics for included metrics, as well as check the respective other metric? 

```

	if includedMetrics.Has(container.PressureMetrics) && includedMetrics.Has(container..CPUMetrics) {
report CPU pressure metrics
```

metrics/prometheus.go
SuperQ: As mentioned in the review comments, it's confusing when you compare `asMicrosecondsToSeconds()`. The base unit is `time.Nanosecond`, so you would have to use `time.Millisecond` to convert microseconds to seconds.

By simply using the float factor, it's consistent to read both functions.

metrics/prometheus.go
SuperQ: No, we only need the totals, as other metrics are derived values. With the totals the end user can derive arbitrary intervals. For example `rate(container_pressure_cpu_stalled_seconds_total[60s])`.

metrics/prometheus.go
xinau: @haircommander I couldn't find any precedent for this. I'll add a check for each of PSI resource (cpu, memory, io). I can't find a strong argument against or for adding such a check.

SuperQ: I don't see why having pressure metrics needs to depend on other metrics. Each metric dataset is independent of each other.

xinau: @SuperQ as PSI metrics are reported as part of the cpu/memory/io controller it might make sense to not report them when a user actively decides against getting metrics for one of these controllers.

It's a bit strange in the case of _io_ as it's reported for block io not only disk.

Like pointed out before I'm undecided about this, it might also make sense to defer this decision to a later point in time when a real use-case arises. This then raises the question which decision provides more backwards compatibility.

haircommander: we can always emit differently in the future, I think this makes sense for now.

metrics/prometheus.go
haircommander: fair, I am not sure that prom query would get exactly the same but true that you could reconstruct those intervals from the total.

haircommander:
I can't remember if I'm powerful enough to do this
/ok-to-test

this LGTM, none of my notes need addressing :)

metrics/prometheus.go
SuperQ: Yea, it's never going to be exactly the same because it depends on exactly the timestamps and values involved.

But it will be within the same tolerance over time.

cwangVT (approved):

PR #3666, opened 2025-03-04
Bump to newer opencontainers/image-spec @ v1.1.1

dims:
There is a tag, can we please switch to it. newer https://github.com/opencontainers/image-spec/releases/tag/v1.1.1

related to some longer term cleanup we identified in https://github.com/kubernetes/kubernetes/issues/128572#issuecomment-2457145817. things need to land here, then back into kubernetes.


dims:
@iwankgb @mrunalp can you please hit the github approve once this goes green. thanks!

PR #3667, opened 2025-03-04
deps: switch to github.com/opencontainers/cgroups, bump runc 

kolyshkin:
This used to be in github.com/opencontainers/runc/libcontainer/cgroups
but was recently moved to a new home in a separate repository.


kolyshkin:
@dims right, thanks!

PR #3674, opened 2025-03-25
Fix grammar mistake in factory logging

louis-lau:
This is kind of a dumb and small PR (luckily that also means it's easy to review), but the obviously wrong sentence I noticed in the logs bothered me :)


louis-lau:
It's just a spelling mistake... I did sign the CLA.

PR #3681, opened 2025-04-04
fix: fix call Errorf with wrong err

alingse:


iwankgb (approved):

iwankgb:
I would love to merge it, but there is still e2e job that won't run.

PR #3676, opened 2025-03-26
Updating docker dependency version to https://github.com/google/cadvisor:v28.0.2+incompatible(latest) to fix security vuln

pbettadapura:
We found vulnerabilities in Go Docker module v26.1.4:

[cadvisor/cmd/go.mod](https://github.com/google/cadvisor/blob/f6e31a3cff918285fd74cb1f20d0af32c3554a68/cmd/go.mod#L67)

Line 67 in [f6e31a3](https://github.com/google/cadvisor/commit/f6e31a3cff918285fd74cb1f20d0af32c3554a68)

 github.com/docker/docker v26.1.4+incompatible // indirect 
More details:
GO (Go) Security Update for github.com/docker/docker (https://github.com/advisories/GHSA-v23v-6jw2-98fq)

A security vulnerability has been detected in certain versions of Docker Engine, which could allow an attacker to bypass authorization plugins (AuthZ) under specific circumstances. The base likelihood of this being exploited is low. This advisory outlines the issue, identifies the affected versions, and provides remediation steps for impacted users.

Refer to Github security advisory [https://github.com/advisories/https://github.com/advisories/GHSA-v23v-6jw2-98fq] https://github.com/advisories/GHSA-v23v-6jw2-98fq for updates and patch information. Patch: Following are links for downloading patches to fix the vulnerabilities: [https://github.com/advisories/https://github.com/advisories/GHSA-v23v-6jw2-98fq] https://github.com/advisories/GHSA-v23v-6jw2-98fq:github.com/docker/docker

Fix for this vuln is to update github.com/docker/docker v26.1.4 to at least v26.1.5](https://github.com/advisories/GHSA-v23v-6jw2-98fq).

PR #3692, opened 2025-06-02
Bump dependencies to latest (June 2, 2025)

dims:
it's been a while since we bumped things we consume. Let's try doing that for next release.


dims:
xref: https://github.com/google/cadvisor/pull/3676

PR #3699, opened 2025-06-16
Let us try to use ghcr.io for container images

dims:
Going to merge and experiment

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -1,0 +1,72 @@
+name: Publish Container Image
+
+on:
+  push:
+    tags:
+      - 'v*'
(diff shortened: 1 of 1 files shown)


VaismanLior:
Hey @dims 
Looks like this PR changed the tagging convention of the cadvisor images in the latest release.
The latest version is tagged as `ghcr.io/google/cadvisor:0.54.0` without the `v` prefix while previous versions had it.
Is this on purpose?

dims:
this is what is coded in our scripts now, so let's just leave it at that?

VaismanLior:
I would fix it to not break backwards compatibility for helm deployments etc.

PR #3717, opened 2025-07-30
Docs: remove note on Debian not supporting memory controller

WilliamDEdwards:
Since Debian 9, the memory controller is enabled by default. As that version has been EOL for 3 years (2022-07-01), I think this note can be removed.


WilliamDEdwards:
I'm not signing a CLA because it requires a Google account, something that open source contributors should not be required to have.

PR #3719, opened 2025-08-05
Fix tcpMaxConn description in comments

l2dy:
According to RFC 2012, the value is -1 in entities where the maximum number of connections is dynamic.


l2dy:
@dims Could you help review this PR?

PR #3720, opened 2025-08-05
Fix comments for TCP prune counters

l2dy:
ref: https://docs.kernel.org/networking/snmp_counter.html#prune


l2dy:
@dims Could you help review this PR?

PR #3729, opened 2025-09-16
feat(summary): add count in percentiles

sambhav-jain-16:
To measure the quality of the derived metric, it is necessary to also store the count of the samples from which the percentile calculation is being done. This commit aims to add this feature in cadvisor.


sambhav-jain-16:
Hi @cwangVT 
I have raised this PR for another issue we are working on which concerns `cadvisor`. PTAL

sambhav-jain-16:
Hi @dims 
Can you also PTAL at this PR ? Thanks

dims:
@sambhav-jain-16 please rebase, also in your experience, will this break any existing folks that consume this percentile? (as we added a new field?)

sambhav-jain-16:
@dims done

> also in your experience, will this break any existing folks that consume this percentile?

Since it is a change in struct, it shouldn't break anything. This change is already deployed in our infra and is running without breaking during rollout.

dims:
@sambhav-jain-16 thanks for confirming

PR #3740, opened 2025-10-13
Update README.md with latest Docker image version and registry

### LINKED ISSUE #3712: Missing docker image for v0.53.0
I cannot find in repo version 0.53.0
Also noticed 0.52.0 has a new creation date than 0.52.1
Not sure if I am looking in the wrong place or there is an issue in the image creation pipeline

Image

eneiss:
Bump image version to 0.53.0 and update container registry to ghcr.io in README.md

Fixes #3712

Let me know if you would prefer using a variable for the container registry or something else instead of a comment on the `docker run` command, I'll be happy to fix it!


vigenere23:
THANK YOU for that PR, I was getting confused as hell. This should definitly get approved.

PR #3750, opened 2025-11-25
memory: Add support for active_anon and inactive_anon

fabiand:
This is exposing active and inactive anonymous memory regions.

Relevant for KubeVirt where guest memory is anonymous.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -828,6 +828,20 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 		ret.Memory.HierarchicalData.Pgmajfault = v
 	}
 
+	inactiveAnonKeyName := "total_inactive_anon"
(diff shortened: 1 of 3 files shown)


dims:
@fabiand still need this?

fabiand:
Yes. This information is still relevant.
For example to approximate performance impact in envs with swap enabled.
As long as active mem fits into RAM, we are good, if we see/expect it to exceed RAW, then we get into SWAP.
We then have PSI to really measure impact. But before PSI, active memory is valuebale.

tl;dr yes. But I don't have capacity to push this fwd.

fabiand:
What else would actually be needed in order to have this merge ready?

michalskrivanek:
@dims in addition, AnonHugePages (anon_thp) would also be quite beneficial to understand how much memory is backed by transparent huge pages.
Is there any objection if I revive it with THP reporting as well?

PR #3640, opened 2025-01-02
feat: add metric for container health check status 

mateuszdrab:
This PR adds the result of the container's health check as a metric

PR #3760, opened 2025-12-04
fix: docker health check status not updating

thielepaul:
I wanted to use the new health check status added in https://github.com/google/cadvisor/pull/3640 and noticed two issues that are fixed in this PR.

- The state of the health status is never updated, instead it is read only once (at container start or at cadvisor start)
- Containers without health check and unhealthy containers both have the value 0 in the metric (I changed it to -1)

### DIFF
--- a/build/integration-in-docker.sh
+++ b/build/integration-in-docker.sh
@@ -58,7 +58,7 @@ function run_tests() {
     --privileged \
     --cap-add="sys_admin" \
     --entrypoint="" \
-    gcr.io/k8s-staging-test-infra/bootstrap \
+    gcr.io/k8s-staging-test-infra/bootstrap:v20250702-52f5173c3a \
     bash -c "export DEBIAN_FRONTEND=noninteractive && \
     apt update && \
     apt install -y $PACKAGES && \


thielepaul:
@dims can you take a look, if these changes make sense to you?

dims:
@thielepaul do you mind adding tests?

thielepaul:
> @thielepaul do you mind adding tests?

I added a test for the value conversion in `prometheus.go`, however, I am unsure how to write a good test for the change in `handler.go`. If you think another test makes sense, please give me some pointers on where to add it.

dims:
here? https://github.com/google/cadvisor/tree/master/integration/tests/api

thielepaul:
> here? https://github.com/google/cadvisor/tree/master/integration/tests/api

Thank you! I added an integration test.
Do you have any other comments?

thielepaul:
@dims can you please trigger the tests again? I added a tag to the gcr.io/k8s-staging-test-infra/bootstrap image, so the same version that was previously is used and the tests hopefully pass.

mateuszdrab:
Thank you for the fixes @thielepaul 
My Golang skills suck! 🤣

PR #3762, opened 2025-12-04
Reduce lock contention in cache/memory package

dgrisonnet:
## Summary

Replace the global `sync.RWMutex`-protected map with `sync.Map` in `InMemoryCache` to eliminate lock contention during concurrent cache access.

## Changes

- Replace `sync.RWMutex` + `map[string]*containerCache` with `sync.Map`
- Use `Load` for lock-free reads in `RecentStats`
- Use `LoadOrStore` pattern in `AddStats` for atomic container creation
- Add `containerCacheMap` typed wrapper to eliminate type assertions at call sites

## Benchmark Results

**Performance improvement: 2.5x faster** for mixed read/write workloads (99% reads, 1% writes)

For the benchmark methodology and detailed performance analysis, see https://github.com/google/cadvisor/pull/3756 which applies the same pattern to the manager package.

### DIFF
--- a/cache/memory/memory.go
+++ b/cache/memory/memory.go
@@ -29,6 +29,38 @@ import (
 // ErrDataNotFound is the error resulting if failed to find a container in memory cache.
 var ErrDataNotFound = errors.New("unable to find data in memory cache")
 
+// containerCacheMap is a typed wrapper around sync.Map that eliminates the need
+// for type assertions at every call site. It stores container name strings
+// mapped to *containerCache values.
+type containerCacheMap struct {
+	m sync.Map
+}
+
+// Load retrieves a container cache by name. Returns nil, false if not found.
+func (c *containerCacheMap) Load(name string) (*containerCache, bool) {
+	v, ok := c.m.Load(name)
+	if !ok {
+		return nil, false
+	}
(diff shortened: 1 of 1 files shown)


haircommander:
LGTM

cache/memory/memory.go
```diff
@@ -67,24 +99,18 @@ func newContainerStore(ref info.ContainerReference, maxAge time.Duration) *conta
...
-		if cstore, ok = c.containerCacheMap[cInfo.ContainerReference.Name]; !ok {
-			cstore = newContainerStore(cInfo.ContainerReference, c.maxAge)
-			c.containerCacheMap[cInfo.ContainerReference.Name] = cstore
-		}
-	}()
+	name := cInfo.ContainerReference.Name
+	cstore, ok := c.containerCacheMap.Load(name)
+	if !ok {
+		newStore := newContainerStore(cInfo.ContainerReference, c.maxAge)
+		cstore, _ = c.containerCacheMap.LoadOrStore(name, newStore)
```
haircommander: part of me wanted to suggest we only Store here, but I suppose two AddStats can race against each other and we mine as well take the old one once created and not replace...

PR #3764, opened 2025-12-04
devicemapper: use atomic.Value for lock-free cache reads

dgrisonnet:
The ThinPoolWatcher cache is read frequently by container housekeeping goroutines (via GetUsage) while being updated every 15 seconds by Refresh(). Using atomic.Value allows lock-free reads, eliminating potential contention between the many readers and the single writer.

### DIFF
--- a/devicemapper/thin_pool_watcher.go
+++ b/devicemapper/thin_pool_watcher.go
@@ -17,19 +17,35 @@ package devicemapper
 import (
 	"fmt"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"k8s.io/klog/v2"
 )
+
+// usageCache is a typed wrapper around atomic.Value that eliminates the need
+// for type assertions at every call site. It stores device ID strings mapped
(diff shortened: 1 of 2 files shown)


thaJeztah:
I'm not too familiar with the code in this repository, but I noticed the devicemapper code, and wondered if it was actually still used; I know docker / moby dropped support for the devicemapper graph driver; is it still supported (and used) by PodMan?

dgrisonnet:
That's a good point, I am not familiar enough with the ecosystem to know. @haircommander do you perhaps know if devicemapper driver is still supported?

haircommander:
Nope not to my understanding. Possibly by kata that was the last remaining user but i doubt it atm

haircommander:
but kata doesn't use cadvisor, so yeah I don't think we need this

thaJeztah:
Thanks! Yeah, so looks like we should be able to start cleaning up some bits. Not the highest priority, but could be a good follow up 👍

PR #3771, opened 2025-12-06
refactor(container): Migrate to std lib context package

Rahul-Lashkari:
- replaced the deprecated `golang.org/x/net/context` with the standard library `context` package
- removed defaultContext() and fixed context leaks
- make format, tidy, test checked

### DIFF
--- a/container/containerd/factory.go
+++ b/container/containerd/factory.go
@@ -15,13 +15,13 @@
 package containerd
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"path"
 	"regexp"
 	"strings"
 
-	"golang.org/x/net/context"
 	"k8s.io/klog/v2"
 
 	"github.com/google/cadvisor/container"
--- a/container/containerd/grpc.go
+++ b/container/containerd/grpc.go
@@ -16,9 +16,11 @@
 package containerd
 
 import (
-	"github.com/google/cadvisor/container/containerd/namespaces"
-	"golang.org/x/net/context"
+	"context"
+
 	"google.golang.org/grpc"
+
+	"github.com/google/cadvisor/container/containerd/namespaces"
 )
 
 type namespaceInterceptor struct {
--- a/container/containerd/handler.go
+++ b/container/containerd/handler.go
@@ -16,6 +16,7 @@
 package containerd
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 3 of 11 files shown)


dims:
ci job fail :( 
```
Run source build/config/plain.sh
>> upgrading golangci-lint from  to 2.6.2
golangci/golangci-lint info checking GitHub for tag 'v2.6.2'
golangci/golangci-lint info found version: 2.6.2 for v2.6.2/linux/amd64
golangci/golangci-lint info installed /home/runner/go/bin/golangci-lint
>> running golangci-lint using configuration at .golangci.yml
Error: container/docker/docker.go:37:7: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
	ctx, _ := context.WithTimeout(context.Background(), dockerTimeout)
	     ^
Error: container/docker/plugin.go:60:8: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
		ctx, _ := context.WithTimeout(context.Background(), startupTimeout)
		     ^
2 issues:
* govet: 2
make: *** [Makefile:89: lint] Error 1
```

thaJeztah:
> ci job fail :(

Looks like it's an existing issue that wasn't detected by the linter due to the legacy context.

But the linter definitely is correct, and the existing implementation doesn't look good. Perhaps the `defaultContext` utility should either be removed, or otherwise updated to return the cancel, so that callers can handle the `defer cancel()`. Given that it's a very shallow wrapper, perhaps removing the utility, and inlining the code where it's used would be more transparent; https://github.com/google/cadvisor/blob/fba17beabca9392a8c1b5d73c7a987a51dce70b2/container/docker/docker.go#L35-L38


@Rahul-Lashkari can you look at making those changes?

Rahul-Lashkari:
> > ci job fail :(
> 
> Looks like it's an existing issue that wasn't detected by the linter due to the legacy context.
> 
> But the linter definitely is correct, and the existing implementation doesn't look good. Perhaps the `defaultContext` utility should either be removed, or otherwise updated to return the cancel, so that callers can handle the `defer cancel()`. Given that it's a very shallow wrapper, perhaps removing the utility, and inlining the code where it's used would be more transparent;
> 
> https://github.com/google/cadvisor/blob/fba17beabca9392a8c1b5d73c7a987a51dce70b2/container/docker/docker.go#L35-L38
> 
> @Rahul-Lashkari can you look at making those changes?

yeah sure on it! thanks for guidance!

Rahul-Lashkari:
PTAL!

thaJeztah (approved):
LGTM, thanks!

Rahul-Lashkari:
thanks for merging!

Issue #3722, opened 2025-08-15
Add `ContainerExitCode` to `info.EventData` in case of container deletion events


sambhav-jain-16:
## Problem?
`cadvisor` has an `EventChannel` where currently 4 types of events are pushed.
Specifically a `EventType.EventContainerDeleteion` event is pushed when `manager` detects a container is deleted and it currently only captures the timestamp of the deletion event and name of the container.
However, since the container can be deleted due to many chaos scenarios, the current information is not sufficient enough to capture why the container is deleted. There's a specific case added for out of memory issues, however rather than accommodating all such scenarios individually, it will be a good idea to also include the exit code of the main process of the container. 

We specifically have a use case where we want to emit a metric when a container is killed with a non-zero exit code and we use `cadvisor` as our main container observability tool. We tried using metrics from orchestrators (k8s) however they are not reliable in our environment. Subscribing to the `EventContainerDeletion` in the event channel looked like a good approach for us, but it lacks the exit code. We do have a work around but it might be better to include this info from the source itself.

Mostly all container runtime provides a way of getting the exit code for their exited containers like docker has `client.ContainerInspect`, containerd API also provide a way to get the exit status using the taskClient.

## Possible Solution
This can be achieved by adding a new field here
```
type EventData struct {
	// Information about an OOM kill event.
	OomKill *OomKillEventData `json:"oom,omitempty"`

	ContainerExitCode int8 `json:"conainer_exit_code,omitempty`
}
```
and then updating the logic here
https://github.com/google/cadvisor/blob/5adb1c3bb38b4c5d50b31f39faf3214a44ae479b/manager/manager.go#L1042-L1052
to

```
	contRef, err := cont.handler.ContainerReference()
	if err != nil {
		return err
	}

	contExitCode, err := cont.handler.GetExitCode() ## New method to the ContainerHandler interface
        if err != nil {
		return err
	}

	newEvent := &info.Event{
		ContainerName: contRef.Name,
		Timestamp:     time.Now(),
		EventType:     info.EventContainerDeletion,
		EventData: info.EventData{
			ContainerExitCode: contExitCode,
		},
	}
```

Then each container runtime can individually implement its way of getting the exit code.

sambhav-jain-16:
Hi @cwangVT,
Wanted to check if you have any thoughts on this issue? TIA

PR #3779, opened 2025-12-11
feat: add exit code in container deletion events

sambhav-jain-16:
## What

Container deletion events now include the container's exit code in the event data.
Fixes: #3722 

  ## Why

Users need to understand why containers terminated (success vs error vs signal). Exit codes provide critical debugging information for container lifecycle management and monitoring.

 ## How

  - Added `GetExitCode()` method to ContainerHandler interface
  - Added `ContainerDeletionEventData` struct with ExitCode field to event system
  - Modified `destroyContainer()` to retrieve exit code before calling `Stop()`
  - Implemented for docker, containerd, and podman (uses runtime inspect APIs)
  - CRI-O and Raw handlers return errors (not supported by those runtimes)
  - Exit code defaults to -1 if unavailable (container still running, retrieval fails, etc.)

  ### Testing

  - Unit tests for docker, containerd, and manager
  - Integration test validates end-to-end flow

  ### API Example
```
  {
    "event_type": "containerDeletion",
    "container_name": "/docker/abc123",
    "event_data": {
      "container_deletion": {
        "exit_code": 1
      }
    }
  }
```

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -63,6 +63,11 @@ type ContainerHandler interface {
 	// Returns the container's ip address, if available
 	GetContainerIPAddress() string
 
+	// GetExitCode returns the container's exit code if available.
+	// Returns an error if the container has not exited, exit codes are not supported
+	// for this handler type, or the container information is unavailable.
+	GetExitCode() (int, error)
+
 	// Returns whether the container still exists.
 	Exists() bool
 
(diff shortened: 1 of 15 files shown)


dims:
@sambhav-jain-16 which container engines did you fully test this on? (containerd only?)

sambhav-jain-16:
@dims 
This is tested on docker and containerd engines and is running already for some time in our setup.

(Not exactly the same code, but we are consuming the current deletion events and then hitting the same docker/containerd APIs to get the exit code)

PR #3789, opened 2025-12-14
fix: modified container healthcheck to support custom port via -port/--port flag

### LINKED ISSUE #3618: cadvisor dockerfile healthchecks do not utilize correct port if --port command is used to change from default 8080
The Dockerfile utilizes a static port `8080` for the healthcheck and expose operations:

https://github.com/google/cadvisor/blob/255cff7808a7d23802852e6fd9a12320b1035bf1/deploy/Dockerfile#L68

If a non-default port is utilized, users have to explicitly set the `CADVISORT_HEALTHCHECK_URL` environment variable to reflect the updated port otherwise the status will remain "unhealthy".

```yaml
  cadvisor:
    image: gcr.io/cadvisor/cadvisor:latest
    container_name: cadvisor
    command:
      - '-port=9092'
    expose:
      - 9092
    ports:
      - 9092:9092
    environment:
      - 'CADVISOR_HEALTHCHECK_URL=http://localhost:9092/healthz'
```

dylandacosta8:
### Problem Statement
The Dockerfile currently has a static healthcheck URL (http://localhost:8080/healthz) that doesn't adapt when users specify a different port using the -port flag. This causes healthchecks to fail when running cAdvisor on non-default ports.

###  Related Issues
Fixes #3618 

### Root Cause
Environment variables set in the entrypoint script don't persist for HEALTHCHECK commands because they run in separate process contexts. Docker doesn't share the entrypoint's environment with the healthcheck process.

### Proposed Solution
Implemented a dedicated healthcheck script that:
1. Reads the cadvisor process command line from `/proc/1/cmdline`
2. Parses the `-port` flag to extract the actual port
3. Performs the healthcheck on the correct port
4. Falls back to 8080 if no custom port is specified

This approach works because Docker healthchecks run inside the container and can access the process information.

### Breaking Changes
None - fully backward compatible

### Test Case 1: Default Port (8080)
```
docker build -f deploy/Dockerfile -t cadvisor-test .
docker run -d --name cadvisor-default cadvisor-test
sleep 35
docker inspect cadvisor-custom --format='{{json .State.Health.Status}}' # Should show "healthy"
```

### Test Case 2: Custom Port (9090)
```
docker run -d --name cadvisor-custom -p 9090:9090 cadvisor-test -port=9090
sleep 35
docker inspect cadvisor-custom --format='{{json .State.Health.Status}}' # Should show "healthy"
```

### Test Case 3: Verify Healthcheck Logic
```
docker exec cadvisor-custom cat /proc/1/cmdline | tr '\0' '\n'  # Should show -port=9090
docker exec cadvisor-custom ./usr/bin/healthcheck.sh  #Should exit with status 1
```

### Note:
Individual CLA submitted


dylandacosta8:
Added the required boilerplate to the shell scripts

dylandacosta8:
Hey, could a maintainer help me understand if any changes are required here?

iwankgb:
Closing just to reopen (I hope to have security scan run successfully).

iwankgb (approved):

dylandacosta8:
Thanks @iwankgb. The CADVISOR_HEALTHCHECK_URL envVar should be obsolete with the next release

lev-blit:
can something be done, like this, to also support `url_base_prefix`?

mtracz:
> The CADVISOR_HEALTHCHECK_URL envVar should be obsolete with the next release

But now when `url_base_prefix` is defined, the healthcheck will fail, because there is no way to pass this prefix to healthcheck.sh.

@lev-blit 
The easiest way is to redefine healthcheck, as shown here: https://github.com/google/cadvisor/issues/3822#issuecomment-3768328487

PR #3802, opened 2025-12-19
manager: fix race condition in destroyContainer causing panic

dims:
The switch from RWMutex to sync.Map in commit 20e39fcd introduced a race condition in destroyContainer(). When multiple goroutines attempt to destroy the same container simultaneously:

1. Thread A: Loads container from sync.Map (succeeds)
2. Thread B: Also loads the same container (succeeds, A hasn't deleted yet)
3. Thread A: Calls Stop(), closes the stop channel, deletes from map
4. Thread B: Calls Stop(), tries to close already-closed channel -> PANIC

The panic manifests as:
  panic: close of closed channel
  goroutine N [running]:
  github.com/google/cadvisor/manager.(*containerData).Stop(...)
          manager/container.go:143 +0x46

Fix this by using LoadAndDelete() which atomically loads and removes the container from the map, ensuring only one goroutine can process the destruction.

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -271,6 +271,17 @@ func (c *containerMap) Delete(name namespacedContainerName) {
 	c.m.Delete(name)
 }
 
+// LoadAndDelete atomically loads and deletes the containerData for the given name.
+// This prevents race conditions where multiple goroutines might try to destroy
+// the same container simultaneously.
+func (c *containerMap) LoadAndDelete(name namespacedContainerName) (*containerData, bool) {
+	v, loaded := c.m.LoadAndDelete(name)
+	if !loaded {
+		return nil, false
+	}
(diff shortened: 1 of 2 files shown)


manager/manager.go
```diff
@@ -1041,8 +1056,7 @@ func (m *manager) destroyContainer(containerName string) error {
 		return err
 	}
 
-	// Remove the container from our records (and all its aliases).
-	m.containers.Delete(namespacedName)
+	// Remove aliases from our records.
```
SergeyKanzhelev: It used to keep the record if error happened. So I think there will be a retry if `Stop` failed. I am wondering if we need to keep this logic in some form? I didn't confirm that retry will happen, but brief look seems to indicate it

Issue #3749, opened 2025-11-11
Failure to get data in Prometheus on latest Docker


zacharyshale95:
Ubuntu 24.04.3 LTS
Docker version 29.0.0, build 3d4129b

> I1111 00:05:15.302059       1 manager.go:233] Version: {KernelVersion:6.14.0-35-generic ContainerOsVersion:Alpine Linux v3.18 DockerVersion: DockerAPIVersion: CadvisorVersion:v0.49.1 CadvisorRevision:6f3f25ba}
I1111 00:05:15.302274       1 factory.go:219] Registration of the podman container factory failed: failed to validate Podman info: response not present: Get "http://d/v1.0.0/info": dial unix /var/run/podman/podman.sock: connect: no such file or directory
I1111 00:05:15.304095       1 factory.go:145] Registering containerd factory
I1111 00:05:15.304109       1 factory.go:221] Registration of the containerd container factory successfully
I1111 00:05:15.304114       1 factory.go:55] Registering systemd factory
I1111 00:05:15.304117       1 factory.go:221] Registration of the systemd container factory successfully
I1111 00:05:15.304220       1 factory.go:219] Registration of the crio container factory failed: Get "http://%2Fvar%2Frun%2Fcrio%2Fcrio.sock/info": dial unix /var/run/crio/crio.sock: connect: no such file or directory
I1111 00:05:15.305482       1 factory.go:219] Registration of the mesos container factory failed: unable to create mesos agent client: failed to get version
I1111 00:05:15.305757       1 factory.go:219] Registration of the docker container factory failed: failed to validate Docker info: failed to detect Docker info: Error response from daemon: client version 1.41 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version
I1111 00:05:15.305951       1 factory.go:103] Registering Raw factory
I1111 00:05:15.305975       1 manager.go:1196] Started watching for new ooms in manager
W1111 00:05:15.305997       1 manager.go:306] Could not configure a source for OOM detection, disabling OOM events: open /dev/kmsg: no such file or directory
I1111 00:05:15.306260       1 manager.go:319] Starting recovery of all containers
I1111 00:05:15.350626       1 manager.go:324] Recovery completed
I1111 00:05:15.361028       1 cadvisor.go:176] Starting cAdvisor version: v0.49.1-6f3f25ba on port 8080

Data for `prom/node-exporter` working fine, but I get no data from cAdvisor on nodes that I have updated docker to 29.0.0

[...]

PR #3805, opened 2025-12-24
Replace references to docker registry `gcr.io` with `ghcr.io`

luckv:
From release of [v0.54.0 - December 2th](https://github.com/google/cadvisor/releases/tag/v0.54.0) (3 weeks ago), newer docker images are pushed to ghcr instead of gcr. I simply updated the docs to reference the new docker registry.

At this moment, is `ghcr.io/google/cadvisor` the official image, or is `gcr.io/cadvisor/cadvisor` and we should expect that to change in the future?

Issue/comment refs:
 - https://github.com/google/cadvisor/issues/3712
 - https://github.com/google/cadvisor/issues/3749#issuecomment-3524798186
 - https://github.com/google/cadvisor/pull/3699


luckv:
close because commit email was not valid with google CLA
Opened there: https://github.com/google/cadvisor/pull/3806

PR #3808, opened 2026-01-06
Add container_creation_time_seconds (previously: container_start_time_seconds); use runtime start time for container_start_time_seconds (podman & docker)

### LINKED ISSUE #3688: Metric container_start_time_seconds reports wrong value
Hello,

Based on my testing, it appears that the `container_start_time_seconds` metric reflects the time since the container was originally created, rather than when it was last (re)started. As a result, accurately tracking container restarts using this metric alone is not feasible in its current form.

Is it possible to add a metric expressing the container's exact uptime?

Image

Thank you 😊

ultinous-akos:
fixes #3688

### DIFF
--- a/container/containerd/client.go
+++ b/container/containerd/client.go
@@ -168,12 +168,16 @@ func (c *client) Version(ctx context.Context) (string, error) {
 
 func containerFromProto(containerpb *containersapi.Container) *containers.Container {
 	var runtime containers.RuntimeInfo
+	var createdAt time.Time
 	// TODO: is nil check required for containerpb
 	if containerpb.Runtime != nil {
 		runtime = containers.RuntimeInfo{
 			Name:    containerpb.Runtime.Name,
 			Options: containerpb.Runtime.Options,
 		}
+	}
+	if containerpb.GetCreatedAt() != nil {
+		createdAt = containerpb.GetCreatedAt().AsTime()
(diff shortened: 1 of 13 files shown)


iwankgb:
Do you mind adding this feature to other handlers too?

ultinous-akos:
> Do you mind adding this feature to other handlers too?


@iwankgb 
For now, i would prefer to keep this PR focused on the Docker and Podman handlers, since that's where I verified the issue and the fix.

dims:
i think at a minimum we'll need containerd too

dims:
we now have a integration test and a way to run it locally as well that covers both `docker` and `containerd` metrics - https://github.com/google/cadvisor/pull/3819

ultinous-akos:
Thanks @dims 
I've added containerd too.

ultinous-akos:
Hello @dims & @iwankgb!

I addressed the requested changes few weeks ago, could you please take another look when you have time, or let me know if anything else is needed? Thanks!

dims:
please squash the commits, i'll then approve the test ci jobs to run

ultinous-akos:
Hello @dims,
I've squashed the commits

Thank you!

eric1tran:
Any idea when this might make it into a release version?

Issue #3813, opened 2026-01-15
/usr/bin/entrypoint.sh and /usr/bin/healthcheck.sh  not found


logopk:
Starting 0.56.0 docker container fails:

```
cadvisor exited with code 255 (restarting)
cadvisor      | exec /usr/bin/entrypoint.sh: no such file or directory
cadvisor      | exec /usr/bin/entrypoint.sh: no such file or directory

```

healtchcheck similar:

```
            "Health": {
                "Status": "unhealthy",
                "FailingStreak": 3,
                "Log": [
                    {
                        "Start": "2026-01-15T15:46:50.598652823+01:00",
                        "End": "2026-01-15T15:46:50.625149191+01:00",
                        "ExitCode": 127,
                        "Output": "/bin/sh: /usr/bin/healthcheck.sh: not found\n"
                    },
```

setting entrypoint in docker-compose to /usr/bin/cadvisor lets me start the container.

The files exist but cannot be started. Permissions are OK

```
▶ docker exec -ti cadvisor sh
/ # ls -l /usr/bin/healthcheck.sh /usr/bin/entrypoint.sh 
-rwxr-xr-x    1 root     root           706 Jan 15 13:56 /usr/bin/entrypoint.sh
-rwxr-xr-x    1 root     root          1119 Jan 15 13:56 /usr/bin/healthcheck.sh
/ # 
/ # /usr/bin/entrypoint.sh 
sh: /usr/bin/entrypoint.sh: not found
/ # /usr/bin/healthcheck.sh
sh: /usr/bin/healthcheck.sh: not found
```

fabianwimberger:
Same here. I think this is a general issue. For now pinned version to the one before.

benji2k2:
Problem is probably, that the scripts require "#!/bin/bash" which is not available in the image. There were changes last week changing this from "#!/bin/sh". I think this is maybe the problem. You can try to change inside container to verify.

logopk:
"#!/bin/sh" works

logopk:
But this truly shows some lack of testing 🫣

benji2k2:
Thank you for verifying. It will work for nobody with the current images i think. But fix is very easy, maybe i find time later to make a pr for this. Or hopefully someone of the maintainers is faster than me...

benji2k2:
I created PR #3814 which hopefully will fix this issue.

dims:
thanks @benji2k2 - i started a fresh tag:
[...]

PR #3814, opened 2026-01-15
Fixing entrypoint- and healthcheck script in container

benji2k2:
Containers are not starting anymore, because the Scripts require "/bin/bash" which is not available in the image. Changing to "/bin/sh" should fix the issue.


benji2k2:
This is a fix for the blocking issue #3813 - the Docker-Images of Version 0.56.0 should work for nobody...

dims:
thanks @benji2k2

PR #3824, opened 2026-01-21
Fix cgroup v2 swap-only memory metrics for KIND

AkshatDudeja77:
This change fixes reporting of swap-only memory metrics for cgroup v2.

Previously, swap usage was either missing or incorrectly derived. This update
adds proper parsing and exposure of swap-only usage so that metrics such as
container_memory_swap and related fields reflect actual cgroup v2 behavior.

Tested locally on cgroup v2 (WSL2) and verified metrics via /metrics endpoint.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -814,7 +814,7 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	if cgroups.IsCgroup2UnifiedMode() {
 		ret.Memory.Cache = s.MemoryStats.Stats["file"]
 		ret.Memory.RSS = s.MemoryStats.Stats["anon"]
-		ret.Memory.Swap = s.MemoryStats.SwapUsage.Usage - s.MemoryStats.Usage.Usage
+		ret.Memory.Swap = s.MemoryStats.SwapOnlyUsage.Usage
 		ret.Memory.MappedFile = s.MemoryStats.Stats["file_mapped"]
 	} else if s.MemoryStats.UseHierarchy {
 		ret.Memory.Cache = s.MemoryStats.Stats["total_cache"]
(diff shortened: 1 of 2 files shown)


AkshatDudeja77:
I’ve signed the Google CLA. Could you please recheck?

dims:
@AkshatDudeja77 the PR is not complete. just adding a file in vendor/ is not going to help from what i can tell.

ajaysundark:
The description is misleading. it is not that cgroupv2 swap/stats collection is not working, but it is not working for "kind".

The file added in this PR is from [opencontainers/cgroups](https://github.com/google/cadvisor/blob/566cb6326fb0804aa9f960723be74993ba4d6dc1/go.mod#L23) which is the module used by cadvisor for stats collection. And from what I can see here: https://github.com/opencontainers/cgroups/blob/e0c56cb31dcb3cb2b3d1554b20dd01ced32e2a2b/fs2/memory.go#L115-L119 cgroupv2 swap collection is correctly handled.

I dont see any code changes in this file, it seems to be just a plain import. did you miss to add your changes in the cadvisor itself?

ajaysundark:
/retitle Fix cgroup v2 swap-only memory metrics for KIND

AkshatDudeja77:
Thanks for the clarification. I agree that this issue is specific to KIND. I’ve updated the PR title accordingly and will add the missing cAdvisor-side changes.

AkshatDudeja77:
This PR fixes incorrect swap reporting under cgroup v2.

Previously, cAdvisor derived swap as SwapUsage - MemoryUsage, which is incorrect for cgroup v2.

The fix ensures swap metrics are sourced from cgroup v2 swap-only accounting, aligning container stats with kernel semantics and Prometheus metrics output.

Issue #3822, opened 2026-01-19
cadvisor healthcheck failed after reverse proxy


4chEr:
Cadvisor after reverse proxy (nginx)
When running in Docker Compose, we used:

```
command:
  - "-url_base_prefix=/cadvisor"
environment:
  - "CADVISOR_HEALTHCHECK_URL=http://localhost:8080/cadvisor/healthz"
```
In release >= 0.56.0 healthcheck fails

https://github.com/google/cadvisor/commit/a2939e760308ebcf5f8b24d125e9d6507b98bb48

4chEr:
Temporary solution

```
    command:
      - "-url_base_prefix=/cadvisor"
    healthcheck:
      test: ["CMD-SHELL", "wget --quiet --tries=1 --spider http://localhost:8080/cadvisor/healthz || exit 1"]
      interval: 30s
      timeout: 3s
      start_period: 5s
      retries: 3
```

mtracz:
Confirm the issue (on 0.56.2), and provided workaround works well.

`CADVISOR_HEALTHCHECK_URL` has been removed in [0.56.0](https://github.com/google/cadvisor/releases/tag/v0.56.0) int this [PR](https://github.com/google/cadvisor/pull/3789) so wont take any effect on healthcheck logic.

dylandacosta8:
I'm putting together a quick fix to account for url_base_prefix

PR #3833, opened 2026-02-02
fix: healthcheck now accounts for url_base_prefix in reverse proxy setups

dylandacosta8:
## Description

Uses the same logic as #3789  to dynamically detect and support the `url_base_prefix` flag in healthchecks for reverse proxy setups.

## Fixes 
- #3822

### DIFF
--- a/deploy/healthcheck.sh
+++ b/deploy/healthcheck.sh
@@ -16,12 +16,13 @@
 
 # Default port
 PORT=8080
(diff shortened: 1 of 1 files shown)


dylandacosta8:
[bump] hope that this fix makes it to the next release. 

@Maintainers, I'd love to understand if there are any blockers or if the PR documentation requires additional information

PR #3834, opened 2026-02-03
Fix healthcheck by respecting --url_base_prefix

ngc7293:
## Problem Statement

Healthchecks fail when using a custom `--url_base_prefix`.

## Root Cause

The healthcheck script does not account for custom URL prefixes. The `/healthz` endpoint is mounted on the root mux which does take it into account.

## Proposed solution

Using the same approach recently added for `--port` (see #3789), parse the cadvisor process' argument list to resolve the proper `--url_base_config.`

## Breaking Changes

None, fixes a regression.

## Test Case 1: Default Port (8080)

docker build -f deploy/Dockerfile -t cadvisor-test .
docker run -d --name cadvisor-default cadvisor-test
sleep 35
docker inspect cadvisor-custom --format='{{json .State.Health.Status}}' # Should show "healthy"

## Test Case 2: Custom URL prefix

docker run -d --name cadvisor-custom cadvisor-test --url_base_config=/cadvisor
sleep 35
docker inspect cadvisor-custom --format='{{json .State.Health.Status}}' # Should show "healthy"

### DIFF
--- a/deploy/healthcheck.sh
+++ b/deploy/healthcheck.sh
@@ -16,6 +16,7 @@
 
 # Default port
 PORT=8080
+BASE_URL=""
 
 # Extract port from the cadvisor process command line
 if [ -f /proc/1/cmdline ]; then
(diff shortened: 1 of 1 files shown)


ngc7293:
Opened too soon, this is a dupe of #3833.

PR #3838, opened 2026-02-11
fix crio deadlock in getting crio sandbox containers

olyazavr:
In https://github.com/cri-o/cri-o/issues/8748, I found that cadvisor v0.48.1 had a bug that caused crio to essentially deadlock if there was a long-terminating container and a kubelet restart. However, with cadvisor v0.49.0 this was fixed. With 0.52.1 it was bugged again. 

I found that https://github.com/google/cadvisor/pull/3457 was the fix, which was then later reverted: https://github.com/google/cadvisor/pull/3565

1.27 has cadvisor v0.47.2 (working)
1.29 has cadvisor v0.48.1 (broken)
1.30 has cadvisor v0.49.0 (working)
1.31 has cadvisor v0.49.0 (working)
1.33 has cadvisor 0.52.1 (broken)

What happens here is that cadvisor finds sandbox containers in cgroups (they exist as cgroup directories), calls cri-o, cri-o returns 404 because sandbox containers aren't returned by the inspect endpoint, and then I suspect something ends up holding a mutex that prevents anything else to go through, which causes kubelet to fail to start because crio/cadvisor are stuck.

This re-does the original PR (https://github.com/google/cadvisor/pull/3457) but addresses the systemd/cgroupfs issue

### DIFF
--- a/container/crio/client.go
+++ b/container/crio/client.go
@@ -45,6 +45,7 @@ type Info struct {
 	StorageDriver string `json:"storage_driver"`
 	StorageRoot   string `json:"storage_root"`
 	StorageImage  string `json:"storage_image"`
+	CgroupDriver  string `json:"cgroup_driver"`
 }
 
 // ContainerInfo represents a given container information
--- a/container/crio/factory.go
+++ b/container/crio/factory.go
@@ -18,11 +18,9 @@ package crio
 
 import (
 	"fmt"
-	"os"
 	"path"
 	"regexp"
 	"strings"
-	"sync"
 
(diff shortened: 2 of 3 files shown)


dims:
@haircommander can you please review?

haircommander:
thanks for picking this up @olyazavr !

haircommander:
LGTM

olyazavr:
@haircommander is there anything else I need to do here? Merging is blocked for me

haircommander:
cc @dims :eyes:

PR #3840, opened 2026-02-17
remove log message when you can't read productName

kannon92:
On IBM Power and IBM s390z, we are getting this error message in our kubelet logs.

Looking at azure and aws, we are not logging an error but returning an error.

Proposing a patch to match these existing cloud vendors.


prb112:
Thank you for the updates!

PR #3842, opened 2026-02-18
crio: add retry to container info call

haircommander:
adapt the containerd code for retry, and factor out the function (with unit tests!)


haircommander:
I have some in-cluster testing to do, but this feels safe generally so I won't put a hold on it. but noting: I haven't technically tested this with cri-o yet

PR #3845, opened 2026-02-22
fix: correct cgroup v2 swap memory calculation (#3823)

### LINKED ISSUE #3823: cAdvisor does not expose swap metrics when running inside container (cgroup v2)
### What happened

When running Kubernetes via kind (kubelet + cAdvisor inside a container), swap is enabled and visible inside the node container, but no swap-related metrics are exposed by cAdvisor.

Verified inside the kind control-plane container:
- `free -h` shows swap
- `/proc/swaps` is populated
- cgroup v2 is in use
- swap cgroup files exist:
  - memory.swap.current
  - memory.swap.max
  - memory.swap.events

However:
- cAdvisor `/metrics` (port 4194) does not emit any swap-related Prometheus metrics
- Consequently, kubelet `/metrics/resource` also does not expose swap metrics

This suggests swap metrics are not being collected or exported by cAdvisor when running inside a containerized environment (likely related to cgroup v2 handling).

---

### What you expected to happen

Swap usage and limits should be exposed as Prometheus metrics by cAdvisor when swap is enabled and available via cgroup v2.

---

### Environment

- cAdvisor: latest (via kind node image)
- Kubernetes: kind v0.31.0
- Runtime: Docker Desktop (WSL2)
- cgroup version: v2
- OS: Linux (inside kind node container)

---

### Context

This was discovered while investigating missing swap metrics in Kubernetes:
https://github.com/kubernetes-sigs/kind/issues/3834

Elvand-Lie:
**Fixes #3823**

**Problem:**
In cgroup v2 environments, `cAdvisor` is currently reporting zero swap metrics even when swap is active. This is due to `handler.go` applying legacy cgroup v1 math (`SwapUsage - Usage`). In v2, `memory.swap.current` is decoupled from RAM usage, so subtracting `Usage` incorrectly zeroes out the metric.

**Solution:**
Updated the `IsCgroup2UnifiedMode()` block to assign `SwapUsage.Usage` directly to `ret.Memory.Swap` without the legacy subtraction.

**Notes:**
- Syntax verified with `go fmt`.
- Shoutout to @AkshatDudeja77 for the initial investigation and identifying the regression point in `handler.go`.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -814,7 +814,8 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	if cgroups.IsCgroup2UnifiedMode() {
 		ret.Memory.Cache = s.MemoryStats.Stats["file"]
 		ret.Memory.RSS = s.MemoryStats.Stats["anon"]
-		ret.Memory.Swap = s.MemoryStats.SwapUsage.Usage - s.MemoryStats.Usage.Usage
+		// In v2, SwapUsage is already exclusively swap. No subtraction needed.
+		ret.Memory.Swap = s.MemoryStats.SwapUsage.Usage
 		ret.Memory.MappedFile = s.MemoryStats.Stats["file_mapped"]
 	} else if s.MemoryStats.UseHierarchy {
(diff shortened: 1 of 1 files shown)


iholder101:
Thank you very much for addressing this!

PR #3854, opened 2026-03-17
feat(manager/container): add configurable initial splay and max jitter factors

sambhav-jain-16:
This PR is an extension of https://github.com/google/cadvisor/pull/2918.
The initial [jitter](https://github.com/google/cadvisor/issues/1053) PR aims to avoid CPU spikes when hundreds of HKs are running. Although it does the job as intended, the actual interval ranges from [HK, 2HK], resulting in higher variance when `cadvisor` metric timestamps are considered for rate calculations in counter metrics. Since the jitter factor is not configurable, it is impossible to reduce the variance in the current scenarios.
Also, a good way to avoid CPU spikes is to spread out the container HKs initially, so that subsequent jitter can be set to zero and we get the same benefit as with random jitter every time.

To keep the change backward-compatible, the default values are set to `1.0`.

Attaching a screenshot showing how this change helped to reduce variance with initial splay at `1.0` and jitter factor at `0.0`, without CPU spikes in `cadvisor`.

image (3)

### DIFF
--- a/manager/container.go
+++ b/manager/container.go
@@ -47,9 +47,13 @@ import (
 	"k8s.io/utils/clock"
 )
 
+const jitterDefault = 1.0
+
 // Housekeeping interval.
 var enableLoadReader = flag.Bool("enable_load_reader", false, "Whether to enable cpu load reader")
 var HousekeepingInterval = flag.Duration("housekeeping_interval", 1*time.Second, "Interval between container housekeepings")
+var InitialSplayFactor = flag.Float64("initial_splay_factor", jitterDefault, "Factor for the initial splay b/w the container housekeepings, default is 1.0. If negative value is passed, the value will be reset to default")
+var JitterFactor = flag.Float64("jitter_factor", jitterDefault, "Factor for the jitters after the initial splay b/w the container housekeepings, default is 1.0. If negative value is passed, the value will be reset to default")
 
 // TODO: replace regular expressions with something simpler, such as strings.Split().
 // cgroup type chosen to fetch the cgroup path of a process.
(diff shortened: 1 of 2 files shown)


sambhav-jain-16:
Hi @dims 
Can you PTAL at this PR? TIA

manager/container.go
```diff
@@ -124,8 +129,8 @@ type containerData struct {
 // to allow clients to avoid converging on periodic behavior.  If maxFactor is 0.0, a
```
dims: fix this comment?

sambhav-jain-16: Done.

sambhav-jain-16:
integration tests failing with 
```
Unable to find image 'gcr.io/k8s-staging-test-infra/bootstrap:v20250702-52f5173c3a' locally
docker: Error response from daemon: manifest for gcr.io/k8s-staging-test-infra/bootstrap:v20250702-52f5173c3a not found: manifest unknown: Failed to fetch "v20250702-52f5173c3a"
```

dims:
> integration tests failing with `Unable to find image 'gcr.io/k8s-staging-test-infra/bootstrap:v20250702-52f5173c3a' locally`

switch to `gcr.io/k8s-staging-test-infra/bootstrap:v20251209-855adc2699`

sambhav-jain-16:
> > integration tests failing with `Unable to find image 'gcr.io/k8s-staging-test-infra/bootstrap:v20250702-52f5173c3a' locally`
> 
> switch to `gcr.io/k8s-staging-test-infra/bootstrap:v20251209-855adc2699`

Created a different PR for this https://github.com/google/cadvisor/pull/3865, will rebase this PR after merging the fix

sambhav-jain-16:
Hi @dims 
LMK if you have other review comments

dims (approved):

Issue #3807, opened 2026-01-01
Change from 0.54.0 -> 0.54.1 broke image name in metrics?


andrewlow:
I stumbled over a few things getting here.

- The repository change from `gcr.io/cadvisor/cadvisor` to `ghcr.io/google/cadvisor` - annoyingly these look very similar, but are not.
- The versioning / tagging change from `v0.53.0` to `0.54.0`

However, after figuring these details out - I got down to being able to see the difference between two versions.

- `ghcr.io/google/cadvisor:0.54.0`
- `ghcr.io/google/cadvisor:0.54.1`

I seem to lose my container names once I step up to 0.54.1 (and beyond - as `latest` is also broken for me)

Here is the entry I care about from the `0.54.0` output - `container_cpu_user_seconds_total` - which I want to graph along with the container name `name="exporter-cadvisor"`

```
container_cpu_user_seconds_total{container_label_org_opencontainers_image_created="2025-12-02T22:00:32.582Z",container_label_org_opencontainers_image_description="Analyzes resource usage and performance characteristics of running containers.",container_label_org_opencontainers_image_licenses="NOASSERTION",container_label_org_opencontainers_image_revision="4b231e63d840f2212b11ed7b9e6150d96d26cd97",container_label_org_opencontainers_image_source="https://github.com/google/cadvisor",container_label_org_opencontainers_image_title="cadvisor",container_label_org_opencontainers_image_url="https://github.com/google/cadvisor",container_label_org_opencontainers_image_version="0.54.0",id="/system.slice/docker-3ee879b1c97ede794502b5d336143141b3d857feb2c14a569e2f268e0c969ddc.scope",image="ghcr.io/google/cadvisor:0.54.0",name="exporter-cadvisor"} 13.897096 1767300502508
```

If I move to `0.54.1` -- I lose this and just get this equivalent line (I think this is the equivalent line.. I'm not 100% sure)

```
container_cpu_user_seconds_total{id="/system.slice/docker-e3792c0987693a8ba1cddddc155b35c9387e888f5fd1b949160b55d3b83ddfd7.scope"} 0.194407 1767300887569
```

It's almost like the version of docker I'm using is no longer compatible?

```
$ docker --version
Docker version 28.5.2, build v28.5.2
```

Looking at the changes from 0.54.0 -> 0.54.1 I don't see anything obvious that I'm doing wrong.

jornae2w32:
I'm seeing this too.  I'm on a Synology NAS with docker.

```
[...]

PR #3857, opened 2026-03-26
fix: add v-prefixed GHCR image tags for release consistency

### LINKED ISSUE #3856: GHCR image tags missing 'v' prefix, inconsistent with GitHub Releases
## Description

There is an inconsistency between GitHub Release tag names and GHCR container image tags:

- **GitHub Releases** use `v` prefix: `v0.55.1`, `v0.56.2`, etc.
- **GHCR image tags** do NOT use `v` prefix: `0.55.1`, `0.56.2`, etc.

This causes confusion because users naturally look at the [Releases page](https://github.com/google/cadvisor/releases) to find the version, then use that exact tag to pull the image — which fails.

## Steps to Reproduce

```bash
# Fails — tag does not exist
docker pull ghcr.io/google/cadvisor:v0.55.1
# Error: ghcr.io/google/cadvisor:v0.55.1: not found

# Works — tag without 'v' prefix
docker pull ghcr.io/google/cadvisor:0.55.1
```

## Impact

This has broken our CI/CD pipeline deployment. We suspect many other users may encounter the same issue, as using `v`-prefixed tags from releases is a very common convention.

Related issues:
- #3712 (Missing docker image for v0.53.0)
- #3807 (Change from 0.54.0 -> 0.54.1 broke image name)

## Suggestion

Either:
1. **Publish GHCR tags with both formats** (`v0.55.1` and `0.55.1`) so both work, OR
2. **Document the correct tag format** in the README — clearly state that GHCR tags do NOT include the `v` prefix

I'm happy to submit a PR for option 2 if the maintainers prefer that approach.

hwan515:
## Summary
- Add `type=semver,pattern=v{{version}}` to the container publish workflow
- This publishes **both** `0.55.1` and `v0.55.1` tags to GHCR
- Fixes the inconsistency where GitHub Releases use `v`-prefixed tags but GHCR images don't

## Problem
Users look at the [Releases page](https://github.com/google/cadvisor/releases) and see `v0.55.1`, then naturally run:
```bash
docker pull ghcr.io/google/cadvisor:v0.55.1
# Error: not found
```
This has broken CI/CD pipelines (see #3856, #3712, #3807).

## Change
One line added to `.github/workflows/publish-container.yml`:
```yaml
type=semver,pattern=v{{version}}
```

This makes `docker/metadata-action` generate an additional tag with the `v` prefix, so both formats work.

## Test plan
- For tag `v0.55.1`, the workflow will now generate:
  - `0.55.1` (existing, `{{version}}`)
  - `v0.55.1` (new, `v{{version}}`)
  - `0.55` (existing, `{{major}}.{{minor}}`)
  - `latest`
  - SHA-based tag

Fixes #3856

### DIFF
--- a/.github/workflows/publish-container.yml
+++ b/.github/workflows/publish-container.yml
@@ -54,6 +54,7 @@ jobs:
           images: ghcr.io/${{ github.repository }}
           tags: |
             type=semver,pattern={{version}}
+            type=semver,pattern=v{{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=raw,value=latest,enable=${{ github.event_name != 'workflow_dispatch' && github.ref == format('refs/tags/{0}', github.ref_name) }}
             type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}


hwan515:
I have signed the Google CLA. Please re-check.

PR #3861, opened 2026-04-01
feat(manager): add constraint data in OOM events

sambhav-jain-16:
Add kernel `Constraint` to OOM events. This helps in identifying the reason for OOM from the event itself.


sambhav-jain-16:
Hi @dims 
Can you PTAL at this PR? TIA

sambhav-jain-16:
Hi @dims 
If everything looks good, can we merge this PR? TIA

dims (approved):

PR #3869, opened 2026-04-23
Also apply rawPrefixWhiteList to rawContainerWatcher and rawContainerHandler

### LINKED ISSUE #3868: Also apply rawPrefixWhiteList to rawContainerWatcher and rawContainerHandler
Now the rawPrefixWhiteList is only be used by rawFactory.CanHandleAndAccept,  but the cadvisor manager also starts rawContainerWatcher to watch all the cgroup paths, and the housekeeping timely calls rawContainerHandler.ListContainers to list  all cgroup paths recursively. if a system has many many cgroup paths under /user.slice, it can exhaust uneccessary performance like cpu&memory, so should we also apply rawPrefixWhiteList to rawContainerWatcher and rawContainerHandler?

249043822:
Fix: https://github.com/google/cadvisor/issues/3868

### DIFF
--- a/container/common/helpers.go
+++ b/container/common/helpers.go
@@ -301,7 +301,19 @@ func readUInt64(dirpath string, file string) uint64 {
 }
 
 // Lists all directories under "path" and outputs the results as children of "parent".
-func ListDirectories(dirpath string, parent string, recursive bool, output map[string]struct{}) error {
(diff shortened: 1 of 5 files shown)


haircommander:
LGTM, since the only caller will filter out not whitelisted directories anyway. Maybe we should change the function name to reflect that ?

haircommander (approved):

PR #3870, opened 2026-05-07
Expose cgroup v2 memory.events as Prometheus metrics

sohankunkerkar:
Kubernetes KEP-2570 (MemoryQoS) uses cgroup v2 memory.high for throttling and memory.min/memory.low for memory protection. To observe the effect of these settings, operators need visibility into memory pressure events. cadvisor currently does not read the memory.events cgroup file. The existing `container_oom_events_total` metric comes from kernel log parsing, not cgroup counters.

Read memory.events on cgroup v2 and expose two new Prometheus counter metrics:

- container_memory_events_high_total: times the container was throttled for breaching memory.high
- container_memory_events_max_total: times the container's usage hit memory.max

xref: https://github.com/kubernetes/enhancements/issues/2570

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/opencontainers/cgroups"
 	"github.com/opencontainers/cgroups/fs2"
+	"github.com/opencontainers/cgroups/fscommon"
 	"k8s.io/klog/v2"
 
 	"github.com/google/cadvisor/container"
@@ -91,6 +92,10 @@ func (h *Handler) GetStats() (*info.ContainerStats, error) {
 		klog.V(4).Infof("Ignoring errors when gathering stats for root cgroup since some controllers don't have stats on the root cgroup: %v", err)
 	}
 	stats := newContainerStats(cgroupStats, h.includedMetrics)
+
+	if cgroups.IsCgroup2UnifiedMode() {
+		setMemoryEvents(h.cgroupManager.Path(""), stats)
+	}
 
 	if h.includedMetrics.Has(container.ProcessSchedulerMetrics) {
 		stats.Cpu.Schedstat, err = h.schedulerStatsFromProcs()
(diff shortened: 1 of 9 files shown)


sohankunkerkar:
cc @haircommander

container/libcontainer/handler.go
```diff
@@ -862,6 +866,32 @@ func setMemoryStats(s *cgroups.Stats, ret *info.ContainerStats) {
 	ret.Memory.WorkingSet = workingSet
 }
 
+func setMemoryEvents(cgroupPath string, ret *info.ContainerStats) {
```
haircommander, line 870: either we could use the existing cgroups library function for this https://github.com/opencontainers/cgroups/blob/main/fs2/fs2.go#L322 
or we can invent a new one there that does this for any number of keys in the events file, but I don't think we should duplicate the file opening/parsing pieces

sohankunkerkar: Good point, I can switch to `fscommon.GetValueByKey` for both keys. It does mean opening and scanning `memory.events` twice, but we're only reading two keys and the file is tiny, so it's not worth optimizing. If we add more keys later (oom, oom_kill) we could add a multi-key variant to the cgroups library at that point, but for two keys `GetValueByKey` seems fine.

sohankunkerkar:
@dims could you take a look at it?

dims:
@sohankunkerkar no tests at all? :( 

- add a unit test in container/libcontainer/handler_test.go? 
- integration/tests/metrics/prometheus_test.go::TestCoreMemoryMetricsExist (add to `memoryMetrics` array)
-  integration/tests/api/event_test.go::TestOomKillEventConstraint (same setup as the OOM test, but poll /metrics until container_memory_events_max_total{name~containerID} > 0.)

sohankunkerkar:
> @sohankunkerkar no tests at all? :(
> 
> * add a unit test in container/libcontainer/handler_test.go?
> * integration/tests/metrics/prometheus_test.go::TestCoreMemoryMetricsExist (add to `memoryMetrics` array)
> * integration/tests/api/event_test.go::TestOomKillEventConstraint (same setup as the OOM test, but poll /metrics until container_memory_events_max_total{name~containerID} > 0.)

@dims I addressed your comments. Could you take a look at it again? Thanks!

dims:
@sohankunkerkar https://github.com/google/cadvisor/actions/runs/25805017191/job/75842992013?pr=3870

dims (approved):

dims:
thanks @sohankunkerkar

sohankunkerkar:
@dims Thanks for reviewing this PR. I had one quick question: do we have any plans to cut a new release of cAdvisor anytime soon? We might need it for testing the MemoryQoS feature in Kubernetes.

dims:
@sohankunkerkar i try to do at least one release of cadvisor to support k8s. Will take stock soon-ish. Do you need this in short order? (weeks? days?)

sohankunkerkar:
> @sohankunkerkar i try to do at least one release of cadvisor to support k8s. Will take stock soon-ish. Do you need this in short order? (weeks? days?)

Thanks for the update! It would be ideal if we could get that once the v1.37 branch opens, or before the feature freeze maybe?

PR #3875, opened 2026-05-23
fix: support space-separated healthcheck port

### LINKED ISSUE #3862: Healthcheck doesn't work if port is specified as "--port XXXX" instead of "--port=XXXX"
As part of #3618 a healthcheck.sh script was introduced which parses the used port from the "--port=XXXX" command line argument.

That doesn't work if the port is specified as "--port XXXX". 

Workaround is obviously to pass the argument as "--port=XXXX", but probably the script should be able to properly parse the possible variants for the invocation.

Haihan-Jiang:
## Summary
- Parse `-port VALUE` and `--port VALUE` in `deploy/healthcheck.sh`, in addition to the existing `-port=VALUE` and `--port=VALUE` forms.
- Add a shell test that stubs `wget` and verifies the default port plus all supported flag forms.
- Run the healthcheck shell test from `make test`.

Fixes #3862

## Tests
- `sh deploy/healthcheck_test.sh`
- `go list ./... | grep -vw integration | xargs go test -race`
- `./build/check_boilerplate.sh`
- `git diff --check`

Note: full `make test` on local macOS reaches the existing `cmd` module step and fails because several cAdvisor packages are Linux-only under the default local build constraints; the new healthcheck test itself passes locally.


Haihan-Jiang:
This is ready for review from my side. Could someone take a look when convenient?

PR #3878, opened 2026-05-24
docker: parse suffixed Docker versions strictly

### LINKED ISSUE #3830: Container metadata labels lost when Docker version contains multiple suffixes
**What happened**:
When using Docker versions with complex suffixes like `v20.10.12-v1.0.2`, cAdvisor fails to parse container metadata labels (e.g., `container`, `pod`, `namespace`, `image`), resulting in empty label values in metrics.

Example metric:
```prometheus
container_cpu_usage_seconds_total{
  container="", 
  pod="", 
  namespace="", 
  image=""
} 0.184540494
```

**What expected to happen** :
Container labels should be correctly populated from Docker metadata, allowing proper association of metrics with Kubernetes pods and containers.

**How to reproduce it** :
1. Install Docker with version containing multiple suffixes (e.g., v 20.10.12-v 1.0.2)
2. Start kubelet and scrape /metrics/cadvisor
3. Observe that container-level metrics have empty values for Kubernetes-related labels

**Root cause**:
cAdvisor uses a strict regex `(\d+)\.(\d+)\.(\d+)` to parse Docker version strings. This regex fails to match versions that contain multiple hyphens (e.g., v 20.10.12-v 1.0.2), causing the Docker-specific handler to fail registration and fall back to the raw handler, which does not enrich metrics with Kubernetes metadata.

[cadvisor/container/docker/docker.go at master · google/cadvisor](https://github.com/google/cadvisor/blob/master/container/docker/docker.go#L185)
```go
func ParseVersion(versionString string, regex *regexp.Regexp, length int) ([]int, error) {
	matches := regex.FindAllStringSubmatch(versionString, -1)
	if len(matches) != 1 {
		return nil, fmt.Errorf("version string \"%v\" doesn't match expected regular expression: \"%v\"", versionString, regex.String())
	}
	versionStringArray := matches[0][1:]
	versionArray := make([]int, length)
	for index, versionStr := range versionStringArray {
		version, err := strconv.Atoi(versionStr)
		if err != nil {
			return nil, fmt.Errorf("error while parsing \"%v\" in \"%v\"", versionStr, versionString)
		}
		versionArray[index] = version
	}
	return versionArray, nil
}
```

**Proposed fix** :
To improve compatibility with custom or extended Docker version strings, I propose the following fix:
Even if the Docker version string contains non-standard suffixes, cAdvisor should still attempt to extract and use the first valid semantic version part (e.g., X.Y.Z) to register the Docker handler. This ensures that container metadata enrichment (e.g., pod, namespace, container, image) continues to function normally, without falling back to the raw handler.

1.  Improve version parsing logic
  - Use the regular expression `(\d+\.\d+\.\d+)` to extract the first matching semantic version part (e.g., 20.10.12).
  - If no valid version is found, return an error.
  - If a match is found, continue to register and use the Docker handler normally.
2.  Add warning logging
  - When a version string contains non-standard suffixes is detected, log a warning message in English for better compatibility with internationalized logging systems.
  - Example log:
```
[WARNING] Detected non-standard Docker version "v20.10.12-v1.0.2", using base version "20.10.12" for handler registration
```

Haihan-Jiang:
Fixes #3830. This is a replacement for the closed #3876 with the maintainer feedback addressed.

Docker server versions can be reported with a leading `v` and build/distribution suffixes such as `v20.10.12-v1.0.2`. cAdvisor currently parses Docker server versions with the generic unanchored `VersionRe`, which rejects that form because it finds multiple semantic-version substrings.

This change adds a Docker-specific server-version regexp that:
- requires the version to start at the beginning of the string, with an optional `v` prefix;
- captures the base `major.minor.patch`;
- allows an optional suffix after the base version;
- rejects arbitrary text such as `random text 1.2.3 and also 9.9.9 build`.

`VersionRe` remains unchanged for kernel-version parsing, so this does not broaden the older generic parser behavior.

Validation:
- `GOOS=linux go test -c ./container/docker -o /tmp/cadvisor-container-docker.test`
- `go test ./container`
- `git diff --check`

Note: `container/docker` is Linux-only, so on macOS I compile-validated the Linux test binary instead of executing it locally.

### DIFF
--- a/container/docker/docker.go
+++ b/container/docker/docker.go
@@ -114,7 +114,7 @@ func ValidateInfo(GetInfo func() (*dockersystem.Info, error), ServerVersion func
 		}
 	}
 
-	version, err := ParseVersion(info.ServerVersion, VersionRe, 3)
+	version, err := ParseVersion(info.ServerVersion, dockerVersionRe, 3)
 	if err != nil {
 		return nil, err
 	}
--- a/container/docker/docker_test.go
+++ b/container/docker/docker_test.go
@@ -30,8 +30,10 @@ func TestParseDockerAPIVersion(t *testing.T) {
 		expected      []int
 		expectedError string
 	}{
-		{"17.03.0", VersionRe, 3, []int{17, 03, 0}, ""},
-		{"17.a3.0", VersionRe, 3, []int{}, `version string "17.a3.0" doesn't match expected regular expression: "(\d+)\.(\d+)\.(\d+)"`},
+		{"17.03.0", dockerVersionRe, 3, []int{17, 03, 0}, ""},
+		{"v20.10.12-v1.0.2", dockerVersionRe, 3, []int{20, 10, 12}, ""},
+		{"17.a3.0", dockerVersionRe, 3, []int{}, `version string "17.a3.0" doesn't match expected regular expression: "^v?(\d+)\.(\d+)\.(\d+)(?:[-+].*)?$"`},
+		{"random text 1.2.3 and also 9.9.9 build", dockerVersionRe, 3, []int{}, `version string "random text 1.2.3 and also 9.9.9 build" doesn't match expected regular expression: "^v?(\d+)\.(\d+)\.(\d+)(?:[-+].*)?$"`},
 		{"1.20", apiVersionRe, 2, []int{1, 20}, ""},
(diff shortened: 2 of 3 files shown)


Haihan-Jiang:
This is ready for review. The Docker version parser now handles suffixed versions strictly, with tests around the factory version gating. The visible checks are green.

PR #3884, opened 2026-05-30
Use Redis storage password

### LINKED ISSUE #3741: Redis storage driver ignores --storage_driver_password, fails against AUTH-required Redis
### What happened?
When running cAdvisor v0.52.1 with `-storage_driver=redis`, `-storage_driver_host=redis:6379`, and `-storage_driver_password=redisPswrd`, no metrics are persisted. Redis logs report a dropped connection with “Possible SECURITY ATTACK” and the `LPUSH` never reaches the database. Manually replaying the connection shows the driver never sends `AUTH`, so Redis immediately rejects the request.

### How to reproduce?
1. Start Redis 7/8 with a password (e.g. `requirepass redisPassword`).
2. Run the stock cAdvisor image (`gcr.io/cadvisor/cadvisor:v0.52.1`) with: 
 -storage_driver=redis -storage_driver_host=redis:6379 -storage_driver_password=redisPswrd

3. Watch Redis via `MONITOR` or check for the `cadvisor` list key.

### What did you expect to happen?
cAdvisor should authenticate using the configured password and push container stats into Redis.

### What actually happened?
Redis blocks the unauthenticated connection. No `AUTH` command is issued by the driver, so the subsequent `LPUSH` never succeeds.

### Anything else we need to know?
- Confirmed by reading `cmd/internal/storage/redis/redis.go`—the driver dials Redis but ignores `storage.ArgDbPassword`.
- Reproduced on Docker Desktop / Debian Bookworm host.
- Workaround is to disable authentication on Redis, which isn’t desirable.

Haihan-Jiang:
## Summary
- authenticate Redis storage connections when `--storage_driver_password` is explicitly set
- keep unauthenticated Redis as the default because the shared storage password flag defaults to `root`
- use `Do("LPUSH", ...)` so writes are flushed and Redis command errors are returned
- add fake Redis server coverage for AUTH and LPUSH behavior

Fixes #3741.

## Testing
- `go test ./internal/storage/redis -count=1` from `cmd/`
- `go test ./internal/storage/...` from `cmd/`
- `GOOS=linux go test -c . -o /tmp/cadvisor-cmd-linux.test` from `cmd/`
- `git diff --check`

Note: `go test .` in `cmd/` cannot run on macOS because Linux-only cAdvisor packages are excluded by build tags, so I used the Linux compile check above for the command package.

### DIFF
--- a/cmd/internal/storage/redis/redis.go
+++ b/cmd/internal/storage/redis/redis.go
@@ -16,6 +16,7 @@ package redis
 
 import (
 	"encoding/json"
+	"flag"
 	"os"
 	"sync"
 	"time"
@@ -56,8 +57,21 @@ func new() (storage.StorageDriver, error) {
 		hostname,
 		*storage.ArgDbName,
 		*storage.ArgDbHost,
+		configuredRedisPassword(),
 		*storage.ArgDbBufferDuration,
 	)
+}
+
+func configuredRedisPassword() string {
+	password := ""
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == "storage_driver_password" {
+			// The shared storage password flag defaults to "root" for other
+			// backends. Redis should authenticate only when the flag is set.
+			password = *storage.ArgDbPassword
+		}
+	})
+	return password
 }
 
 func (s *redisStorage) defaultReadyToFlush() bool {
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. The visible checks are green; could someone take a look when convenient?

PR #3886, opened 2026-06-01
Exposing additional cgroup v2 memory.stat metrics

egorikas:
### What:
Publishing six additional cgroup v2 `memory.stat` fields that cAdvisor already collects but doesn't surface:
- `container_memory_file_dirty_bytes`
- `container_memory_file_writeback_bytes`
- `container_memory_pgscan_total`
- `container_memory_pgsteal_total`
- `container_memory_workingset_refault_file_total`
- `container_memory_workingset_refault_anon_total`
### Why:
Right now there's no easy way to see when a container is under memory pressure or doing a lot of disk writes. These fields cover that. When a container gets close to its memory limit, the kernel tries to free up space by dropping or writing out pages. 
- `pgscan/pgsteal` show how hard it's working at that (and if the numbers are far apart, it's struggling).
- `workingset_refault_*` show pages that were dropped and then needed again. That's a sign the container doesn't have enough memory.
### How:
- Read the values from the existing `MemoryStats.Stats` map in `setMemoryStats` (cgroup v2 only; left at zero on v1, where the keys don't exist).
- Add the fields to `info/v1.MemoryStats` and emit them under the existing memory metric set in `metrics/prometheus.go`.
- Add a unit test, update the Prometheus golden files, and add an integration test.


egorikas:
Hello @dims! @sambhav-jain-16 advised me to seek your review for that PR. Could you take a look? That's a production-proven implementation :)

roman-vynar:
This is good and not. 
`container_memory_*` metric grew so much that it produces high cardinality if you scrape from multiple nodes.
The only way to get rid of unnecessary(not for all) stuff is to drop on scrape on Prometheus side, e.g.:
```
metric_relabel_configs:
  - source_labels: [__name__]
    regex: 'container_memory_(failures_total|failcnt|mapped_file|max_usage_bytes|working_set_bytes|events_(high|max)_total|file_(dirty|writeback)_bytes|pg(scan|steal)_total|workingset_refault_(anon|file)_total)'
    action: drop
```

Can we make it configurable instead of adding a ton of stuff to the same metric out of control?
Bear with me it is mainly for troubleshooting/debugging.

egorikas:
> This is good and not. `container_memory_*` metric grew so much that it produces high cardinality if you scrape from multiple nodes. The only way to get rid of unnecessary(not for all) stuff is to drop on scrape on Prometheus side, e.g.:
> 
> ```
> metric_relabel_configs:
>   - source_labels: [__name__]
>     regex: 'container_memory_(failures_total|failcnt|mapped_file|max_usage_bytes|working_set_bytes|events_(high|max)_total|file_(dirty|writeback)_bytes|pg(scan|steal)_total|workingset_refault_(anon|file)_total)'
>     action: drop
> ```
> 
> Can we make it configurable instead of adding a ton of stuff to the same metric out of control? Bear with me it is mainly for troubleshooting/debugging.

Memory hardware is crazy expensive right now. And unfortunately, any memory optimization effort requires a pile of memory metrics.

In this particular case, I was open-sourcing the internal patch that helped a lot with the memory efficiency initiative. So, while I tend to agree with the idea, that cadvisor needs a more flexible configuration what to emit and what. In the same time, I tend to disagree, that memory metrics are useless/only-for-debugging.

PR #3888, opened 2026-06-09
Expose cgroup v2 memory.events OOM metrics

nils-stegra:
### What:
Follow-up to #3870. Expose the remaining cgroup v2 `memory.events` OOM counters as Prometheus metrics:
- `container_memory_events_oom_total`
- `container_memory_events_oom_kill_total`
- `container_memory_events_oom_group_kill_total`

### Why:
`memory.events` already contains these OOM counters on cgroup v2 systems. Exposing them gives operators direct cgroup-level visibility into allocation failures and OOM kills. This is separate from the existing `container_oom_events_total` metric, which comes from the OOM event path rather than the cgroup v2 `memory.events` counters.

### How:
- Read `oom`, `oom_kill`, and `oom_group_kill` in the existing `setMemoryEvents` path.
- Add the fields to `info/v1.MemoryEvents` and emit them under the existing memory metric set.
- Update unit tests, integration metric checks, docs, and Prometheus golden files.

### DIFF
--- a/container/libcontainer/handler.go
+++ b/container/libcontainer/handler.go
@@ -895,6 +895,15 @@ func setMemoryEvents(cgroupPath string, ret *info.ContainerStats) {
 	if val, err := fscommon.GetValueByKey(cgroupPath, "memory.events", "max"); err == nil {
 		ret.Memory.Events.Max = val
 	}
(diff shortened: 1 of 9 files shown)


nils-stegra:
Closing in favor of reopening from a personal GitHub account for CLA handling.

PR #3889, opened 2026-06-11
Expose remaining cgroup v2 memory.events counters (oom, oom_kill, oom_group_kill, low)

note89:
### What:
Follow-up to #3870. Expose the remaining cgroup v2 `memory.events` counters as Prometheus metrics:
- `container_memory_events_oom_total`
- `container_memory_events_oom_kill_total`
- `container_memory_events_oom_group_kill_total`
- `container_memory_events_low_total` (added on request from @egorikas)

With these, all six `memory.events` counters (`low`, `high`, `max`, `oom`, `oom_kill`, `oom_group_kill`) are exposed.

### Why:
`memory.events` already contains these counters on cgroup v2 systems. Exposing them gives operators direct cgroup-level visibility into allocation failures, OOM kills, and memory.low protection breaches. This is separate from the existing `container_oom_events_total` metric, which comes from the OOM event path rather than the cgroup v2 `memory.events` counters.

### How:
- Read `low`, `oom`, `oom_kill`, and `oom_group_kill` in the existing `setMemoryEvents` path.
- Add the fields to `info/v1.MemoryEvents` and emit them under the existing memory metric set.
- Update unit tests, integration metric checks, docs, and Prometheus golden files.


note89:
@egorikas I saw your comment, added low, tnx!

egorikas:
> @egorikas I saw your comment, added low, tnx!

Oops, I've somehow deleted it :) Thanks for adding a metric per my request :)

PR #3907, opened 2026-07-08
Update containerd ttrpc and typeurl

liggitt:
Picks up https://github.com/containerd/typeurl/pull/51 and https://github.com/containerd/ttrpc/pull/239

Helps in dropping out the unmaintained gogo dependency


liggitt:
cc @dims

liggitt:
sweet... if we can tag this (v0.60.4?) we can pull it into k/k and drop another link to gogo

dims:
will do a quick scan of anything else that needs to land then tag @liggitt

PR #3912, opened 2026-07-10
lib/manager: allow disabling container discovery

dgrisonnet:
When Kubelet has CRI stats enabled, the container runtime handles container tracking and cAdvisor is only queried for the root container stats. Discovering and watching all containers at startup is unnecessary overhead in that mode.

Add a DisableContainerDiscovery option to HousekeepingConfig that skips plugin initialization, cgroup watchers, subcontainer scanning, and the global housekeeping loop. The root container is still created at startup to serve node-level resource and filesystem stats. Other containers are created lazily via getContainer() on first query.


dgrisonnet:
This is the PR that pulls that change: https://github.com/kubernetes/kubernetes/pull/140429

I added performance tests results in the description.

dgrisonnet:
cc @dims

dims:
thanks @dgrisonnet Damien!

PR #3918, opened 2026-07-20
fs/zfs: fall back to VFS when zfs stats are unavailable

### LINKED ISSUE #3904: zfsPlugin.GetStats does not work in unprivileged containers
### Summary

When launching kubelet v1.36.1 in unprivileged LXC containers with a zfs root volume, kubelet fails with the following error:

```
09:10:26 capn-default-unprivileged-2fnr-47zq6-cbrt2 kubelet[697]: E0518 09:10:26.166372     697 container_manager_linux.go:987] "Unable to get rootfs data from cAdvisor interface" err="cannot find filesystem info for device \"zfs/containers/capn-default-unprivileged-2fnr-47zq6-cbrt2\""
```

### Details

The same issue does not appear on older Kubernetes versions, and I've found that:
- Kubernetes v1.36.1 uses cadvisor `v0.56.2` https://github.com/kubernetes/kubernetes/blob/release-1.36/go.mod
- Kubernetes v1.35.5 uses cadvisor `v0.53.0` https://github.com/kubernetes/kubernetes/blob/release-1.35/go.mod

The issues seems to be related with a behavior change during the [migration to plugins for cadvisor filesystems](https://github.com/google/cadvisor/pull/3794/changes). 

[Previous behavior was](https://github.com/google/cadvisor/blob/v0.53.0/fs/fs.go#L433-L440):
- `os.Stat("/dev/zfs")` is attempted, and zfs-lib code path is used if it returns an error that is `os.IsExist(err)` (which seems not correct?)
- in other cases, it falls back to the vfs plugin

New behavior is: 
- `os.Stat("/dev/zfs")` is attempted, and if it returns an `os.IsNotExist(err)`, then it falls back to the vfs plugin
- otherwise, it attempts to read from zfslib, and returns an error otherwise.
- this fails in unprivileged containers, as they do not have permissions to read from /dev/zfs.

We are finding this issue with `cluster-api-provider-incus`, which currently falls back to [non-zfs storage for Kubernetes v1.36.1](https://github.com/lxc/cluster-api-provider-incus/blob/main/test/e2e/shared/hack_unprivileged.go#L27-L53)

### How to address

It appears that there are two possible paths:
1. fallback to vfs if the GetZfsStats(...) call fails.
2. (if cadvisor rejects the above) attempt to mask the `/dev/zfs` in unprivileged containers, such that the `os.Stat(..)` call fails and cadvisor falls back to vfs.

Can create a PR for (1), but looking for maintainer opinions as well.

locker95:
#### Description

Unprivileged containers (e.g. LXC) with a ZFS root often still have `/dev/zfs`, but the process cannot use the zfs module. After the filesystem plugin migration (#3794), cAdvisor selected the ZFS plugin whenever the device node existed and returned an error from `GetZfsStats` instead of falling back to VFS. That leaves kubelet without rootfs capacity (`cannot find filesystem info for device "zfs/..."`).

This restores the practical pre-migration behavior for those hosts:

- If `/dev/zfs` is missing **or unusable**, fall back to VFS via `ErrFallbackToVFS`
- If `zfs list` fails for any reason (including EPERM), fall back to VFS as well

Privileged hosts with working ZFS continue to use native zfs stats.

#### Related Issue

Fixes #3904

#### Testing Done

```
cd lib && go test ./fs/zfs/ -count=1 -v
```

- `TestGetStatsFallsBackWhenDevMissing`
- `TestGetStatsFallsBackOnZfsError` (simulates EPERM after device present)
- `TestGetStatsSuccess`
- existing `TestParseZfsListUsage`

### DIFF
--- a/lib/fs/zfs/plugin.go
+++ b/lib/fs/zfs/plugin.go
@@ -24,6 +24,12 @@ import (
 
 type zfsPlugin struct{}
 
+// Overridable for tests.
+var (
+	zfsDevicePath = "/dev/zfs"
+	getZfsStats   = GetZfsStats
+)
+
 // NewPlugin creates a new ZFS filesystem plugin.
 func NewPlugin() fs.FsPlugin {
 	return &zfsPlugin{}
@@ -44,15 +50,19 @@ func (p *zfsPlugin) Priority() int {
 }
 
 // GetStats returns filesystem statistics for ZFS.
+//
+// When /dev/zfs is missing, or when zfs tooling cannot collect stats (common
+// in unprivileged containers that still see the device node but lack
+// permission to use it), fall back to VFS so mountpoint capacity remains
+// available via statfs.
 func (p *zfsPlugin) GetStats(device string, partition fs.PartitionInfo) (*fs.FsStats, error) {
-	// Check if ZFS is available - if /dev/zfs doesn't exist, fall back to VFS
-	if _, err := os.Stat("/dev/zfs"); os.IsNotExist(err) {
+	if _, err := os.Stat(zfsDevicePath); err != nil {
 		return nil, fs.ErrFallbackToVFS
 	}
 
-	capacity, free, avail, err := GetZfsStats(device)
+	capacity, free, avail, err := getZfsStats(device)
 	if err != nil {
-		return nil, err
+		return nil, fs.ErrFallbackToVFS
 	}
 
 	return &fs.FsStats{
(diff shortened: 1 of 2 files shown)


locker95:
cla: yes

locker95:
I have signed the Google CLA with this GitHub account (locker95) and authorized email 862469039@qq.com.

cla: yes

Please recheck.

Issue #3828, opened 2026-01-24
Downgrade warning for NUMA nodes without CPUs (valid Linux topology)


bon77:
On my aarch64 architecture node (RaspberryPi) I get the following warning:
```
alloy[669]: ts=2026-01-24T22:30:39.860434727Z level=warn msg="Found node without any CPU, nodeDir: /sys/devices/system/node/node1, number of cpuDirs 0, err: <nil>" component_path=/ component_id=prometheus.exporter.cadvisor.integrations_cadvisor func=Warningf
```
I think the issue is that on ARM systems, Linux may expose NUMA node directories that only contain cpulist / cpumap **files** and no per-CPU **directories**. 

I suggest to change the [code](https://github.com/google/cadvisor/blob/566cb6326fb0804aa9f960723be74993ba4d6dc1/utils/sysinfo/sysinfo.go#L227) from `klog.Warning` to a lower log level or to change the logic to also work with this cpu setup.

Additional info about this specific CPU:
```
 $ uname -m
aarch64
 $ cat /proc/cpuinfo | grep -m1 "CPU architecture"
CPU architecture: 8
 $ ls -1d /sys/devices/system/node/node*
/sys/devices/system/node/node0
/sys/devices/system/node/node1
 $ for n in /sys/devices/system/node/node*; do
>   echo "== $n =="
>   ls "$n" | grep '^cpu' || echo "(no CPUs)"
> done
== /sys/devices/system/node/node0 ==
cpu0
cpu1
cpu2
cpu3
cpulist
cpumap
== /sys/devices/system/node/node1 ==
cpulist
cpumap
 $ for c in /sys/devices/system/cpu/cpu*/topology/physical_package_id; do
>   echo "$c: $(cat "$c")"
> done
/sys/devices/system/cpu/cpu0/topology/physical_package_id: 0
/sys/devices/system/cpu/cpu1/topology/physical_package_id: 0
/sys/devices/system/cpu/cpu2/topology/physical_package_id: 0
/sys/devices/system/cpu/cpu3/topology/physical_package_id: 0
 $ cat /sys/devices/system/cpu/cpu0/topology/numa_node
cat: /sys/devices/system/cpu/cpu0/topology/numa_node: No such file or directory

```

PR #3931, opened 2026-09-08
Treat memory-only NUMA nodes as valid topology

joejulian:
Memory-only NUMA nodes are valid Linux topology, but `GetNodesInfo` currently emits a warning whenever a node has no CPU directories. On systems that intentionally expose memory-only nodes this warning repeats on every topology refresh.

Treat an empty CPU set as a valid memory-only node and move that diagnostic to V(4). Return actual `GetCPUsPaths` failures instead of conflating them with the valid empty result.

This follows up #3828.

Tests added for:
- retaining a memory-only NUMA node and its memory data without a warning-level failure path
- propagating a real CPU-path discovery error

Validation:
- `go test ./utils/sysinfo` (from `lib/`)
- `go test -race ./utils/sysinfo` (from `lib/`)
- `go vet ./...` (root and `lib/`)
- `gofmt -l .`


joejulian:
@dims since you touched this file last, would you please take a look?

dims (approved):

